PR #28, opened 2023-05-26
feat: modernize the Benchmark report

triceo:
- Convert to use the Timefold webjar.
- Convert to Bootstrap 5.
- Charts are now generated in the browser by JavaScript, not by JFreeChart.
- Address several long-standing bugs identified during the process.


benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/WebsiteResourceUtils.java
```diff
@@ -5,33 +5,54 @@
...
-        // copyResource(benchmarkReportDirectory, "twitterbootstrap/js/bootstrap.min.js");
-        copyResource(benchmarkReportDirectory, "twitterbootstrap/js/jquery.js");
-        // copyResource(benchmarkReportDirectory, "twitterbootstrap/js/jquery.min.js");
-        copyResource(benchmarkReportDirectory, "twitterbootstrap/js/prettify.js");
-        // Website resources
-        copyResource(benchmarkReportDirectory, "website/css/benchmarkReport.css");
-        copyResource(benchmarkReportDirectory, "website/img/timefold-logo.png");
+        // Extract webjars.
+        File webjarDirectory = new File(benchmarkReportDirectory, "website/webjars/");
+        new WebJarAssetLocator()
```
triceo, line 17: This copies *the entire webjars*, incl. licenses and possibly other stuff we don't need.
As a result, the benchmark report now carries ~20 MB of stuff.
Maybe we want to write code to prune this before we merge?

triceo:
@mcimbora I see you recently fixed a bug in the benchmarker. Maybe you want to give this a try before we merge it? Any feedback appreciated.

mcimbora:
Zdarec @triceo, I checked this out briefly from user perspective. The report looks a lot better 👍 JFreeChart is a great library, but the charts have a bit of 90s vibe. Perhaps they've added new renderers meanwhile? Haven't checked.

A couple of points/glitches, feel free to pick what's relevant:
- output folder size 40 MB (new) vs 1.6 MB (old). Will make difference if you have 50 results in the output folder.
- the charts are responsive, but one way only. If you descale the window and resize it back to original size, the charts don't rescale back, e.g.:
image

- new charts sometimes don't include origin (see vertical axis on chart below) and sometimes add negative range values (see horizontal axis)
image

vs:

image

- maybe provide better indication that the labels are clickable. Figured it out by chance :-)
- the winner line is thicker and sometimes obscures other configurations, e.g.:
image
image

- section labels could be aligned left:
image

triceo:
Díky, @mcimbora ! I'll see what I can do about the things you mentioned.

triceo:
I think I addressed most of the issues.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
```diff
@@ -131,8 +130,7 @@ public BenchmarkAggregatorFrame(BenchmarkAggregator benchmarkAggregator) {
         this.benchmarkAggregator = benchmarkAggregator;
         benchmarkResultIO = new BenchmarkResultIO();
         plannerBenchmarkResultList = Collections.emptyList();
-        Locale locale = benchmarkAggregator.getBenchmarkReportConfig().determineLocale();
-        millisecondsSpentNumberFormat = new MillisecondsSpentNumberFormat(locale);
+        millisecondsSpentNumberFormat = new MillisecondsSpentNumberFormat();
```
ge0ffrey, line 133: A) Either the benchmark report has a locale.
B) Or the locale is the eye of the beholder. Basically, the browser's locale determines if it's  1'000,05 (BE) or 1,000.05 (US).

If B) is possible, I prefer that. A big advantage over the JfreeChart approach is that the locale isn't hard coded in the report output.

Anyway:
- In A), the MillisecondsSpentNumberFormat should still use the locale. Even if it's just for tables in html.
- In B), the millisecondsSpentNumberFormat can be removed, as no number is formatted during report generation (only during report rendering).

ge0ffrey (approved):
LGTM, see soft suggestions

PR #68, opened 2023-06-05
perf: make CS flattenLast() perform better on collections

triceo:
Represents a marginal improvement of Nurse Rostering.
A related improvement to Project Job Scheduling increases performance there by ~300 %.


examples/src/main/java/ai/timefold/solver/examples/projectjobscheduling/score/ProjectJobSchedulingConstraintProvider.java
```diff
@@ -43,9 +40,7 @@ protected Constraint renewableResourceCapacity(ConstraintFactory constraintFacto
                 .filter(ResourceRequirement::isResourceRenewable)
                 .join(Allocation.class,
                         Joiners.equal(ResourceRequirement::getExecutionMode, Allocation::getExecutionMode))
-                .flattenLast(a -> IntStream.range(a.getStartDate(), a.getEndDate())
-                        .boxed()
-                        .collect(Collectors.toList()))
+                .flattenLast(Allocation::getBusyDates)
```
triceo, line 43: This improves PJS by about ~300 %.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/AbstractFlattenLastNode.java
```diff
@@ -103,6 +123,29 @@ public void update(InTuple_ tuple) {
...
     }
 
+    private List<FlattenedItem_> iterableToList(Iterable<FlattenedItem_> iterable) {
+        if (iterable instanceof Collection<FlattenedItem_> collection) {
+            // Optimization for Collection, where we know the size.
+            int size = collection.size();
+            if (size == 0) {
+                return Collections.emptyList();
+            }
+            List<FlattenedItem_> result = new ArrayList<>(size);
```
rsynek, line 133: In this case, it is safe to cast the `iterable` to a `Collection` and pass it here directly. It could bring an additional minor perf improvement, if we skip the `forEach` and the method reference (no proof).

rsynek (approved):
LGTM, one small suggestion to consider.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/AbstractFlattenLastNode.java
triceo: I actually tried that and, for some reason, sizing the collection and filling it like this is marginally *faster* than initializing the list with the collection directly. Overall, the difference wasn't big enough for me to pursue this further.

PR #69, opened 2023-06-05
Create semantic.yml

triceo:


.github/semantic.yml
```diff
@@ -0,0 +1,17 @@
...
+  - feat
+  - fix
+  - docs
+  - style
+  - refactor
+  - perf
+  - test
+  - build
+  - ci
+  - chore
```
rsynek, line 15: What's the difference between style, refactor and chore? And what about build vs. ci? The fewer keywords we use, the bigger chance we use them right.

triceo: image

https://gist.github.com/brianclements/841ea7bffdb01346392c

rsynek: Thanks; I like "refactor" more than "chore", but in general, the categories are clear now.

PR #81, opened 2023-06-12
ci: Run timefold-quickstarts in PRs

triceo:


.github/workflows/downstream_quickstarts.yml
```diff
@@ -0,0 +1,54 @@
+name: Downstream - Timefold Quickstarts
+
+on:
+  pull_request:
+    branches: [main]
```
triceo, line 5: Should we support `.x` branches?
How should it behave then?

rsynek: It does not hurt setting it up already. I think it should already work with you script for PRs over both repositories. For a PR to solver only, you have to provide the right default branch (the same `.x` branch) from the target organization and repository.

.github/workflows/downstream_quickstarts.yml
```diff
@@ -0,0 +1,54 @@
+name: Downstream - Timefold Quickstarts
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
rsynek: That's a question. Cloning the PR author repositories without merging with the target repositories does not protected from stale PR, which might, in theory, result in issues only after merging the PR.

If we want to cover this corner case, we have to either A) merge the forks with origin repos before the build or B) set up a GHA that checks the repositories again after merging.

Obviously, A) is more complex, but we get the result before merging the PR, while B) is easier, but also wasteful (we run the tests twice) and comes a bit too late.

rsynek (approved):
LGTM, thanks for addressing the stale forks!

PR #85, opened 2023-06-13
perf: speed up CS consecutive collector

triceo:
Improves nurse rostering performance by ~15 %.

### DIFF
--- a/examples/src/main/java/ai/timefold/solver/examples/common/experimental/impl/ComparableValue.java
+++ b/examples/src/main/java/ai/timefold/solver/examples/common/experimental/impl/ComparableValue.java
@@ -1,0 +1,53 @@
+package ai.timefold.solver.examples.common.experimental.impl;
+
+import java.util.TreeMap;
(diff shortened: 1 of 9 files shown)


examples/src/main/java/ai/timefold/solver/examples/common/experimental/impl/ComparableValue.java
```diff
@@ -0,0 +1,32 @@
...
+        implements
+            Comparable<ComparableValue<Value_, Point_>> {
+
+    @Override
+    public int compareTo(ComparableValue<Value_, Point_> other) {
+        if (this == other) {
+            return 0;
+        }
+        Point_ point1 = this.index;
+        Point_ point2 = other.index;
```
triceo, line 13: This would previously do a hashmap lookup, which is extremely expensive for a comparator.

rsynek (approved):
Good to merge when the CI is green. There seem to be some test failures.

PR #90, opened 2023-06-14
fix: address various issues identified during benchmarking

triceo:
Includes the following fixes:

- NPE in the benchmark report when no subsingles used.
- Exception in 2-opt in some VRP cases.
- More appropriate (but much slower) mutation counting in entities.
- Mutation chart correctly drops to zero on the next tick.
- Charts correctly respond to when datasets are removed.
- Screenshots in the documentation match what the benchmarker will actually generate.


core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/mutation/MutationCounter.java
```diff
@@ -35,11 +37,34 @@ public int countMutations(Solution_ a, Solution_ b) {
...
+                        List<Object> aValues = listVariableDescriptor.getListVariable(aEntity);
+                        List<Object> bValues = listVariableDescriptor.getListVariable(bEntity);
+                        int aSize = aValues.size();
+                        int bSize = bValues.size();
+                        if (aSize != bSize) {
+                            // First add mutations for the elements that are missing in one list.
+                            mutationCount += Math.abs(aSize - bSize);
+                        }
+                        // Then iterate over the list and count every item that is different.
+                        for (int i = 0; i < aSize; i++) {
```
rsynek, line 52: Maybe slightly more readable:

```
int shorterListSize = Math.min(aSize, bSize);
for (int i = 0; i < shorterListSize; i++) {
```

rsynek (approved):
Looks good; one small suggestion.

PR #99, opened 2023-06-16
fix: allow min/max constraint collectors on planning variables

triceo:
The pre-existing way of doing min/max constraint collectors relied on a user-supplied comparator.
The values would then be put straight into a `TreeMap`, sorted according to this comparator.

However, if the comparator relied on a planning variable, there would be issues.
During insertion, the value would be put into a map.
But then the value's comparison would change due to the planning variable changing.
As a result, on value removal from the map, the value would fail to be removed, leading to exceptions.

The new code no longer relies on user-supplied comparator. 
Instead, it asks the user to supply a function to return the comparable object and creates the comparator internally.
This allows the collector to store the value that was inserted, and on removal, remove that same value; not the current value returned by the user-supplied function.

This performs worse, but it is safer and therefore preferrable.
The old methods were deprecated, as there is no way to make them safe.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
@@ -732,7 +732,7 @@ public final class ConstraintCollectors {
      * incremental score calculation.
      * <p>
      * For example, {@code [Ann(age = 20), Beth(age = 25), Cathy(age = 30), David(age = 30), Eric(age = 20)]} with
-     * {@code .groupBy(max(Person::name, Person::age))} returns {@code Ann} or {@code Eric},
+     * {@code .groupBy(min(Person::name, Person::age))} returns {@code Ann} or {@code Eric},
      * as both have the same age.
      * <p>
      * The default result of the collector (e.g. when never called) is {@code null}.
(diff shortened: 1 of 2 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -718,19 +718,130 @@ private static <Result> Runnable innerSum(MutableReference<Result> resultContain
...
+     * The elements will be compared according to the value returned by the comparable function.
+     * <p>
+     * Important: The {@link Comparable}'s {@link Comparable#compareTo(Object)} must be <i>consistent with equals</i>,
+     * such that {@code e1.compareTo(e2) == 0} has the same boolean value as {@code e1.equals(e2)}.
+     * In other words, if two elements compare to zero, any of them can be returned by the collector.
+     * It can even differ between 2 score calculations on the exact same {@link PlanningSolution} state, due to
+     * incremental score calculation.
+     * <p>
+     * For example, {@code [Ann(age = 20), Beth(age = 25), Cathy(age = 30), David(age = 30), Eric(age = 20)]} with
+     * {@code .groupBy(max(Person::name, Person::age))} returns {@code Ann} or {@code Eric},
```
rsynek, line 735:
```suggestion
     * {@code .groupBy(min(Person::name, Person::age))} returns {@code Ann} or {@code Eric},
```

rsynek (approved):
I thinks there is one copy-paste error, otherwise LGTM.

PR #103, opened 2023-06-16
fix: when there are no possible values, skip solving

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/AbstractFromPropertyValueRangeDescriptor.java
```diff
@@ -127,15 +127,7 @@ protected ValueRange<?> readValueRange(Object bean) {
...
         }
-        valueRange = doNullInValueRangeWrapping(valueRange);
-        if (valueRange.isEmpty()) {
-            throw new IllegalStateException("The @" + ValueRangeProvider.class.getSimpleName()
-                    + " annotated member (" + memberAccessor
-                    + ") called on bean (" + bean
-                    + ") must not return an empty valueRange (" + valueRangeObject + ").\n"
-                    + "Maybe apply overconstrained planning as described in the documentation.");
-        }
-        return valueRange;
```
triceo, line 138: Previously, we *enforced* no empty value ranges.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -71,10 +72,33 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+                    }
+                    return false;
+                });
+        if (entityWithoutPossibleValues != null) {
+            logger.info(
+                    "Skipped all phases ({}): there are no possible values for planning variable(s) on one or more of planning entities. "
+                            + "See previous log messages for more information.",
+                    phaseList.size());
             return;
         }
```
triceo, line 103: Now, we check every entity for empty ranges, at some cost to solver startup.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -566,9 +566,7 @@ void solveChainedEmptyEntityList() {
         assertThat(solution.getScore().isSolutionInitialized()).isTrue();
     }
 
-    // TODO https://issues.redhat.com/browse/PLANNER-1738
     @Test
-    @Disabled("We currently don't support an empty value list yet if the entity list is not empty.")
     void solveEmptyValueList() {
```
triceo, line 570: And to my surprise, there already was a test for this. :-)

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -992,9 +1020,11 @@ public long getMaximumValueCount(Solution_ solution) {
     }
 
     /**
+     * @deprecated This was never implemented; any code that depends on this will throw {@link UnsupportedOperationException}.
      * @param solution never null
      * @return {@code >= 0}
      */
+    @Deprecated(forRemoval = true, since = "1.0.0")
     public int getValueCount(Solution_ solution) {
```
triceo, line 1028: Normally I'd just remove it, but that would require removing value ratio tabu.
It *doesn't work* and it never worked, throwing an exception.
But technically, it is a backwards incompatible change.

ge0ffrey, line 1028: SolutionDescription isn't public API nor documention nor really know.
I'd just remove the method.

triceo: See my explanatory comment on this PR.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -838,6 +839,12 @@ public void validateConstraintWeight(String constraintPackage, String constraint
     // Extraction methods
     // ************************************************************************
 
+    public Collection<Object> getAllEntities(Solution_ solution) {
```
ge0ffrey, line 842: Is this method used somewhere in the new code?

triceo: No, will delete.

triceo:
Turned this PR back into a draft; Geoffrey and I had a long discussion, and we decided to increase the scope of this task.

PR #115, opened 2023-06-20
ci: improve release automation

triceo:


.github/workflows/release.yml
```diff
@@ -1,16 +1,27 @@
...
+        description: 'Next version after release for release branch (e.g. 1.0.1, -SNAPSHOT will be added automatically)'
+        required: true
+      sourceBranch:
+        description: 'Branch to cut the release from'
+        default: main
+        required: true
+      releaseBranch:
+        description: 'Release branch to create (e.g. 1.0.x for version 1.0.0; once created, branch protection rules apply)'
+        default: dry_run
+        required: true
```
rsynek, line 21: Let's move these two before the "Next version..." parameters to define these branches before referring to them.

.github/workflows/release.yml
```diff
@@ -19,79 +30,94 @@ jobs:
...
-            timefold-solver/target/jreleaser/output.properties
+            target/jreleaser/trace.log
+            target/jreleaser/output.properties
 
-      - name: Set snapshot version timefold-solver
-        working-directory: ./timefold-solver
-        run: mvn --no-transfer-progress --batch-mode -Dfull versions:set -DnewVersion=${{ github.event.inputs.nextVersion }}-SNAPSHOT
+      - name: Set micro snapshot version on the release branch
+        run: |
+          mvn --no-transfer-progress --batch-mode -Dfull versions:set -DnewVersion=${{ github.event.inputs.nextMicroVersion }}-SNAPSHOT
```
rsynek, line 91: Something for consideration: define some MAVEN_OPTIONS (?) env var to keep the `-no-transfer-progress --batch-mode` in a single place.

.github/workflows/release.yml
```diff
@@ -19,79 +30,94 @@ jobs:
...
+          git add docs/src/antora.yml
+          find . -name 'pom.xml' | xargs git add
+          git commit -m "chore: move to ${{ github.event.inputs.nextVersion }}-SNAPSHOT"
+
+      # JReleaser should push the tag on its own, but it does not due to a bug.
+      # https://github.com/jreleaser/jreleaser/issues/1425
+      - name: Push branches to git
+        run: |
+          git push origin ${{ github.event.inputs.releaseBranch }}
+          git push origin ${{ github.event.inputs.releaseBranch }}-bump
```
rsynek, line 114: In the last stage of the release, this branch should be deleted. Apart from just having a stale branch in the git repository, if you want to do a micro release from already existing release branch, there will likely be a conflict with the existing release-bump branch.

rsynek (approved):
Nice work! I left there some suggestions, but nothing really blocks merging.

.github/workflows/release.yml
triceo: This needs to exist for the PR. The PR is responsible for deleting that branch; in fact, it happens automatically. I'll add it to the manual PR steps.

PR #128, opened 2023-06-25
ci: enable CodeQL scanning

triceo:


triceo:
Doesn't support Java 17.

PR #129, opened 2023-06-25
ci: fix MAVEN_OPTS in release CI

triceo:


triceo:
Harmless and necessary to make progress, so merged.

PR #141, opened 2023-06-28
fix: make Quarkus and Spring not fail without global solver termination

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -579,6 +580,19 @@ public SolverConfig withMonitoringConfig(MonitoringConfig monitoringConfig) {
     // Smart getters
     // ************************************************************************
 
+    public boolean willTerminate() {
```
triceo, line 583: I'm thinking maybe a rename?
`willTerminate` sounds a lot like a solution to the stopping problem, which it certainly isn't. :-)

rsynek (approved):

PR #143, opened 2023-06-28
fix: don't attempt to count entities which do not match the descriptor

triceo:


test/src/main/java/ai/timefold/solver/test/api/solver/change/MockProblemChangeDirector.java
```diff
@@ -81,7 +81,7 @@ public <EntityOrProblemFact> EntityOrProblemFact lookUpWorkingObjectOrFail(Entit
     @Override
     public <EntityOrProblemFact> Optional<EntityOrProblemFact>
             lookUpWorkingObject(EntityOrProblemFact externalObject) {
-        return Optional.of((EntityOrProblemFact) lookUpTable.get(externalObject));
+        return Optional.ofNullable((EntityOrProblemFact) lookUpTable.get(externalObject));
```
triceo, line 84: Unrelated, but I put it in as it was too small otherwise.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -632,14 +624,6 @@ public Map<String, MemberAccessor> getEntityCollectionMemberAccessorMap() {
         return entityCollectionMemberAccessorMap;
     }
 
-    public List<String> getEntityMemberAndEntityCollectionMemberNames() {
```
triceo, line 635: Dtto.

rsynek (approved):

PR #173, opened 2023-07-14
fix: fail fast on sorted selection with step cache

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/AbstractSelectorFactory.java
```diff
@@ -24,6 +24,22 @@ protected void validateCacheTypeVersusSelectionOrder(SelectionCacheType resolved
                 break;
             case SORTED:
             case SHUFFLED:
```
triceo, line 26: Tests currently think that STEP+SHUFFLED is supported.

That said, I am 100 % certain that the situation described for STEP+SORTED will happen there too. (They use the same parent class, therefore have the same problem.)

ge0ffrey: step caching with shuffled is a theoretically possible use case, no?
You generate all your moves at the beginning of a step (because you can't do it at phase cache because the generated moves change depending on the variables state), and then you want to go through them shuffled (unlike random get every element at most once).

Sorted has the same logic, no?

triceo: The comment in the code explains what exactly is wrong.
And yes, Sorted and Shuffled share the exact same code, so the NPE will be the exact same.

triceo:
Closing as outdated.
The underlying issue is not a priority at the moment.

PR #195, opened 2023-08-03
fix: bail out of a never-ending loop when no move is doable

triceo:
Move selector has supported bail-out for a long time.
However, it did not apply to non-doable moves, so when all the moves were non-doable, the bail-out would never happen.
This change makes bail-out happen for non-doable moves as well, by filtering out non-doable moves directly in the move selector. This requires to force an arbitrary limit on the number of moves that will be evaluated, as not every move selector provides its size.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
@@ -10,13 +10,13 @@ import ai.timefold.solver.core.impl.heuristic.selector.move.AbstractMoveSelector
 import ai.timefold.solver.core.impl.heuristic.selector.move.MoveSelector;
 import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
 
-public class FilteringMoveSelector<Solution_> extends AbstractMoveSelector<Solution_> {
+public final class FilteringMoveSelector<Solution_> extends AbstractMoveSelector<Solution_> {
 
-    protected final MoveSelector<Solution_> childMoveSelector;
-    protected final SelectionFilter<Solution_, Move<Solution_>> filter;
-    protected final boolean bailOutEnabled;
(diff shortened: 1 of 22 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
```diff
@@ -96,16 +96,29 @@ protected Move<Solution_> createUpcomingSelection() {
...
         }
-        return childMoveSelector.getSize() * 10L;
+        try {
+            return childMoveSelector.getSize() * 10L;
+        } catch (Exception ex) {
+            /*
+             * Some move selectors throw an exception when getSize() is called.
+             * In this case, we choose to disregard it and pick a large-enough bail-out size anyway.
+             */
+            long bailOutSize = Short.MAX_VALUE * 10L;
```
triceo, line 110: Interesting spot.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/LocalSearchDecider.java
```diff
@@ -94,15 +94,9 @@ public void decideNextStep(LocalSearchStepScope<Solution_> stepScope) {
         for (Move<Solution_> move : moveSelector) {
             LocalSearchMoveScope<Solution_> moveScope = new LocalSearchMoveScope<>(stepScope, moveIndex, move);
             moveIndex++;
-            // TODO use Selector filtering to filter out not doable moves
```
triceo, line 97: Geoffrey obviously expected something like this PR would eventually get done.
I'm assuming it never got done because of the size issues above.

triceo, line 97: Geoffrey obviously expected something like this PR would eventually get done.
I'm assuming it never got done because of the size issues above.

rsynek (approved):

PR #205, opened 2023-08-14
perf: read value range size instead of the value range

triceo:
To read an entire value range to get its size,
the value range needs to be first converted to a collection. 
This is costly and we therefore implement a shortcut, 
allowing us to read the size without creating a collection.


rsynek (approved):

triceo:
@rsynek Please re-review, I decided to add one more change. (Deprecating uncountable, because that really barely ever works and never should have been added.)

rsynek (approved):

PR #206, opened 2023-08-14
deps: bump Quarkus to 3.2.4.Final

triceo:


quarkus-integration/quarkus-benchmark/integration-test/pom.xml
```diff
@@ -81,4 +81,42 @@
...
+          <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-1.2-api</artifactId>
+            <version>${version.org.apache.logging.log4j}</version>
+          </dependency>
+        </dependencies>
+      </dependencyManagement>
+      <properties>
+        <quarkus.package.type>native</quarkus.package.type>
+        <quarkus.native.additional-build-args>--initialize-at-run-time=freemarker.ext.jython.JythonWrapper</quarkus.native.additional-build-args>
```
rsynek, line 117: Good as a temporary workaround, but please at least file an issue to investigate this further. Ideally, this should be reported to Quarkus.

If this module is not supposed to work in native, we should not lose time by native compilation in every CI run.

rsynek (approved):
Good to merge after linking an issue to that we don't forget to investigate the native build activation.

quarkus-integration/quarkus-benchmark/integration-test/pom.xml
triceo: Agreed off-line that no issue will be filed.

PR #215, opened 2023-08-15
fix: propagate retracts before updates in order to have clean indexes

triceo:
Backports 0b47ed1ec6332bcbd1ef91075326304deb7bc61b from main.
The backport is done via moving the entire CS impl from main to 0.8.x,
so technically all the improvements in CS since branching are now in 0.8.x as well.
(Except for new features such as expand(), because I did not touch public API.)

Also upgrades Drools, Quarkus and Spring.
Turtle tests pass.


rsynek (approved):
LGTM.

I've reviewed mainly the other commits in this PR, as the propagation has been copied from the `main` branch and tests pass.

PR #223, opened 2023-08-17
fix: address some bugs suggested by Sonar

triceo:


rsynek (approved):
I love seeing that SonarCloud brings results :)

PR #226, opened 2023-08-19
refactor: Changes code to use Java 17's instanceof pattern matching

timtebeek:
Hi! Figured create a PR to adopt some Java 17 language features that hadn't been picked up yet. Hope that helps!

### DIFF
--- a/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
+++ b/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/swingui/BenchmarkAggregatorFrame.java
@@ -416,15 +416,15 @@ public class BenchmarkAggregatorFrame extends JFrame {
                                 "New benchmark's name cannot be empty.",
                                 "Warning", JOptionPane.WARNING_MESSAGE);
                     } else {
-                        if (benchmarkResult instanceof PlannerBenchmarkResult) {
-                            ((PlannerBenchmarkResult) benchmarkResult).setName(newBenchmarkResultName);
+                        if (benchmarkResult instanceof PlannerBenchmarkResult result) {
+                            result.setName(newBenchmarkResultName);
                             mixedCheckBox.setText(newBenchmarkResultName);
                             ((DefaultTreeModel) checkBoxTree.getModel()).nodeChanged(treeNode);
-                        } else if (benchmarkResult instanceof SolverBenchmarkResult) {
+                        } else if (benchmarkResult instanceof SolverBenchmarkResult result) {
                             mixedCheckBox.setText(newBenchmarkResultName + " ("
-                                    + ((SolverBenchmarkResult) benchmarkResult).getRanking() + ")");
+                                    + result.getRanking() + ")");
                             ((DefaultTreeModel) checkBoxTree.getModel()).nodeChanged(treeNode);
-                            solverBenchmarkResultNameMapping.put((SolverBenchmarkResult) benchmarkResult,
+                            solverBenchmarkResultNameMapping.put(result,
                                     newBenchmarkResultName);
                         }
                         dispose();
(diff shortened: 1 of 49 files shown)


triceo (commented):
Thank you for the contribution! (We do usually make changes like these to files as we touch them for other reasons, so it takes a while to convert the entire codebase.)

The pattern matching changes are largely beneficial, and I like them. Except maybe where it uses var names such as `class1`, I'd prefer if they were `castClass` instead, because that is the pattern we generally use in other places in the code.


However, I am not too excited about the string-related changes.

- I don't see much difference between `"...".formatted(...)` and `String.format(...)`. In fact, one could argue that choosing one over the other is a matter of personal preference. As such, I'd leave those as is; there appears to be no reason for a change.
- Text block are a nice feature, but I question if they bring benefits in many of the cases we see here. Yes, it does remove string concat and `\n`, but it often adds `\` instead. So it doesn't really remove boilerplate, just replaces one with another. As in the previous case, without strong evidence for *needing* this change, I'd leave it out.

How hard would it be for you to provide a PR which only introduces pattern matching, and leaves the strings untouched?

triceo:
Also @timtebeek please see

    Error:  Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.23.0:validate (default) on project timefold-solver-core-impl: File '/home/runner/work/timefold-solver/timefold-solver/core/core-impl/src/main/java/ai/timefold/solver/core/impl/io/jaxb/GenericJaxbIO.java' has not been previously formatted. Please format file (for example by invoking `mvn -f core/core-impl net.revelc.code.formatter:formatter-maven-plugin:2.23.0:format`) and commit before running validation! -> [Help 1]

The error message says what to do about this, but a simple `mvn clean verify` will suffice.

triceo (approved):
Thank you!
I'll merge when CI is green.

timtebeek:
Hi @triceo ! Thanks for the quick & honest review; I agree with you that the text block and formatted changes included before can be a stylistic choice; I've updated the PR to now only apply the instanceof pattern matching changes.

These changes are almost completely automated, and as such quite easy to make. The only manual work then are customizations such as the requested `class1` change and running the tests, and some git wrangling to target the same branch as before.

timtebeek:
I understand you usually apply these changes as you go; personally I like to apply these changes as a quick one off run, such that it's easier to review, and there's less distractions in the code base when working on features. To each their own of course, but if you'd like some more automated changes there's another quick few that are easy to do, based on [Common static analysis issues](https://docs.openrewrite.org/recipes/staticanalysis/commonstaticanalysis).

```
    151 "ExplicitInitialization"
     56 "UnnecessaryParentheses"
     55 "NeedBraces"
     48 "FinalClass"
     28 "EqualsAvoidsNull"
     26 "UseDiamondOperator"
     26 "RenameLocalVariablesToCamelCase"
     24 "RemoveExtraSemicolons"
     22 "ModifierOrder"
     19 "ReplaceLambdaWithMethodReference"
      8 "IsEmptyCallOnCollections"
      5 "InlineVariable"
      4 "SimplifyBooleanReturn"
      3 "UnnecessaryExplicitTypeArguments"
      3 "RenamePrivateFieldsToCamelCase"
      3 "MinimumSwitchCases"
      2 "MultipleVariableDeclarations"
      2 "MethodNameCasing"
      2 "LambdaBlockToExpression"
      1 "WriteOctalValuesAsDecimal"
      1 "UpperCaseLiteralSuffixes"
```
The above table how often each recipe would have made code changes to timesolver; the link above has more detail on each individual change. Some of these are again stylistic choices that I don't want to force upon you, but others like using the diamond operator are easy to fix and nice to get in.

Feel free to tell me which you'd like to apply, or run a selected subset yourself through https://app.moderne.io/recipes/org.openrewrite.staticanalysis.CommonStaticAnalysis

triceo:
Interesting. We already use OpenRewrite in our `migration` module, but I'll look into using it for this kind of thing. Thanks for the pointers!

PR #242, opened 2023-08-29
docs: refactor docs for a cleaner ToC

triceo:
I compiled and checked both the PDF and the Antora website.
Content looks good.
I didn't see any new warnings while doing so.


rsynek (approved):
Nice improvement!

I am thinking whether we should go a bit further (not necessarily in this PR):
- split "Configuring Timefold Solver" into "Modelling an optimization problem" and "Running and configuring Timefold Solver"
- The former would focus on domain design, from simpler concepts to more advanced and would end with "Design patterns"
- The latter would cover the Solver and SolverManager API and configs.

triceo:
The split probably makes sense. This chapter - and the chapter on optimization algorithms - are far too long already.

PR #246, opened 2023-09-01
fix: avoid fail-fast to fix SolutionManager.update() on chained var

triceo:
There is a fail fast to detect that an inverse var points to the incorrect inverse.
But update() is used to fix those cases, so the fail-fast must not trigger there.


rsynek (approved):
Thanks for the fix.

PR #257, opened 2023-09-06
fix: avoid metaspace leak by no longer using the LambdaMetaFactory

triceo:
The leak came from the fact that once the factory creates the lambda, it becomes pinned to a classloader and there is no way to get rid of it, unless we get rid of the entire classloader. This is effectively a metaspace leak and people who create many solvers within the same JVM run into an OOP relatively quickly.

The JVM guys are aware of this issue and have recently changed their position on it. They argue it is not a bug, and that it only exists because we are using an API we shouldn't have been using. According to them, when used in the proper context, this issue doesn't happen. The only relevant solution for us, however, is to remove the use of the API altogether.

(For more details on the Java issue, see https://bugs.openjdk.org/browse/JDK-8302154.)

The biggest performance drop as a result of this change was seen in Cloud Balancing incremental score calculator. (-6 %.) This is not surprising, because with its vastly superior speed, it is the most exposed to these low-level calls. On the other hand, CS shows overall positive results (+2 % median), with two outliers worsening by 4 %.

These impacts would likely be less visible in a real solver run as opposed to a micro-benchmark focused on this. I think the trade-off is worth it.


core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
```diff
@@ -15,15 +16,27 @@ public final class ReflectionBeanPropertyMemberAccessor extends AbstractMemberAc
...
         this(getterMethod, false);
     }
 
     public ReflectionBeanPropertyMemberAccessor(Method getterMethod, boolean getterOnly) {
         this.getterMethod = getterMethod;
-        getterMethod.setAccessible(true); // Performance hack by avoiding security checks
+        MethodHandles.Lookup lookup = MethodHandles.lookup();
+        try {
+            getterMethod.setAccessible(true); // Performance hack by avoiding security checks
+            this.getherMethodHandle = lookup.unreflect(getterMethod)
```
rsynek, line 32: There is also `unreflectGetter` and `unreflectSetter`, but these operate on `Field` instances, while here we already have the method.

rsynek (approved):
Could you please explain where the memory leak this PR fixes was?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
triceo: Exactly, which is why I can't use those specialized methods here. There is a member accessor to deal specifically with fields.

PR #262, opened 2023-09-10
fix: impact assertion in constraint verifier

RobJohansen:
# Issue
Constraint Verifier gives unexpected result when verifying a constraint terminating with impact() (i.e. MIXED score type).

# Resolution
- Invert PENALTY and REWARD branches of switches.
- Clarify the test case data naming.

### DIFF
--- a/test/src/main/java/ai/timefold/solver/test/impl/score/stream/DefaultSingleConstraintAssertion.java
+++ b/test/src/main/java/ai/timefold/solver/test/impl/score/stream/DefaultSingleConstraintAssertion.java
@@ -113,13 +113,13 @@ public final class DefaultSingleConstraintAssertion<Solution_, Score_ extends Sc
         if (actualScoreImpactType == ScoreImpactType.MIXED) {
             // Impact means we need to check for expected impact type and actual impact match.
             switch (scoreImpactType) {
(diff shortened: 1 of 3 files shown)


triceo:
Thanks for the contribution @RobJohansen!

I'll have to investigate this carefully, as I have a suspicion that this breaks backwards compatibility for users of `impact(...)`. Even if the old behavior is a bug, people have been relying on it to behave the way it has - stay tuned.

RobJohansen:
@triceo no problem, been working with optaplanner for years now and decided to start making some contributions 👍 

FYI, it doesn't affect the behaviour of impact, just the verification, but I could see how that would still confuse if changed.

triceo:
@RobJohansen We're happy to have you. :-)

In this case, I really don't think we can merge it as is. It would break people's tests, and that's not something we can afford to do in a minor version.

The thing we can do in a minor version is deprecate everything `impact(...)` in the CS API and create a new overload, one which doesn't suffer from this. This is the clean solution to the problem, and probably something I want to do, but I can't currently promise when.

triceo:
As explained above, we can not merge this due to backwards incompatibility.
We're still open to accepting another fix to the problem as proposed.

PR #280, opened 2023-09-20
build: remove Javadoc workaround

triceo:


Christopher-Chianelli (approved):
`mvn clean install -Dfull` passes locally, although some JavaDoc warnings did appear regarding missing descriptions.

triceo:
I haven't gone over the warnings in detail, but there were plenty of warnings before and the build no longer fails, so I consider that an improvement. Thanks for the review!

PR #297, opened 2023-09-29
feat: Add concat operation to constraint streams

Christopher-Chianelli:
The concat operation acts as an or between two constraint streams, propogating both stream tuples' downstream. It appears update/retract may be called before an insert, meaning the out tuple stored inside the in tuple is not set. As such, for update, if the out tuple is not set, it does an insert instead, and for retract, it simply ignores the operation.


core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/BavetConstraintSessionFactory.java
```diff
@@ -140,6 +141,13 @@ private long determineLayerIndex(AbstractNode node, NodeBuildHelper<Score_> buil
             var leftParentNode = buildHelper.findParentNode(leftParent);
             var rightParentNode = buildHelper.findParentNode(rightParent);
             return Math.max(leftParentNode.getLayerIndex(), rightParentNode.getLayerIndex()) + 1;
+        } else if (node instanceof AbstractConcatNode<?> concatNode) {
+            var nodeCreator = (BavetJoinConstraintStream<?>) buildHelper.getNodeCreatingStream(concatNode);
```
triceo, line 145: Why would concat implement join? The two operations have barely anything in common.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetAbstractBiConstraintStream.java
```diff
@@ -359,6 +359,19 @@ public BiConstraintStream<A, B> distinct() {
...
     }
 
+    @Override
+    public BiConstraintStream<A, B> concat(BiConstraintStream<A, B> otherStream) {
+        var other = (BavetAbstractBiConstraintStream<Solution_, A, B>) otherStream;
+        var leftBridge = new BavetForeBridgeBiConstraintStream<>(constraintFactory, this);
+        var rightBridge = new BavetForeBridgeBiConstraintStream<>(constraintFactory, other);
+        var concatStream = new BavetConcatBiConstraintStream<>(constraintFactory, leftBridge, rightBridge);
+        return constraintFactory.share(concatStream, concatStream_ -> {
+            // Connect the bridges upstream, as it is an actual new join.
```
triceo, line 369: This is not a join; let's not call it like that.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetBiConcatNode.java
```diff
@@ -0,0 +1,19 @@
...
+
+    BavetBiConcatNode(TupleLifecycle<BiTuple<A, B>> nextNodesTupleLifecycle, int inputStoreIndexLeftOutTupleList,
+            int inputStoreIndexRightOutTupleList,
+            int outputStoreSize) {
+        super(nextNodesTupleLifecycle, inputStoreIndexLeftOutTupleList, inputStoreIndexRightOutTupleList, outputStoreSize);
+    }
+
+    @Override
+    protected BiTuple<A, B> getOutTuple(BiTuple<A, B> inTuple) {
+        return new BiTuple<>(inTuple.factA, inTuple.factB, outputStoreSize);
```
triceo, line 17: Since concat does not modify the tuples, it doesn't need to create new ones.
I'm betting it will be much more efficient if it's rewritten as a pass-through node, without node creation.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/AbstractConcatNode.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    public final void retractLeft(Tuple_ tuple) {
+        Tuple_ outTuple = tuple.getStore(inputStoreIndexLeftOutTupleList);
+        if (outTuple == null) {
+            return;
+        }
+
+        TupleState state = outTuple.state;
+        if (!state.isActive()) {
+            // Impossible because they shouldn't linger in the indexes.
```
triceo, line 59: There aren't any indexes.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/AbstractConcatNode.java
```diff
@@ -0,0 +1,103 @@
...
+import ai.timefold.solver.constraint.streams.bavet.common.tuple.LeftTupleLifecycle;
+import ai.timefold.solver.constraint.streams.bavet.common.tuple.RightTupleLifecycle;
+import ai.timefold.solver.constraint.streams.bavet.common.tuple.TupleLifecycle;
+import ai.timefold.solver.constraint.streams.bavet.common.tuple.TupleState;
+
+public abstract class AbstractConcatNode<Tuple_ extends AbstractTuple>
+        extends AbstractNode
+        implements LeftTupleLifecycle<Tuple_>, RightTupleLifecycle<Tuple_> {
+    private final int inputStoreIndexLeftOutTupleList;
+    private final int inputStoreIndexRightOutTupleList;
```
triceo, line 17: These vars need better naming, as there are no indexes and no out tuple list.

triceo (commented):
Nice! 
I think we can still make the node itself more efficient.

Also, some minor comments. (Please make sure to apply them consistently in all cardinalities of streams.)

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/AbstractConcatNode.java
```diff
@@ -0,0 +1,103 @@
...
+        this.inputStoreIndexRightOutTupleList = inputStoreIndexRightOutTupleList;
+        this.outputStoreSize = outputStoreSize;
+    }
+
+    protected abstract Tuple_ getOutTuple(Tuple_ inTuple);
+
+    @Override
+    public final void insertLeft(Tuple_ tuple) {
+        Tuple_ outTuple = getOutTuple(tuple);
+        tuple.setStore(inputStoreIndexLeftOutTupleList, outTuple);
```
triceo, line 36: If you store the node-specific tuple state in the store instead of the outtuple, you won't need to create the outtuple just to store the state.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/BavetConstraintSessionFactory.java
Christopher-Chianelli: Largely for a fact that like `join`, `concat` has two parents, and the only two methods on `BavetJoinConstraintStream` are getters for the two parents. IMO, it would be pointless for two classes to have exactly the same methods, although those two methods could be extracted into a super interface `BavetBinaryOperatorResultStream` and have `BavetJoinConstraintStream` extend that (essentially making `BavetJoinConstraintStream` a marker interface).

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetBiConcatNode.java
Christopher-Chianelli: Tried that, the queue did not like it since it sees the inTuple was already inserted.

Christopher-Chianelli: Turn out we do need new tuples, since each node can only have 1 tuple source, and since concat has two parents, it must be a tuple source, and a tuple source must produce new tuples, not pass through existing ones.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/uni/BavetConcatUniConstraintStream.java
```diff
@@ -0,0 +1,104 @@
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        BavetConcatUniConstraintStream<?, ?> other = (BavetConcatUniConstraintStream<?, ?>) o;
+        /*
+         * Bridge streams do not implement equality because their equals() would have to point back to this stream,
+         * resulting in StackOverflowError.
+         * Therefore we need to check bridge parents to see where this join node comes from.
```
triceo, line 74: Not a join node.
(Same comment also likely in Bi, Tri and Quad.)

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/BavetConstraintSessionFactory.java
```diff
@@ -140,6 +142,13 @@ private long determineLayerIndex(AbstractNode node, NodeBuildHelper<Score_> buil
             var leftParentNode = buildHelper.findParentNode(leftParent);
             var rightParentNode = buildHelper.findParentNode(rightParent);
             return Math.max(leftParentNode.getLayerIndex(), rightParentNode.getLayerIndex()) + 1;
+        } else if (node instanceof AbstractConcatNode<?> concatNode) {
+            var nodeCreator = (BavetConcatConstraintStream<?>) buildHelper.getNodeCreatingStream(concatNode);
+            var leftParent = nodeCreator.getLeftParent();
+            var rightParent = nodeCreator.getRightParent();
+            var leftParentNode = buildHelper.findParentNode(leftParent);
+            var rightParentNode = buildHelper.findParentNode(rightParent);
+            return Math.max(leftParentNode.getLayerIndex(), rightParentNode.getLayerIndex()) + 1;
```
triceo, line 151: This code seems to repeat a lot now. Maybe we extract it?

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,47 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
 See <<constraintStreamsDealingWithDuplicateTuplesUsingDistinct,`distinct()`>> for how to deal with duplicate tuples.
 ====
 
+[[constraintStreamsConcat]]
+==== Concat
+
+Concat allows you to create a constraint stream containing tuples of two constraint streams of the same <<constraintStreamsCardinality,cardinality>>.
+If <<constraintStreamsJoin,join>> acts like a cartesian product of two lists, concat acts like a concatenation of two lists.
+Unlike union of sets, concatenation of lists repeat duplicated elements.
```
triceo, line 1519:
```suggestion
Unlike union of sets, concatenation of lists repeats duplicated elements.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,47 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+If the two constraint stream parents share tuples, they will be repeated downstream.
+If this is undesired, use <<constraintStreamsDealingWithDuplicateTuplesUsingDistinct,distinct>>.
```
triceo, line 1521:
```suggestion
If this is undesired, use the <<constraintStreamsDealingWithDuplicateTuplesUsingDistinct,`distinct` building block>>.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,47 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+
+For example, to count the total number of cats and dogs:
+
+[source,java,options="nowrap"]
+----
+    private Constraint countCatsAndDogs(ConstraintFactory constraintFactory) {
+        return constraintFactory.forEach(Employee.class)
+         .filter(employee -> employee.hasDog())
+         .concat(constraintFactory.forEach(Employee.class)
+                     .filter(employee -> employee.hasCat()))
```
triceo, line 1533: I am not sure this is a good example. The first question once a smart person sees this is: why not just use a much simple `.filter(hasCat || hasDog)`?

I think we need an example which doesn't have an obvious simpler faster solution.

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,47 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+         .groupBy(ConstraintCollectors.count())
+         .penalize(HardSoftScore.ONE_SOFT, count -> count)
+         .asConstraint("Count cats and dogs");
+    }
+----
+
+This correctly count cats and dogs when an Employee has both a cat and a dog.
```
triceo, line 1540:
```suggestion
This correctly counts cats and dogs when an Employee has both a cat and a dog.
```

triceo, line 1540:
```suggestion
This correctly counts cats and dogs when an Employee has both a cat and a dog.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
Christopher-Chianelli: The obvious example would be to implement group by on an outer join, like this:

```java
factory.forEach(A.class)
    .join(B.class, ...)
    .concat(factory.forEach(A.class)
                               .ifNotExists(B.class, ...)
                               .expand(a -> null))
    .groupBy((a,b) -> a, conditionally((a,b) -> b != null, countBi())
    ...
```

Although the example would lead to the question "why no outer join" since that would simplify the code too

```java
factory.forEach(A.class)
    .outerJoin(B.class, ...)
    .groupBy((a,b) -> a, conditionally((a,b) -> b != null, countBi())
    ...
```

The other thing it can be used for is or'ing Joiners, which is unrelated to outer joins

```java
factory.forEach(A.class)
    .join(B.class, equal(a -> a.value, b -> b.value))
    .concat(factory.forEach(A.class)
                               .join(B.class, equal(a -> a.group, b -> b.group)))
    ...
```

Christopher-Chianelli:
TODO: change text in image to use `UNION ALL` instead of `OR` (https://www.techonthenet.com/sql/union_all.php)

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
triceo: If we're doing concat for the minimal constraint, why not show that? Maybe there's a simplified variant that isn't too much code while still showing the point?

I'm not worried if people start asking about outer join after they see this. First of all, we have a decent answer, this is faster because it doesn't need to do any sort of indexing. Second, if they do, and if there's enough of them, then maybe we eventually do it? It's an opportunity to coax some more real-world use cases out of people, and that's a good thing.

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,57 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
 See <<constraintStreamsDealingWithDuplicateTuplesUsingDistinct,`distinct()`>> for how to deal with duplicate tuples.
 ====
 
+[[constraintStreamsConcat]]
+==== Concat
+
+Concat allows you to create a constraint stream containing tuples of two constraint streams of the same <<constraintStreamsCardinality,cardinality>>.
```
triceo, line 1517:
```suggestion
The `concat` building block allows you to create a constraint stream containing tuples of two constraint streams of the same <<constraintStreamsCardinality,cardinality>>.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,57 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+If <<constraintStreamsJoin,join>> acts like a cartesian product of two lists, concat acts like a concatenation of two lists.
```
triceo, line 1518:
```suggestion
If <<constraintStreamsJoin,join>> acts like a cartesian product of two lists, `concat` acts like a concatenation of two lists.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,57 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+Unlike union of sets, concatenation of lists repeats duplicated elements.
+If the two constraint stream parents share tuples, they will be repeated downstream.
```
triceo, line 1520:
```suggestion
If the two constraint stream parents share tuples, which happens eg. when the streams being concatenated come from the same source of data, the tuples will be repeated downstream.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,57 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+                   conditionally((employee, shift) -> shift != null,
+                                 countBi())
+         )
+         .filter((employee, shiftCount) -> shiftCount < employee.minimumAssignedShifts)
+         .penalize(HardSoftScore.ONE_SOFT, (employee, shiftCount) -> employee.minimumAssignedShifts - shiftCount)
+         .asConstraint("Minimum number of assigned shifts");
+    }
+----
+
+This correctly count the number of shifts each Employee has, *even when the Employee has no shifts*.
```
triceo, line 1547:
```suggestion
This correctly counts the number of shifts each Employee has, *even when the Employee has no shifts*.
```

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1511,6 +1511,57 @@ the tuple `(SomePerson, USER)` is sent downstream twice.
...
+If it was implemented without concat like this:
```
triceo, line 1548:
```suggestion
If it was implemented without `concat` like this:
```

triceo (approved):
I added the final set of comments as I re-read the docs.
I'll run my benchmarks overnight to see how it behaves, and then we can merge.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetBiConcatNode.java
```diff
@@ -16,4 +16,10 @@ public final class BavetBiConcatNode<A, B> extends AbstractConcatNode<BiTuple<A,
     protected BiTuple<A, B> getOutTuple(BiTuple<A, B> inTuple) {
         return new BiTuple<>(inTuple.factA, inTuple.factB, outputStoreSize);
     }
+
+    @Override
+    protected void updateOutTuple(BiTuple<A, B> inTuple, BiTuple<A, B> outTuple) {
+        outTuple.factA = inTuple.factA;
```
triceo, line 22: Tuples have `updateInDifferent(...)` just for this case.
Turns out it is a common operation, so it was extracted already.

PR #308, opened 2023-10-08
feat: improve support for Java records

triceo:
Does several things:

1. Records can serve as planning variables; planning ID annotation on record components no longer throws.
2. Records can no longer be solutions or entities, as these need to be mutable.
3. Records are no longer deep-cloned, receiving the same treatment as enums.
4. Throws a better exception when cloning fails due to an exception in constructor.

The second and third change are technically backwards incompatible behavior but we introduce it anyway, as the old behavior was erroneous.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
@@ -458,6 +458,16 @@ public class ConfigUtils {
         }
     }
 
+    /**
+     * This method is heavy, and it is effectively a computed constant.
+     * It is recommended that its results are cached at call sites.
+     *
+     * @param clazz never null
+     * @param memberAccessorFactory never null
+     * @param domainAccessType never null
+     * @return null if no accessor found
+     * @param <C> the class type
+     */
     public static <C> MemberAccessor findPlanningIdMemberAccessor(Class<C> clazz,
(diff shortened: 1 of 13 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -487,10 +487,28 @@ private static <C> Member getSingleMember(Class<C> clazz, Class<? extends Annota
...
-                    + annotationClass.getSimpleName() + " annotation.");
+        int size = memberList.size();
+        if (clazz.isRecord()) {
+            /*
+             * A record has a field and a getter for each record component.
+             * When the component is annotated with @PlanningId,
+             * the annotation ends up both on the field and on the getter.
+             */
+            if (size == 2) { // The getter is used to retrieve the value of the record component.
+                return memberList.get(1);
```
rsynek, line 498: Is it guaranteed that the second member in the list is the getter and not the field?

rsynek (approved):
Nice fix, thanks!

core/core-impl/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
triceo: I don't think there is such guarantee. In fact, I'd expect that this is totally up to the discretion of every JDK implementation. Maybe I can figure out which is which from the type...

PR #331, opened 2023-10-12
fix: Always propagate map's update even if mapped value does not change

Christopher-Chianelli:
If map maps to a planning entity, then update MUST always be propagated, since variables on the planning entity might of changed.


triceo (approved):

triceo:
Replaced with https://github.com/TimefoldAI/timefold-solver/pull/335

PR #337, opened 2023-10-13
docs: Document planning value type requirements

mcimbora:


docs/src/modules/ROOT/pages/configuration/configuration.adoc
```diff
@@ -1030,9 +1030,14 @@ public class StartTimeUpdatingVariableListener implements VariableListener<TaskA
 ==== Planning value
 
 A planning value is a possible value for a genuine planning variable.
-Usually, a planning value is a problem fact, but it can also be any object, for example a ``double``.
+Usually, a planning value is a problem fact, but it can also be any object, for example a ``Double``.
```
triceo, line 1033:
```suggestion
Usually, a planning value is a problem fact, but it can also be any object, for example an ``Integer``.
```

docs/src/modules/ROOT/pages/configuration/configuration.adoc
```diff
@@ -1030,9 +1030,14 @@ public class StartTimeUpdatingVariableListener implements VariableListener<TaskA
...
 It can even be another planning entity or even an interface implemented by both a planning entity and a problem fact.
 
+[NOTE]
+====
+Primitive types (such as ``double``) are not allowed.
```
triceo, line 1038:
```suggestion
Primitive types (such as ``int``) are not allowed.
```

triceo (approved):
I'm making some suggestions to move away from mentioning any kind of floating-point numbers.

PR #340, opened 2023-10-16
feat: introduce CS concat() overloads for different cardinalities

triceo:
Squash before merging.


core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/common/inliner/ConstraintMatchSupplier.java
```diff
@@ -65,7 +65,7 @@ private static RuntimeException createJustificationException(Constraint constrai
 
     private static String factsToString(Object... facts) {
         return Arrays.stream(facts)
-                .map(Object::toString)
+                .map(s -> s == null ? null : s.toString())
```
Christopher-Chianelli, line 68: Maybe `Objects::toString`? https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-

docs/src/modules/ROOT/pages/constraints-and-score/constraints-and-score.adoc
```diff
@@ -1514,10 +1514,14 @@ See <<constraintStreamsDealingWithDuplicateTuplesUsingDistinct,`distinct()`>> fo
 [[constraintStreamsConcat]]
 ==== Concat
 
-The `concat` building block allows you to create a constraint stream containing tuples of two constraint streams of the same <<constraintStreamsCardinality,cardinality>>.
-If <<constraintStreamsJoin,join>> acts like a cartesian product of two lists, `concat` acts like a concatenation of two lists.
+The `concat` building block allows you to create a constraint stream containing tuples
+of two constraint streams of the same <<constraintStreamsCardinality,cardinality>>.
```
Christopher-Chianelli, line 1518: Inaccurate, since now multiple cardinalities are allowed

Christopher-Chianelli (approved):
Overall LGTM. Some minor comments. Nitpick: it not that the overloads are less garbage-intensive, but that they automatically expand lower cardinality streams, eliminating the need for expand (which generates the garbage).

PR #341, opened 2023-10-16
fix: Deep clone unknown classes in GIZMO solution cloner

### LINKED ISSUE #339: Gizmo cloning corruption with subclass
**Describe the bug**

I have a planning entity with a list of shadow variables. The list type is an abstract class with one concrete subclass. If I run my example without gizmo, it works as expected. However, with gizmo I get this exception:

```
Exception in thread "main" java.lang.IllegalStateException: Cloning corruption: the same entity (example.BinaryShadowEntity@89ff02e) is present in both the original and the clone.
So when a planning variable in the original solution changes, the cloned solution will change too.
Check the SolutionCloner.
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.score.director.AbstractScoreDirector.lambda$cloneSolution$2(AbstractScoreDirector.java:256)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor.lambda$visitAllEntities$9(SolutionDescriptor.java:863)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor.visitAllEntities(SolutionDescriptor.java:877)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor.visitAllEntities(SolutionDescriptor.java:863)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.score.director.AbstractScoreDirector.cloneSolution(AbstractScoreDirector.java:249)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.solver.scope.SolverScope.setWorkingSolutionFromBestSolution(SolverScope.java:242)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.solver.AbstractSolver.solvingStarted(AbstractSolver.java:64)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.solver.DefaultSolver.solvingStarted(DefaultSolver.java:225)
	at ai.timefold.solver.core.impl@1.2.0/ai.timefold.solver.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:192)
	at example/example.Main.main(Main.java:33)
```

**Expected behavior**
It works the same as without gizmo -- no exception and a solution is printed at the end, like this:

```
[INFO ] [2023-10-14 11:50:40,102] [main] [a.t.s.c.i.s.DefaultSolver] Solving ended: time spent (99), best score (0hard/-2soft), score calculation speed (1030/sec), phase total (2), environment mode (FULL_ASSERT), move thread count (NONE).
Explanation of score (0hard/-2soft):
    Constraint match totals:
        -1soft: constraint (Prefer assigned values) has 1 matches:
            -1soft: justifications ([example.MyPlanningEntity@5c48c0c0])
        -1soft: constraint (Prefer shadow entities whose planning entities are fully bound) has 1 matches:
            -1soft: justifications ([example.BinaryShadowEntity@674c583e, [example.MyPlanningEntity@1d9bec4d, example.MyPlanningEntity@5c48c0c0]])
    Indictments:
        -1soft: indicted object (example.MyPlanningEntity@5c48c0c0) has 1 matches:
            -1soft: constraint (Prefer assigned values)
        -1soft: indicted object (example.BinaryShadowEntity@674c583e) has 1 matches:
            -1soft: constraint (Prefer shadow entities whose planning entities are fully bound)
        -1soft: indicted object ([example.MyPlanningEntity@1d9bec4d, example.MyPlanningEntity@5c48c0c0]) has 1 matches:
            -1soft: constraint (Prefer shadow entities whose planning entities are fully bound)

Solution: 
example.MyPlanningVariable@3f2ef586
example.MyPlanningVariable@611df6e3
example.MyPlanningVariable@52500920
null
```

**Actual behavior**
I get the exception above.

**To Reproduce**

Here is an example demonstrating the issue:

[gizmoExample.zip](https://github.com/TimefoldAI/timefold-solver/files/12906628/gizmoExample.zip)

The key features that appear to cause the issue are the fact that `MyShadowEntity` is abstract and has a subclass `BinaryShadowEntity` and that the environment mode is `FULL_ASSERT` (it does not happen in the default environment mode). It may be that my model is in fact incorrect and that this is a correct exception. In that case, the bug would be that this exception isn't thrown when not using gizmo.

This is modeled after my real system, which I can't share. As such, it's possible it's more complex than necessary. E.g., does it really matter that this involves a shadow variable? I'm not sure. I can say that the exception still occurs even if there are no constraints.

## Environment

Timefold 1.2.0

openjdk version "17.0.8.1" 2023-08-24
OpenJDK Runtime Environment Temurin-17.0.8.1+1 (build 17.0.8.1+1)
OpenJDK 64-Bit Server VM Temurin-17.0.8.1+1 (build 17.0.8.1+1, mixed mode, sharing)

Microsoft Windows 11 Pro
10.0.22621 Build 22621

## Additional information

My actual project has a lot more "stuff" in the relevant classes (getters/setters, equals/hashcode), but that stuff doesn't seem to make any difference in this example so I stripped it all out.

Christopher-Chianelli:
If a deep cloned class that is not accessible from the SolutionDescriptor (for instance, a class that extends a @PlanningEntity but adds no new variables) is inside the solution, it would incorrectly be shallowly cloned. The GIZMO solution cloner will now fall back to the reflection solution cloner if it encounters an unknown class.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
@@ -62,6 +62,27 @@ public final class FieldAccessingSolutionCloner<Solution_> implements SolutionCl
         return cloneSolution;
     }
 
+    /**
+     * Used by GIZMO when it encounters an undeclared entity class, such as when an abstract planning entity is extended.
+     */
+    public Object gizmoFallbackDeepClone(Object originalValue, Map<Object, Object> originalToCloneMap) {
+        if (originalValue == null) {
+            return null;
+        }
+        Queue<Unprocessed> unprocessedQueue = new ArrayDeque<>();
+        Class<?> fieldType = originalValue.getClass();
+        if (originalValue instanceof Collection<?> collection) {
+            return cloneCollection(fieldType, collection, originalToCloneMap, unprocessedQueue);
+        } else if (originalValue instanceof Map<?, ?> map) {
+            return cloneMap(fieldType, map, originalToCloneMap, unprocessedQueue);
+        } else if (originalValue.getClass().isArray()) {
+            return cloneArray(fieldType, originalValue, originalToCloneMap, unprocessedQueue);
+        } else {
+            return clone(originalValue, originalToCloneMap, unprocessedQueue,
+                    retrieveClassMetadata(originalValue.getClass()));
+        }
+    }
+
     private Object process(Unprocessed unprocessed, Map<Object, Object> originalToCloneMap,
             Queue<Unprocessed> unprocessedQueue) {
         Object originalValue = unprocessed.originalValue;
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionCloner.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionCloner.java
@@ -1,0 +1,8 @@
+package ai.timefold.solver.core.impl.domain.solution.cloner.gizmo;
+
+import ai.timefold.solver.core.api.domain.solution.cloner.SolutionCloner;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+public interface GizmoSolutionCloner<Solution_> extends SolutionCloner<Solution_> {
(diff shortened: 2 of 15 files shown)


triceo (approved):
LGTM, assuming native checks pass.

PR #346, opened 2023-10-17
perf: GIZMO should not check for unknown classes in Quarkus

Christopher-Chianelli:
In Quarkus, we know all subclasses, so we cannot encounter an unknown class.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
@@ -28,7 +28,8 @@ public final class GizmoSolutionClonerFactory {
                     ") the classpath or modulepath must contain io.quarkus.gizmo:gizmo.\n" +
                     "Maybe add a dependency to io.quarkus.gizmo:gizmo.");
         }
-        return GizmoSolutionClonerImplementor.createClonerFor(solutionDescriptor, gizmoClassLoader);
+        return new GizmoSolutionClonerImplementor().createClonerFor(solutionDescriptor,
+                gizmoClassLoader);
     }
 
     // ************************************************************************
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
@@ -52,7 +52,10 @@ public class GizmoSolutionClonerImplementor {
             Object.class);
     private static final MethodDescriptor PUT_METHOD = MethodDescriptor.ofMethod(Map.class, "put", Object.class,
             Object.class, Object.class);
+    private static final String FALLBACK_CLONER = "fallbackCloner";
     public static final boolean DEBUG = false;
+
+    public static boolean IS_QUARKUS = false;
 
     /**
      * Return a comparator that sorts classes into instanceof check order.
@@ -125,8 +128,10 @@ public class GizmoSolutionClonerImplementor {
         SortedSet<Class<?>> deepCloneClassesThatAreNotSolutionSortedSet = new TreeSet<>(instanceOfComparator);
         deepCloneClassesThatAreNotSolutionSortedSet.addAll(deepCloneClassesThatAreNotSolutionSet);
 
-        classCreator.getFieldCreator("fallbackCloner", FieldAccessingSolutionCloner.class)
-                .setModifiers(Modifier.PRIVATE | Modifier.STATIC);
+        if (!IS_QUARKUS) {
+            classCreator.getFieldCreator(FALLBACK_CLONER, FieldAccessingSolutionCloner.class)
+                    .setModifiers(Modifier.PRIVATE | Modifier.STATIC);
+        }
         createConstructor(classCreator);
         createSetSolutionDescriptor(classCreator, solutionDescriptor);
         createCloneSolution(classCreator, solutionDescriptor);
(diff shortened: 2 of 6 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -52,8 +52,11 @@ public class GizmoSolutionClonerImplementor {
             Object.class);
     private static final MethodDescriptor PUT_METHOD = MethodDescriptor.ofMethod(Map.class, "put", Object.class,
             Object.class, Object.class);
+    private static final String FALLBACK_CLONER = "fallbackCloner";
     public static final boolean DEBUG = false;
 
+    public static boolean IS_QUARKUS = false;
```
triceo, line 58: Let's figure out a different way to pass this information. Static mutable fields are fundamentally wrong, and we've seen it numerous times already.

Do we need a Quarkus-specific subtype? So be it.
Quarkus-specific constructor? Sure.
Just not this, please.

In my opinion, the fact that there are ifs for particular situations screams for two distinct implementations with a shared supertype, or the lambda/strategy pattern.

triceo (changes requested):
Thanks for the adjustment. Please see the comment and let's figure out a way to do the switch better.
Also, CI is failing.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
```diff
@@ -28,7 +28,8 @@ public static <T> SolutionCloner<T> build(SolutionDescriptor<T> solutionDescript
                     ") the classpath or modulepath must contain io.quarkus.gizmo:gizmo.\n" +
                     "Maybe add a dependency to io.quarkus.gizmo:gizmo.");
         }
-        return GizmoSolutionClonerImplementor.createClonerFor(solutionDescriptor, gizmoClassLoader);
+        return new GizmoSolutionClonerImplementor().createClonerFor(solutionDescriptor,
```
triceo, line 31: Why not:

```suggestion
        return GizmoSolutionClonerImplementor.createClonerFor(solutionDescriptor,
```

Seems like the instance creation is just boilerplate, it can happen internally. People need not know what's happening under the covers.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -100,14 +101,20 @@ public static Comparator<Class<?>> getInstanceOfComparator(Set<Class<?>> deepClo
                 .thenComparing(Class::getName).reversed();
     }
 
+    protected void createFields(ClassCreator classCreator) {
+        classCreator.getFieldCreator(FALLBACK_CLONER, FieldAccessingSolutionCloner.class)
+                .setModifiers(Modifier.PRIVATE | Modifier.STATIC);
+    }
```
triceo, line 107: Could have stayed static.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -417,6 +419,76 @@ public String generateGizmoBeanFactory(ClassOutput classOutput, Set<Class<?>> be
         return generatedClassName;
     }
 
+    private static class QuarkusGizmoSolutionClonerImplementor extends GizmoSolutionClonerImplementor {
```
triceo, line 422: The class is IMO a big too big to be an inner class. Consider making it a top-level class; it's already static, so you don't need the reference to the outer class.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerTest.java
```diff
@@ -69,7 +69,7 @@ protected <Solution_> SolutionCloner<Solution_> createSolutionCloner(SolutionDes
                             generateGizmoSolutionOrEntityDescriptor(solutionDescriptor, clazz));
                 });
 
-        GizmoSolutionClonerImplementor.defineClonerFor(classCreator, solutionDescriptor,
+        new GizmoSolutionClonerImplementor().defineClonerFor(classCreator, solutionDescriptor,
```
triceo, line 72: Same comment as above; I think it could've been made static and hidden the instance.

triceo (approved):
Some of these newly instance methods might have been kept static, I think - let's do that.  
Also some more smaller comments.

Once resolved, good to merge.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: No, this method is overridden in Quarkus to do nothing, which make it a mandatory to be virtual.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerTest.java
Christopher-Chianelli: Issue is, Quarkus need to pass its own subclass as an instance instead of using the default one; I prefer `new Subclass().defineClonerFor(args)` to `Implementor.defineClonerFor(new Subclass(), args)`

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerFactory.java
Christopher-Chianelli: See comment below

core/core-impl/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerTest.java
triceo: > Issue is, Quarkus need to pass its own subclass as an instance instead of using the default one; I prefer `new Subclass().defineClonerFor(args)` to `Implementor.defineClonerFor(new Subclass(), args)`

In the example you give, I can see how that is a question of personal preference.
But on line 72, where I put my comment, you're creating the instance without assigning it to any variable; why have the user create the instance then? It just looks odd.

PR #369, opened 2023-10-27
perf: Implement node sharing for ifExists, expand, map, groupby

Christopher-Chianelli:
- Created a class ConstantLambdaUtils to store common lambdas used in CS methods so we are guaratee to get the same instance and not depending on JDK specific behaviour.

- DefaultConstraintCollector now supports hashCode and equality. It takes an ConstraintCollectors.ConstraintCollectorKind  and Object[] equalityArgs in its constructor. If the kind and equalityArgs in two different DefaultConstraintCollectors are equal, the two DefaultConstraintCollectors are considered equal.

- groupBy checks reference equality of keyMappers and object equality of constraint collectors.

- ifExists and ifNotExists use the same logic as join

- expand and map check reference equality of passed functions.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
@@ -1,0 +1,59 @@
+package ai.timefold.solver.core.impl.util;
+
+import java.util.Objects;
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.ToIntBiFunction;
+import java.util.function.ToIntFunction;
+
+import ai.timefold.solver.core.api.function.QuadFunction;
+import ai.timefold.solver.core.api.function.ToIntQuadFunction;
+import ai.timefold.solver.core.api.function.ToIntTriFunction;
+import ai.timefold.solver.core.api.function.TriFunction;
+
+public final class ConstantLambdaUtils {
+    @SuppressWarnings("rawtypes")
+    public static final Function IDENTITY = Function.identity();
+    @SuppressWarnings("rawtypes")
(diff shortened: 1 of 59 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.function.Function;
+import java.util.function.ToIntBiFunction;
+import java.util.function.ToIntFunction;
+
+import ai.timefold.solver.core.api.function.QuadFunction;
+import ai.timefold.solver.core.api.function.ToIntQuadFunction;
+import ai.timefold.solver.core.api.function.ToIntTriFunction;
+import ai.timefold.solver.core.api.function.TriFunction;
+
+public final class ConstantLambdaUtils {
```
triceo, line 15: If these functions are returned via static methods, you can get the casting done inside the methods, as opposed to plaguing the entire codebase with it.

Example: 

    public static final <A, B> BiPredicate<A, B> notEquals() {
        return (BiPredicate<A, B>) NOT_EQUALS; 
    }

triceo: Also, let's add a Javadoc to the class, explaining why we need to do something like this. Every time there is such non-standard behavior, there should be an explanation as to why it needs to be like that.

triceo (changes requested):
Chris, I'm going to ask you to do things a bit differently this time. Let's do this bit by bit.

Let's start with the constraint lambda utils changes for now, and move on to the others in separate PRs when we're done with this one. Build it up gradually.

This is an invasive change and I want to make sure nothing gets lost just because the PR is large.

Christopher-Chianelli:
Superseded by #374

PR #371, opened 2023-10-27
fix: invalidate computed variables in Project Job Scheduling correctly

triceo:
Due to incorrectly invalidated busy dates for allocations, no new best solutions were found, and therefore the screen was never refreshed.


mcimbora (approved):
Thanks for the quick fix.

My gut feeling - this is one of our most complex examples. This caching makes it even more complex. My first impression was - there must be a score corruption somewhere :-)
Anyway, I'd rather accept a bit lower performance with the benefit of having an easy-to-understand example. Your call. I think this example needs some lifting anyway (quickstart even?).

triceo:
Thanks for the review. On the whole, I agree - this example is complex, and it is a use case which we should be able to show off better. 

Taking into account the priorities, I do not think it will happen any time soon.
Examples are in maintenance mode, and have been for a very long time now.

PR #374, opened 2023-10-27
perf: Use lambda constants in CS to allow for more node sharing

Christopher-Chianelli:


Christopher-Chianelli:
This contains only the lambda constants; I will be breaking the original PR into four parts:

- Lambda Constants + initial node sharing tests (this)
- Node Share IfExists/IfNotExists
- Node Share Map/Expand/FlattenLast
- Node Share GroupBy

core/constraint-streams/src/test/java/ai/timefold/solver/constraint/streams/common/ConstraintStreamFunctionalTest.java
```diff
@@ -252,4 +252,12 @@ default void expandToQuad() {
 
     void impactNegativeBigDecimalCustomJustifications();
 
+    // ************************************************************************
```
triceo, line 255: Let's start a new `NodeSharingTest` for this.

core/constraint-streams/src/test/java/ai/timefold/solver/constraint/streams/common/bi/AbstractBiConstraintStreamTest.java
```diff
@@ -3000,4 +3008,190 @@ public void joinerEqualsAndSameness() {
...
+                            factory.forEach(TestdataEntity.class).join(TestdataEntity.class);
+                    Function<TestdataEntity, TestdataEntity> function = a -> a;
+                    BiFunction<TestdataEntity, TestdataEntity, TestdataEntity> bifunction = (a, b) -> a;
+                    Function<TestdataEntity, Iterable<String>> iterableFunction = a -> Collections.emptyList();
+                    BiPredicate<TestdataEntity, TestdataEntity> bipredicate = (a, b) -> true;
+                    TriPredicate<TestdataEntity, TestdataEntity, TestdataEntity> tripredicate = (a, b, c) -> true;
+
+                    if (base.join(TestdataEntity.class) != base.join(TestdataEntity.class)) {
+                        failureList.add("join(no-joiners)");
+                    }
```
triceo, line 3032: This is needlessly complicated. Since you never need to call those constraint streams, you don't need `buildScoreDirector` or `failureList`.

Let's change the test to look like this:

    Assertions.assertThat(base.join(...))
        .as(...)
        .isSameAs(base.join(...)

Maybe you'll need to figure out how to get a ConstraintFactory instance, but it will be better than using this `failureList` hack. It is also part of my reasoning for having this in a new test class, because this will require new approaches and generally be unlike the other tests in the current class.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
```diff
@@ -0,0 +1,216 @@
...
+import ai.timefold.solver.core.api.function.TriFunction;
+
+/**
+ * A class that hold common lambdas that are guarantee to be the same across method calls. In most JDK's, stateless lambdas are
+ * bound to a {@link java.lang.invoke.ConstantCallSite} inside the method that define them, but that
+ * {@link java.lang.invoke.ConstantCallSite} is not shared across methods (even for methods in the same class). Thus, when
+ * lambda reference equality is important (such as for node sharing in Constraint Streams), the lambdas in this class should be
+ * used.
+ */
+public final class ConstantLambdaUtils {
```
triceo, line 22: Let's add a private constructor to prevent instances.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.function.ToIntBiFunction;
+import java.util.function.ToIntFunction;
+
+import ai.timefold.solver.core.api.function.QuadFunction;
+import ai.timefold.solver.core.api.function.ToIntQuadFunction;
+import ai.timefold.solver.core.api.function.ToIntTriFunction;
+import ai.timefold.solver.core.api.function.TriFunction;
+
+/**
+ * A class that hold common lambdas that are guarantee to be the same across method calls. In most JDK's, stateless lambdas are
```
triceo, line 16:
```suggestion
 * A class that holds common lambdas that are guaranteed to be the same across method calls. In most JDK's, stateless lambdas are
```

core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/ConstantLambdaUtils.java
```diff
@@ -0,0 +1,216 @@
...
+    private static final QuadFunction QUAD_PICK_FOURTH = (a, b, c, d) -> d;
+
+    @SuppressWarnings("rawtypes")
+    private static final ToIntFunction UNI_CONSTANT_ONE = (a) -> 1;
+    @SuppressWarnings("rawtypes")
+    private static final ToIntBiFunction BI_CONSTANT_ONE = (a, b) -> 1;
+    @SuppressWarnings("rawtypes")
+    private static final ToIntTriFunction TRI_CONSTANT_ONE = (a, b, c) -> 1;
+
+    @SuppressWarnings("rawtypes")
```
triceo, line 63: Let's have consistent whitespace. Either use newlines, or don't.
If you want to separate blocks of lambdas, do so, but do it consistently.

triceo (changes requested):
Some comments. Most importantly, the tests should be simplified.

core/constraint-streams/src/test/java/ai/timefold/solver/constraint/streams/common/bi/AbstractBiConstraintStreamTest.java
triceo: In fact if each such assertions becomes its own well named test method, even better, because we will be able to introduce an interface for all 4 cardinalities, enforcing the kinds of things we want to test in each.

triceo (approved):

PR #377, opened 2023-10-27
perf: Implement node sharing for groupBy

Christopher-Chianelli:
Depends on #374

### DIFF
--- a/core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetAbstractBiConstraintStream.java
+++ b/core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/bi/BavetAbstractBiConstraintStream.java
@@ -169,7 +169,8 @@ public abstract class BavetAbstractBiConstraintStream<Solution_, A, B> extends B
         GroupNodeConstructor<UniTuple<Result_>> nodeConstructor =
                 of((groupStoreIndex, undoStoreIndex, tupleLifecycle, outputStoreSize,
                         environmentMode) -> new Group0Mapping1CollectorBiNode<>(groupStoreIndex, undoStoreIndex, collector,
-                                tupleLifecycle, outputStoreSize, environmentMode));
+                                tupleLifecycle, outputStoreSize, environmentMode),
+                        collector);
         return buildUniGroupBy(nodeConstructor);
     }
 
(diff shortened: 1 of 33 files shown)


Christopher-Chianelli:
- Too much new boilerplate inside collectors. Agree, although it mostly because each collector implement the same class, meaning we need a unique identifier to identify what kind of collector it is. Additionally, since each collector is the same class, we cannot use collector specific fields to identify what kind of collector it is. Third, the collectors construct lambda's using parameters, meaning we cannot rely on their reference equality to check for equals.
- The NodeConstructionFunction is a lambda that uses parameters, meaning we cannot rely on their reference equality to check for equals.
- The user does not need to use our own collectors; they just need to implement equals and hashCode on their custom ConstraintCollector to take advantage of groupBy node sharing

triceo:
May I suggest we leave the `groupBy` node sharing last?

For a while, I've had an idea for a large refactor of the collectors, which will help here. (Main points: remove all of the impls from `ConstraintCollectors` and create a hierarchy of private classes which share code and can put equality where it belongs.)

Wrt. node constructors - maybe the node constructor mechanism has outlived its usefulness and we need a better one. Nobody forces us to use this; it was good for me once, maybe it's not good anymore.

Wrt. user collector equality - indeed, but that needs to be mentioned in the `Uni/Bi...Collector` javadocs.

Christopher-Chianelli:
No problem with me (it being last was my original intention, but I accidentally did the rebasing out-of-order).

triceo:
Ok, let's start here: let's minimize the `ConstraintCollectors` class. At this point, the IDE has a hard time even highlighting things in that class, it's too complex.

- Create impl classes for each kind of collector; not in public API.
- Have subtypes for uni/bi/... Ideally hide constructors or even the entire classes, have a factory static class. (Even if not in public API, the impls would still be on the classpath and therefore visible; people would start using them directly.)
- Equality goes into those classes too.
- `ConstraintCollectors` only exists to create and return instances of those types.

It will be a lot of manual effort, but IMO a very necessary cleanup.
Once we have that, we see what complexity remains to be solved to get us to node sharing.

WDYT?

Christopher-Chianelli:
Sounds good to me; should it be in a separate PR or this one?

triceo:
Separate one is better. `chore`, unrelated to performance.

Christopher-Chianelli:
Superseded by #401

PR #392, opened 2023-11-01
chore: Move implementations of Collectors outside of ConstraintCollectors

Christopher-Chianelli:


triceo (changes requested):
What a job, Chris! I like how you break down the collectors in smaller pieces, and I like the new hierarchy.

Unfortunately, there'll have to be some changes, see my comments. Please let me know if you want to discuss any of these - I'm available as always.

Also, I will run benchmarks on this over the weekend, just to double-check there was no negative impact. So I expect to merge this next week, still in time for 1.4.0.

core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1547,14 +1121,14 @@ private static <A, B, C, D, Mapped> QuadConstraintCollector<A, B, C, D, SortedMa
...
      */
     public static <A extends Comparable<A>> UniConstraintCollector<A, ?, SortedSet<A>> toSortedSet() {
-        return toSortedSet(a -> a);
+        return toSortedSet(ConstantLambdaUtils.<A> identity());
     }
 
     /**
      * As defined by {@link #toSortedSet()}, only with a custom {@link Comparator}.
      */
     public static <A> UniConstraintCollector<A, ?, SortedSet<A>> toSortedSet(Comparator<? super A> comparator) {
```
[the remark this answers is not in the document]
Christopher-Chianelli: To be tested;  I am pretty sure the TreeMap would be corrupted if something like this occurs:

```
ConstraintCollectors.toSortedSet(Comparator.comparing(Entity::myPlanningVariable));
// ... sorted-set = {1, 2, 3, 4, 5, 6, 7, 8}, entity.myPlanningVariable=4
beforeVariableChange(entity, "myPlanningVariable");
entity.setMyPlanningVariable(10);
afterVariableChange(entity, "myPlanningVariable");
```
In particular, the remove call will have entity.myPlanningVariable=10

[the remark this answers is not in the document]
triceo: Yes, I think so too - but it's always best to test, not guess.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/IntCounter.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public void increment() {
+        count++;
+    }
+
+    public void decrement() {
+        count--;
+    }
+
+    public Integer result() {
```
triceo, line 14: Since there are no generics here, can we return `int`? Let's return what we have, and deal with boxing elsewhere.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/LongCounter.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public void increment() {
+        count++;
+    }
+
+    public void decrement() {
+        count--;
+    }
+
+    public Long result() {
```
triceo, line 14: `long` instead of `Long`?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/ToMapPerKeyCounter.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public Value_ merge(BinaryOperator<Value_> mergeFunction) {
+        // Rebuilding the value from the collection is not incremental.
+        // The impact is negligible, assuming there are not too many values for the same key.
+        return counts.entrySet()
+                .stream()
+                .flatMap(e -> Stream.generate(e::getKey).limit(e.getValue()))
+                .reduce(mergeFunction)
+                .orElseThrow(() -> new IllegalStateException("Programming error: Should have had at least one value."));
```
triceo, line 40:
```suggestion
                .orElseThrow(() -> new IllegalStateException("Impossible state: Should have had at least one value."));
```

"Impossible state" seems to be the convention for this.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/ComposeFourBiCollector.java
```diff
@@ -0,0 +1,114 @@
...
+                fourthFinisher.apply(resultHolder.getD()));
+    }
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object)
+            return true;
+        if (object == null || getClass() != object.getClass())
+            return false;
+        ComposeFourBiCollector<?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?> that =
```
triceo, line 101: I'd argue we can avoid this craziness with `var`.
The type is already on the right-hand side of the assignment, no need to duplicate it.

Let's try to apply it consistently.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/CountLongBiCollector.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof CountLongBiCollector;
+    }
+
+    @Override
+    public int hashCode() {
+        return 0;
+    }
```
triceo, line 41: If you make argless `count...()` a singleton, you get to avoid this, and also needless instances.
(Not that it'd ever be a problem.)

triceo (approved):
A few more minor comments; if we're touching this, we may as well clean it up.
(Let's try to apply them consistently in all the copy-paste places.)

I'll merge this after I've run the benchmarks.
Nice one, @Christopher-Chianelli!

triceo:
So far I have confirmed that none of the CS changes had impact on performance, positive or negative.
Tonight I will benchmark this PR.

triceo:
Results of the benchmarks were largely positive.

PR #401, opened 2023-11-06
perf: Implement node sharing for groupBy

Christopher-Chianelli:


Christopher-Chianelli:
Supersedes #401

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/GroupNodeConstructor.java
```diff
@@ -7,17 +7,138 @@
...
+    }
+
+    static <CollectorA_, CollectorB_, CollectorC_, Tuple_ extends AbstractTuple> GroupNodeConstructor<Tuple_>
+            zeroKeysGroupBy(CollectorA_ collectorA, CollectorB_ collectorB, CollectorC_ collectorC,
+                    NoKeysThreeCollectorsGroupByBuilder<CollectorA_, CollectorB_, CollectorC_, Tuple_> builder) {
+        return new GroupNodeConstructorWithAccumulate<>(Triple.of(collectorA, collectorB, collectorC),
+                (groupStoreIndex, undoStoreIndex, nextNodesTupleLifecycle, outputStoreSize, environmentMode) -> builder.build(
+                        groupStoreIndex, undoStoreIndex, collectorA, collectorB, collectorC,
+                        nextNodesTupleLifecycle, outputStoreSize, environmentMode));
+    }
```
triceo, line 40: This confused me at first.

- `CollectorA_` et al. are not `...ConstraintCollector` subtypes. (Because that would cause further explosion in the number of interfaces you'd have to create for this.)
- `NoKeysThreeCollectorsGroupByBuilder` also doesn't reference constraint collectors.
- But then the builder itself is `Group...Node::new`, which *requires* a collector implementation. `CollectorA_` is not a collector subtype.

How does this compile?

It compiles because `Group...Node::new` is covariant with `build(int groupStoreIndex, int undoStoreIndex, CollectorA_ collectorA, CollectorB_ collectorB, CollectorC_ collectorC, TupleLifecycle<Tuple_> nextNodesTupleLifecycle, int outputStoreSize, EnvironmentMode environmentMode)`. The compiler is fooled and if the generic type in the end doesn't match the expected collector type, we'll get a cast error at runtime.

This is not at all obvious.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/GroupNodeConstructor.java
```diff
@@ -7,17 +7,138 @@
 import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.score.stream.ConstraintStream;
 import ai.timefold.solver.core.config.solver.EnvironmentMode;
+import ai.timefold.solver.core.impl.util.Pair;
+import ai.timefold.solver.core.impl.util.Quadruple;
+import ai.timefold.solver.core.impl.util.Triple;
 
 public interface GroupNodeConstructor<Tuple_ extends AbstractTuple> {
```
triceo, line 14: It appears we no longer need the tuple here.
(Which also means we can now replace all declarations of `GroupNodeConstructor` variables with `var`.)

Actually, I checked the history and this was the case on `main` already.
Please make the changes anyway - it's a good thing to leave the code in a better shape than the one we found it in.

triceo (approved):
Left some small suggestions.
Will benchmark overnight.

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/GroupNodeConstructor.java
Christopher-Chianelli: It appears the `<Tuple_ extends AbstractTuple>` is actually an aid to the compiler; since it is used in the return type of the group by node constructors, it informs the compiler that `NodeConstructorWithAccumulate` uses `Tuple_` in the lambdas, allowing the `GroupNodeConstructorWithAccumulate` to accept it without a cast. Removing it makes a tons of casts necessary.  The alternative is to replace all `GroupNodeConstructor` usage with `GroupNodeConstructorWithAccumulate` or `GroupNodeConstructorWithoutAccumulate` (which does use the tuple type, in a constructor and a field).

triceo: Interesting. IntelliJ shows it to me as unnecessary. Must be a bug then.

Christopher-Chianelli: IntelliJ is not wrong; it is unused in GroupNodeConstructor, but `GroupNodeConstructorWith(out)Accumulate` does this: `GroupNodeConstructorWith(out)Accumulate<Tuple_ extends AbstractTuple> implements GroupNodeConstructor<Tuple_>`, and then uses Tuple_ in a field and its constructor. So when you remove the `Tuple_` from `GroupNodeConstructor` in the return type, the compiler no longer knows the generic signature of `GroupNodeConstructorWith(out)Accumulate` to use, making a ton of casting required in GroupNodeConstructor and all the ConstraintStream(s).

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/GroupNodeConstructor.java
```diff
@@ -36,6 +180,128 @@ AbstractNode apply(int groupStoreIndex, TupleLifecycle<Tuple_> nextNodesTupleLif
...
+    interface GroupBy0Mapping3CollectorNodeBuilder<CollectorA_, CollectorB_, CollectorC_, Tuple_ extends AbstractTuple> {
+        AbstractNode build(int groupStoreIndex, int undoStoreIndex,
+                CollectorA_ collectorA,
+                CollectorB_ collectorB,
+                CollectorC_ collectorC,
+                TupleLifecycle<Tuple_> nextNodesTupleLifecycle, int outputStoreSize, EnvironmentMode environmentMode);
+    }
+
+    @FunctionalInterface
+    interface GroupBy0Mapping4CollectorNodeBuilder<CollectorA_, CollectorB_, CollectorC_, CollectorD_, Tuple_ extends AbstractTuple> {
```
triceo, line 208: One thing just crossed my mind... if we don't need `Collector_` to extend `...ConstraintCollector`, do we need `Tuple_` to extend `AbstractTuple`?
If not, even more boilerplate to remove right here.

Christopher-Chianelli: It should not be needed, but it allows for type safety. Unlike Tuple(s), the ConstraintCollector interfaces do not have a common superinterface that we can use here.

triceo: Fair enough.

PR #415, opened 2023-11-13
chore: support TSPLIB95 datasets in the TSP example

triceo:


examples/src/main/java/ai/timefold/solver/examples/tsp/domain/location/Location.java
```diff
@@ -98,6 +98,10 @@ public double getAngle(Location location) {
         return Math.atan2(latitudeDifference, longitudeDifference);
     }
 
+    protected static long adjust(double distance) {
+        return (long) (distance + 0.5); // +0.5 to avoid floating point rounding errors
```
rsynek, line 102: Is it any different from `Math.round()` ?

rsynek (approved):

examples/src/main/java/ai/timefold/solver/examples/tsp/domain/location/Location.java
triceo: This is what the TSPLIP specifies, so this is what I implemented. :-)

PR #416, opened 2023-11-14
feat: add constraint weight to ScoreAnalysis

triceo:
Introduces constraint weight to ScoreAnalysis.
Sorts constraint matches by constraint weight first.


rsynek (approved):
Just to make sure, as the PR title speaks about an added feature, does this PR only rename weight -> constraintWeight ?

rsynek (approved):

PR #420, opened 2023-11-15
docs: Example on how to use asciidoctor-tabs in docs

pieterdeschepper:
Split up `Maven` and `Gradle` examples in tabs as an example on how to use `asccidoctor-tabs` that is now added as an option to the docs on timefold.ai website.


triceo (approved):
Looks decent in the PDF too.
Thanks, @pieterdeschepper!

PR #426, opened 2023-11-17
feat: promote consecutive sequence collector to public API

triceo:
- Moves the collector from examples to public API.
- Introduces `collectAndThen()` collector to improve a potential new common case.
- Refactors and extends collector coverage.

TODO docs


core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1829,6 +1879,248 @@ public static <A, ResultContainer_, Result_> UniConstraintCollector<A, ResultCon
...
+        return InnerUniConstraintCollectors.toConsecutiveSequences(indexMap);
+    }
+
+    /**
+     * Contains info regarding the consecutive sequences and breaks in a collection of points.
+     *
+     * @param <Value_> The type of value in the sequence.
+     * @param <Difference_> The type of difference between values in the sequence.
+     */
+    public interface SequenceChain<Value_, Difference_ extends Comparable<Difference_>> {
```
Christopher-Chianelli, line 1915: I wonder if these interfaces should be top-level

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/ConsecutiveSequencesBiConstraintCollector.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public Supplier<SequenceCalculator<Result_>> supplier() {
+        return () -> new SequenceCalculator<>(indexMap);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o instanceof ConsecutiveSequencesBiConstraintCollector<?, ?, ?> other) {
```
Christopher-Chianelli, line 29: You can call `super.equals()` to check if mappers are the same

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/ObjectCalculatorBiCollector.java
```diff
@@ -10,9 +10,9 @@
 
 abstract sealed class ObjectCalculatorBiCollector<A, B, Input_, Output_, Calculator_ extends ObjectCalculator<Input_, Output_>>
         implements BiConstraintCollector<A, B, Calculator_, Output_>
-        permits AverageReferenceBiCollector, CountDistinctIntBiCollector, CountDistinctLongBiCollector,
-        SumReferenceBiCollector {
-    private final BiFunction<? super A, ? super B, ? extends Input_> mapper;
+        permits AverageReferenceBiCollector, ConsecutiveSequencesBiConstraintCollector, CountDistinctIntBiCollector,
+        CountDistinctLongBiCollector, SumReferenceBiCollector {
+    protected final BiFunction<? super A, ? super B, ? extends Input_> mapper;
```
Christopher-Chianelli, line 15: Can remain private; call `super.equals` in subclasses' `equals` method.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/score/stream/ConsecutiveSetTreeTest.java
```diff
@@ -79,9 +101,9 @@ void testConsecutiveNumbers() {
         tree.add(atomic(8), 8);
 
         IterableList<Sequence<AtomicInteger, Integer>> sequenceList = new IterableList<>(tree.getConsecutiveSequences());
-        assertThat(sequenceList).hasSize(2);
+        Assertions.assertThat(sequenceList).hasSize(2);
```
Christopher-Chianelli, line 104: Why use `Assertions.assertThat` when `import static org.assertj.core.api.Assertions.assertThat;` is in the file?

Christopher-Chianelli (approved):
Code LGTM; For the collector implementation, mapper can remain private; call `super.equals()` and `super.hashCode()`.

core/core-impl/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
triceo: I was actually considering three options:

1. Make them inner interfaces.
2. Put them in the same package as `ConstraintCollectors`. Seemed wrong; these types are not as universal as the other types in that package.
3. Put them in a different package.

Since this was a judgement call and I couldn't really see a big difference between 1 and 3, I decided to go for 1. Seeing as you prefer 1, I made the change.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/ConsecutiveSequencesBiConstraintCollector.java
triceo: This is intentional.

I believe that equals/hashCode are supposed to be simple and easy to understand. One shouldn't have to go to superclasses to understand what equals will do; it should be obvious.

If the superclass had many fields with complicated relationships, I'd understand wanting to reuse that code. In that case, this aspect would IMO win over readability. But in this case, we're really talking about just one field in the superclass, and that's IMO not enough to obfuscate these simple methods.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/stream/bi/ObjectCalculatorBiCollector.java
triceo: See above for explanation.

PR #428, opened 2023-11-20
Use more records and create less instances

triceo:
This PR does several things:

- Many classes, like `PairImpl` become records, removing boilerplate code.
- `PairImpl` becomes `Pair`; we get rid of virtual calls and needless abstraction.
- `EqualsIndexer`s are more efficient because each maps to a single composite key, as opposed to iterating over multiple keys and creating the composite keys on the fly.

The performance of this PR has been benchmarked to be within the margin of error of the old code. Considering that this code has less boilerplate, it is better.


core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/index/IndexerFactory.java
```diff
@@ -1,115 +1,568 @@
...
-            joinerTypes[i] = joiner.getJoinerType(i);
         }
     }
 
     public boolean hasJoiners() {
-        return joinerTypes.length > 0;
+        return joiner.getJoinerCount() > 0;
+    }
+
+    public <A> Function<A, IndexProperties> buildUniLeftMapping() {
```
triceo, line 91: This is a lot of code duplication (uni, bi, tri, quad), the code is essentially the same. Yet it can not be de-duplicated, on account of the different number of arguments of the functions used. The usual CS issue. :-(

core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/index/IndexerFactory.java
```diff
@@ -1,115 +1,568 @@
...
+        var joinerCount = joiner.getJoinerCount();
+        if (joinerCount < 2) {
+            joinerTypeMap = null;
+        } else {
+            joinerTypeMap = new TreeMap<>();
+            for (var i = 1; i <= joinerCount; i++) {
+                var joinerType = i < joinerCount ? joiner.getJoinerType(i) : null;
+                var previousJoinerType = joiner.getJoinerType(i - 1);
+                if (joinerType != JoinerType.EQUAL || previousJoinerType != joinerType) {
+                    // Equal joiner may build composite keyFunction with preceding equal joiner(s).
```
rsynek, line 80: The condition and the comment made me think there is a bug at the first glance; after carefully looking in the code I understand we intentionally skip the map for Equal joiners so that the index increases through the iteration.
Maybe a bit more explanatory comment would help.

rsynek (approved):

PR #432, opened 2023-11-21
feat: Introduce the Recommended Fit API

triceo:


docs/src/modules/ROOT/pages/responding-to-change/responding-to-change.adoc
```diff
@@ -522,6 +522,109 @@ A `BestSolutionChangedEvent` does not guarantee that every `ProblemChange` has b
...
+ShiftAssignment unassignedShift = new ShiftAssignment(...); // A new shift needs to be assigned.
+nurseRoster.getShiftAssignmentList().add(unassignedShift);
+----
+
+The `SolutionManager` is then used to retrieve the recommended fit for the uninitialized entity:
+
+[source,java,options="nowrap"]
+----
+SolutionManager<NurseRoster, HardSoftScore> solutionManager = ...;
+List<RecommendedFit<Employee, HardSoftScore>> recommendations =
```
rsynek, line 568: What happens if the planning entity has two planning variables of different types? Do we call the API twice (for every variable with different proposition functions)?

docs/src/modules/ROOT/pages/responding-to-change/responding-to-change.adoc
```diff
@@ -522,6 +522,109 @@ A `BestSolutionChangedEvent` does not guarantee that every `ProblemChange` has b
...
+or any values that will mutate as the solver tries to find the best fit.
+In that case, there are some caveats to consider.
+
+The solver will try to find the best fit for the uninitialized entity,
+and it will start from the solution it received on input.
+Before trying the next value to assign, it will first return to that original solution.
+The consequence of this is that if our proposition function returns any values that change during this process,
+those changes will also affect the previously processed propositions.
+
+In other words, if we decide to return the entire entity from the proposition function,
```
rsynek, line 613: Can you maybe think of an example to show this situation?

docs/src/modules/ROOT/pages/responding-to-change/responding-to-change.adoc
triceo: Good point that I should mention that this requires a construction heuristic configuration in the solver config.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/Fitter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public List<RecommendedFit<Out_, Score_>> apply(InnerScoreDirector<Solution_, Score_> scoreDirector) {
+        var solutionDescriptor = scoreDirector.getSolutionDescriptor();
+        var initializationStatistics = solutionDescriptor.computeInitializationStatistics(originalSolution);
+        var uninitializedCount =
+                initializationStatistics.uninitializedEntityCount() + initializationStatistics.unassignedValueCount();
+        if (uninitializedCount > 1) {
+            throw new IllegalStateException("""
```
mcimbora, line 37: Trying to understand this check. Let's say I've got the following solution with nullable variables (which is very common use case, since you might not be able to assign all the shifts everytime the solver runs).

Shift A - null
Shift B - employee1
Shift C - employee2

Now I call recommend(solution_with_shift_D, Shift D). This will contain 2 unitialized shifts (Shift A, Shift D). Does the new API handle this case, or throws an exception?

triceo: Nullable variables count as initialized. So if the variable is nullable, the total uninitialized count will be 1.
If the variable is not nullable, the exception will be thrown, as there'd be 2 uninitialized entities.

There is another way of thinking about this - the `init` score. If the init score was anything but zero when you started, the API will fail fast.

triceo: I edited the previous comment, as I said something that wasn't true and corrected it later.

mcimbora: Sounds good, thanks for the explanation.

mcimbora (approved):
Smoke tested this version locally with one of our models, all good.

PR #433, opened 2023-11-21
feat: Report missing before/after calls on undo corruption

Christopher-Chianelli:


triceo (changes requested):
Right now, I think it's safe to assume that your needs for the current task will be evolving.

Let's wait until you've finished your work and then go through this code, figure out which parts we need, and what is the best way to package and deliver them.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -63,7 +67,18 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
...
+    private AllVariablesAssert<Solution_> afterMoveSnapshot;
+    private Solution_ afterMoveSolution;
+    private final List<Pair<VariableDescriptor<Solution_>, Object>> beforeVariableChangedForwardEvents = new ArrayList<>();
+    private final List<Pair<VariableDescriptor<Solution_>, Object>> afterVariableChangedForwardEvents = new ArrayList<>();
+    private final List<Pair<VariableDescriptor<Solution_>, Object>> beforeVariableChangedUndoEvents = new ArrayList<>();
+    private final List<Pair<VariableDescriptor<Solution_>, Object>> afterVariableChangedUndoEvents = new ArrayList<>();
+
+    // These points to either ...ForwardEvents or ...UndoEvents, or null if not in an assert mode
+    private List<Pair<VariableDescriptor<Solution_>, Object>> beforeVariableChangedEvents = null;
+    private List<Pair<VariableDescriptor<Solution_>, Object>> afterVariableChangedEvents = null;
```
triceo, line 81: This complicates the score director far too much.

Let's think of a way of encapsulating it in a single type. All of this work should happen within that type, with the interactions between `ScoreDirector` and the other type being kept at a minimum;  thinking out loud here, perhaps some event-based system could do.

Once we have this in a separate class independent of the score director, we can test that it behaves properly in all cases. (Environment modes are largely untested, and with the code becoming really complicated, we need to change that.)

core/core-impl/src/test/java/ai/timefold/solver/core/config/solver/EnvironmentModeTest.java
```diff
@@ -147,15 +155,23 @@ void corruptedUndoShadowVariableListener(EnvironmentMode environmentMode) {
 
         switch (environmentMode) {
             case FULL_ASSERT:
-            case FAST_ASSERT:
```
triceo, line 150: We need a name for the new environment mode, something more descriptive than `SUPER_FULL_ASSERT`. Or maybe not?

Your original numbers suggest that this is not *that much* slower than `FULL_ASSERT`. If you can first fix the score calculation numbers to give us *correct* data, we can assess those speeds again and if we're talking about a drop of a couple percent, we really don't need a whole new env mode for this.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -644,53 +693,166 @@ public void assertExpectedUndoMoveScore(Move<Solution_> move, Score_ beforeMoveS
...
+            out.append(expectedSnapshotVariable.getEntity());
+            out.append(") expected (");
+            out.append(expectedSnapshotVariable.getValue());
+            out.append(") actual (");
+            out.append(actualSnapshotVariable.getValue());
+            out.append(")\n");
+            violationCount++;
+            if (violationCount >= limit) {
+                return;
+            }
```
triceo, line 820: Same comment on readable strings.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -644,53 +693,166 @@ public void assertExpectedUndoMoveScore(Move<Solution_> move, Score_ beforeMoveS
...
-                    + differentShadowVariables
+                    + differentVariables
                     + "  1) Enable EnvironmentMode " + EnvironmentMode.FULL_ASSERT
                     + " (if you haven't already) to fail-faster in case there's a score corruption or variable listener corruption.\n"
                     + "  2) Check the Move.createUndoMove(...) method of the moveClass (" + move.getClass() + ")."
                     + " The move (" + move + ") might have a corrupted undoMove (" + undoMoveText + ").\n"
                     + "  3) Check your custom " + VariableListener.class.getSimpleName() + "s (if you have any)"
                     + " for shadow variables that are used by score constraints that could cause"
-                    + " the scoreDifference (" + scoreDifference + ").");
+                    + " the scoreDifference (" + scoreDifference + ").",
```
triceo, line 717: IMO we can also do a lot here to make the string much more readable, raw strings were made for this.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -644,53 +693,166 @@ public void assertExpectedUndoMoveScore(Move<Solution_> move, Score_ beforeMoveS
...
-        } else if (undoShadowVariableViolations != null) {
-            return "Shadow variables have different values when recalculated from scratch after undo:\n" +
-                    undoShadowVariableViolations + "\n";
-        } else if (beforeShadowVariableViolations != null) {
-            return "Shadow variables have different values when recalculated from scratch before undo:\n"
-                    + beforeShadowVariableViolations
-                    + "\n";
-        } else {
-            return "Shadow variables agrees with from scratch calculations before and after undo.\n";
+    private void appendVariableChangedEventViolations(StringBuilder out, String prefix,
```
triceo, line 824: Let's make this a function without side effects; return `String`, don't accept `out`.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/AllVariablesAssert.java
```diff
@@ -0,0 +1,54 @@
...
+
+import ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+
+/**
+ * Serves for detecting undo move corruption. When a snapshot is created, it records the state of all variables (genuine and
+ * shadow) for all entities.
+ */
+public class AllVariablesAssert<Solution_> {
```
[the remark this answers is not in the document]
triceo: Also, is this an `Assert` if it doesn't actually assert anything? I'd call it `VariableSnapshotTotal` or something like that.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/VariableSnapshot.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public VariableId<Solution_> getVariableId() {
+        return variableId;
+    }
+
+    public boolean isDifferentFrom(VariableSnapshot<Solution_> other) {
+        return !Objects.equals(value, other.value);
+    }
+}
```
triceo, line 48: Why not use `equals(...)` for this? Isn't that what it's for?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -203,13 +211,16 @@ private void assertNonNullPlanningId(Object fact) {
 
     @Override
     public Score_ doAndProcessMove(Move<Solution_> move, boolean assertMoveScoreFromScratch) {
-        if (assertMoveScoreFromScratch) {
-            beforeMoveSnapshot = ShadowVariablesAssert.takeSnapshot(getSolutionDescriptor(), workingSolution);
+        if (solutionTracker != null) {
```
triceo, line 214: I like the previous approach, where this condition was hidden behind a variable. The variable makes the purpose of the check obvious, while hiding what the check actually is.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -607,7 +621,8 @@ public void assertShadowVariablesAreNotStale(Score_ expectedWorkingScore, Object
                     + "s were triggered without changes to the genuine variables"
                     + " after completedAction (" + completedAction + ").\n"
                     + "But all the shadow variable values are still the same, so this is impossible.\n"
-                    + "Maybe run with " + EnvironmentMode.FULL_ASSERT + " if you aren't already, to fail earlier.");
+                    + "Maybe run with " + EnvironmentMode.FULL_ASSERT_WITH_TRACKING
+                    + " if you aren't already, to fail earlier.");
```
triceo, line 625: Since we're touching this code, let's convert this nastiness to the new formatting pattern.

Remember: if you can, and it is not too disruptive, always leave the code in better shape than the one you found it in.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -667,53 +682,41 @@ public void assertExpectedUndoMoveScore(Move<Solution_> move, Score_ beforeMoveS
...
-            String differentShadowVariables = buildShadowVariableDiff(afterUndoSnapshot);
+            String corruptionDiagnosis = "";
+            if (solutionTracker != null) {
+                solutionTracker.setFromScratchSolution(workingSolution);
+                corruptionDiagnosis = solutionTracker.buildScoreCorruptionMessage();
+            }
             String scoreDifference = undoScore.subtract(beforeMoveScore).toShortString();
-
-            throw new IllegalStateException("UndoMove corruption (" + scoreDifference
+            String corruptionMessage = ("UndoMove corruption (" + scoreDifference
```
triceo, line 698: Dtto., formatting pattern.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
```diff
@@ -68,6 +69,14 @@ public void setAssertClonedSolution(boolean assertClonedSolution) {
         this.assertClonedSolution = assertClonedSolution;
     }
 
+    public boolean isTrackingWorkingSolution() {
```
triceo, line 72: This deserves a Javadoc, pointing to someplace that describes what it will do. Perhaps a reference to the relevant `EnvironmentMode`?

core/core-impl/src/test/java/ai/timefold/solver/core/config/solver/EnvironmentModeTest.java
```diff
@@ -88,17 +87,16 @@ void determinism(EnvironmentMode environmentMode) {
         Solver<TestdataSolution> solver2 = SolverFactory.<TestdataSolution> create(solverConfig).buildSolver();
 
         switch (environmentMode) {
-            case NON_REPRODUCIBLE:
+            case NON_REPRODUCIBLE -> {
```
triceo, line 90: I like we're doing this!

Just please double check the entire codebase for switching over environment modes and apply this consistently. (If you haven't already.)

core/core-impl/src/test/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorSemanticsTest.java
```diff
@@ -141,6 +141,8 @@ void constraintPresentEvenIfNoMatches() {
             scoreDirector.beforeVariableChanged(entity, "value");
             entity.setValue(null);
             scoreDirector.afterVariableChanged(entity, "value");
+            // FULL_ASSERT add variable listeners for tracking
+            scoreDirector.triggerVariableListeners();
             var score2 = scoreDirector.calculateScore();
```
triceo, line 146: This concerns me.
Are you saying we need to do something *new* in order to use this functionality? 
Why would this code need it, if it doesn't run any of the assert modes?

triceo (changes requested):
I have added a whole bunch of comments, some to functionality, some to general code quality and cleanliness.

IMO we should discuss some general best practices we will follow throught every PR. We have some, but I feel like the list should be somewhat extended and modernized.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/EnvironmentMode.java
```diff
@@ -19,6 +19,26 @@
...
+     * {@link ai.timefold.solver.core.api.solver.exception.UndoScoreCorruptionException}, which can be used to inspect the
+     * working solution state before, after and after the undo of a {@link Move}.
+     * <p>
+     * This mode is reproducible (see {@link #REPRODUCIBLE} mode).
+     * <p>
+     * This mode is intrusive because it calls the {@link InnerScoreDirector#calculateScore()} more frequently than a non assert
+     * mode.
+     * <p>
+     * This mode is appallingly slow.
+     */
```
triceo, line 39: Besides the general comment on line breaks and formatting, I think maybe we're a bit more evolved now to use words such as "horrible" and "appaling" in user-facing documentation. :-) 
In this case, I suggest "This mode is by far the slowest of all the modes."

Also, somewhere in the body of the text, I would reference that it does everything that `FULL_ASSERT` does, plus X and Y. If you can reference an information that is already somewhere else, I prefer you do so - that way, you still have only a single point of truth for that information.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/EnvironmentMode.java
```diff
@@ -19,6 +19,26 @@
...
+    FULL_ASSERT_WITH_TRACKING(true),
+
     /**
      * This mode turns on all assertions
```
triceo, line 43: This no longer turns on _all_ assertions.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/VariableId.java
```diff
@@ -0,0 +1,6 @@
+package ai.timefold.solver.core.impl.domain.variable.listener.support.violation;
+
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+
+public record VariableId<Solution_>(VariableDescriptor<Solution_> variableDescriptor, Object entity) {
```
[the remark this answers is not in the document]
Christopher-Chianelli: Well, the snapshots are taken from the same working solution, meaning the entities will be the same.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/AllVariablesAssert.java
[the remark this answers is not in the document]
Christopher-Chianelli: Counterpoint: `ShadowVariableAssert`.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/ListVariableTracker.java
```diff
@@ -0,0 +1,107 @@
...
+                        + variableDescriptor.getVariableName() + ").");
+            }
+        }
+        beforeVariableChangedEntityList.clear();
+        afterVariableChangedEntityList.clear();
+        return out;
+    }
+
+    public TrackerDemand demand() {
+        return new TrackerDemand();
```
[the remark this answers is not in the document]
Christopher-Chianelli: No two demands for these classes will ever be equal. It a hack in order to get the variable listeners set up, since they need a Demand to get put into the logic.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/AllVariablesAssert.java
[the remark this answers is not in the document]
triceo: Good point. That said, if we're not already fixing mistakes of the past, at least we can not make new ones. WDYT?

core/core-impl/src/test/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorSemanticsTest.java
Christopher-Chianelli: This code was introduced before `FULL_ASSERT` and `TRACKED_FULL_ASSERT` were seperated. `TRACKED_FULL_ASSERT` adds it own variable listeners, so it need the variable listeners to be trigerred to not throw an exception about missed before... and after... calls.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/EnvironmentMode.java
Christopher-Chianelli: It does turn on all assertions; `TRACKED_FULL_ASSERT` will only fail when `FULL_ASSERT` fails, the main difference is it tells you why it failed.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorSemanticsTest.java
triceo: Which effectively means it will *break* everything which didn't already do this.
`ScoreDirector` is public, its behavior is set in stone.
Therefore this is not acceptable.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/exception/UndoScoreCorruptionException.java
```diff
@@ -0,0 +1,100 @@
+package ai.timefold.solver.core.api.solver.exception;
```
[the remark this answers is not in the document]
Christopher-Chianelli: This exception purely exists for the user; the internals do not use it. I myself has already used the exception to debug shadow variables. It allows you to do things like:

```
ObjectMapper mapper = ...;
try {
    ...
} catch (UndoScoreCorruptionException e) {
    Files.write("reproducer-solution.json", mapper.writeAsJson(e.getBeforeMoveSolution()));
    Files.write("reproducer-move.json", mapper.writeAsJson(e.getMove()));
}
```

to save the reproducer, or

```
try {
    ...
} catch (UndoScoreCorruptionException e) {
    // Debug point here
    e.reproduce();
}
```
To reproduce the exception and examine your own code when things went horribly wrong.
Strictly speaking, the reason why the solutions are cloned are purely for this exception; the snapshots has all the information we need for the exception message.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/VariableSnapshot.java
triceo: In that case, yes, I agree on the exception. Failing fast is good, let's not be like a certain rule engine. ;-)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/exception/UndoScoreCorruptionException.java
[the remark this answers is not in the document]
triceo: Fair enough; my argument still stands though - having something in the public API means supporting it forever, and that means thinking a lot about all the possible implications of it.

If we put it outside of the public API, we can observe how people use it in real world, and _change it_ according to that usage. Think of this as the experimental package.

Please remove this from the public API.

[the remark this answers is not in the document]
Christopher-Chianelli: We do need a place for experimental features we want people to use but understand they are not stable (yet). `impl` -> "Please do not use, unstable", `api` -> "Please use, stable", `?` -> "Please use, unstable"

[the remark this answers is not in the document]
triceo: I agree in principle. In practice, it becomes more complex, as for example extensions to existing APIs (such as new methods on `SolutionManager`) make this more complicated.

Deserves a proper discussion, not part of this PR I'd argue.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/NormalVariableTracker.java
```diff
@@ -0,0 +1,99 @@
...
+
+import ai.timefold.solver.core.api.domain.variable.VariableListener;
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.listener.SourcedVariableListener;
+import ai.timefold.solver.core.impl.domain.variable.supply.Demand;
+import ai.timefold.solver.core.impl.domain.variable.supply.Supply;
+import ai.timefold.solver.core.impl.domain.variable.supply.SupplyManager;
+
+public class NormalVariableTracker<Solution_>
```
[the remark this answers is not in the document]
Christopher-Chianelli: Well this tracks both Genuine and Shadow variables (basically everything but List Variable)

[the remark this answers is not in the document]
triceo: Then I'd argue we don't need the `Normal` prefix, and we do need a descriptive Javadoc.

triceo (approved):

triceo:
@Christopher-Chianelli One thing I totally forgot - we should add this to the docs. AFAIK there is a whole section on assert modes, the new one should show up there too, with an explanation of what it's for, ideally with an example. Please do that, and then I'll merge it.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
 
 The FULL_ASSERT mode is horribly slow (because it does not rely on incremental score calculation).
 
+[#environmentModeTracedFullAssert]
+=== `TRACED_FULL_ASSERT`
+
+The TRACED_FULL_ASSERT mode turns on all the <<environmentModeFullAssert, FULL_ASSERT>> assertions and additionally track changes to the working solution.
```
triceo, line 108:
```suggestion
The TRACED_FULL_ASSERT mode turns on all the <<environmentModeFullAssert, FULL_ASSERT>> assertions and additionally tracks changes to the working solution.
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
...
+This allows the solver to report exactly what variables were corrupted and what variable listener events are missing.
+
+For instance, when this incorrect move is evaluated
```
triceo, line 111:
```suggestion
Consider the following incorrectly implemented move:
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
...
+
+    @Override
+    protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
+        scoreDirector.beforeVariableChanged(entity, "value");
+        entity.setValue(newValue);
+        scoreDirector.afterVariableChanged(entity, "value");
+    }
+```
+
+the exception thrown by the solver will have the following in its message:
```
triceo, line 127:
```suggestion
When it is evaluated by the solver, an exception will be thrown with the following in its message:
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
...
+
+```
+Variables that are different between before and undo:
+  - Actual value (newValue) of variable value on EntityClass entity (entity) differs from expected (oldValue)
+```
+
+In particular, the solver will recalculate all shadow variables from scratch on the solution after the undo and then report:
+
+- Genuine and shadow variables that are different between before and undo.
```
triceo, line 136:
```suggestion
- Genuine and shadow variables that are different between "before" and "undo".
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
...
+After an undo move is evaluated, it is expected to exactly match the original working solution.
+
+- Variables that are different between from scratch and before.
+This is to detect if the solution was corrupted before the move was evaluated due to shadow variable corruption.
+
+- Variables that are different between from scratch and undo.
```
triceo, line 142:
```suggestion
- Variables that are different between "from scratch" and "undo".
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -102,6 +102,59 @@ This mode is reproducible (see the reproducible mode). It is also intrusive beca
...
+
+- Missing variable listener events for actual move.
+Any variable that changed between the before move solution and after move solution without either a
+`beforeVariableChanged` or `afterVariableChanged` would be reported here.
+
+- Missing variable listener events for undo move.
+Any variable that changed between the after move solution and after undo move solution without either a
+`beforeVariableChanged` or `afterVariableChanged` would be reported here.
+
+This mode is reproducible (see the reproducible mode).
```
triceo, line 153: The reproducible mode deserves a direct reference. (Unless it doesn't have an ID.)

triceo (approved):
I added some suggestions for improving the clarity of the doc, otherwise LGTM.

PR #437, opened 2023-11-22
Refactor certain sections of docs

triceo:


docs/src/modules/ROOT/nav.adoc
```diff
@@ -6,16 +6,20 @@
 ** xref:quickstart/spring-boot/spring-boot-quickstart.adoc[leveloffset=+1]
 * xref:use-cases-and-examples/use-cases-and-examples.adoc[leveloffset=+1]
 * xref:configuration/configuration.adoc[leveloffset=+1]
```
rsynek, line 8: What is the difference between this configuration chapter and the `using-timefold-solver/configuration.adoc` ? I would rather avoid duplicate menu entries.

rsynek (approved):

docs/src/modules/ROOT/nav.adoc
triceo: Interesting, I thought I'd deleted it. Will delete it, wasn't supposed to be there.

PR #459, opened 2023-12-01
chore: maintenanceScheduling diagrams

ge0ffrey:


triceo (changes requested):
I believe we no longer use `readyDate` and `dueDate`. Please fix the picture.

ge0ffrey:
@triceo For now, in the maintenance scheduling quickstart, we still have ready and due.
In the long run, we will all need to standarize that to minStart and maxEnd (it's more flexible and more clear).
But until than, I guess it's good as is?

triceo:
Hmm, interesting, I swear I saw a PR changing it. I guess there's still some quickstarts where it remains. Fair enough.

triceo (approved):

PR #485, opened 2023-12-12
fix: filter trivial 2-opt moves and make KOptListMove public

### LINKED ISSUE #480: Trivial K-Opt moves can not be filtered out
See https://stackoverflow.com/questions/77619677/timefold-is-it-possible-to-filter-out-all-trivial-2opt-moves-for-planninglistva

Two changes we need to make:
- Make sure moves are public, so that they can be used in move filters.
- Make sure the trivial moves are filtered out by default.

Christopher-Chianelli:
Fixes #480


triceo (approved):
LGTM, thank you!
(But there appears to be a Javadoc error.)

PR #491, opened 2023-12-14
feat: pinning support for list variable

triceo:


triceo:
@Christopher-Chianelli Before I move on to implementing the planning pin index, I want you to do a review first so that we're both satisfied with this approach before I make it proliferate. 

You implemented the K-Opt move, so your knowledge of the selectors makes you the ideal person.

Side note: One test is failing, it is one of those you introduced recently. Please take a look at fixing it. (More info in a DM on Slack.)

rsynek (approved):
Did a very quick check of the API and tested the feature shortly (not a deep code review).

PR #515, opened 2023-12-21
feat: Overriding Solver termination configuration

triceo:
This pull request introduces a new feature that enables users to override solver termination settings without recreating the `SolverFactory` or `SolverManager`.

The `SolverManager` now has a builder API for job configuration and execution. Additionally, `SolverFactory` can build a solver and override the termination config settings.


core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.UUID;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+
+/**
+ * Provides a fluid contract that allows customization and submission of planning problems to solve. It differs {@link SolverJobFactory}
```
triceo, line 11:
```suggestion
 * Provides a fluent API that allows customization and submission of planning problems to solve. It differs {@link SolverJobFactory}
```

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * because it allows to listen to intermediate best solutions.
+ * <p>
+ * To create a {@link SolverJob}, use {@link #solveAndListen()}.
+ *
+ * @param <Solution_>  the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of a submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public interface SolverAndListenerJobFactory<Solution_, ProblemId_> extends SolverJobFactory<Solution_, ProblemId_> {
```
triceo, line 19: Be prepared that there will be a naming discussion and these names are most likely to change. :-)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
```diff
@@ -0,0 +1,41 @@
...
+    // With methods
+    // ************************************************************************
+    @Override
+    SolverAndListenerJobFactory<Solution_, ProblemId_> withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
+
+    @Override
+    SolverAndListenerJobFactory<Solution_, ProblemId_> withExceptionHandler(BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler);
+
+    @Override
+    SolverAndListenerJobFactory<Solution_, ProblemId_> withTerminationConfig(TerminationConfig terminationConfig);
```
triceo, line 31: Please replace this with a method accepting a `SolverConfigOverride`, which will have this method on it.
The reason is that we will then use the same override in both the `Solver` and the `SolverManager` builders, without having to copy each of the methods.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Provides a fluid contract that allows customization and submission of planning problems to solve.
+ * <p>
+ * To create a {@link SolverJob}, use {@link #solve()}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of a submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public interface SolverJobFactory<Solution_, ProblemId_> {
```
triceo, line 18: I understand why you decided to make this two interfaces, since you want to - correctly - share some of these methods.

In this case, there is another factor to consider - in the public API, we want to have interfaces which are easy to read and understand. If an interface requires another interface, that ease of reading goes down. I'd prefer if we avoided inheritance in this case.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -268,11 +364,46 @@ default SolverJob<Solution_, ProblemId_> solveAndListen(ProblemId_ problemId,
...
+     * @param configOverride sometimes null, includes settings that override the default configuration
+     * @return never null
+     */
     SolverJob<Solution_, ProblemId_> solveAndListen(ProblemId_ problemId,
             Function<? super ProblemId_, ? extends Solution_> problemFinder,
             Consumer<? super Solution_> bestSolutionConsumer,
             Consumer<? super Solution_> finalBestSolutionConsumer,
-            BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler);
+            BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler,
+            SolverConfigOverride configOverride);
```
triceo, line 406: The point of the builder is that we would no longer be creating new overloads. This method, therefore, should not exist.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverConfigOverride.java
```diff
@@ -0,0 +1,51 @@
+package ai.timefold.solver.core.config.solver;
+
+import java.util.function.Consumer;
+
+import ai.timefold.solver.core.config.AbstractConfig;
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+import ai.timefold.solver.core.config.util.ConfigUtils;
+
+public class SolverConfigOverride extends AbstractConfig<SolverConfigOverride> {
```
triceo, line 9: I wouldn't implement `AbstractConfig`. This class is not an actual config, it doesn't serialize to XML, it belongs to the `solver` package. For the same reason, I'd remove the getters and setters and just leave the `with` method; but you may need the getters later, so I'd at least make them package-private.

Let's not give people two ways of doing the same thing.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -98,6 +99,65 @@ static <Solution_, ProblemId_> SolverManager<Solution_, ProblemId_> create(
...
+     *
+     * @param problemId never null, a ID for each planning problem. This must be unique.
+     *        Use this problemId to {@link #terminateEarly(Object) terminate} the solver early,
+     *        {@link #getSolverStatus(Object) to get the status} or if the problem changes while solving.
+     * @param problemFinder never null, function that returns a {@link PlanningSolution}, usually with uninitialized planning
+     *                      variables
+     * @return never null
+     */
+    SolverJobFactory<Solution_, ProblemId_> builder(ProblemId_ problemId,
+                                                    Function<? super ProblemId_, ? extends Solution_> problemFinder);
```
triceo, line 130: You have no way of knowing this, so just for context: we hate the fact that the user has to provide `problemId`. It complicates the API, forces the user to have to think about things they may not need.

How about creating another builder overload which doesn't require a `problemId` and instead generates `long` IDs internally? We can not get rid of the `problemId`, but we can at least help hide it if the user doesn't care.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -98,6 +99,65 @@ static <Solution_, ProblemId_> SolverManager<Solution_, ProblemId_> create(
...
+     * @param problemId never null, a ID for each planning problem. This must be unique.
+     *        Use this problemId to {@link #terminateEarly(Object) terminate} the solver early,
+     * @param problem never null, a {@link PlanningSolution} usually with uninitialized planning variables
+     * @param bestSolutionConsumer never null, called multiple times, on a consumer thread
+     * @return never null
+     */
+    default SolverAndListenerJobFactory<Solution_, ProblemId_> builder(ProblemId_ problemId, Solution_ problem,
+                                                               Consumer<? super Solution_> bestSolutionConsumer) {
+        return builder(problemId, (problemId_) -> problem, bestSolutionConsumer);
+    }
```
triceo, line 145: I'd argue put `bestSolutionConsumer` on the builder, remove this overload.

core/core-impl/src/test/java/ai/timefold/solver/core/api/solver/SolverFactoryTest.java
```diff
@@ -128,10 +137,24 @@ void createFromInvalidXmlFile_failsShowingBothPathAndReason() throws IOException
...
-        SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class);
-        SolverFactory<TestdataSolution> solverFactory = SolverFactory.create(solverConfig);
-        Solver<TestdataSolution> solver = solverFactory.buildSolver();
-        assertThat(solver).isNotNull();
+        { // No config override
+            SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class);
+            SolverFactory<TestdataSolution> solverFactory = SolverFactory.create(solverConfig);
+            Solver<TestdataSolution> solver = solverFactory.buildSolver();
+            assertThat(solver).isNotNull();
+        }
```
triceo, line 145: Although these explicit code blocks don't hurt, we do not use that patterns.
It's an extra code which IMO makes no difference.

core/core-impl/src/test/java/ai/timefold/solver/core/api/solver/SolverFactoryTest.java
```diff
@@ -128,10 +137,24 @@ void createFromInvalidXmlFile_failsShowingBothPathAndReason() throws IOException
...
+
+        { // Config override
+            SolverConfig solverConfig = PlannerTestUtils.buildSolverConfig(TestdataSolution.class, TestdataEntity.class);
+            SolverFactory<TestdataSolution> solverFactory = SolverFactory.create(solverConfig);
+            SolverConfigOverride configOverride = mock(SolverConfigOverride.class);
+            TerminationConfig terminationConfig = new TerminationConfig();
+            terminationConfig.withSpentLimit(Duration.ofSeconds(60));
+            doReturn(terminationConfig).when(configOverride).getTerminationConfig();
+            Solver<TestdataSolution> solver = solverFactory.buildSolver(configOverride);
```
triceo, line 154: Arguably, we also want to support the override directly on `solver.solve(...)`. But let's see how hard that would be; maybe we decide not to do that in the end.

core/core-impl/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
```diff
@@ -213,6 +217,59 @@ void solveGenerics() throws ExecutionException, InterruptedException {
         solverJob.getFinalBestSolution();
     }
 
+    @Test
+    void solveWithOverride() throws InterruptedException {
+        CountDownLatch countDownLatch = new CountDownLatch(2);
+        PhaseConfig<?> customPhaseConfig = new CustomPhaseConfig().withCustomPhaseCommands(
+                scoreDirector -> {
+                    await().pollDelay(Duration.ofSeconds(5L)).until(() -> true);
+                    countDownLatch.countDown();
```
triceo, line 226: We typically try to avoid unit tests which run for seconds. Unit tests are supposed to be cheap and quick, otherwise people stop running them. Arguably you can have the wait much shorter, and then in the end check for how long the solver actually ran.

(Note: every time you play with time in tests, you inevitably invite the test to be flaky.)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -196,10 +264,35 @@ default SolverJob<Solution_, ProblemId_> solve(ProblemId_ problemId,
...
+     * @return never null
+     */
     SolverJob<Solution_, ProblemId_> solve(ProblemId_ problemId,
-            Function<? super ProblemId_, ? extends Solution_> problemFinder,
-            Consumer<? super Solution_> finalBestSolutionConsumer,
-            BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler);
+                                           Function<? super ProblemId_, ? extends Solution_> problemFinder,
+                                           Consumer<? super Solution_> finalBestSolutionConsumer,
+                                           BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler,
+                                           SolverConfigOverride configOverride);
```
triceo, line 295: Leave the config override business to the builder.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -98,6 +99,65 @@ static <Solution_, ProblemId_> SolverManager<Solution_, ProblemId_> create(
...
+     * Creates a Builder that allows to customize and submit a planning problem to solve.
+     *
+     * @param problemId never null, a ID for each planning problem. This must be unique.
+     *        Use this problemId to {@link #terminateEarly(Object) terminate} the solver early,
+     * @param problem never null, a {@link PlanningSolution} usually with uninitialized planning variables
+     * @return never null
+     */
+    default SolverJobFactory<Solution_, ProblemId_> builder(ProblemId_ problemId, Solution_ problem) {
+        return builder(problemId, (problemId_) -> problem);
+    }
```
triceo, line 116: Personally, I'd leave the `solution` etc. all the way for the `solve(solution)` and `solveAndListen(solution)`. Keep the number of overloads in the `SolverManager` to a minimum.

triceo (commented):
First batch of comments are in.
I like what you did, but I'm thinking we can simplify it for the user.

core/core-impl/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
zepfred: Yes, it makes sense. The goal is to ensure the termination time is overridden, but the internal state is not fetchable. I'll try to think of something else.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobFactory.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: Let's wait for the design meeting to define the next steps here.

core/core-impl/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
zepfred: I simplified the test but had to expose `Termination<Solution_>` on `DefaultSolverJob`.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
zepfred: What do you think about hiding the config override class from the user? 
I suggest simplifying the user's experience by only dealing with the Termination configuration.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobFactory.java
zepfred: Done!

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: Done!

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: Done!

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
zepfred: The new design includes only two methods:

1. `<Solution_, ProblemId_> buildSolver();`
2. `<Solution_, ProblemId_> solve(SolverJobConfig<Solution_, ProblemId_> solverJobConfig);`

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Provides a fluid contract that allows customization and submission of planning problems to solve.
+ * <p>
+ * To create a {@link SolverJob}, use {@link #run()}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public interface SolverJobBuilder<Solution_, ProblemId_> {
```
triceo, line 19: The methods in this class need Javadocs.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+    // ************************************************************************
+    // With methods
+    // ************************************************************************
+    SolverJobBuilder<Solution_, ProblemId_> withProblemId(ProblemId_ id);
```
triceo, line 24: We've been moving away from this comment pattern in new code.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+            withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
+
+    SolverJobBuilder<Solution_, ProblemId_>
+            withExceptionHandler(BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler);
+
+    // ************************************************************************
+    // Overriding settings
+    // ************************************************************************
+    SolverJobBuilder<Solution_, ProblemId_> withTerminationConfig(TerminationConfig terminationConfig);
+
```
triceo, line 44: This method shouldn't exist. Instead, there should be the one that accepts `SolverConfigOverride`.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -98,6 +99,17 @@ static <Solution_, ProblemId_> SolverManager<Solution_, ProblemId_> create(
...
+    // ************************************************************************
+    // Builder method
+    // ************************************************************************
+
+    /**
+     * Creates a Builder that allows to customize and submit a planning problem to solve.
+     *
+     * @return never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> buildSolver();
```
triceo, line 111: I think we agreed that this method would be called `solveBuilder()`. To be consistent with the `solve()` methods.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -196,10 +224,30 @@ default SolverJob<Solution_, ProblemId_> solve(ProblemId_ problemId,
...
+     * The planning problem is solved on a solver {@link Thread}, as soon as one is available.
+     * To retrieve the final best solution, use {@link SolverJob#getFinalBestSolution()}.
+     * <p>
+     * All settings needed by {@link ai.timefold.solver.core.api.solver.SolverManager} for solving a problem is given
+     * by {@link SolverJobConfig}.
+     *
+     * @param solverJobConfig never null, settings required to solve the problem
+     * @return never null
+     */
+    SolverJob<Solution_, ProblemId_> solve(SolverJobConfig<Solution_, ProblemId_> solverJobConfig);
```
triceo, line 251: Please don't expose this method on a public API.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverConfigOverride.java
```diff
@@ -0,0 +1,38 @@
...
+
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+
+/**
+ * Includes settings to override default {@link ai.timefold.solver.core.api.solver.Solver} configuration.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+public class SolverConfigOverride<Solution_> {
```
triceo, line 11: It is questionable if this class belongs in `config` package. As you can see, it caused it to be needlessly inserted in the XSD. I'd move this to `api/solver`, which is where it is primarily used.

Also, let's make it `final`. I am a big friend of removing freedom from people; if you allow people to do something, in this case extending the class, they will. And then you have to live with that forever.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverConfigOverride.java
```diff
@@ -0,0 +1,38 @@
...
+    private TerminationConfig terminationConfig = null;
+
+    // ************************************************************************
+    // Constructors and simple getters/setters
+    // ************************************************************************
+
+    public SolverConfigOverride() {
+    }
+
+    public SolverConfigOverride(SolverJobConfig<Solution_, ?> executionConfig) {
```
triceo, line 22: `SolverJobConfig` should not exist in public API.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverJobConfig.java
```diff
@@ -0,0 +1,101 @@
...
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+
+/**
+ * Settings used by {@link ai.timefold.solver.core.api.solver.SolverManager} to submit and solve problems. It also
+ * includes settings to override default {@link ai.timefold.solver.core.api.solver.Solver} configuration.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of a submitted problem, such as {@link Long} or {@link UUID}.
+ */
+public class SolverJobConfig<Solution_, ProblemId_> {
```
triceo, line 18: I understand why you decided to introduce it, but from the point of view of the user, this is boilerplate.
If you want to keep this class, please move it outside of `config`/`api` packages, where it becomes public API, and make sure public API does not reference it in any API contract.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -120,7 +123,8 @@ public Solver<Solution_> buildSolver() {
                 scoreDirectorFactory.getInitializingScoreTrend(),
                 solutionDescriptor,
                 ClassInstanceCache.create()).build();
-        var terminationConfig = Objects.requireNonNullElseGet(solverConfig.getTerminationConfig(), TerminationConfig::new);
+        var terminationConfig = ofNullable(configOverride).map(SolverConfigOverride::getTerminationConfig)
+                .orElseGet(() -> Objects.requireNonNullElseGet(solverConfig.getTerminationConfig(), TerminationConfig::new));
```
triceo, line 127: You don't know this yet, but we try to avoid `Optional` _inside_ code.

Optional is great for contracts, where it says things about nullity. But inside methods, it is pointless - null checks work just as well, and read easier. 

Don't be afraid of dealing with null directly, introduce an if statement. If you encapsulate this code in a new method, it will be more readable.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJobSession.java
```diff
@@ -0,0 +1,82 @@
...
+ * <p>
+ * To solve a planning problem, set the problem configuration: {@link #withProblemId(Object)},
+ * {@link #withProblemFinder(Function)} and {@link #withProblem(Object)}. Then solve it by calling {@link #run()}.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <ProblemId_> the ID type of submitted problem, such as {@link Long} or {@link UUID}.
+ *
+ * @see DefaultSolverManager
+ */
+public class DefaultSolverJobSession<Solution_, ProblemId_> implements SolverJobBuilder<Solution_, ProblemId_> {
```
triceo, line 29: If you're implementing `SolverJobBuilder`, typically we call the implementation `DefaultSolverJobBuilder`.

In this case, it is questionable whether we need the interface at all - why not put the make this the `SolverJobBuilder` directly? Make the constructors package-private and there are no issues.

core/core-impl/src/main/resources/solver.xsd
```diff
@@ -1435,6 +1435,18 @@
       
   </xs:complexType>
     
+  <xs:complexType name="solverConfigOverride">
```
triceo, line 1438: If done correctly, there will be no changes to this file. See other comments on this PR.

triceo (commented):
Good start! I left several comments, feel free to discuss if you disagree.

Some things not yet included in this PR:
- Tests for Quarkus and Spring Boot.
- Migration from the deprecated methods.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverFactory.java
```diff
@@ -115,6 +116,15 @@ static <Solution_> SolverFactory<Solution_> create(SolverConfig solverConfig) {
...
+        return this.buildSolver(null);
+    }
 
+    /**
+     * As defined by {@link #buildSolver()}.
+     *
+     * @param configOverride sometimes null, includes settings that override the default configuration
+     * @return never null
+     */
+    Solver<Solution_> buildSolver(SolverConfigOverride<Solution_> configOverride);
```
triceo, line 129: I question whether this is the right way. If you have the argument, and you also have an overload which does not need the argument, then this method should not accept `null`. You shouldn't have two ways of doing the same obvious thing.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverAndListenerJobFactory.java
triceo: Let's expose the override for now. We can always simplify later.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
triceo: I don't think that's correct.
As far as I remember, we agreed there would be:

- solve(...) with the most basic arguments.
- solveAndListen(...) with the most basic arguments.
- solveBuilder().

Everything else would be deprecated.

zepfred: Upon reviewing the code, I noticed that the method `solve(...)` already includes the `solveAndListen(...)` logic. The only necessary addition would be best consumer validation. Do you still think we should add the method `<Solution_, ProblemId_> solveAndListen(SolverJobConfig<Solution_, ProblemId_> solverJobConfig);`?

triceo: No. We `SolverJobConfig` should not be the argument of any method in the public API.

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverConfigOverride.java
zepfred: > It is questionable if this class belongs in `config` package. As you can see, it caused it to be needlessly inserted in the XSD. I'd move this to `api/solver`, which is where it is primarily used.
> 

Yes, I was unsure about how to deal with this XSD auto-generation, and it makes sense to move it to `api/solver.`

> Also, let's make it `final`. I am a big friend of removing freedom from people (in code); if you allow people to do something, in this case extending the class, they will. And then you have to live with that forever.
> 
> We can always allow it if we see the need in the future. We can never forbid it once we've allowed it.

Done!

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/SolverJobConfig.java
zepfred: I removed this class as it had become unused after updating the public API contract.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJobSession.java
zepfred: > If you're implementing `SolverJobBuilder`, typically we call the implementation `DefaultSolverJobBuilder`.

Done!

> In this case, it is questionable whether we need the interface at all - why not make this the `SolverJobBuilder` directly? Make the constructors package-private and there are no issues.

The interface belongs to the public API, and `DefaultSolverJobBuilder` needs to access `DefaultSolverManager` in order to submit the problem. So, it seems not correct to access `DefaultSolverManager` from the public API. Does that make sense?

triceo: I see your point. Technically, you're correct - it creates dependency from public API to private code, and it would be cleaner to separate the API from the impl as you've done.

That said, this (anti)pattern is already *everywhere* in the Solver, and often it can not even be avoided. Therefore I don't mind if we do it here as well. I don't mind if you want to keep it like that, but in that case I see there is a very long Javadoc on the class which should only be on the interface, as the user-facing description. (Javadocs on the implementation should only deal with the details of the implementation.)

core/core-impl/src/main/resources/solver.xsd
zepfred: Yes, I confirmed that no changes have been made to this file.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverFactory.java
zepfred: I have made changes to the logic to ensure that the configOverride is now required.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /**
+     * Sets the solver {@link TerminationConfig}.
+     *
+     * @param terminationConfig allows overriding the default termination config of {@link Solver}
+     * @return this, never null
+     */
+    public SolverConfigOverride<Solution_> withTerminationConfig(TerminationConfig terminationConfig) {
+        this.terminationConfig = terminationConfig;
```
triceo, line 26: Let's require non-null here.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.UUID;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+
+/**
+ * Provides a fluid contract that allows customization and submission of planning problems to solve.
```
triceo, line 11: `fluent`

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+ * <p>
+ * A {@link SolverManager} can solve multiple planning problems and can be used across different threads. The builder is
+ * responsible for constructing and solving a specific configuration problem. Hence, it is possible to
+ * have multiple distinct build configurations that are scheduled to run by the {@link SolverManager} implementation.
```
triceo, line 15: s/implementation/instance

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+ * <p>
+ * To solve a planning problem, set the problem configuration: {@link #withProblemId(Object)},
+ * {@link #withProblemFinder(Function)} and {@link #withProblem(Object)}. Then solve it by calling {@link #run()}.
```
triceo, line 18: Typically we put each sentence on a new line, to make the comment read more easily. If you install the Grazie plugin in your IDE, it will suggest where to put line breaks. (Among other things.)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param problemFinder never null, a function that returns a {@link PlanningSolution}, usually with uninitialized
+     *        planning variables
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> withProblemFinder(Function<? super ProblemId_, ? extends Solution_> problemFinder);
+
+    /**
+     * Sets the best solution consumer, which may be called multiple times during the solving process.
+     *
+     * @param bestSolutionConsumer called multiple times for each new best solution on a consumer thread
```
triceo, line 55: Does not having "never null" here mean we actually accept null here? Why would we?

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> withBestSolutionConsumer(Consumer<? super Solution_> bestSolutionConsumer);
+
+    /**
+     * Sets the final best solution consumer, which is called at the end of the solving process and return the final
+     * best solution.
+     *
+     * @param finalBestSolutionConsumer called only once at the end of the solving process on a consumer thread
```
triceo, line 64: Dtto.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Sets the best solution consumer, which may be called multiple times during the solving process.
+     *
+     * @param bestSolutionConsumer called multiple times for each new best solution on a consumer thread
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> withBestSolutionConsumer(Consumer<? super Solution_> bestSolutionConsumer);
+
+    /**
+     * Sets the final best solution consumer, which is called at the end of the solving process and return the final
```
triceo, line 61: return*s*

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param finalBestSolutionConsumer called only once at the end of the solving process on a consumer thread
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_>
+            withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
+
+    /**
+     * Sets the custom exception handler.
+     *
+     * @param exceptionHandler called if an exception or error occurs. If null it defaults to logging the exception as
```
triceo, line 73: Dtto.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+     *        an error.
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_>
+            withExceptionHandler(BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler);
+
+    /**
+     * Sets the solver config override.
+     *
+     * @param solverConfigOverride allows overriding the default behavior of {@link Solver}
```
triceo, line 83: Dtto.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverManager.java
```diff
@@ -118,10 +129,15 @@ static <Solution_, ProblemId_> SolverManager<Solution_, ProblemId_> create(
      * @param problemId never null, a ID for each planning problem. This must be unique.
      *        Use this problemId to {@link #terminateEarly(Object) terminate} the solver early,
      * @param problem never null, a {@link PlanningSolution} usually with uninitialized planning variables
+     * @deprecated It is recommended to use {@link #solveBuilder()}
      * @return never null
      */
+    @Deprecated(forRemoval = true)
```
triceo, line 135: For completeness, I recommend also adding `since = "1.6.0"`

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJobBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+    private Function<? super ProblemId_, ? extends Solution_> problemFinder;
+    private Consumer<? super Solution_> bestSolutionConsumer;
+    private Consumer<? super Solution_> finalBestSolutionConsumer;
+    private BiConsumer<? super ProblemId_, ? super Throwable> exceptionHandler;
+    private SolverConfigOverride<Solution_> solverConfigOverride;
+
+    public DefaultSolverJobBuilder(DefaultSolverManager<Solution_, ProblemId_> solverManager) {
+        this.solverManager = Objects.requireNonNull(solverManager, "The SolverManager (" + solverManager + ") cannot be null.");
+        // The config is required by SolverFactory and it is always initialized
+        this.solverConfigOverride = new SolverConfigOverride<>();
```
triceo, line 31: Personally, I'd create this instance when we need it. If the user doesn't provide one, create in on-demand in `run()`.

quarkus-integration/quarkus/integration-test/src/main/java/ai/timefold/solver/quarkus/it/TimefoldTestResource.java
```diff
@@ -39,4 +39,27 @@ public String solveWithSolverFactory() {
             throw new IllegalStateException("Solving failed.", e);
         }
     }
+
+    @POST
+    @Path("/solver-factory/builder")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String solveWithSolverBuilderFactory() {
```
triceo, line 46: The intention of this method was to test the override. Quarkus has some very special logic that deals with terminations (you'll find out very soon), and we want to ensure that if we provide the termination override, things work as expected.

quarkus-integration/quarkus/reflection-integration-test/src/main/java/ai/timefold/solver/quarkus/it/reflection/TimefoldTestResource.java
```diff
@@ -40,4 +40,28 @@ public String solveWithSolverFactory() {
             throw new IllegalStateException("Solving failed.", e);
         }
     }
+
+    @POST
+    @Path("/solver-factory/builder")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String solveWithSolverBuilderFactory() {
```
triceo, line 47: I think the first Quarkus test was enough.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfigurationTest.java
```diff
@@ -279,6 +279,30 @@ void solve() {
                 });
     }
 
+    @Test
+    void solveWithBuilder() {
```
triceo, line 283: Same comment as with Quarkus - please actually test the override.

triceo (commented):
Final batch of comments.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
zepfred: The only case I can think of is if we want to reuse a builder and update the configuration, which seems an anti-pattern.

zepfred:
@triceo, the PR is ready for your review.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorGizmoKitchenSinkTest.java
```diff
@@ -72,7 +72,10 @@ void solve() throws ExecutionException, InterruptedException {
                 Collections.emptyList(),
                 HardSoftLongScore.ZERO);
 
-        SolverJob<TestDataKitchenSinkSolution, Long> solverJob = solverManager.solve(1L, problem);
+        SolverJob<TestDataKitchenSinkSolution, Long> solverJob = solverManager.solveBuilder()
+                .withProblemId(1L)
+                .withProblem(problem)
+                .run();
```
triceo, line 78: Due to this PR, I realized this is used *a lot*. It seems to be *everywhere*. Maybe we un-deprecate this method on `SolverManager`?

triceo:
@zepfred Check out the `migration` module. We want to automate this migration for our users, so you need to implement it there. (You can then use the migration script on the quickstarts, to replace any uses of SolverManager deprecated methods, if there are any.)

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorGizmoKitchenSinkTest.java
zepfred: Agreed. Should we keep using the builder or revert changes for this method?

triceo: I know it's extra work, but I'd prefer reverting this method back.

quarkus-integration/quarkus/integration-test/src/test/java/ai/timefold/solver/quarkus/it/TimefoldTestResourceTest.java
```diff
@@ -27,4 +27,17 @@ void solveWithSolverFactory() throws Exception {
                         "0hard/5soft"));
     }
 
+    @Test
+    @Timeout(600)
+    void solveWithTimeOverride() throws Exception {
```
triceo, line 32: I got this error:

    [ERROR] ai.timefold.solver.quarkus.it.TimefoldTestResourceTest.solveWithTimeOverride -- Time elapsed: 10.98 s <<< FAILURE!
    java.lang.AssertionError:
    1 expectation failed.
    Response body doesn't match expectation.
    Expected: is "0hard/5soft,0.50"
    Actual: 0hard/5soft,0,50

Likely a locale difference between my computer and yours. Please make the test locale-independent.

triceo (commented):
LGTM!

Before merging:
- Please resolve the final comments.
- Make sure CI is green.
- Check docs and quickstarts for the use of the deprecated methods.

Issue #526, opened 2023-12-27
Score corruption analysis fails with ClassCastException because MatchAnalysis doesn


thimmwork:
**Describe the bug**
This probably wasn't detected earlier because it only occurs on actual score corruption:
When appending the analysis, the scoreDirector performs an unparameterized `Set<MatchAnalysis>.stream().sorted()` which would require MatchAnalysis to implement Comparable, but it doesn't.

**Expected behavior**
The score analysis should not fail.
The actually intended behavior is debatable. I'll provide a PR that removes sorting.
I assume that the sorting won't bring much value when analyzing the corruption, because you have to compare the sets of excess and missing constraint matches anyway.
Feel free to suggest otherwise.

**Actual behavior**
a `ClassCastException` is thrown with a stacktrace like this:
```
java.lang.ClassCastException: class ai.timefold.solver.core.api.score.analysis.MatchAnalysis cannot be cast to class java.lang.Comparable (ai.timefold.solver.core.api.score.analysis.MatchAnalysis is in unnamed module of loader 'app'; java.lang.Comparable is in module java.base of loader 'bootstrap')

	at java.base/java.util.Comparators$NaturalOrderComparator.compare(Comparators.java:47)
	at java.base/java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
	at java.base/java.util.TimSort.sort(TimSort.java:220)
	at java.base/java.util.Arrays.sort(Arrays.java:1308)
	at java.base/java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:353)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:528)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
[...]

PR #527, opened 2023-12-27
Issue #526: Fix ClassCastException of MatchAnalysis to Comparable

thimmwork:
when performing score corruption analysis with more than one excess/missing MatchAnalysis element.


triceo (approved):
LGTM, thank you @thimmwork!

PR #533, opened 2024-01-01
build: replace formatter and impsort with spotless

triceo:
One plugin to rule them all.

Follow-up:
https://github.com/TimefoldAI/timefold-solver-enterprise/pull/64


zepfred (approved):
Looks great to me!

zepfred:
@triceo, it looks like we need to add the goal `check` to [bind it ](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#binding-to-maven-phase)to the Maven `verify` phase.

zepfred:
> @triceo, it looks like we need to add the goal `check` to [bind it ](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#binding-to-maven-phase)to the Maven `verify` phase.

Please ignore my comment. The `verify` phase will also run the goal `apply.`

PR #536, opened 2024-01-02
fix: avoid exception when analysing uninitialized solution

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolutionManager.java
```diff
@@ -126,7 +126,7 @@ default ScoreAnalysis<Score_> analyze(Solution_ solution, ScoreAnalysisFetchPoli
      * Calculates and retrieves information about which constraints contributed to the solution's score.
      * This is a faster, JSON-friendly version of {@link #explain(Object)}.
      *
-     * @param solution never null, must be fully initialized otherwise an exception is thrown
+     * @param solution never null
```
triceo, line 129: Note: this is technically a change to the public API, but because we're *removing* an exception, it is allowed; this change will not break anyone.

zepfred (changes requested):
The refactoring seems correct, but I am missing unit tests to automatize the refactoring validation and avoid regressions. My suggestion is to update `ai.timefold.solver.core.impl.score.director.AbstractScoreDirectorSemanticsTest` and test all three new scopes: `DEFAULT`, `SCORE_CORRUPTION`, and `RECOMMENDATION_API`.

triceo:
@zepfred Test coverage added for the normal code paths.
Recommendation API and score analysis was already tested in other places.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -111,6 +120,96 @@ void solve() {
         assertThat(solution.getScore().isSolutionInitialized()).isTrue();
     }
 
+    @Test
```
zepfred, line 123: Nice!

core/core-impl/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -865,4 +964,74 @@ void solveWithMultipleChainedPlanningEntities() {
         assertThat(solution.getScore().isSolutionInitialized()).isTrue();
     }
 
+    public static class CorruptedEasyScoreCalculator implements EasyScoreCalculator<TestdataSolution, SimpleScore> {
```
zepfred, line 967: If only the method `calculatScore` is called during the solving process, then mocking it would make the test simpler.

triceo: True. But since I need to go through the entire Solver hierarchy, I need an actual `class` to pass to the `SolverConfig`. An mock instance of the class will not be enough. Therefore I decided to do it like this.

zepfred (approved):
Thanks for including the tests! I have one final suggestion regarding mocking the score classes.

PR #540, opened 2024-01-04
perf: Change records back to classes for GraalVM native image performance

Christopher-Chianelli:
GraalVM uses reflection!!! for records equals and hashCode (see https://github.com/oracle/graal/issues/4348), degrading performance by 100x compared to normal classes. Use normal classes instead of records for classes used during solving.

Draft as there some test failures and some improvements still to be made.

### DIFF
--- a/core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/index/SingleIndexProperties.java
+++ b/core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/index/SingleIndexProperties.java
@@ -1,6 +1,18 @@
 package ai.timefold.solver.constraint.streams.bavet.common.index;
 
-record SingleIndexProperties<A>(A property) implements IndexProperties {
+import java.util.Objects;
+
+final class SingleIndexProperties<A> implements IndexProperties {
+
+    private final A property;
+
+    public SingleIndexProperties(A property) {
+        this.property = property;
+    }
+
+    public A property() {
+        return property;
+    }
 
     @Override
     public <Type_> Type_ toKey(int id) {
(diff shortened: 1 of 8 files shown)


core/constraint-streams/src/main/java/ai/timefold/solver/constraint/streams/bavet/common/index/SingleIndexProperties.java
```diff
@@ -10,4 +22,18 @@ public <Type_> Type_ toKey(int id) {
...
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        SingleIndexProperties<?> that = (SingleIndexProperties<?>) o;
+        return Objects.equals(property, that.property);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(property);
```
triceo, line 37: This is an anti-pattern.

`Object.hash(...)` is a varargs method, and as such, every invocation creates a needless array that is immediately thrown away. This causes a major performance regression and should never be used on the hot path.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/util/Pair.java
```diff
@@ -8,6 +10,35 @@
  * @param <Key_>
  * @param <Value_>
  */
-public record Pair<Key_, Value_>(Key_ key, Value_ value) {
+public class Pair<Key_, Value_> {
```
triceo, line 13: This causes Javadoc plugin to fail. The params need to be on the constructor, not on the class. (As it's no longer a record with a canonical constructor.)

triceo (changes requested):
I am blocking this for now, pending an outcome of an internal discussion on the pros and cons.

triceo:
We agreed we wouldn't merge this as there is an upstream fix coming for this.

PR #546, opened 2024-01-08
feat: syntactic sugar for ScoreAnalysis

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
```diff
@@ -130,8 +128,17 @@ public ScoreAnalysis<Score_> diff(ScoreAnalysis<Score_> other) {
                             return ConstraintAnalysis.diff(constraintRef, constraintAnalysis, otherConstraintAnalysis);
                         },
                         (constraintRef, otherConstraintRef) -> constraintRef,
-                        TreeMap::new));
+                        HashMap::new));
```
zepfred, line 131: I wonder if this change could break existing user tests.

zepfred (approved):
LGTM!

core/core-impl/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
triceo: No. Because then the `ScoreAnalysis` constructor rebuilds the map anyway.
I made this change because sorting the map here was entirely pointless. :-)

PR #560, opened 2024-01-10
perf: reduce allocations in CS

triceo:
Various CS nodes could allocate gigabytes of `ElementAwareListIterator` per minute.
We refactor the code so that iterator instances are shared, removing this needless source of GC pressure.
Improves performance of some benchmarks by up to 6 %, while reducing their memory consumption by ~20 %.


rsynek (approved):
Nice optimization.

PR #564, opened 2024-01-12
feat: Support multiple SolverManager instances

zepfred:
This pull request updates the Quarkus extension to support multiple instances of Solver in an application. The new logic utilizes a map to keep track of the configurations, and nothing changes for the current behavior for single solvers, which are mapped by default to the key name `default`.

A new build step that executes at runtime has been included to fetch an updated properties instance from `TimefoldRuntimeConfig`. The reason we need to register the managed beans at runtime is because `Arc.container().instance()` does not return an instance of `TimefoldRuntimeConfig` when using interfaces instead of classes. Defining configuration properties as interfaces is the only way to use `@WithUnnamedKey`. This is the default approach documented in both Quarkus and Smallrye pages.

When using multiple solver properties, only the `SolverManager` can be injected by default. Also, the `SolverManager` can be injected using the following parameterized types because the `ProblemId_` domain is not restricted.

- String
- Short
- Integer
- Long
- Float
- Double


zepfred:
@triceo I am planning to submit a separate PR to update the documentation.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolverPropertiesTest.java
```diff
@@ -0,0 +1,312 @@
...
+    ScoreManager<TestdataQuarkusSolution, HardMediumSoftBigDecimalScore> hardMediumSoftBigDecimalScoreManager1;
+    @Inject
+    @Named("solver1ScoreManager")
+    ScoreManager<TestdataQuarkusSolution, BendableScore> bendableScoreManager1;
+    @Inject
+    @Named("solver1ScoreManager")
+    ScoreManager<TestdataQuarkusSolution, BendableLongScore> bendableLongScoreManager1;
+    @Inject
+    @Named("solver1ScoreManager")
+    ScoreManager<TestdataQuarkusSolution, BendableBigDecimalScore> bendableBigDecimalScoreManager1;
```
triceo, line 108: `ScoreManager` is deprecated. We need not and should not support it in this new feature.
Also, aren't we missing the `SolverManager` here?

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolverPropertiesTest.java
```diff
@@ -0,0 +1,312 @@
...
+            .overrideConfigKey("quarkus.timefold.solver.\"solver1\".termination.best-score-limit", "0")
+            .overrideConfigKey("quarkus.timefold.solver.\"solver2\".daemon", "false")
+            .overrideConfigKey("quarkus.timefold.solver.\"solver2\".domain-access-type", "REFLECTION")
+            .overrideConfigKey("quarkus.timefold.solver.\"solver2\".termination.spent-limit", "1h")
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(TestdataQuarkusEntity.class, TestdataQuarkusSolution.class,
+                            TestdataQuarkusConstraintProvider.class));
+
+    @Inject
+    @Named("solver1Config")
```
triceo, line 63: Are the `Config`/`Factory`/`Manager` suffixes necessary?
They are boilerplate, I'd like to avoid them.

quarkus-integration/quarkus-benchmark/deployment/src/main/java/ai/timefold/solver/benchmark/quarkus/deployment/TimefoldBenchmarkProcessor.java
```diff
@@ -66,11 +63,35 @@ void registerAdditionalBeans(BuildProducer<AdditionalBeanBuildItem> additionalBe
...
+     * {@link ai.timefold.solver.quarkus.config.TimefoldRuntimeConfig}.
+     * <p>
+     * The reason we need to register the managed beans at runtime is because {@code Arc.container().instance()} does
+     * not return an instance of {@link ai.timefold.solver.quarkus.config.TimefoldRuntimeConfig} when using interfaces
+     * instead of classes. Defining configuration properties as interfaces is the only way to use {@code @WithUnnamedKey}.
+     * This is the default approach documented in both Quarkus and Smallrye pages.
+     * <p>
+     * Finally, recording the bean at runtime is necessary to use updated instances of configuration properties, and
+     * annotating {@link ai.timefold.solver.quarkus.config.TimefoldRuntimeConfig} with {@code @StaticInitSafe} has no effect.
+     */
```
triceo, line 83: This seems like a defensive comment. You're justifying why you did it like this, as if someone was going to say you've done it wrong.

In my opinion, after this code is merged, this comment no longer has any value. We should remove it.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/SolverBuildTimeConfig.java
```diff
@@ -18,37 +17,40 @@
  * See also {@link SolverRuntimeConfig}
  */
 @ConfigGroup
-public class SolverBuildTimeConfig {
+public interface SolverBuildTimeConfig {
```
triceo, line 20: Have you made sure none of the pre-existing behavior have changed? (In other words: is this 100 % backwards compatible?)

Have you ensured that for the single instance, the config is still properly overriden when using the config XML as well as these properties?

Have you ensured the same with the benchmark config?

triceo (commented):
I am providing initial feedback, mostly in the form of questions.

I don't yet understand the approach well enough, but I think I already see enough to understand that the approach may have to change.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolverPropertiesTest.java
zepfred: Okay. I'll remove them. The `SolveManager` instances are injected at the lines 70 and 156.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolverPropertiesTest.java
zepfred: I wasn't happy either. I did it because of issues with duplicated names when registering the synthetic beans. Let me double-check that and ensure that I haven't overlooked anything.

zepfred: The new approach now aligns with the decision to inject only SolverManager for multiple solvers.

quarkus-integration/quarkus-benchmark/deployment/src/main/java/ai/timefold/solver/benchmark/quarkus/deployment/TimefoldBenchmarkProcessor.java
zepfred: The goal was to document the design, but I am fine with discarding it.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/SolverBuildTimeConfig.java
zepfred: In short, we have integration tests to validate all changes regarding the default/old and new behaviors.

Also, I've used the quickstart project to validate the changes, including the benchmark.

quarkus-integration/quarkus/runtime/src/main/java/ai/timefold/solver/quarkus/config/TimefoldRuntimeConfig.java
```diff
@@ -1,24 +1,39 @@
...
+
+    String DEFAULT_SOLVER_NAME = "default";
 
-@ConfigRoot(name = "timefold", phase = ConfigPhase.RUN_TIME)
-public class TimefoldRuntimeConfig {
     /**
-     * During run time, this is translated into Timefold's {@link SolverConfig}
-     * runtime properties.
+     * During run time, this is translated into Timefold's {@link SolverConfig} runtime properties per solver. If a solver
+     * name is not explicitly specified, the solver name will default to {@link #DEFAULT_SOLVER_NAME}.
```
triceo, line 25:
```suggestion
     * During run time, this is translated into Timefold's {@link SolverConfig} runtime properties per solver. 
     * If a solver name is not explicitly specified, the solver name will default to {@link #DEFAULT_SOLVER_NAME}.
```

quarkus-integration/quarkus-benchmark/deployment/src/main/java/ai/timefold/solver/benchmark/quarkus/deployment/TimefoldBenchmarkProcessor.java
```diff
@@ -40,16 +40,20 @@ HotDeploymentWatchedFileBuildItem watchSolverBenchmarkConfigXml() {
...
-            SolverConfigBuildItem solverConfigBuildItem,
-            TimefoldBenchmarkRecorder recorder) {
-        if (solverConfigBuildItem.getSolverConfig() == null) {
+            SolverConfigBuildItem solverConfigBuildItem) {
+        // We don't support benchmarking for multiple solvers
+        if (solverConfigBuildItem.getAllSolverConfigurations().size() > 1) {
+            throw new ConfigurationException("When defining multiple solvers, the benchmark feature is not enabled. " +
+                    "Consider using separate <solverBenchmark> instances for evaluating different solver configurations. " +
+                    "You can find additional information by visiting the Timefold documentation page and navigating to the " +
+                    "\"Benchmarking and tweaking\" section.");
```
triceo, line 51:
```suggestion
            throw new ConfigurationException("When defining multiple solvers, the benchmark feature is not enabled. " +
                    "Consider using separate <solverBenchmark> instances for evaluating different solver configurations.");
```

I wouldn't go so far as to recommend specific sections of the documentation.
Documentation changes, sections move around, titles are renamed... these comments very quickly get obsolete.

Also, please use raw string (`"""`) and put each sentence on its own line. That way, you get to avoid the ugly string contatenation where it's not necessary.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/SolverBuildTimeConfig.java
```diff
@@ -18,37 +19,53 @@
...
 
     /**
      * What constraint stream implementation to use. Defaults to {@link ConstraintStreamImplType#BAVET}.
      *
      * @deprecated Not used anymore.
      */
-    @ConfigItem
     @Deprecated(forRemoval = true, since = "1.4.0")
-    public Optional<ConstraintStreamImplType> constraintStreamImplType;
+    Optional<ConstraintStreamImplType> constraintStreamImplType();
```
triceo, line 69: Any chance to not include the deprecated field for the new config, only keep it for the old way? (If it's too hard to split the two, or requires code duplication, forget about it.)

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/TimefoldBuildTimeConfig.java
```diff
@@ -1,32 +1,47 @@
...
      * If this property isn't specified, that solverConfig.xml is optional.
      */
-    @ConfigItem
-    public Optional<String> solverConfigXml;
+    Optional<String> solverConfigXml();
 
     /**
-     * Configuration properties that overwrite Timefold's {@link SolverConfig}.
+     * Configuration properties that overwrite Timefold's {@link SolverConfig} per Solver. If a solver name is not
+     * explicitly specified, the solver name will default to {@link #DEFAULT_SOLVER_NAME}.
```
triceo, line 32:
```suggestion
     * Configuration properties that overwrite the {@link SolverConfig} per Solver. 
     * If a solver name is not explicitly specified, the solver name will default to {@link #DEFAULT_SOLVER_NAME}.
```

When writing about Timefold, please remember:

- Timefold is the company.
- Timefold Solver is the product.

In this case, no need to mention Timefold at all; everyone knows that it's inside the Timefold Solver code.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorSingleDefaultSolverPropertiesTest.java
```diff
@@ -0,0 +1,115 @@
...
+import ai.timefold.solver.quarkus.testdata.normal.domain.TestdataQuarkusSolution;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+class TimefoldProcessorSingleDefaultSolverPropertiesTest {
```
triceo, line 40: This test - a test for the old behavior - should still test the deprecated `ScoreManager`. It's only the new `@Named` behavior where we don't support it.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/rest/TimefoldProcessorHotReloadMultipleSolversTest.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Do not remove the public modifier from this class. Otherwise, the test will fail with the following:
+ * <p>
+ * Caused by: java.lang.IllegalAccessException: class io.quarkus.test.QuarkusDevModeTest cannot access a member of class
+ * ai.timefold.solver.quarkus.rest.TimefoldProcessorHotReloadMultipleSolversTest with modifiers ""
+ * at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
+ * at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:489)
+ * at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
+ * at io.quarkus.test.QuarkusDevModeTest.createTestInstance(QuarkusDevModeTest.java:222)
+ */
```
triceo, line 30: No need to be so verbose. Enough to say that removing public will cause an exception. If someone decides to try even though they've been warned, they'll find out what exception it is. :-)

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/rest/TimefoldProcessorHotReloadTest.java
```diff
@@ -1,21 +1,32 @@
...
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import io.quarkus.test.QuarkusDevModeTest;
 import io.restassured.RestAssured;
 
+/**
+ * Do not remove the public modifier from this class. Otherwise, the test will fail with the following:
```
triceo, line 18: Dtto. I'm even considering if we need to warn about this at all; we typically don't.

quarkus-integration/quarkus/devui-integration-test/src/test/java/ai/timefold/solver/quarkus/it/devui/TimefoldDevUIMultipleSolversTest.java
```diff
@@ -0,0 +1,111 @@
...
+                            + "  </scoreDirectorFactory>\n"
+                            + "  <termination/>\n"
+                            + "</solver>");
+        });
+    }
+
+    @Test
+    void testModelPage() throws Exception {
+        JsonNode modelResponse = super.executeJsonRPCMethod("getModelInfo");
+        List.of("solver1", "solver2").forEach(key -> {
```
triceo, line 82: For sake of simplicity, I'd refactor this to avoid the lambda.
Make the assertions be inside a method, have the string as an argument to the method, call the method twice.

The stack trace in case of a failure will be nicer, and more self-explanatory.

quarkus-integration/quarkus/devui-integration-test/src/test/java/ai/timefold/solver/quarkus/it/devui/TimefoldDevUIMultipleSolversTest.java
```diff
@@ -0,0 +1,111 @@
...
+            assertThat(modelResponse.get(key).get("entityClassToShadowVariableListMap")
+                    .get(TestdataStringLengthShadowEntity.class.getCanonicalName()))
+                    .containsExactly(new TextNode("length"));
+        });
+    }
+
+    @Test
+    void testConstraintsPage() throws Exception {
+        JsonNode constraintsResponse = super.executeJsonRPCMethod("getConstraints");
+        List.of("solver1", "solver2").forEach(key -> {
```
triceo, line 102: Dtto.

triceo (commented):
Adding some comments, overall looks good.

triceo:
Out of curiosity, I wanted to check out the dev UI and I got this:

Object?

The other tabs (model / constraints) are empty.
No exception in the log.

I used the `employee-scheduling` quickstart.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/SolverBuildTimeConfig.java
zepfred: Can you please provide more details regarding this comment?

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorSingleDefaultSolverPropertiesTest.java
zepfred: I have completely removed the support for it and need to revert those changes and bring back the support.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/rest/TimefoldProcessorHotReloadTest.java
zepfred: The comment was simplified. It took me some time to figure out the test was failing because of the missing modifier, and I think having it is a good idea.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorSingleDefaultSolverPropertiesTest.java
triceo: I'm sorry if I wasn't clear enough earlier, my bad. So, just to be completely safe:

- The old behavior with default solver config needs to support `ScoreManager` - it is deprecated, but not removed, so this would've broken backwards compatibility.
- But the new `@Named` functionality - on account of being brand new - can forget about `ScoreManager`.

I hope it makes sense.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/config/SolverBuildTimeConfig.java
triceo: This option is deprecated, which means we don't want new code and new users to be seeing it/using it.

For the "old" default solver config, we need to keep this option, because removing it would break backwards compatibility.
But technically, for the `@Named` configs (which are a new feature), we have no backwards compatibility to worry about, and therefore we can possibly expose an interface to those which does not include this deprecated option.

That's in theory. In practice, I worry this would've caused a lot of code duplication or otherwise some kind of a hack, so if that's the case, let's keep it as is and forget about this comment.

zepfred:
> > Also, the SolverManager can be injected using the following parameterized types because the ProblemId_ domain is not restricted.
> 
> Please explain what you mean by this. Give a few examples of what is not possible to do. That way, we can discuss if that's OK, or if we need to do something more about it.

In the original approach, the accepted `ProblemId_` types were considered required to create a synthetic bean. Otherwise, it would fail to locate the managed bean. Let's consider the synthetic bean definition below:
```
SyntheticBeanBuildItem.configure(SolverManager.class)
  .scope(Singleton.class)
  .supplier(recorder.solverManager(key, value, runtimeConfig,
      GizmoMemberAccessorEntityEnhancer.getGeneratedGizmoMemberAccessorMap(recorderContext,
         solverConfigBuildItem
             .getGeneratedGizmoClasses().generatedGizmoMemberAccessorClassSet),
      GizmoMemberAccessorEntityEnhancer.getGeneratedSolutionClonerMap(recorderContext,
            solverConfigBuildItem
             .getGeneratedGizmoClasses().generatedGizmoSolutionClonerClassSet)))
  .setRuntimeInit()
  .named(key)
  .done();
```

Defining a bean without a parameterized type would cause an injection to fail because the problem type is not solved during injection time.
```
@Inject
SolverManager<Solution,String> manager;
```

After conducting further investigation, I discovered a method to accept any type of `Problem_id` by using `TypeVariable.create(Object.class.getName())`.
```
SyntheticBeanBuildItem.configure(SolverManager.class)
  .scope(Singleton.class)
  .addType(ParameterizedType.create(DotName.createSimple(SolverManager.class.getName()),
          Type.create(DotName.createSimple(value.getSolutionClass().getName()),
                 Type.Kind.CLASS),
           TypeVariable.create(Object.class.getName())))
   .supplier(recorder.solverManager(key, value, runtimeConfig,
              GizmoMemberAccessorEntityEnhancer.getGeneratedGizmoMemberAccessorMap(recorderContext,
                    solverConfigBuildItem
                        .getGeneratedGizmoClasses().generatedGizmoMemberAccessorClassSet),
              GizmoMemberAccessorEntityEnhancer.getGeneratedSolutionClonerMap(recorderContext,
                   solverConfigBuildItem
                       .getGeneratedGizmoClasses().generatedGizmoSolutionClonerClassSet)))
  .setRuntimeInit()
  .named(key)
  .done()
```

zepfred:
> Out of curiosity, I wanted to check out the dev UI and I got this:
> 
> Object?
> 
> The other tabs (model / constraints) are empty. No exception in the log.
> 
> I used the `employee-scheduling` quickstart.

Since this is only available in dev mode, I have modified the JSON structure to include multiple solvers. Let me check the UI.

zepfred:
@triceo the DevUI components are fixed.

## Default behavior
config1
config3
config5

# New behavior
config2
config4
config6

zepfred:
@triceo, the DevUI changes are done!

cngotho:
@zepfred and @triceo much appreciation for this effort. I have question with regard to this effort which I have not understood well. Will this capability of having multiple named solvers allow the capability to have multiple @PlanningSolution and ConstraintProvider implementations for the different solver configurations?

I see from this code snippet from TimefoldProcessor will not allow more than one @PlanningSolution. The example solverConfig.xml files `solver1` and `solver2` the same entries have been used for both. 

```
TimefoldProcessor.java@489
   private Class<?> findSolutionClass(IndexView indexView) {
        Collection<AnnotationInstance> annotationInstances = indexView.getAnnotations(DotNames.PLANNING_SOLUTION);
        if (annotationInstances.size() > 1) {
            throw new IllegalStateException("Multiple classes (" + convertAnnotationInstancesToString(annotationInstances)
                    + ") found with a @" + PlanningSolution.class.getSimpleName() + " annotation.");
        }
        ....
    }    
```

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/rest/TimefoldProcessorHotReloadTest.java
```diff
@@ -27,11 +28,11 @@ public class TimefoldProcessorHotReloadTest {
     @Test
     void solverConfigHotReload() {
```
triceo, line 29: This test (and its `MultipleSolvers` counterpart) seems to throw a lot of exceptions into the sysout, even though it is passing. Let's fix that.

triceo:
@cngotho Thanks for the feedback. 

The fail-fasts you see exist in cases where there are multiple solution classes / constraint providers, neither of which is attached to any solver config. This makes them ambiguous and we *have to* fail in that case. 

We will not fail if you correctly specify which solver config should use which of those classes. We have test coverage for that, and we will improve the error messages to make this more obvious.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -229,18 +230,27 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
         ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
 
         Map<String, SolverConfig> allSolverConfig = new HashMap<>();
```
triceo, line 232: According to the convention, this would be a `solverConfigMap`.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -268,9 +278,79 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
+            throw new IllegalStateException(
+                    "No classes found with a @%s annotation.".formatted(PlanningSolution.class.getSimpleName()));
+        }
+        // Multiple classes and single solver
+        if (annotationInstances.size() > 1 && allSolverConfig.size() == 1) {
+            throw new IllegalStateException("Multiple classes (%s) found with a @%s annotation.".formatted(
+                    convertAnnotationInstancesToString(annotationInstances), PlanningSolution.class.getSimpleName()));
+        }
+        // Multiple classes and at least one solver config does not specify the solution class
+        List<String> emptySolutionClasses = allSolverConfig.entrySet().stream()
```
triceo, line 295: `emptySolutionClassList`

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -268,9 +278,79 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
+                            Multiple classes (%s) found with a @%s annotation and some solver configs ([%s]) does not specify it.
+                            Maybe set the XML config file to the related solver configs, or add the missing solution classes to the XML files,
+                            or remove the unnecessary solution classes from the classpath.
+                            """
+                            .formatted(convertAnnotationInstancesToString(annotationInstances),
+                                    PlanningSolution.class.getSimpleName(),
+                                    String.join(", ", emptySolutionClasses)));
+        }
+        // Unused solution classes
+        List<String> unusedSolutionClasses = annotationInstances.stream()
```
triceo, line 311: `unusedSolutionClassList`

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -268,9 +278,79 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
+                        .anyMatch(
+                                c -> c.getSolutionClass().getName().equals(planningClass.target().asClass().name().toString())))
+                .map(AnnotationInstance::target)
+                .filter(t -> t.kind() != AnnotationTarget.Kind.CLASS)
+                .map(t -> t.asClass().name().toString())
+                .toList();
+        if (!invalidTargetClasses.isEmpty()) {
+            throw new IllegalStateException("The targets classes ([%s]) with a @%s must be a class."
+                    .formatted(String.join(", ", invalidTargetClasses), PlanningSolution.class.getSimpleName()));
+        }
```
triceo, line 333: What is a target? Nothing in the solver is called a target. The exception should not mention a "target" because the user does not know what that is.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -268,9 +278,79 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
+        // Multiple classes and at least one solver config does not specify the solution class
+        List<String> emptySolutionClasses = allSolverConfig.entrySet().stream()
+                .filter(e -> e.getValue().getSolutionClass() == null)
+                .map(Map.Entry::getKey)
+                .toList();
+        if (annotationInstances.size() > 1 && !emptySolutionClasses.isEmpty()) {
+            throw new IllegalStateException(
+                    """
+                            Multiple classes (%s) found with a @%s annotation and some solver configs ([%s]) does not specify it.
```
triceo, line 302: I'd rephrase the first sentence. "Some solver configs (...) don't specify a @PlanningSolution class, yet there are multiple available (...) on the classpath."

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -268,9 +278,79 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
+    private void assertSolverConfigEntityClasses(IndexView indexView) {
+        Collection<AnnotationInstance> annotationInstances = indexView.getAnnotations(DotNames.PLANNING_ENTITY);
+        if (annotationInstances.isEmpty()) {
+            throw new IllegalStateException(
+                    "No classes found with a @%s annotation.".formatted(PlanningEntity.class.getSimpleName()));
+        }
+        List<AnnotationTarget> targetList = annotationInstances.stream()
+                .map(AnnotationInstance::target)
+                .toList();
+        if (targetList.stream().anyMatch(target -> target.kind() != AnnotationTarget.Kind.CLASS)) {
```
triceo, line 345: Targets again.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
zepfred: The validation process involves checking the Java type of the solution class, which uses the `AnnotationInstance#target()` method. The target word comes from this method and it will be removed from the error message.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -121,10 +127,17 @@ private static void registerSpi(Class<?> serviceClass, BuildProducer<ServiceProv
     }
 
     @BuildStep
-    HotDeploymentWatchedFileBuildItem watchSolverConfigXml() {
-        String solverConfigXML = timefoldBuildTimeConfig.solverConfigXml
+    void watchSolverConfigXml(BuildProducer<HotDeploymentWatchedFileBuildItem> hotDeploymentWatchedFiles) {
+        String solverConfigXML = timefoldBuildTimeConfig.solverConfigXml()
                 .orElse(TimefoldBuildTimeConfig.DEFAULT_SOLVER_CONFIG_URL);
-        return new HotDeploymentWatchedFileBuildItem(solverConfigXML);
+        Set<String> solverConfigXMLFiles = new HashSet<>();
```
triceo, line 133: Please be consistent; `solverCongigXmlFileSet`.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -225,29 +218,330 @@ SolverConfigBuildItem recordAndRegisterBeans(TimefoldRecorder recorder, Recorder
                     + "application.properties entries (quarkus.index-dependency.<name>.group-id"
                     + " and quarkus.index-dependency.<name>.artifact-id).");
             additionalBeans.produce(new AdditionalBeanBuildItem(UnavailableTimefoldBeanProvider.class));
-            return new SolverConfigBuildItem(null);
+            Map<String, SolverConfig> solverConfig = new HashMap<>();
```
triceo, line 221: `solverConfigMap`

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -225,29 +218,330 @@ SolverConfigBuildItem recordAndRegisterBeans(TimefoldRecorder recorder, Recorder
...
+            }
+
+            if (!declaringClass.annotationsMap().containsKey(DotNames.PLANNING_ENTITY)) {
+                throw new IllegalStateException(prefix + "with a @" +
+                        annotationInstance.name().withoutPackagePrefix() +
+                        " annotation is in a class (" + declaringClass.name()
+                        + ") that does not have a @" + PlanningEntity.class.getSimpleName() +
+                        " annotation.\n" +
+                        "Maybe add a @" + PlanningEntity.class.getSimpleName() +
+                        " annotation on the class (" + declaringClass.name() + ").");
```
triceo, line 315: And here as well.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -225,29 +218,330 @@ SolverConfigBuildItem recordAndRegisterBeans(TimefoldRecorder recorder, Recorder
...
+                .filter(e -> e.getValue().getSolutionClass() == null)
+                .map(Map.Entry::getKey)
+                .toList();
+        if (annotationInstances.size() > 1 && !solverConfigWithoutSolutionClassList.isEmpty()) {
+            throw new IllegalStateException(
+                    """
+                            Some solver configs (%s) don't specify a %s class, yet there are multiple available (%s) on the classpath.
+                            Maybe set the XML config file to the related solver configs, or add the missing solution classes to the XML files,
+                            or remove the unnecessary solution classes from the classpath.
+                            """
```
triceo, line 341:
```suggestion
                            or remove the unnecessary solution classes from the classpath."""
```

Be mindful that, if the `"""` is on a new line, the exception message will include an empty line. You don't want that.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -225,29 +218,330 @@ SolverConfigBuildItem recordAndRegisterBeans(TimefoldRecorder recorder, Recorder
...
+                            .formatted(String.join(", ", solverConfigWithoutSolutionClassList),
+                                    PlanningSolution.class.getSimpleName(),
+                                    convertAnnotationInstancesToString(annotationInstances)));
+        }
+        // Unused solution classes
+        List<String> unusedSolutionClasses = annotationInstances.stream()
```
triceo, line 347: `unusedSolutionClassNameSet`.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -225,29 +218,330 @@ SolverConfigBuildItem recordAndRegisterBeans(TimefoldRecorder recorder, Recorder
...
+            throw new IllegalStateException(errorMessage.formatted(
+                    incrementalScoreClasses.stream().map(c -> c.name().toString()).collect(Collectors.joining(", ")),
+                    IncrementalScoreCalculator.class.getSimpleName()));
+        }
+        // Multiple classes and at least one solver config does not specify the score class
+        errorMessage = """
+                Some solver configs (%s) don't specify a %s score class, yet there are multiple available (%s) on the classpath.
+                Maybe set the XML config file to the related solver configs, or add the missing score classes to the XML files,
+                or remove the unnecessary score classes from the classpath.
+                """;
```
triceo, line 408: Another newline at the end.

triceo (approved):
LGTM, with some minor comments.
Will merge after addressed.

PR #590, opened 2024-01-24
feat: Support multiple SolverManager instances (Spring-boot)

zepfred:
This pull request updates the Spring boot extension to support multiple instances of Solver in an application. The new logic utilizes a map to keep track of the configurations, and nothing changes for the current behavior for single solvers, which are mapped by default to the key name `default`.

### DIFF
--- a/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
+++ b/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
@@ -1,0 +1,115 @@
+package ai.timefold.solver.spring.boot.autoconfigure;
+
+import static java.util.Collections.emptyList;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+
+import org.springframework.beans.factory.annotation.AnnotatedBeanDefinition;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.boot.autoconfigure.AutoConfigurationPackages;
+import org.springframework.boot.autoconfigure.domain.EntityScanPackages;
+import org.springframework.boot.autoconfigure.domain.EntityScanner;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.core.type.filter.AssignableTypeFilter;
+import org.springframework.util.ClassUtils;
+
+public class IncludeAbstractClassesEntityScanner extends EntityScanner {
+
+    private final ApplicationContext context;
+
+    public IncludeAbstractClassesEntityScanner(ApplicationContext context) {
+        super(context);
+        this.context = context;
+    }
+
+    public <T> Class<? extends T> findFirstImplementingClass(Class<T> targetClass) {
+        return Optional.ofNullable(findImplementingClassList(targetClass)).filter(classes -> !classes.isEmpty())
+                .map(classes -> classes.get(0)).orElse(null);
+    }
+
+    public <T> List<Class<? extends T>> findImplementingClassList(Class<T> targetClass) {
+        if (!AutoConfigurationPackages.has(context)) {
+            return emptyList();
+        }
+        ClassPathScanningCandidateComponentProvider scanner = new ClassPathScanningCandidateComponentProvider(false);
+        scanner.setEnvironment(context.getEnvironment());
+        scanner.setResourceLoader(context);
+        scanner.addIncludeFilter(new AssignableTypeFilter(targetClass));
+
+        EntityScanPackages entityScanPackages = EntityScanPackages.get(context);
+
+        Set<String> packages = new HashSet<>();
+        packages.addAll(AutoConfigurationPackages.get(context));
+        packages.addAll(entityScanPackages.getPackageNames());
+        return packages.stream()
+                .flatMap(basePackage -> scanner.findCandidateComponents(basePackage).stream())
+                // findCandidateComponents can return the same package for different base packages
+                .distinct()
+                .sorted(Comparator.comparing(BeanDefinition::getBeanClassName))
+                .map(candidate -> {
+                    try {
+                        return (Class<? extends T>) ClassUtils.forName(candidate.getBeanClassName(), context.getClassLoader())
+                                .asSubclass(targetClass);
+                    } catch (ClassNotFoundException e) {
+                        throw new IllegalStateException("The " + targetClass.getSimpleName() + " class ("
+                                + candidate.getBeanClassName() + ") cannot be found.", e);
+                    }
+                })
+                .collect(Collectors.toList());
+    }
+
+    public boolean hasSolutionOrEntityClasses() {
+        try {
+            return !scan(PlanningSolution.class).isEmpty() || !scan(PlanningEntity.class).isEmpty();
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningSolution.class.getSimpleName()
+                    + " and @" + PlanningEntity.class.getSimpleName() + " annotations failed.", e);
+        }
+    }
+
+    public Class<?> findFirstSolutionClass() {
+        Set<Class<?>> solutionClassSet;
+        try {
+            solutionClassSet = scan(PlanningSolution.class);
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningSolution.class.getSimpleName()
+                    + " annotations failed.", e);
+        }
+        return solutionClassSet.iterator().next();
+    }
+
+    public List<Class<?>> findEntityClassList() {
+        Set<Class<?>> entityClassSet;
+        try {
+            entityClassSet = scan(PlanningEntity.class);
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningEntity.class.getSimpleName() + " failed.", e);
+        }
+        return new ArrayList<>(entityClassSet);
+    }
+
+    @Override
+    protected ClassPathScanningCandidateComponentProvider
+            createClassPathScanningCandidateComponentProvider(ApplicationContext context) {
+        return new ClassPathScanningCandidateComponentProvider(false) {
+            @Override
+            protected boolean isCandidateComponent(AnnotatedBeanDefinition beanDefinition) {
+                AnnotationMetadata metadata = beanDefinition.getMetadata();
+                // Do not exclude abstract classes nor interfaces
+                return metadata.isIndependent();
+            }
+        };
+    }
+
+}
--- a/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
+++ b/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
@@ -83,7 +83,7 @@ public class IncludeAbstractClassesEntityScanner extends EntityScanner {
     }
 
     @SafeVarargs
-    public final List<Class<?>> findAnnotationsWithRepeatable(Class<? extends Annotation>... annotations) {
+    public final List<Class<?>> findClassesWithAnnotation(Class<? extends Annotation>... annotations) {
         if (!AutoConfigurationPackages.has(context)) {
             return emptyList();
         }
--- a/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
+++ b/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
@@ -1,12 +1,19 @@
 package ai.timefold.solver.spring.boot.autoconfigure;
 
+import static ai.timefold.solver.spring.boot.autoconfigure.util.LambdaUtils.rethrowFunction;
 import static java.util.Collections.emptyList;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Comparator;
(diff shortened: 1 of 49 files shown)


spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.boot.autoconfigure.AutoConfigurationPackages;
+import org.springframework.boot.autoconfigure.domain.EntityScanPackages;
+import org.springframework.boot.autoconfigure.domain.EntityScanner;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.core.type.filter.AssignableTypeFilter;
+import org.springframework.util.ClassUtils;
+
+public class IncludeAbstractClassesEntityScanner extends EntityScanner {
```
triceo, line 27: Why was introducing this necessary?
What abstract classes do we need to find?

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+    private final ApplicationContext context;
+
+    public IncludeAbstractClassesEntityScanner(ApplicationContext context) {
+        super(context);
+        this.context = context;
+    }
+
+    public <T> Class<? extends T> findFirstImplementingClass(Class<T> targetClass) {
+        return Optional.ofNullable(findImplementingClassList(targetClass)).filter(classes -> !classes.isEmpty())
+                .map(classes -> classes.get(0)).orElse(null);
```
triceo, line 38: Let's avoid `Optional` if the only purpose of it is to avoid a local `null` check.

I'd argue the following code reads much better:

    var implementingClassList = findImplementingClassList(targetClass);
    if (implementingClassList == null || implementingClassList.isEmpty()) {
        return null;
    } else {
        return implementingClassList.get(0);
    }
    
`Optional` is useful to communicate nullness information with outside code. But if we can keep the `null` local inside of a method, using `Optional` is just a more difficult way to write a `null` check.

Also, why are we bothering with `null` at all? `findImplementingClassList(...)` doesn't seem to ever return `null`.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+            return emptyList();
+        }
+        ClassPathScanningCandidateComponentProvider scanner = new ClassPathScanningCandidateComponentProvider(false);
+        scanner.setEnvironment(context.getEnvironment());
+        scanner.setResourceLoader(context);
+        scanner.addIncludeFilter(new AssignableTypeFilter(targetClass));
+
+        EntityScanPackages entityScanPackages = EntityScanPackages.get(context);
+
+        Set<String> packages = new HashSet<>();
```
triceo, line 52: `packageSet`.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+                .flatMap(basePackage -> scanner.findCandidateComponents(basePackage).stream())
+                // findCandidateComponents can return the same package for different base packages
+                .distinct()
+                .sorted(Comparator.comparing(BeanDefinition::getBeanClassName))
+                .map(candidate -> {
+                    try {
+                        return (Class<? extends T>) ClassUtils.forName(candidate.getBeanClassName(), context.getClassLoader())
+                                .asSubclass(targetClass);
+                    } catch (ClassNotFoundException e) {
+                        throw new IllegalStateException("The " + targetClass.getSimpleName() + " class ("
```
triceo, line 65: Please apply the new convention to formatting error messages.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+                    }
+                })
+                .collect(Collectors.toList());
+    }
+
+    public boolean hasSolutionOrEntityClasses() {
+        try {
+            return !scan(PlanningSolution.class).isEmpty() || !scan(PlanningEntity.class).isEmpty();
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningSolution.class.getSimpleName()
```
triceo, line 76: Dtto.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+                    + " and @" + PlanningEntity.class.getSimpleName() + " annotations failed.", e);
+        }
+    }
+
+    public Class<?> findFirstSolutionClass() {
+        Set<Class<?>> solutionClassSet;
+        try {
+            solutionClassSet = scan(PlanningSolution.class);
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningSolution.class.getSimpleName()
```
triceo, line 86: Dtto.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return solutionClassSet.iterator().next();
+    }
+
+    public List<Class<?>> findEntityClassList() {
+        Set<Class<?>> entityClassSet;
+        try {
+            entityClassSet = scan(PlanningEntity.class);
+        } catch (ClassNotFoundException e) {
+            throw new IllegalStateException("Scanning for @" + PlanningEntity.class.getSimpleName() + " failed.", e);
```
triceo, line 97: Dtto.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
```diff
@@ -64,62 +70,422 @@
...
     }
 
-    @Bean
-    @ConditionalOnMissingBean
-    public SolverConfig solverConfig() {
-        String solverConfigXml = timefoldProperties.getSolverConfigXml();
+    @Override
+    public void setEnvironment(Environment environment) {
+        // postProcessBeanFactory runs before creating any bean, but we need TimefoldProperties. Therefore, we use the
+        // Environment to load the properties
```
triceo, line 98:
```suggestion
        // postProcessBeanFactory runs before creating any bean, but we need TimefoldProperties. 
        // Therefore, we use the Environment to load the properties.
```

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
```diff
@@ -64,62 +70,422 @@
...
+        // If the config map is empty, we build the config using the default solver name
+        if (timefoldProperties.getSolver() == null || timefoldProperties.getSolver().isEmpty()) {
+            solverConfigMap.put(TimefoldProperties.DEFAULT_SOLVER_NAME,
+                    createSolverConfig(timefoldProperties, TimefoldProperties.DEFAULT_SOLVER_NAME));
+        } else {
+            timefoldProperties.getSolver().keySet()
+                    .forEach(solverName -> solverConfigMap.put(solverName, createSolverConfig(timefoldProperties, solverName)));
+        }
+        // Step 2 - validate all SolverConfig definitions
+        // TODO - Should we assert planning members belongs to a PlanningEntity Quarkus#assertNoMemberAnnotationWithoutClassAnnotation?
```
triceo, line 128: I suggest we should have the same behavior in Spring Boot and Quarkus, where possible.
What would be the reasons _not_ to do this here?

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
```diff
@@ -64,62 +70,422 @@
...
         SolverConfig solverConfig;
-        if (solverConfigXml != null) {
-            if (beanClassLoader.getResource(solverConfigXml) == null) {
-                throw new IllegalStateException("Invalid timefold.solverConfigXml property (" + solverConfigXml
-                        + "): that classpath resource does not exist.");
+        if (solverConfigXml.isPresent()) {
+            String solverUrl = solverConfigXml.get();
+            if (beanClassLoader.getResource(solverUrl) == null) {
+                String message =
+                        "\"Invalid timefold.solverConfigXml property (%s): that classpath resource does not exist."
```
triceo, line 169: Why the quote inside the string?

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
```diff
@@ -64,62 +70,422 @@
...
+        // No Entity class
+        String emptyListErrorMessage = """
+                No classes were found with a @%s annotation.
+                Maybe your @%s annotated class(es) are not in a subpackage of your @%s annotated class's package.
+                Maybe move your planning entity classes to your application class's (sub)package(or use @%s)."""
+                .formatted(
+                        PlanningEntity.class.getSimpleName(), PlanningEntity.class.getSimpleName(),
+                        SpringBootApplication.class.getSimpleName(), EntityScan.class.getSimpleName());
+        assertEmptyInstances(entityScanner, PlanningEntity.class, emptyListErrorMessage);
+        // TODO - Should we validate target types?
```
triceo, line 338: Let's not leave the TODO here, let's fix it instead.

What would be the pros and cons?
What does Quarkus do?

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/config/SolverProperties.java
```diff
@@ -1,12 +1,24 @@
...
 import ai.timefold.solver.core.api.domain.common.DomainAccessType;
 import ai.timefold.solver.core.api.score.stream.ConstraintStreamImplType;
 import ai.timefold.solver.core.config.solver.EnvironmentMode;
 
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
 
 public class SolverProperties {
+    public static final Set<String> VALID_FIELD_NAMES_SET =
+            Set.of("solver-config-xml", "environment-mode", "daemon", "move-thread-count", "domain-access-type",
+                    "constraint-stream-impl-type", "termination");
```
triceo, line 15: It is a good idea to have a constant for this.
However, I'd go further and have constants even for the names themselves - that way, you don't need to repeat them later in this class.

triceo (commented):
Looks good! Leaving some comments.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
zepfred: `IncludeAbstractClassesEntityScanner` is the original class name, and I did not change it. I don't prefer internal classes because they make code less readable and hinder concern separation. The class's functionality is similar to that of Quarkus `IndexView`.

zepfred: Should we change the name or move it back to the Spring configuration class?

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
zepfred: Okay, I'll remember that for next time. Also, you're right about not bothering with the value since it will never be null.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
zepfred: My only concern is whether using reflection can cause issues while generating the native image. Let me replicate the behavior of Quarkus and evaluate it later.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/IncludeAbstractClassesEntityScanner.java
triceo: No, that's OK. I guess I just didn't notice that the class existed before.
Avoiding inner classes makes sense, I don't much like them either.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
zepfred: It was a typo.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldAutoConfiguration.java
zepfred: Quarkus logic requires that all planning entities be implemented as Java classes.

zepfred: Let me implement the same logic for Spring boot

triceo (approved):

PR #596, opened 2024-01-26
doc: Adding documentation for multiple resource injection (Quarks and Spring Boot)

zepfred:
This pull request contains documentation regarding a new feature that enables injecting multiple SolverManager instances in the same application.

Also, there is a specific section to describe the available managed resources for a single solver configuration.


docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
+////
+Quarkus and Spring Boot support the same configuration properties.
```
triceo, line 2: Good idea.

One thing that's not entirely obvious to me is the `<solverName>` part. You explain it in another section, and that is fine. But do you ever mention that this is optional? That `<solverName>` can be exluded if you only ever inject one thing?

I think this needs to be described in more detail, to give more clarity.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+All the properties are in this file, which can then be included multiple times.
+The {property_prefix} attribute is used for Quarkus properties.
+////
+
+{property_prefix}timefold.solver.<solverName>.solver-config-xml::
+A classpath resource to read the solver configuration XML.
+Defaults to `solverConfig.xml`.
+If this property isn't specified, that file is optional.
```
triceo, line 10: What do you mean "optional"? If the property is not specified, but the file `solverConfig.xml` is still found, will it be used? And if the file is not found/not used, where do the defaults come frome?

(I'm not saying we should _change_ existing behavior, but we should _describe_ it.)

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+{property_prefix}timefold.solver.<solverName>.environment-mode::
+Enable runtime assertions to detect common bugs in your implementation during development.
```
triceo, line 13: Could use a link to the chapter discussing those.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
+////
+Quarkus and Spring Boot support the same configuration properties.
+All the properties are in this file, which can then be included multiple times.
+The {property_prefix} attribute is used for Quarkus properties.
+////
+
+{property_prefix}timefold.solver.<solverName>.solver-config-xml::
+A classpath resource to read the solver configuration XML.
```
triceo, line 8: Add a link to a chapter on solver configuration.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+{property_prefix}timefold.solver.<solverName>.domain-access-type::
+How Timefold Solver should access the domain model.
+See xref:using-timefold-solver/configuration.adoc#domainAccess[the domain access section] for more details.
+ifeval::["{property_prefix}" == "quarkus."]
+Defaults to `GIZMO`.
+The other possible value is `REFLECTION`.
+endif::[]
+ifeval::["{property_prefix}" == ""]
+Defaults to `REFLECTION`.
+The other possible value is `GIZMO`.
```
triceo, line 37: Since this is for Spring Boot, I recommend adding a note on Gizmo limitations outside of Quarkus.
If we have this in the docs already, a link will be enough.
Otherwise we really should explain how GIZMO - for some minor performance benefits - requires you to open your domain model, making fields public.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+endif::[]
+
+{property_prefix}timefold.solver.<solverName>.termination.spent-limit::
+How long the solver can run.
+For example: `30s` is 30 seconds. `5m` is 5 minutes. `2h` is 2 hours. `1d` is 1 day.
```
triceo, line 42: I am not entirely sure, but don't we support the ISO duration format [1] here?
If so, we might as well provide a link.

[1] https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+{property_prefix}timefold.solver.<solverName>.termination.unimproved-spent-limit::
+How long the solver can run without finding a new best solution after finding a new best solution.
+For example: `30s` is 30 seconds. `5m` is 5 minutes. `2h` is 2 hours. `1d` is 1 day.
```
triceo, line 46: Dtto.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+{property_prefix}timefold.solver.<solverName>.termination.best-score-limit::
+Terminates the solver when a specific or higher score has been reached.
```
triceo, line 49: I'd say "when a specific score (or better) has been reached".

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,52 @@
...
+For example: `0hard/-1000soft` terminates when the best score changes from `0hard/-1200soft` to `0hard/-900soft`.
+Wildcards are supported to replace numbers.
```
triceo, line 51: Is this true? Wow, I didn't know. Cool.

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,743 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
 
+=== Timefold configuration properties
+
 Following properties are supported in the Quarkus `application.properties`:
 
 :property_prefix: quarkus.
 include::config-properties.adoc[]
 
+[#integrationWithQuarkusManagedResources]
+=== Auto-managed resources
```
triceo, line 282: I'd call this "injecting managed resources".

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,743 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
+    @Inject
+    ScoreManager<Timetable, SimpleScore> simpleScoreManager; // <1>
+
+    @Inject
+    ConstraintVerifier<TimetableConstraintProvider, Timetable> constraintVerifier;
+
+    ...
+}
+----
+<1> The following score types are available: `SimpleScore`, `SimpleLongScore`, `SimpleBigDecimalScore`, `HardSoftScore`, `HardSoftLongScore`, `HardSoftBigDecimalScore`, `HardMediumSoftScore`, `HardMediumSoftLongScore`, `HardMediumSoftBigDecimalScore`, `BendableScore`, `BendableLongScore`, and `BendableBigDecimalScore`.
```
triceo, line 324: Don't we have a list of score types elsewhere in the docs? If so, I'd just reference that. No point in duplicating content.

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,743 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
 include::config-properties.adoc[]
 
+[#integrationWithQuarkusManagedResources]
+=== Auto-managed resources
+
+The Quarkus integration allows the injection of several managed resources, including `SolverConfig`, `SolverFactory`,
+`SolverManager`, `SolutionManager`, `ScoreManager`, and `ConstraintVerifier`.
+
+The `SolverConfig` resource is constructed by reading the `application.properties` file and classpath. Therefore, Domain entities
+(solution, entities, and constraint classes) and customized properties (spent-limit, domain-access-type, etc) for the
```
triceo, line 288:
```suggestion
(solution, entities, and constraint classes) and customized properties (`spent-limit`, `domain-access-type` etc.) for the
```

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,743 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
+--
+====
+
+[NOTE]
+====
+Consider using xref:#integrationWithQuarkusMultiStage[Multi-stage planning] for multiple solver configurations instead of manually creating resources.
+====
+
+[#integrationWithQuarkusMultiStage]
+=== Multi-stage planning
```
triceo, line 427: I am not entirely sure why we're selling this as multi-stage planning. This feature can be used for other uses as well. It is a generic feature.

Personally, I'd remove references to multi-stage, turn it into a generic chapter on injecting multiple instances of `SolverManager`, and maybe mention in a note that this feature could be used for multi-stage planning.

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,743 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
+
+Following properties are supported per each solver:
+
+:property_prefix: quarkus.
+include::config-solver-properties.adoc[]
+
+==== Working with multiple Solvers
+
+The different solver settings are configured in the `application.properties` file. For example, two different time
+settings for spent-limit are defined as follows:
```
triceo, line 442:
```suggestion
The different solver settings are configured in the `application.properties` file. For example, two different time
settings for `spent-limit` are defined as follows:
```

triceo (changes requested):
Good start! I have provided some comments; let's discuss on Slack if necessary.
Most of the comments apply to both Quarkus and Spring Boot. When you address them, please make sure to apply your changes to both the Quarkus and Spring parts.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
zepfred: It means something must be present on the classpath if something is specified.
I've changed the description to be more consistent.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
zepfred: This description was copied from the existing file `docs/src/modules/ROOT/pages/integration/config-properties.adoc`. Let me validate it.

zepfred: @triceo, I've confirmed the description is valid.

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,708 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
+----
+--
+====
+
+[NOTE]
+====
+Consider using xref:#integrationWithQuarkusMultiStage[multiple solver configurations] instead of manually creating resources.
+====
+
+[#integrationWithQuarkusMultiStage]
```
triceo, line 427: Please change the ref to match the title.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
```diff
@@ -0,0 +1,61 @@
+////
+Quarkus and Spring Boot support the same configuration properties.
+All the properties are in this file, which can then be included multiple times.
+The {property_prefix} attribute is used for Quarkus properties.
+////
+
+{property_prefix}timefold.solver.<solverName>.solver-config-xml::
```
triceo, line 7: I'm thinking... this is content duplication with the `config-properties.adoc` file.
Can we somehow define a property (just like we do `property_prefix`) which will only inject `<solverName>.` when necessary by the context?

That way, we'll be able to use the same file for all three places where it's used.

docs/src/modules/ROOT/pages/integration/integration.adoc
```diff
@@ -271,22 +271,708 @@ To use Timefold Solver with Quarkus, read the xref:quickstart/quarkus/quarkus-qu
...
+        this.lessonToRoomTimeslotSolverManager = lessonToRoomTimeslotSolverManager
+    }
+
+    ...
+}
+----
+--
+====
+
+It is worth noting that **multi-stage** planning can be accomplished by using a separate solver configuration for each optimization stage.
```
triceo, line 617:
```suggestion
Multi-stage planning can also be accomplished by using a separate solver configuration for each optimization stage.
```

triceo (approved):
LGTM, with some extra comments.

docs/src/modules/ROOT/pages/integration/config-solver-properties.adoc
zepfred: Let me check it

zepfred: That was a great idea!
It is possible to address your request.

PR #600, opened 2024-01-30
fix: Ensuring the spent termination property is read in the Quarkus environment

zepfred:
This pull request fixes a bug in the Quarkus Benchmark extension that was causing the termination properties to be ignored.


triceo (approved):
Maybe I wouldn't check for the entire long message of the exception, just the relevant part.

But LGTM! Please double-check if we perhaps have the same issue in Spring.

zepfred:
> Maybe I wouldn't check for the entire long message of the exception, just the relevant part.
> 
> But LGTM! Please double-check if we perhaps have the same issue in Spring.

The Spring Boot module is not affected by this bug.

triceo:
Formatting is broken.

PR #601, opened 2024-01-30
perf: Wait until calculateScore() to update changed entities in a Bavet session

Christopher-Chianelli:
Bavet cannot take advantage of variable infomation to only update some nodes, so updating it on afterVariableChanged is pointless and can degrade performance if an entity has multiple geninue or shadow variables. Instead, we can record what has changed in afterVariableChanged, and then update the entities in calculateScore. This should improve performance for models with multiple shadow/geninue variables.


Christopher-Chianelli:
Turn out this behaviour is already present; `update` does not automatically propagate the update, but instead add it to the PropagationQueue(s) of affected nodes.  The propagation begins in `calculateScore`. There no performance improvement (and in fact, a slight degradation) in the use cases I tested.

PR #614, opened 2024-02-05
fix: nearby selection pinning support

triceo:


core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -175,6 +175,33 @@ public void afterElementUnassigned(ListVariableDescriptor<Solution_> variableDes
...
+            int firstUnpinnedIndex = pinningStatus.firstUnpinnedIndex();
+            if (fromIndex < firstUnpinnedIndex || toIndex < firstUnpinnedIndex) {
+                throw new IllegalStateException(
+                        """
+                                Attempting to change list variable (%s) on an entity (%s) in range [%d, %d), but the variable's first unpinned index is (%d).
+                                This is most likely a bug in a move.
+                                Maybe you are using an improperly implemented custom move?"""
+                                .formatted(variableDescriptor, entity, fromIndex, toIndex, firstUnpinnedIndex));
+            }
+        }
```
triceo, line 199: Very useful, but needs benchmarking to ensure it doesn't slow the hot path too much.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -244,7 +271,7 @@ private void simulateGenuineVariableChange(Object entity) {
             if (variableDescriptor.isGenuineListVariable()) {
                 var listVariableDescriptor = (ListVariableDescriptor<Solution_>) variableDescriptor;
                 int size = listVariableDescriptor.getListVariable(entity).size();
-                beforeListVariableChanged(listVariableDescriptor, entity, 0, size);
+                beforeListVariableChangedUnchecked(listVariableDescriptor, entity, 0, size);
```
triceo, line 274: This is a call to artifically trigger all listeners. The check above must be skipped, otherwise this would never pass. It is harmless, as it only exists to update all possible shadow vars.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/RandomListChangeIterator.java
```diff
@@ -42,13 +42,30 @@ protected Move<Solution_> createUpcomingSelection() {
...
+        }
         return new ListChangeMove<>(
                 listVariableDescriptor,
                 inverseVariableSupply.getInverseSingleton(upcomingValue),
                 indexVariableSupply.getIndex(upcomingValue),
                 destination.entity(),
                 destination.index());
     }
+
+    public static ElementRef findUnpinnedDestination(Iterator<ElementRef> destinationIterator,
```
Christopher-Chianelli, line 57: Should this be a static method in `UpcomingSelectionIterator`?

Christopher-Chianelli (commented):
I like the new pinning check. Maybe the Javadoc for
https://github.com/TimefoldAI/timefold-solver/blob/f9901d0d6e24921fe6c8e66ed1bad876e23c90c1/core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/IterableSelector.java#L8 or https://github.com/TimefoldAI/timefold-solver/blob/f9901d0d6e24921fe6c8e66ed1bad876e23c90c1/core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/MoveSelector.java#L11 should explicitly mention the selector can return null.

winklerm (approved):
Looks good from the FSR model perspective, thank you!

PR #618, opened 2024-02-07
fix: Recommended Fit API needs to ignore any unassigned entities/values

### LINKED ISSUE #581: Recommend fit sometimes checks fittedEntityOrElement not passed via SolutionManager.recommendFit
Using solver v 1.6.0. See screenshots below.

Reproducible with a model that uses nullable planning variable (Shift.employee). When some of the shifts in the input have no employee assigned, the fitter internally explores moves that include Shifts other than the one passed to `SolutionManager.recommendFit` (id 99999), yet the proposition function still operates on that shift (99999). This causes always-null employee assignment for the shift 99999.

Can be workarounded by pinning all shifts but the one with id 99999 before invoking `SolutionManager.recommendFit`.

image
image

triceo:
With nullable vars, the solution may contain many unassigned entities/values.
But the user only wants recommendations for the one that they specified.
Therefore we need to ignore every other one.

I also marginally improve composite filters, deduplicating and decomposing where possible.

Closes #581.

### DIFF
--- a/core/core-impl/src/test/java/ai/timefold/solver/core/impl/testdata/domain/nullable/TestdataNullableIncrementalScoreCalculator.java
+++ b/core/core-impl/src/test/java/ai/timefold/solver/core/impl/testdata/domain/nullable/TestdataNullableIncrementalScoreCalculator.java
@@ -10,7 +10,6 @@ import java.util.stream.Stream;
 
 import ai.timefold.solver.core.api.score.buildin.simple.SimpleScore;
 import ai.timefold.solver.core.api.score.calculator.ConstraintMatchAwareIncrementalScoreCalculator;
-import ai.timefold.solver.core.api.score.constraint.ConstraintMatch;
 import ai.timefold.solver.core.api.score.constraint.ConstraintMatchTotal;
 import ai.timefold.solver.core.api.score.constraint.ConstraintRef;
 import ai.timefold.solver.core.api.score.constraint.Indictment;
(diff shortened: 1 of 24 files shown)


triceo:
@Christopher-Chianelli Please review at your earliest convenience. I'm available for a call if you have any questions.

core/core-impl/src/test/java/ai/timefold/solver/core/impl/testdata/domain/nullable/TestdataNullableIncrementalScoreCalculator.java
```diff
@@ -10,50 +10,28 @@
...
         implements ConstraintMatchAwareIncrementalScoreCalculator<TestdataNullableSolution, SimpleScore> {
 
-    private int score = 0;
-    private DefaultConstraintMatchTotal<SimpleScore> constraintMatchTotal;
+    private TestdataNullableSolution workingSolution;
     private Map<Object, Indictment<SimpleScore>> indictmentMap;
 
     @Override
     public void resetWorkingSolution(TestdataNullableSolution workingSolution) {
-        score = 0;
```
Christopher-Chianelli, line 30: Why was this change necessary?

Christopher-Chianelli (approved):
Code changes look good to me, provided that `TestdataNullableIncrementalScoreCalculator` wasn't changed to work around a quirk/bug.

PR #624, opened 2024-02-09
feat: Add ability to enable lambda sharing in ConstraintProvider for Enterprise users

Christopher-Chianelli:


benchmark/src/main/resources/benchmark.xsd
```diff
@@ -443,6 +443,9 @@
           <xs:element minOccurs="0" name="constraintStreamImplType" type="tns:constraintStreamImplType"/>
                               
           
+          <xs:element minOccurs="0" name="constraintStreamShareLambdas" type="xs:boolean"/>
```
triceo, line 446: Not sure about the name. Maybe we should avoid implementation details?

- `constraintStreamAutomaticNodeSharing` (true/false)
- `constraintStreamNodeSharing` (auto/none)

triceo (commented):
Not much to see here, other than the naming.

triceo:
Also, needs docs in the enterprise section.

benchmark/src/main/resources/benchmark.xsd
Christopher-Chianelli: I prefer `constraintStreamAutomaticNodeSharing`, since technically, node sharing can be done (and will be done) already, just in a way that very annoying for the user.

core/core-impl/src/main/java/ai/timefold/solver/core/enterprise/TimefoldSolverEnterpriseService.java
```diff
@@ -90,7 +94,8 @@ <Solution_> DestinationSelector<Solution_> applyNearbySelection(DestinationSelec
     public enum Feature {
         MULTITHREADED_SOLVING("Multi-threaded solving", "remove moveThreadCount from solver configuration"),
         PARTITIONED_SEARCH("Partitioned search", "remove partitioned search phase from solver configuration"),
-        NEARBY_SELECTION("Nearby selection", "remove nearby selection from solver configuration");
+        NEARBY_SELECTION("Nearby selection", "remove nearby selection from solver configuration"),
+        LAMBDA_SHARING("Lambda sharing", "remove lambda sharing from solver configuration");
```
triceo, line 98:
```suggestion
        LAMBDA_SHARING("Automatic node sharing", "remove automatic node sharing from solver configuration");
```

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -716,5 +717,215 @@ the host is likely to hang or freeze,
...
+[#automaticNodeSharing]
+=== Automatic node sharing
 
+[NOTE]
+====
+This feature is a commercial feature of Timefold Solver Enterprise Edition.
+It is not available in the Community Edition.
+====
+
+When using xref:constraints-and-score/score-calculation.adoc#constraintStreams[constraint streams], each xref:constraints-and-score/score-calculation.adoc#constraintStreamsBuildingBlocks[building block] forms a node in the score calculation network.
```
triceo, line 729: Think like a user who wants to use this feature. What do they want?

- What do I get out of it?
- How do I enable it?

Everything else is secondary. Start with a simple paragraph explaining what it's for (improves score speeds when many constraints). If you can give a ballpark estimate for the improvements (10%, 50%, ...), do that. Then show the configuration tab. Then mention the downside (debugging).

Only after that, separate sub-sections can explain all the gory details like what it does, what's the downside. Most people won't read that, they just want to get stuff done. They will be happy with the first paragraph. The other paragraphs are for people who like to understand, like us.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -716,5 +717,215 @@ the host is likely to hang or freeze,
...
+    Constraint b(ConstraintFactory constraintFactory) {
+        return factory.forEach(Shift.class)
+                      .filter($predicate1)
+                      .penalize(SimpleScore.ONE)
+                      .asConstraint("b");
+    }
+}
+----
+
+IMPORTANT: This transformation prevents breakpoints from being placed inside the ConstraintProvider.
```
triceo, line 861: As explain above, this needs to be up there. But because it can not mention this transformation, it needs to be restated. Something like "Debugging breakpoints put inside your constraints will not be respected, because the `ConstraintProvider` class will be transformed when this feature is enabled".

Then when you're back down here, you can mention it again - and now they will understand _why_. Those that will care. Which, again, most won't.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -716,5 +717,215 @@ the host is likely to hang or freeze,
...
++
+[source,xml,options="nowrap"]
+----
+<!-- ... -->
+<dependencies>
+  <dependency>
+    <groupId>ai.timefold.solver.enterprise</groupId>
+    <artifactId>timefold-solver-enterprise-asm</artifactId>
+  </dependency>
+</dependencies>
```
triceo, line 880: This should not be necessary. This is boilerplate, we need Enterprise to be easy. Make `enterprise-core` depend on ASM by default.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -716,5 +717,215 @@ the host is likely to hang or freeze,
...
+                      .asConstraint("b");
+    }
+}
+----
+
+IMPORTANT: This transformation prevents breakpoints from being placed inside the ConstraintProvider.
+
+This allows building blocks to share the same nodes without needing the `ConstraintProvider` to be written in an awkward way.
+
+==== Configuration
```
triceo, line 865: I'm thinking if we should enable this by default everywhere. Enterprise ought to be simple. Maybe not right now, but eventually, yes.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -716,5 +717,215 @@ the host is likely to hang or freeze,
...
+<!-- ... -->
+<scoreDirectorFactory>
+  <constraintProviderClass>org.acme.MyConstraintProvider</constraintProviderClass>
+  <constraintStreamAutomaticNodeSharing>true</constraintStreamAutomaticNodeSharing>
+</scoreDirectorFactory>
+<!-- ... -->
+----
+
+Quarkus::
+Automatic node sharing is enabled by default.
```
triceo, line 924: If we're not enabling it by default everywhere, then I'd argue it shouldn't be the default anywhere.

triceo (changes requested):
The docs need a better structure.
Besides that, we need to agree whether to enable by default or not. I'm currently leaning towards not, with a change later once proven.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
triceo: I was proposing something less drastic, I think - Make `asm` JAR a dependency of the `enterprise-core` JAR. If they can add it as a dependency anyway, there's no reason why we shouldn't do it for them.

But maybe we should first decide if we want this feature enabled by default. If not, then obviously we should not add this default dependency.

Christopher-Chianelli:
Superseded by #685

PR #637, opened 2024-02-14
chore: Improving loading error messages for YAML files

zepfred:
This PR improves the loading error messages for YAML files. We now enforce the kebab-case format from version 1.7.0 onwards. This may affect the Spring Boot module as older versions did not enforce kebab-case format.


triceo (approved):
Well done!

PR #641, opened 2024-02-15
chore: Enabling SonarCloud for forks

zepfred:
This pull request modifies the SonarCloud workflow to function with forked repositories as well. The `pull_request_target` event works similarly to `pull_request`, but runs on the base branch and allows using secrets. Therefore, the PR ref version needs to be specified in the workflow script.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - main
-  # PRs from forks are unsupported, therefore this is pointless.
-  # pull_request:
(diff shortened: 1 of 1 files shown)


zepfred:
@triceo, the current SonarCloud configuration will trigger the workflow for opened and reopened PR or every time the PR is updated. Is this the expected behavior?

triceo:
Yes, it is. Should't I be seeing Sonar here now?

zepfred:
> Yes, it is. Should't I be seeing Sonar here now?

I don't think so. It should be merged first.

PR #656, opened 2024-02-19
feat: Adding penalty methods for `long` and `BigDecimal` values without match weighter

zepfred:
This pull request introduces two new penalty methods for `long` and `BigDecimal` values without a match weighter.


triceo (approved):
Good when green.

PR #670, opened 2024-02-26
ci: enable Enterprise and test it with some Sonar-specific fixes

triceo:


.github/workflows/downstream_enterprise.yml
```diff
@@ -1,11 +1,14 @@
-# This workflow currently doesn't work as a PR can not check out a private repo due to no secrets being available.
 name: Downstream - Timefold Solver Enterprise Edition
 
 on:
-  push:
-    branches: [main]
-  pull_request:
+  # Enables the workflow to run on PRs from forks;
+  # token sharing is safe here, because enterprise is a private repo and therefore fully under our control.
+  pull_request_target:
```
zepfred, line 6: Have you checked if the action ran after submitting a PR in your fork?

I suspect that the CI may not run if we open a PR due to the `branches` property: `branches: [main, '*.x']`. You can replicate it by making this modification to the main branch of your fork and submitting a pull request to your fork.

triceo: I've tested that the action triggers on a PR to `main` branch.

zepfred (approved):
Thanks for checking the approach!

PR #671, opened 2024-02-26
feat: Expose entity, variable and value counts in the logs and metrics

### LINKED ISSUE #604: Show the number of entities and values in the INFO log and expose it through micrometer
Motivation:
Not every dataset is the same. Some are a lot bigger than others. It's important to make the dataset size transparent.

- [ ] INFO logging: the "Solving ended" (and maybe "solving started"?) should mention the number of planning entities and planning values, as an indication of the size of the planning problem. 
- [ ] Through Micrometer it should be possible to track not just the number of solvers that ran last night (as is possible today), but also the number of planning entities and planning values that ran in those solvers.
- [ ] The benchmarker already reports this information. But maybe that code can reuse the code above?
- [ ] Document the counting system clearly

We'll need a design meeting to figure out what to do with multiple variables, multiple entity classes, etc.

Design meeting question: Is all other information in the "solving ended" message useful? Is the number of phases actually useful?

The benchmarker info, for illustration:
image

Christopher-Chianelli:
Closes #604.

The benchmarker still uses the original code, since it wants to get the entity/variable/value count of its solution, not the working solution. This difference is important, since the working solution will not be set if the Solver fails to initialize for any reason (i.e. invalid model/bad SolverConfig).


core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
```diff
@@ -0,0 +1,12 @@
+package ai.timefold.solver.core.api.solver;
+
+/**
+ * The statistics of a given problem submitted to a {@link Solver}.
+ *
+ * @param entityCount The number of genuine entities defined by the problem.
+ * @param variableCount The number of genuine variables defined by the problem.
+ * @param maximumValueRangeSize The number of possible assignments for the genuine variable with the largest range.
+ * @param problemScale An approximate log of the solution space.
```
triceo, line 9: I think this last line deserves a bit more description. What do you mean by `log`, a logarithm? A natural logarithm? What is a solution space? Maybe it'd be best if we just say what the formula is.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/SolverJob.java
```diff
@@ -95,6 +95,14 @@ public interface SolverJob<Solution_, ProblemId_> {
      */
     long getScoreCalculationCount();
 
+    /**
+     * Return the {@link ProblemStatistics} for the {@link PlanningSolution problem} submitted to the
+     * {@link SolverManager}.
```
triceo, line 100: I'm wondering if we should describe the volatility of this value.
It may be something else at solver start, during solver run, and at the end. (`ProblemChange` again.)

core/core-impl/src/main/java/ai/timefold/solver/core/config/solver/monitoring/SolverMetric.java
```diff
@@ -36,7 +36,10 @@ public enum SolverMetric {
     PICKED_MOVE_TYPE_BEST_SCORE_DIFF("timefold.solver.move.type.best.score.diff", new PickedMoveBestScoreDiffStatistic<>(),
             true),
     PICKED_MOVE_TYPE_STEP_SCORE_DIFF("timefold.solver.move.type.step.score.diff", new PickedMoveStepScoreDiffStatistic<>(),
-            false);
+            false),
+    PROBLEM_ENTITY_COUNT("timefold.solver.problem.entities", false),
+    PROBLEM_VARIABLE_COUNT("timefold.solver.problem.variables", false),
+    PROBLEM_VALUE_COUNT("timefold.solver.problem.values", false);
```
triceo, line 42: Since these metrics are enabled by default, I'd put them amongst the others which are enabled by default, at the top. The enum is not inside public API, so I don't worry about ordinals staying constant.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1039,6 +1039,7 @@ public long getMaximumValueCount(Solution_ solution) {
     /**
      * Calculates an indication on how big this problem instance is.
      * This is intentionally very loosely defined for now.
+     * This is explicitly NOT the solution space size.
```
triceo, line 1042: I see that this PR makes no changes to any of the scale computations. So we are happy with what is currently there for list and chained? I recall there were some PLANNER JIRA references there; at the very least they should be removed, since we've been over that code and the new version is something we want to move forward with.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -234,18 +228,52 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+        Metrics.gauge(SolverMetric.PROBLEM_VALUE_COUNT.getMeterId(), solverScope.getMonitoringTags(),
+                solverScope.getProblemStatistics(), ProblemStatistics::maximumValueRangeSize);
+        solverScope.getSolverMetricSet().forEach(solverMetric -> solverMetric.register(this));
+    }
+
+    private void unregisterSolverSpecificMetrics() {
+        for (var metric : List.of(SolverMetric.SCORE_CALCULATION_COUNT,
+                SolverMetric.PROBLEM_ENTITY_COUNT,
+                SolverMetric.PROBLEM_VARIABLE_COUNT,
+                SolverMetric.PROBLEM_VALUE_COUNT)) {
```
triceo, line 267: Shouldn't we unregister the metrics that are registered, instead of relying on a list that needs to be updated every time we add a new default metric?

This is a bug in the making.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -224,6 +226,22 @@ public long getTimeMillisSpent() {
...
+        this.problemStatistics = problemStatistics;
+    }
+
+    public ProblemStatistics calculateProblemStatistics(Solution_ problem) {
+        var solutionDescriptor = getSolutionDescriptor();
+        return new ProblemStatistics(solutionDescriptor.getGenuineEntityCount(problem),
+                solutionDescriptor.getGenuineVariableCount(problem),
+                solutionDescriptor.getMaximumValueRangeSize(problem),
+                solutionDescriptor.getProblemScale(problem));
+    }
```
triceo, line 243: This might as well be on the `SolutionDescriptor` directly. It only uses stuff from there.

triceo (commented):
Leaving some comments.
Please update the docs with the new default metrics.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -302,16 +330,20 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
...
                 + "phase total ({}), environment mode ({}), move thread count ({}).",
                 solverScope.getTimeMillisSpent(),
                 solverScope.getBestScore(),
                 solverScope.getScoreCalculationSpeed(),
+                solverScope.getProblemStatistics().entityCount(),
+                solverScope.getProblemStatistics().variableCount(),
+                solverScope.getProblemStatistics().maximumValueRangeSize(),
                 phaseList.size(),
                 environmentMode.name(),
                 moveThreadCountDescription);
```
triceo, line 344: I'm thinking that maybe we introduce a new logging line for problem statistics?
It seems to me that the counts and sizes are a different kind of information than the env mode, move thread count etc. - one is problem-specific, one is solver configuration.

Not entirely sure about it, but the current approach seems to me like information overload. Too much information at the same place.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
```diff
@@ -0,0 +1,12 @@
...
+ */
+public record ProblemStatistics(long entityCount, long variableCount, long maximumValueRangeSize, long problemScale) {
```
triceo, line 11: Since this is public API, maybe we need to think about this a but more.
Shouldn't it be `planningEntityCount`? `planningVariableCount`?

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
```diff
@@ -0,0 +1,12 @@
+package ai.timefold.solver.core.api.solver;
+
+/**
+ * The statistics of a given problem submitted to a {@link Solver}.
+ *
+ * @param entityCount The number of genuine entities defined by the problem.
+ * @param variableCount The number of genuine variables defined by the problem.
+ * @param maximumValueRangeSize The number of possible assignments for the genuine variable with the largest range.
```
triceo, line 8: We already have to iterate over every entity to get this number. So I'm wondering... why not replace this number with a total number of distinct values?

Could be a problem for int value ranges, which are uncountable. How do you handle that case right now?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
Christopher-Chianelli: This is a list of all solve level Solver-specific metrics (notice the solve duration longer timer is not deregistered as it not solver specific, nor are the best solution/other metrics since they are handled via event listeners).

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
Christopher-Chianelli: Was waiting on @ge0ffrey opinion since changing this code would result in problem scale changes in benchmark. I have changes locally that make this calculation more well defined:

- log is in base maximumValueRangeSize (for basic variables, most entities will either add 0 (pinned) or 1 (not pinned) to the count. If the value range for the entity is entity sourced, it will add a number between 0 and 1.
- For list/chained, the formula is `(V! * (V * E))/E!`, where `V` the value count and `E` the entity count (There are `V!` ways to permute a list of `V` elements. You can place `E` markers on a list of size `V` `V * E` ways. You can arrange `E` partitions of a list `E!` ways. So there are `V! * (V * E)` ways to create `E` partitions for all permutations of a list of size `V`, of which   `V! * (V * E)/E!` are unique.

The base of the log in the record is dynamic (i.e. based on `maximumValueRangeSize`) to preserve accuracy, but can be converted to a common base using the formula `log_b(x) = log_a(x) / log_a(b)`

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
Christopher-Chianelli: Currently, `maximumValueRangeSize` is used for the base of the log, so it will result in the `problemScale` being less accurate (since instead of having values commonly be `0` or `1`, they be a decimal)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
Christopher-Chianelli: Pending @ge0ffrey decision to change the meaning of `problemScale`  from `entityCount * valueCount` (that the current definition) to `log(product(value_range_size_for_entity)) = log(value_range_size_for_entity_1) +  log(value_range_size_for_entity_2) + ... ~= log(value_range_size) * entityCount ` (this is for basic variables, list/chained variables have a different counting system to discard invalid assignments)

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemStatistics.java
Christopher-Chianelli: There an issue with `valueCount`:

- Is it the number of distinct values defined by the problem? If so, to prevent double counting, we would need to put all the values into a set (value ranges can share values).
- Or is it the sum of possible assignments per variable?

triceo: Yeah, I was worried about that. I don't think it's practical to put the items in a set. There could just be prohibitively too many. Double counting it is; the number is an _estimate_ and it needs to be made clear.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/SubSingleBenchmarkRunner.java
```diff
@@ -121,12 +119,10 @@ public SubSingleBenchmarkRunner<Solution_> call() {
...
-            SolverScope<Solution_> solverScope = solver.getSolverScope();
-            SolutionDescriptor<Solution_> solutionDescriptor = solverScope.getSolutionDescriptor();
-            problemBenchmarkResult.registerScale(solutionDescriptor.getGenuineEntityCount(solution),
-                    solutionDescriptor.getGenuineVariableCount(solution),
-                    solutionDescriptor.getMaximumValueCount(solution),
-                    solutionDescriptor.getProblemScale(solution));
+            var solverScope = solver.getSolverScope();
+            var solutionDescriptor = solverScope.getSolutionDescriptor();
+            problemBenchmarkResult
+                    .registerScale(solutionDescriptor.getProblemSizeStatistics(solverScope.getScoreDirector(), solution));
```
triceo, line 125: Aren't we mixing terms here? One speaks of scale, the other speaks of size.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/result/ProblemBenchmarkResult.java
```diff
@@ -419,45 +420,42 @@ private void determineWinningScoreDifference() {
...
      * Called multiple times, for every {@link SingleBenchmarkResult} of this {@link ProblemBenchmarkResult}.
      *
-     * @param registeringEntityCount {@code >= 0}
-     * @param registeringVariableCount {@code >= 0}
-     * @param registeringProblemScale {@code >= 0}
+     * @param problemStatistics never null
      */
-    public void registerScale(long registeringEntityCount, long registeringVariableCount,
-            long registeringMaximumValueCount, long registeringProblemScale) {
+    public void registerScale(ProblemSizeStatistics problemStatistics) {
```
triceo, line 425: Arguably `problemSizeStatistics` now. 
Please looks for all uses of `problemStatistics` (incl. in getter/setter names) and make it consistent - looking at the PR, there are many places where you kept the old name, and I won't name all of them.

core/core-impl/src/main/java/ai/timefold/solver/core/api/solver/ProblemSizeStatistics.java
```diff
@@ -0,0 +1,50 @@
...
+            return "~%s".formatted(BASIC_FORMATTER.format(Math.pow(10d, approximateProblemSizeLog)));
+        }
+        // The actual number will often be too large to fit in a double, so cannot use normal
+        // formatting.
+        // Separate the exponent into its integral and fractional parts
+        // Use the integral part as the power of 10, and the fractional part as the significant digits.
+        double exponentPart = Math.floor(approximateProblemSizeLog);
+        double remainderPartAsExponent = approximateProblemSizeLog - exponentPart;
+        double remainderPart = Math.pow(10, remainderPartAsExponent);
+        return "~%se%s".formatted(
```
triceo, line 46: I'd prefer going with a more human-readable `A * 10^B`, unless there is a specific reason why we can't do that.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -684,22 +687,57 @@ public long getMaximumValueCount(Solution_ solution, Object entity) {
...
+                                    continue;
+                                }
+                                tracker.visitedAnchorSet().add(value);
+                                tracker.listTotalEntityCount().increment();
+                                tracker.listMovableEntityCount().increment();
+                                // Assumes anchors are not pinned
+                            }
+                        }
+                    } else {
+                        // Assume 0 entities (this might be impossible)
```
triceo, line 719: Uncountable value ranges already fail fast elsewhere. Please confirm and if that's the case, please also fail fast here. People shouldn't be using them.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1026,7 +1030,39 @@ public long getGenuineVariableCount(Solution_ solution) {
         return result.longValue();
     }
 
-    public long getMaximumValueCount(Solution_ solution) {
+    /**
+     * @param solution never null
+     * @return {@code >= 0}
+     */
+    public long getApproximateValueCount(Solution_ solution) {
+        Set<GenuineVariableDescriptor<Solution_>> genuineVariableDescriptorSet =
```
triceo, line 1038: IMO this is precisely the kind of code that benefits from `var` the most. We have `Set` in the var name, we have `Set` in the method call on the right, no need to repeat the entire type on the left. (Even if it means putting the generic type directly to the diamond of `IdentityHashMap`.)

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1036,22 +1072,64 @@ public long getMaximumValueCount(Solution_ solution) {
                 .orElse(0L);
     }
 
+    public record ProblemScaleTracker(long logBase,
```
triceo, line 1075: This is abusing the record a bit, isn't it? Records are supposed to be immutable, and to get that, we introduce deeply mutable types. :-) Personally, I think this should just be a class, which in turn would allow us to get rid of the `MutableLong`.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1036,22 +1072,64 @@ public long getMaximumValueCount(Solution_ solution) {
...
+            MutableLong listTotalValueCount) {
+        public void addBasicProblemScale(long count) {
+            basicProblemScaleLog.add(MathUtils.getScaledApproximateLog(MathUtils.LOG_PRECISION, logBase, count));
+        }
+    }
+
     /**
      * Calculates an indication on how big this problem instance is.
-     * This is intentionally very loosely defined for now.
+     * This is approximately the base 10 log of the solution space size.
```
triceo, line 1089: Solution space size... Maybe we use "search space size" instead? It's closer to the industry lingo. (Assuming that search space size is what you mean here.)

core/core-impl/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -234,16 +226,34 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+        registerSolverSpecificMetrics();
+        logger.info("Solving {}: time spent ({}), best score ({}), "
+                + "environment mode ({}), move thread count ({}), random ({}).",
                 (startingSolverCount == 1 ? "started" : "restarted"),
                 solverScope.calculateTimeMillisSpentUpToNow(),
                 solverScope.getBestScore(),
                 environmentMode.name(),
                 moveThreadCountDescription,
                 (randomFactory != null ? randomFactory : "not fixed"));
+        logger.info(
```
triceo, line 240: I keep thinking... isn't it enough to print this just once, either at start or finish?
It will end up in the log, which means it's just duplication if the same information is there twice.

triceo (commented):
Leaving some comments inline based on the final iteration.

Also:

- Please introduce docs for the new metrics.
- Check out Sonar and fix what makes sense. (Some of those checks we may want to disable entirely.)
- Run the benchmark manually, making sure that the new scale numbers don't break the code and the final report that is rendered, where those numbers are shown. (This is not tested at all.)

After that, LGTM.

triceo (approved):

PR #673, opened 2024-02-28
feat: Adding Nearby configuration by exception

zepfred:
This pull request adds easy configuration of the Nearby Selection feature and updates the local search phase factory to provide default selectors when custom configuration is not provided. The configuration for the selectors is listed below:
1. Basic variables selectors: Change, Swap, Change with nearby selection, and Swap with nearby selection.
2. List variables selectors: Change, Swap, Change with nearby selection, Swap with nearby selection, 2-OPT with nearby selection.
3. Chained variables selectors: Change, Swap, Change with nearby selection, Swap with nearby selection, Tail Chain Swap with nearby selection.

The following XML configuration can be used to enable Nearby selection quickly:

`<nearbyDistanceMeterClass>...</nearbyDistanceMeterClass>`

If there are multiple entities, we will fail fast.


core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,18 +198,112 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+                                new ListSwapMoveSelectorConfig()
+                                        .withValueSelectorConfig(new ValueSelectorConfig()
+                                                .withId("swapMoveSelector"))
+                                        .withSecondaryValueSelectorConfig(new ValueSelectorConfig()
+                                                .withNearbySelectionConfig(new NearbySelectionConfig()
+                                                        .withOriginValueSelectorConfig(new ValueSelectorConfig()
+                                                                .withMimicSelectorRef("swapMoveSelector"))
+                                                        .withNearbyDistanceMeterClass(
+                                                                configPolicy.getNearbyDistanceMeterClass()))),
+                                new KOptListMoveSelectorConfig()
```
triceo, line 231: I'm asking myself... if we're adding this move, why only for nearby? Does it make sense to add it for both?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,18 +198,112 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+                                    new SwapMoveSelectorConfig()
+                                            .withEntitySelectorConfig(new EntitySelectorConfig()
+                                                    .withId("swapMoveSelector"))
+                                            .withSecondaryEntitySelectorConfig(new EntitySelectorConfig()
+                                                    .withNearbySelectionConfig(new NearbySelectionConfig()
+                                                            .withOriginEntitySelectorConfig(new EntitySelectorConfig()
+                                                                    .withMimicSelectorRef("swapMoveSelector"))
+                                                            .withNearbyDistanceMeterClass(
+                                                                    configPolicy.getNearbyDistanceMeterClass()))),
+                                    new TailChainSwapMoveSelectorConfig()
```
triceo, line 267: Same question here.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,18 +198,112 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+                                                            .withNearbyDistanceMeterClass(
+                                                                    configPolicy.getNearbyDistanceMeterClass()))));
+                }
+            }
         } else {
             /*
              * We have a mix of basic and list variables.
              * The "old" move selector configs handle this situation correctly but they complain of it being deprecated.
              *
+             * Combining essential variables with list variables in a single entity is not supported. Therefore,
```
triceo, line 304: How can this even happen? List and basic vars are mutually incompatible - so shouldn't this `else` branch fail fast instead? Or is there a place where this actually works?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,18 +198,112 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+            if (configPolicy.getNearbyDistanceMeterClass() == null) {
+                return new UnionMoveSelectorConfig()
+                        .withMoveSelectors(new ListChangeMoveSelectorConfig(), new ListSwapMoveSelectorConfig());
+            } else {
+                return new UnionMoveSelectorConfig()
+                        .withMoveSelectors(new ListChangeMoveSelectorConfig(),
+                                new ListSwapMoveSelectorConfig(),
+                                new ListChangeMoveSelectorConfig()
+                                        .withValueSelectorConfig(new ValueSelectorConfig()
+                                                .withId("changeMoveSelector"))
```
triceo, line 215: I'm thinking... I don't remember the specific behavior, but aren't these supposed to be unique? Maybe we want to generate the selector names with some randomness? `changeMoveSelector-f4edab` or something like that.

triceo (commented):
Leaving some questions that we should consider.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
zepfred: The selectors are created for an empty configuration, and their uniqueness is guaranteed. Anyway, let me introduce some randomness to it.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,21 +199,137 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
              *
+             * Combining essential variables with list variables in a single entity is not supported. Therefore,
+             * default selectors do not support enabling Nearby Selection with multiple entities.
+             *
              * TODO Improve so that list variables get list variable selectors directly.
              * TODO PLANNER-2755 Support coexistence of basic and list variables on the same entity.
              */
+            if (configPolicy.getNearbyDistanceMeterClass() != null) {
+                throw new IllegalArgumentException(
+                        "The configuration contains basic and list variables that are incompatible with using the nearbyDistanceMeterClass (%s)."
```
triceo, line 330:
```suggestion
                        """
                        The configuration contains both basic and list variables, which makes it incompatible with using a top-level nearbyDistanceMeterClass (%s).
                        Specify move selectors manually or remove the top-level nearbyDistanceMeterClass from your solver config."""
```

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,21 +199,137 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+                                            .withEntitySelectorConfig(new EntitySelectorConfig()
+                                                    .withId(tailChainSelectorName))
+                                            .withValueSelectorConfig(new ValueSelectorConfig()
+                                                    .withNearbySelectionConfig(new NearbySelectionConfig()
+                                                            .withOriginEntitySelectorConfig(new EntitySelectorConfig()
+                                                                    .withMimicSelectorRef(tailChainSelectorName))
+                                                            .withNearbyDistanceMeterClass(
+                                                                    configPolicy.getNearbyDistanceMeterClass()))));
+                } else {
+                    String changeSelectorName = "changeMoveSelector-%s".formatted(UUID.randomUUID().toString().substring(0, 8));
```
triceo, line 293: I recommend turning the code for randomness into its own method - it is repeated far too often. Something like `addRandomSuffix(...)`.
Also, any chance we could make this randomness deterministic, sharing a pre-determined random seed across all the invocations?

core/core-impl/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -191,21 +199,137 @@ private UnionMoveSelectorConfig determineDefaultMoveSelectorConfig(HeuristicConf
...
+            if (configPolicy.getNearbyDistanceMeterClass() == null) {
+                if (hasChainedVariable) {
+                    return new UnionMoveSelectorConfig()
+                            .withMoveSelectors(new ChangeMoveSelectorConfig(), new SwapMoveSelectorConfig(),
+                                    new TailChainSwapMoveSelectorConfig());
+                } else {
+                    return new UnionMoveSelectorConfig()
+                            .withMoveSelectors(new ChangeMoveSelectorConfig(), new SwapMoveSelectorConfig());
+                }
+            } else {
```
triceo, line 256: The method is getting crazy long. I suggest we split it into parts - perhaps each of these inner-most branches gets its own method?

triceo (approved):
LGTM after final comments are resolved.

PR #684, opened 2024-03-06
feat: Enabling Nearby Selection for non-empty move selectors configuration

zepfred:
This pull request makes the Nearby Selection feature easy to configure and updates the local search phase settings to provide the default Nearby configuration when the user specifies move selectors.

1. Only a subset of the move selectors accepts Nearby autoconfiguration.
2. We fail fast if there is any pre-defined Nearby configuration.
3. We fail fast if Nearby is enabled and the Cartesian product move selector is defined.

### DIFF
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
@@ -2,6 +2,7 @@ package ai.timefold.solver.core.config.heuristic.selector.move;
 
 import java.util.Comparator;
 import java.util.List;
+import java.util.Random;
 import java.util.function.Consumer;
 
 import jakarta.xml.bind.annotation.XmlSeeAlso;
@@ -32,6 +33,7 @@ import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.Selectio
 import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionProbabilityWeightFactory;
 import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionSorter;
 import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionSorterWeightFactory;
+import ai.timefold.solver.core.impl.heuristic.selector.common.nearby.NearbyDistanceMeter;
 
 /**
  * General superclass for {@link ChangeMoveSelectorConfig}, etc.
@@ -279,5 +281,34 @@ public abstract class MoveSelectorConfig<Config_ extends MoveSelectorConfig<Conf
         fixedProbabilityWeight = ConfigUtils.inheritOverwritableProperty(
                 fixedProbabilityWeight, inheritedConfig.getFixedProbabilityWeight());
     }
+
+    protected String addRandomSuffix(String name, Random random) {
+        StringBuilder value = new StringBuilder(name);
+        value.append("-");
+        random.ints(97, 122) // ['a', 'z']
+                .limit(4) // 4 letters
+                .forEach(value::appendCodePoint);
+        return value.toString();
+    }
+
+    /**
+     * Check if the selector factory accepts Nearby selection autoconfiguration.
+     *
+     * @return false by default
+     */
+    public boolean acceptNearbySelectionAutoConfiguration() {
+        return false;
+    }
+
+    /**
+     * Enables the Nearby Selection autoconfiguration.
+     *
+     * @return new instance with the Nearby Selection settings properly configured
+     */
+    public Config_ enableNearbySelection(Class<NearbyDistanceMeter<?, ?>> distanceMeter, Random random) {
+        throw new UnsupportedOperationException();
+    }
+
+    public abstract boolean hasNearbySelectionConfig();
 
 }
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
@@ -292,7 +292,7 @@ public abstract class MoveSelectorConfig<Config_ extends MoveSelectorConfig<Conf
     }
 
     /**
-     * Check if the selector factory accepts Nearby selection autoconfiguration.
+     * Check if the move selector accepts Nearby autoconfiguration.
      *
      * @return false by default
      */
@@ -309,6 +309,9 @@ public abstract class MoveSelectorConfig<Config_ extends MoveSelectorConfig<Conf
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Verifies if the current configuration has any Nearby Selection settings.
+     */
     public abstract boolean hasNearbySelectionConfig();
 
 }
--- a/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
+++ b/core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
@@ -280,9 +280,4 @@ public abstract class MoveSelectorConfig<Config_ extends MoveSelectorConfig<Conf
                 fixedProbabilityWeight, inheritedConfig.getFixedProbabilityWeight());
     }
 
-    /**
-     * Verifies if the current configuration has any Nearby Selection settings.
-     */
-    public abstract boolean hasNearbySelectionConfig();
-
 }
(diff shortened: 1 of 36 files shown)


core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -280,4 +282,33 @@ private void inheritCommon(MoveSelectorConfig<?> inheritedConfig) {
                 fixedProbabilityWeight, inheritedConfig.getFixedProbabilityWeight());
     }
 
+    protected String addRandomSuffix(String name, Random random) {
```
triceo, line 285: Should be `static`, no?

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -280,4 +282,33 @@ private void inheritCommon(MoveSelectorConfig<?> inheritedConfig) {
...
+    }
+
+    /**
+     * Check if the selector factory accepts Nearby selection autoconfiguration.
+     *
+     * @return false by default
+     */
+    public boolean acceptNearbySelectionAutoConfiguration() {
+        return false;
+    }
```
triceo, line 301: I'd consider working without this. You can call `enableNearbySelection(...)`, which by default returns `null`. If null, nothing happens.

(I'm driven by eliminating boilerplate code; the selectors already have plenty without adding more.)

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -280,4 +282,33 @@ private void inheritCommon(MoveSelectorConfig<?> inheritedConfig) {
...
+    /**
+     * Enables the Nearby Selection autoconfiguration.
+     *
+     * @return new instance with the Nearby Selection settings properly configured
+     */
+    public Config_ enableNearbySelection(Class<? extends NearbyDistanceMeter<?, ?>> distanceMeter, Random random) {
+        throw new UnsupportedOperationException();
+    }
+
+    public abstract boolean hasNearbySelectionConfig();
```
triceo, line 312: Maybe this could return `false` by default, so that you only need to override it when `true` needs to be returned?

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
triceo: We could (maybe) even go as far as to create a new sub-class for nearby-selection enabled selectors, so that it is absolutely obvious on a type level whether or not something supports nearby selection.

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/generic/list/SubListChangeMoveSelectorConfig.java
```diff
@@ -154,6 +154,18 @@ public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
         }
     }
 
+    @Override
+    public boolean hasNearbySelectionConfig() {
+        return (subListSelectorConfig != null && ((subListSelectorConfig.getNearbySelectionConfig() != null)
```
triceo, line 159: This condition is kinda insane.
Maybe simplify it by introducing boolean variables for its components?
The code will be longer, but people will actually be able to understand what the condition does.

(There are similar conditions elsewhere in this PR.)

core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/UnionMoveSelectorFactory.java
```diff
@@ -22,8 +23,24 @@ public UnionMoveSelectorFactory(UnionMoveSelectorConfig moveSelectorConfig) {
...
     protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<Solution_> configPolicy,
             SelectionCacheType minimumCacheType, boolean randomSelection) {
-        List<MoveSelector<Solution_>> moveSelectorList = buildInnerMoveSelectors(config.getMoveSelectorList(),
-                configPolicy, minimumCacheType, randomSelection);
+        List<MoveSelectorConfig> moveSelectorConfigList = new LinkedList<>(config.getMoveSelectorList());
+        if (configPolicy.getNearbyDistanceMeterClass() != null) {
+            for (MoveSelectorConfig selectorConfig : config.getMoveSelectorList().stream()
+                    .filter(MoveSelectorConfig::acceptNearbySelectionAutoConfiguration).toList()) {
+                if (selectorConfig.hasNearbySelectionConfig()) {
+                    throw new IllegalArgumentException(
```
triceo, line 31: Maybe explain _why_ in a comment?

The problem with cartesian is that if there are two selectors, it will have A*B moves. So if there are four selectors (2 non-nearby + 2 nearby), it will be A*B*C*D moves.

Whereas union will always be A+B+C+D.

core/core-impl/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/UnionMoveSelectorFactory.java
```diff
@@ -22,8 +23,24 @@ public UnionMoveSelectorFactory(UnionMoveSelectorConfig moveSelectorConfig) {
...
+                                    The selector configuration (%s) already includes the Nearby Selection setting, making it incompatible with the top-level property nearbyDistanceMeterClass (%s).
+                                    Remove the Nearby setting from the selector configuration or remove the top-level nearbyDistanceMeterClass."""
+                                    .formatted(selectorConfig, configPolicy.getNearbyDistanceMeterClass()));
+                }
+                // Add a new configuration with Nearby Selection enabled
+                moveSelectorConfigList.add(selectorConfig.enableNearbySelection(configPolicy.getNearbyDistanceMeterClass(),
+                        configPolicy.getRandom()));
+            }
+        }
+        List<MoveSelector<Solution_>> moveSelectorList =
```
triceo, line 42: Cautious use of `var` wouldn't hurt here, I think.
The name already says `list`.

triceo (commented):
Good progress. Providing initial batch of comments.

core/core-impl/src/test/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfigTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class MoveSelectorConfigTest {
+
+    @Test
+    void assertEnableNearbyForChangeMoveSelectorConfig() {
+        // Default configuration
+        ChangeMoveSelectorConfig config = new ChangeMoveSelectorConfig();
+        assertFalse(config.hasNearbySelectionConfig());
```
triceo, line 42: We use AssertJ, not JUnit assertions.

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
zepfred: I liked the second approach and will proceed with implementing it.

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/generic/list/SubListChangeMoveSelectorConfig.java
zepfred: Let me try to simplify it.

core/core-impl/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/composite/UnionMoveSelectorConfig.java
```diff
@@ -162,6 +166,31 @@ public void visitReferencedClasses(Consumer<Class<?>> classVisitor) {
...
     }
 
+    @Override
+    public UnionMoveSelectorConfig enableNearbySelection(Class<? extends NearbyDistanceMeter<?, ?>> distanceMeter,
+            Random random) {
+        UnionMoveSelectorConfig nearbyConfig = copyConfig();
+        List<MoveSelectorConfig> updatedMoveSelectorList = new LinkedList<>();
+        for (NearbyAutoConfigurationMoveSelectorConfig<?> selectorConfig : moveSelectorConfigList.stream()
+                .filter(s -> NearbyAutoConfigurationMoveSelectorConfig.class.isAssignableFrom(s.getClass()))
+                .map(s -> (NearbyAutoConfigurationMoveSelectorConfig<?>) s).toList()) {
```
triceo, line 176: Id argue that the following would read better, also coincidentally doesn't need to create an extra list or iterate twice:

   for (... all configs ...) {
       if (not nearby) {
           continue;
       }
       ...
   }

Also, the `selectorConfig` variable is another nice use of `var` - all the necessary context is already there.

triceo (approved):

triceo:
You'll have to update to latest `main`, as a change to solver config XSD recently landed there.

zepfred:
> You'll have to update to latest `main`, as a change to solver config XSD recently landed there.

I had to update the XSD files with the new `NearbyAutoConfigurationMoveSelectorConfig` configuration class.

Issue #690, opened 2024-03-07
Configured class loader not used when loading services


marinier:
**Describe the bug**
When creating a `SolverConfig` you can set a class loader on it like this:

```java
solverConfig.setClassLoader(this.getClass().getClassLoader());
```

This class loader eventually gets passed to `ScoreDirectorFactoryFactory::decideMultipleScoreDirectorFactories`, but it is not passed to the `ServiceLoader.load` call on line 69 of `ScoreDirectorFactoryFactory.java`. That is, instead of using the provided class loader, it ends up using the default class loader. In my application, this does not work -- the provided class loader needs to be used.

**Expected behavior**
Uses the class loaded provided by the `SolverConfig` when loading services.

**Actual behavior**
Uses the default class loader when loading services. (Specifically, `ServiceLoader` defaults to `Thread.currentThread().getContextClassLoader()`.)

**To Reproduce**
Reproducing my problem where the service loader fails requires access to the specific environment I'm working in, which is not possible. However, you can easily check which class loader is being used by setting a breakpoint and stepping in the `ServiceLoader.load` call on line 69 of `ScoreDirectorFactoryFactory.java`.

## Environment

**Timefold Solver Version or Git ref**: 

1.7.0

**Output of `java -version`:**

openjdk version "17.0.10" 2024-01-16
OpenJDK Runtime Environment Temurin-17.0.10+7 (build 17.0.10+7)
OpenJDK 64-Bit Server VM Temurin-17.0.10+7 (build 17.0.10+7, mixed mode, sharing)

**Output of `uname -a` or `ver`:**

OS Name	Microsoft Windows 11 Pro
Version	10.0.22631 Build 22631

## Additional information

The fix should be very simple, as the class loader is already passed as an argument to `ScoreDirectorFactoryFactory::decideMultipleScoreDirectorFactories`. Simply change:

```java
ServiceLoader.load(ScoreDirectorFactoryService.class)
```
 to
```java 
ServiceLoader.load(ScoreDirectorFactoryService.class, classLoader)
```

I'm happy to provide a pull request if desired.

triceo:
Hello @marinier - PRs are welcome. Even if we eventually remove the service loader from constraint streams (as discussed [here](https://github.com/TimefoldAI/timefold-solver/issues/664)), it will still remain for enterprise edition, so this will still make sense.

marinier:
Added a PR here: https://github.com/TimefoldAI/timefold-solver/pull/692

PR #692, opened 2024-03-07
fix: pass provided class loader to service loader

marinier:
This PR is for issue #690. It's a very simple 1-line change.


marinier:
The pipeline build failed in the quarkus deploy step. This happens on my machine as well even with a fresh checkout and no changes. Is the build just broken, or is there something I'm supposed to do to fix this?

triceo:
@marinier All tests are passing on `main` - this means that there is some aspect of the problem that the PR didn't get right. (At the moment, I don't have the time to help.)

marinier:
It was a silly mistake -- I didn't realize the class loader could be null at that point in the code. Added a null check, and now things seem to work.

triceo (approved):

PR #714, opened 2024-03-12
feat: Adding down-sampling process for Benchmark

zepfred:
This pull request updates the Line chart of the benchmark module and adds a down-sampling process based on `Largest-Triangle-Three-Buckets` to compress the data point list size.

The implementation is based on the original work, which is available [here](https://github.com/sveinn-steinarsson/flot-downsample). It compresses the list of data points without changing the chart's shape. The sample size is `1280`, the default chart width setting.

## Non-compressed
Screenshot from 2024-03-12 08-14-37
## Compressed
Screenshot from 2024-03-12 08-14-48
## Non-compressed
Screenshot from 2024-03-12 08-14-59
## Compressed
Screenshot from 2024-03-12 08-15-07
## All lines
Screenshot from 2024-03-12 08-15-16
## Non-compressed
Screenshot from 2024-03-12 08-15-45
## Compressed
Screenshot from 2024-03-12 08-15-53
## All lines
Screenshot from 2024-03-12 08-16-04


zepfred:
@triceo, I will let it run for one hour and then share the charts again.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/LineChart.java
```diff
@@ -173,7 +187,7 @@ public void writeToFile(Path parentFolder) {
     }
 
     public static final class Builder<X extends Number & Comparable<X>, Y extends Number & Comparable<Y>> {
-
+        private static final int MAX_CHART_WIDTH = 1280;
```
triceo, line 190: Seems low, no? I'd increase it up to 4K width, which is 3840.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/LineChart.java
```diff
@@ -209,25 +223,35 @@ public LineChart<X, Y> build(String fileName, String title, String xLabel, Strin
              * This allows Chart.js to only draw the absolute minimum necessary points.
              */
             data.values().forEach(map -> {
-                List<Map.Entry<X, Y>> entries = map.entrySet().stream().toList();
+                List<Entry<X, Y>> entries = map.entrySet().stream().toList();
```
triceo, line 226: I'd really like us to start using `var` for these generic-heavy things.
The declaration brings no additional context here. You already know it's a list (`toList()`) and you already know it is an entry (`entrySet()`).

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/LineChart.java
```diff
@@ -209,25 +223,35 @@ public LineChart<X, Y> build(String fileName, String title, String xLabel, Strin
...
                 }
             });
-            // Then find all points on the X axis across all data sets.
+            /*
+             * Sometimes, when the dataset size is large, it can cause the browser to freeze or use excessive memory
+             * while rendering the line chart. To solve the issue of a large volume of data points, we use the
+             * Largest-Triangle-Three-Buckets algorithm to down-sample the data.
+             */
+            Map<String, Map<X, Y>> datasetMap = new LinkedHashMap<>(data.size());
+            for (Entry<String, NavigableMap<X, Y>> entry : data.entrySet()) {
```
triceo, line 248: Another great place for `var`. Makes the code much more readable.
In fact, I'd go as far as saying that we should make `var` the default, and use the full type if we have a reason to do that.

benchmark/src/main/resources/ai/timefold/solver/benchmark/impl/report/chart-line.js.ftl
```diff
@@ -34,7 +31,7 @@ var ${chartId} = new Chart(document.getElementById('${chartId}'), {
                     borderWidth: 1
                   </#if>,
                   data: [
-                    <#list dataset.data() as datum><#if datum??>${datum?cn}</#if><#sep>, </#sep></#list>
+                    <#list chart.points(dataset.label()) as datum>{x: ${datum.key()?cn}, y: ${datum.value()?cn}}<#sep>, </#sep></#list>
```
triceo, line 34: If I may recommend - don't touch the format of JS files.

I spent days tweaking chart.js to avoid all of its bugs and quirks. Sometimes it required weird choices in data set format. Unless you want to manually check every single supported chart, including tooltips, scaling and different browsers, I recommend not touching the data format.

benchmark/src/test/resources/ai/timefold/solver/benchmark/impl/result/resultValues.txt
```diff
@@ -0,0 +1 @@
.../-177368.00,1984698.00/-177307.00,1985432.00/-177240.00,1985684.00/-177176.00,1985932.00/-177097.00,1986561.00/-177075.00,1987304.00/-176939.00,1989111.00/-176869.00,1990041.00/-176703.00,1990258.00/-176613.00,1991035.00/-176487.00,1992117.00/-176389.00,2523485.00/-176360.00,2597077.00/-176217.00,2597595.00/-176096.00,2600931.00/-176005.00,2601430.00/-175817.00,2604282.00/-175755.00,2604596.00/-175664.00,2604935.00/-175576.00,2605731.00/-175450.00,2606843.00/-175412.00,2607241.00/-175379.00,2607902.00/-175290.00,2610423.00/-175145.00,2868654.00/-174996.00,2868848.00/-174878.00,2869944.00/-174636.00,2871747.00/-174568.00,2872171.00/-174510.00,2873069.00/-174246.00,2873352.00/-174097.00,2874196.00/-174067.00,2875034.00/-174014.00,2875714.00/-173917.00,2876790.00/-173857.00,2878191.00/-173748.00,2878412.00/-173669.00,2883184.00/-173600.00,2885176.00/-173568.00,2885601.00/-173501.00,2886618.00/-173390.00,2887661.00/-173351.00,2888256.00/-173279.00,2889078.00/-173135.00,2889566.00/-172886.00,2889891.00/-172796.00,2890498.00/-172623.00,2896148.00/-172536.00,2896411.00/-172515.00,2896817.00/-172423.00,2897498.00/-172287.00,2897693.00/-172191.00,3589914.00/-172070.00,3600000.00/-172046.00,
```
triceo, line 1: Ideally this file would have the same format as the input files.
And all those input files would be called not `.txt`, but `.csv`, because that is what they are - comma-separated values.

triceo (approved):
Leaving some comments. Overall LGTM.
Looking forward to results on large data sets!

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/LineChart.java
zepfred: Yes, we can increase it. Should we also change the chart setting?

triceo: I'd keep it what it is - but when a user scales it up manually, they should get all the extra points. Therefore changing the maximum constant.

PR #723, opened 2024-03-15
ci: only trigger Dependabot once per week

triceo:


.github/dependabot.yml
```diff
@@ -3,26 +3,12 @@ updates:
   - package-ecosystem: maven
     directory: "/"
     schedule:
-      interval: daily
-      time: '05:00'
+      interval: weekly
```
zepfred, line 6: Should we keep the scheduling time?

zepfred (approved):

.github/dependabot.yml
triceo: I chose the time totally at random to prevent all the jobs happening at the same time. Now that it's done weekly, I think it doesn't matter anymore - let Github do its best.

PR #726, opened 2024-03-18
chore: Remove conference scheduling code

zepfred:
This pull request removes the code related to conference scheduling and replaces it with a new quickstart.


docs/src/modules/ROOT/pages/use-cases-and-examples/conference-scheduling/conference-scheduling.adoc
```diff
@@ -10,7 +10,7 @@ Assign each conference talk to a timeslot and a room, after the talks have been
 
 image::use-cases-and-examples/conference-scheduling/conferenceSchedulingMilestonesTimeline.png[align="center"]
```
triceo, line 11: I think this entire adoc file can disappear entirely. It is a description of an example which we no longer have. (In fact, I will be removing the entire "use cases and examples" chapter very soon.)

triceo (approved):

PR #727, opened 2024-03-19
chore: Add internal interface for planning cloning atypical types

Christopher-Chianelli:
Sometimes, the domain model has atypical types for fields (for instance, UserList instead of List or ArrayList). The default planning cloner originally failed fast when the field cannot be assigned a typical value. Now, it checks if the collection implements the internal interface. If it does, it uses it to create an empty collection/map. This allows the default planning cloner to be used when the user has control over UserList.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
@@ -108,8 +108,14 @@ public final class FieldAccessingSolutionCloner<Solution_> implements SolutionCl
         if (existingClone != null) {
             return existingClone;
         }
+
         Class<C> declaringClass = (Class<C>) original.getClass();
-        C clone = constructClone(declaringClass);
+        C clone;
+        if (original instanceof PlanningCloneable<?> planningCloneable) {
+            clone = (C) planningCloneable.createNewInstance();
+        } else {
+            clone = constructClone(declaringClass);
+        }
         originalToCloneMap.put(original, clone);
         copyFields(declaringClass, original, clone, unprocessedQueue, declaringClassMetadata);
         return clone;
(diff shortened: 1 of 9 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/PlanningCloneable.java
```diff
@@ -0,0 +1,5 @@
+package ai.timefold.solver.core.impl.domain.solution.cloner;
+
+public interface PlanningCloneable<T> {
+    T createEmptyInstance();
```
triceo, line 4: I'd like this interface to have wider use. It should not only serve for collections, it should serve for any type. 

To that end, the method should be called `createNewInstance()`, and the cloner should be adapted to handle it. As a test case, if it can clone a cloneable planning entity inside of a non-cloneable planning solution, that will be enough proof.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
```diff
@@ -228,12 +232,16 @@ private <K, V> Map<K, V> cloneMap(Class<?> expectedType, Map<K, V> originalMap,
...
+    @SuppressWarnings("unchecked")
     private static <K, V> Map<K, V> constructCloneMap(Map<K, V> originalMap) {
         // Normally, a Map will never be selected for cloning, but extending implementations might anyway.
         if (originalMap instanceof SortedMap<K, V> map) {
             var setComparator = map.comparator();
             return new TreeMap<>(setComparator);
         }
+        if (originalMap instanceof PlanningCloneable<?> planningClonable) {
+            return (Map<K, V>) planningClonable.createEmptyInstance();
+        }
```
triceo, line 244: Why shouldn't this go first? Why does `TreeMap` get precedence?

triceo (changes requested):
Overall LGTM, but the scope should be increased. It is counter-intuitive that this should only ever apply to collections.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/PlanningCloneable.java
```diff
@@ -0,0 +1,5 @@
+package ai.timefold.solver.core.impl.domain.solution.cloner;
+
+public interface PlanningCloneable<T> {
```
triceo, line 3: The interface needs a good description of what it's for, what will happen when used, and how the method should be implemented.

triceo (approved):

PR #728, opened 2024-03-19
ci: introduce Java 22 and split Native into its own job

triceo:
Will only become green after Java 22 is universally available.


rsynek (approved):
LGTM

PR #732, opened 2024-03-21
docs: Spring and Quarks multithread documentation

zepfred:
This pull request includes documentation on how to enable multithreaded solving for Spring and Quarks, along with benchmark comparisons.

This solves TimefoldAi/timefold-solver-enterprise#77


triceo:
Before I do a thorough review, I propose we do the following first:

- Remove "one thread". No point in showing that to people, that setting makes no sense in production.
- Don't show the charts, keep the tables. The charts are confusing - they show "one thread", and in one of them "one thread" is even the favorite! (What???) The tables will be enough, we simply want to show the benefit.
- Move the Quarkus/Spring multi-threaded sections to the Enterprise part of the docs. Then you'll don't have to explain how to switch to Enterprise edition, because that's already part of that doc.
- I'd still reference the enterprise docs from the `integration` chapter; if people land in that chapter, let's make sure they find their way to the enterprise docs.

triceo:
Replaced by https://github.com/TimefoldAI/timefold-solver/pull/734

PR #749, opened 2024-03-26
fix: add rebase support to NoChangeMove and ListUnassignMove

triceo:
Otherwise construction heuristics is going
to fail with multi-threaded solving,
when list variable with unassigned values is used.


rsynek (approved):
LGTM

PR #759, opened 2024-03-28
chore: address CodeQL warnings and fix release script

triceo:


core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/CompositeValueRangeDescriptor.java
```diff
@@ -52,39 +52,45 @@ public boolean isEntityIndependent() {
 
     @Override
     public ValueRange<?> extractValueRange(Solution_ solution, Object entity) {
-        List<CountableValueRange<?>> childValueRangeList = new ArrayList<>(childValueRangeDescriptorList.size());
-        for (ValueRangeDescriptor<Solution_> valueRangeDescriptor : childValueRangeDescriptorList) {
-            childValueRangeList.add((CountableValueRange) valueRangeDescriptor.extractValueRange(solution, entity));
+        return innerExtractValueRange(solution, entity);
```
zepfred, line 55: Do we have tests to validate the changes?

triceo: This isn't really a change - the underlying method is exactly the same, I just a compile-time type check to avoid some compiler warnings.

But to answer your question - no, it seems that `CompositeValueRange` is entirely untested. One of the 10+ years old pieces of code which I didn't even know we had.

zepfred: Okay. Assuming you have tested it and it is working, my point is that even minor changes could introduce bugs.

zepfred (approved):

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/CompositeValueRangeDescriptor.java
triceo: You're right to be concerned. I checked what I could, I believe we'll be fine.

PR #761, opened 2024-03-28
fix: Benchmark duplicated metrics

zepfred:
This pull request resolves a bug in the benchmark module where metrics are duplicated when running it multiple times, leading to inconsistent results.

The problem is caused by a combination of an existing benchmark config that is created after the first run and the property `inheritedSolverBenchmark`. Therefore,  the method `ai.timefold.solver.benchmark.impl.DefaultPlannerBenchmarkFactory#buildEffectiveSolverBenchmarkConfigList` will duplicate the values when inheriting the configuration at the line `#163`.

The proposed solution fixes the problem by avoiding the duplicates in the config class `ProblemBenchmarksConfig`.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
@@ -18,10 +18,12 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -204,6 +204,20 @@ public static <T> List<T> inheritMergeableListProperty(List<T> originalList, Lis
...
+        if (inheritedList == null) {
+            return originalList;
+        } else if (originalList == null) {
+            // Shallow clone due to modifications after calling inherit
+            return new ArrayList<>(inheritedList);
+        } else {
+            // The inheritedList should be before the originalList
+            List<T> mergedList = new ArrayList<>(inheritedList);
+            mergedList.addAll(originalList.stream().filter(v -> !mergedList.contains(v)).toList());
+            return mergedList;
```
triceo, line 217: I think it'd read easier if it were instead:

    Set<T> mergedSet = new LinkedHashSet<>(originalList);
    mergedSet.addAll(inheritedList);
    return new ArrayList<>(mergedSet);

Maintains insertion order, doesn't need to use any streams, and absolute efficiency is not required here.

triceo (approved):
Nice, simple fix! Left a suggestion.

PR #766, opened 2024-04-02
fix: Add fail-fast when node sharing is enabled in SolverConfig instead of application.properties

Christopher-Chianelli:


quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -330,6 +331,21 @@ Some solver configs (%s) don't specify a %s class, yet there are multiple availa
...
 
+    private void assertNodeSharingDisabled(Map<String, SolverConfig> solverConfigMap) {
+        for (var entry : solverConfigMap.entrySet()) {
+            var solverConfig = entry.getValue();
+            if (solverConfig.getScoreDirectorFactoryConfig() != null &&
+                    Boolean.TRUE
+                            .equals(solverConfig.getScoreDirectorFactoryConfig().getConstraintStreamAutomaticNodeSharing())) {
+                throw new IllegalStateException(
+                        ("SolverConfig %s enabled automatic node sharing via SolverConfig, which is not allowed." +
+                                " Enable automatic node sharing with the property %s instead.")
```
triceo, line 342: You probably wanted to get the string on a single line, but I don't think it's necessary.
In fact, the suggestion typically follows on the second line; in which case, you can use the raw string and can avoid the parens and the concat.

triceo (changes requested):
I left a minor comment. But more importantly, isn't the vast majority of node-sharing code in Enterprise? I think the fail fast should be there too.

We have the config in core, because there is no other way to do this, but let's keep the related code as much in Enterprise as possible. Ideally, I'd like us to get to a situation that - when I look for something that's related to Enterprise - I immediately know I'll find it in the Enterprise repo.

Christopher-Chianelli:
The code must be validated before `SolverConfig` becomes available as a `SolverConfigBuildItem`, which means we must do it here (which is where we construct the `SolverConfigBuildItem`)

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
Christopher-Chianelli: /me forgot we had multi-line strings in Java 17

triceo:
> The code must be validated before `SolverConfig` becomes available as a `SolverConfigBuildItem`, which means we must do it here (which is where we construct the `SolverConfigBuildItem`)

This is going to bite us one day. Is there no chance of making the enterprise quarkus extension, well, extend the community quarkus extension? Do they need to be two different processors?

Christopher-Chianelli:
Must be two separate processors due to different dependencies.
SolverConfig validation is done in community, see
https://github.com/TimefoldAI/timefold-solver/blob/bc2d5a24201bab3c7d659b726f0656c711849215/quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java#L696-L706
which is also done in community despite being an enterprise feature.

triceo (approved):
Ok, nothing to do then, LGTM.

PR #770, opened 2024-04-04
fix: Simplified moves need to list their entities and values

### LINKED ISSUE #764: List variable with unassigned values doesn't support tabu search
See https://stackoverflow.com/questions/78260554/timefold-list-model-how-to-set-to-use-tabu-search

triceo:
Closes #764


Christopher-Chianelli (commented):
The changes LGTM, but need tests. Maybe a solve on list variables with Tabu local search?

triceo:
Test added @Christopher-Chianelli.

Christopher-Chianelli (approved):

PR #779, opened 2024-04-08
chore: address some Sonar-reported reliability issues

triceo:


core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -28,31 +29,44 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public class SolverScope<Solution_> {
-    protected Set<SolverMetric> solverMetricSet;
-    protected Tags monitoringTags;
-    protected int startingSolverCount;
-    protected Random workingRandom;
-    protected InnerScoreDirector<Solution_, ?> scoreDirector;
+
+    private static AtomicLong resetAtomicLongTimeMillis(AtomicLong atomicLong) {
```
zepfred, line 33: It seems off to have a method before the variable definitions. I suggest to move it after the constructor.

zepfred (approved):

PR #782, opened 2024-04-09
ci: give up on automatic start.spring.io updates

triceo:


.github/workflows/downstream_quickstarts.yml
```diff
@@ -1,8 +1,6 @@
 name: Downstream - Timefold Quickstarts
 
 on:
-  push:
-    branches: [main]
```
triceo, line 5: No point, also breaks stale repo checks.

rsynek (approved):

PR #786, opened 2024-04-11
chore: allow to mark objects as immutable

triceo:
For Python solver, we need a way to mark certain objects as immutable.
We introduce a new internal interface for that purpose, and we make the cloner and the lookups honor the request.


core/src/main/java/ai/timefold/solver/core/impl/domain/lookup/LookUpStrategyResolver.java
```diff
@@ -81,17 +38,17 @@ public LookUpStrategyResolver(DescriptorPolicy descriptorPolicy, LookUpStrategyT
      */
     public LookUpStrategy determineLookUpStrategy(Object object) {
         return decisionCache.computeIfAbsent(object.getClass(), objectClass -> {
-            if (objectClass.isEnum()) {
+            if (DeepCloningUtils.isImmutable(objectClass)) {
```
triceo, line 41: I'm wondering... should we check for arrays here as well? Those can't possible be annotated.

Christopher-Chianelli: I  think value ranges that return arrays are rare, but probably not impossible. Arrays are mutable though, and may contain reference to mutable values (i.e. PlanningEntities).

Christopher-Chianelli (approved):
Changes LGTM

core/src/main/java/ai/timefold/solver/core/impl/domain/lookup/LookUpStrategyResolver.java
triceo: That's true, but do they need to be looked up?
(The array check wouldn't be inside of `isImmutable`.)

Christopher-Chianelli: Yes, they will be looked up in `ChangeMove.rebase/SwapMove.rebase`.

PR #800, opened 2024-04-18
feat: Add the concurrent usage collector

### LINKED ISSUE #60: Move ExperimentalConstraintCollectors to core?
ExperimentalConstraintCollectors class provides a nice set of collectors that help formulate consecutive* constraints (e.g. max n consecutive working days) especially for employee rostering use cases.

If I want to use the collectors now, I have 2 options:
- copy the package `ai/timefold/solver/examples/common/experimental/*` to my project
- depend on `examples` module, which pulls in a bunch of transitive dependencies

Is it possible to provide these classes as a part of `core` module? I understand these might change with a new release, but the name clearly indicates this is not a stable API anyway.

TODO before we can do this:

- [x] Figure out a better name for `ConsecutiveInfo`.
- [ ] Figure out an API to supply custom `Point_`, not just `Integer`.
- [x] `ConsecutiveInfo` needs to give access to first and last `Sequence`, not just to `Iterable<Sequence>`.

Christopher-Chianelli:
- Renamed the consecutive interval collector to the concurrent usage collector and move it to core
- Added methods for finding maximum and minimum concurrent usage in the cluster
- Fix bugs relating to duplicate intervals in the concurrent usage collector
- Add docs


core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1955,6 +1957,263 @@ public static <A, ResultContainer_, Result_> UniConstraintCollector<A, ResultCon
...
+     *
+     * For instance, {@code [Equipment from=2, to=4] [Equipment from=3, to=5] [Equipment from=6, to=7] [Equipment from=7, to=8]}
+     * returns the following information:
+     *
+     * <pre>
+     * {@code
+     * ConcurrentUsage: [minConcurrentUsage: 1, maxConcurrentUsage: 2,
+     *                  [Equipment from=2, to=4] [Equipment from=3, to=5]],
+     *                  [minConcurrentUsage: 1, maxConcurrentUsage: 1,
+     *                  [Equipment from=6, to=7] [Equipment from=7, to=8]]
```
triceo, line 1974: Looking at the example, shouldn't the `to` rather be exclusive?
Java APIs concerning ranges typically follow conventions `[min, max)`.

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsage.java
```diff
@@ -1,11 +1,15 @@
...
-public interface IntervalCluster<Interval_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
+public interface ConcurrentUsage<Interval_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
         extends Iterable<Interval_> {
     int size();
 
     boolean hasOverlap();
 
+    int getMinimumConcurrentUsage();
+
+    int getMaximumConcurrentUsage();
```
triceo, line 11: The first method speak of overlap.
The other two methods speak of concurrent usage.
We should standardize this; personally, `minimumOverlap` is more self-explanatory to me.

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsageInfo.java
```diff
@@ -0,0 +1,16 @@
+package ai.timefold.solver.core.api.score.stream.common;
+
+public interface ConcurrentUsageInfo<Interval_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>> {
+
+    /**
+     * @return never null, an iterable that iterates through the interval clusters
+     *         contained in the collection in ascending order
+     */
+    Iterable<ConcurrentUsage<Interval_, Point_, Difference_>> getConcurrentUsages();
```
triceo, line 9: This, in my opinion, is another place where the new name "concurrent usage" doesn't really work. "get concurrent usages" sounds to me a bit like something a non-native speaker would produce.

Overlapping intervals probably wasn't ideal either. How about overlapping ranges?

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsageInfo.java
```diff
@@ -0,0 +1,16 @@
...
+
+    /**
+     * @return never null, an iterable that iterates through the breaks contained in
+     *         the collection in ascending order
+     */
+    Iterable<Break<Point_, Difference_>> getBreaks();
```
triceo, line 15: Just out of curiosity... can you think of a term for a "break" inbetween two ranges of numbers? I'm thinking maybe maths has a term for this.

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/IntervalBreak.java
```diff
@@ -11,21 +11,21 @@ public interface IntervalBreak<Interval_, Point_ extends Comparable<Point_>, Dif
     /**
      * @return never null, the interval cluster leading directly into this
      */
-    IntervalCluster<Interval_, Point_, Difference_> getPreviousIntervalCluster();
+    ConcurrentUsage<Interval_, Point_, Difference_> getPreviousConcurrentUsage();
```
triceo, line 14: getPreviousRange?

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/IntervalBreak.java
```diff
@@ -11,21 +11,21 @@ public interface IntervalBreak<Interval_, Point_ extends Comparable<Point_>, Dif
...
 
     /**
      * Return the end of the sequence before this break. For the
      * break between 6 and 10, this will return 6.
      *
      * @return never null, the item this break is directly after
      */
-    default Point_ getPreviousIntervalClusterEnd() {
-        return getPreviousIntervalCluster().getEnd();
+    default Point_ getPreviousConcurrentUsageEnd() {
```
triceo, line 27: If we're changing the difference function to be `[start, end)`, we should consider changing this too.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ConcurrentUsageBiConstraintCollector.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (!(o instanceof ConcurrentUsageBiConstraintCollector<?, ?, ?, ?, ?> that))
+            return false;
+        if (!super.equals(o))
+            return false;
+        return Objects.equals(startMap, that.startMap) && Objects.equals(endMap,
+                that.endMap) && Objects.equals(differenceFunction, that.differenceFunction);
```
triceo, line 42: Let's start learning the "new" equals...

    return o instanceof ConcurrentUsageBiConstraintCollector<?, ?, ?, ?, ?> that 
        && Objects.equals(startMap, that.startMap)
        && ...

People have started showing that the JVM doesn't really like the way everyone typically does equals.

triceo (commented):
Let's discuss the API. My two main comments:

1. The end of the range. Should it be exclusive, as is often the case in Java APIs?

2. The name "concurrent usages". I appreciate that you tried to give the collector some semantics, indicate what it's for - I think that's a good instinct. At the same time, I'm wondering if:

2a. Aren't we boxing ourselves in? If people use it for something else than concurrent resource use, they'll be confused. Maybe there is a better generic naming?

2b. Even if we want to keep semantics in the name, "concurrent usage" doesn't seem the best. Especially in the part "getConcurrentUsages", it really doesn't sound like proper English. Maybe this is a "concurrent use" collector then?

Finally - Sonar, please.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
Christopher-Chianelli: (in particular, putting `[6, 7)` and `[7, 8)` into two seperate clusters means creating a break `[7, 7)`, which has length `0`)

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/IntervalBreak.java
Christopher-Chianelli: Huh, I thought I deleted this file; it should be unused (in particular, the collector uses the same break interface as consecutive)

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsageInfo.java
Christopher-Chianelli: We probably do not want to go to full-on `math` for this; math terms can be non-intuitive. What seems to be common (although without a definition defining it formally) is "gap": https://oeis.org/search?q=gap&language=english&go=Search)

However, it should be consistent with the consecutive collector, which uses "breaks".

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsageInfo.java
Christopher-Chianelli: (`connected` is also what the term is called in math, see https://www.math.toronto.edu/ivan/mat327/docs/notes/18-connected.pdf ) (in particular, it tracks the connected components of the space defined by the union of all intervals: https://en.wikipedia.org/wiki/Connected_space#Connected_components )

triceo: Connected ranges, connected intervals, I like either of them.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
triceo: But isn't that correct?
`[6, 7)` and `[7, 8)` aren't overlapping, so there _should_ be a break between them. (Even though the length is technically zero.)

What happens with `[6, 7]` and `[7, 8]` - is that an overlap?
And how do you model a situation where you don't _want_ an overlap in that situation?

IMO we need to be very clear about the conceptual model of this thing; no matter what it is.

core/src/main/java/ai/timefold/solver/core/api/score/stream/common/ConcurrentUsageInfo.java
triceo: On the contrary, I'd actually argue that the two collectors don't need to have anything in common; they are different collectors for different use cases, they don't need to share naming. Since we're going for `connected ranges`, I think `gaps` makes sense in this context. 

I have "break" more associated with a gap between two periods of time, which is needlessly specific for this case, I think.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connectedRanges/ConnectedRangeImpl.java
```diff
@@ -1,24 +1,29 @@
-package ai.timefold.solver.examples.common.experimental.impl;
+package ai.timefold.solver.core.impl.score.stream.collector.connectedRanges;
```
triceo, line 1: I'd go for just `ranges` - we generally don't do camel case in package names.

Christopher-Chianelli: `ranges` is too vague I think. `connected_ranges` ? (from `testdata/domain/allows_unassigned` and friends, which I misremberred as being `camelCase` instead of `snake_case`).

triceo: OK.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -1955,6 +1957,263 @@ public static <A, ResultContainer_, Result_> UniConstraintCollector<A, ResultCon
...
+     *
+     * @param <A> type of the first mapped fact
+     * @param <PointType_> temporal type of the endpoints
+     * @param startMap Maps the fact to its start
+     * @param endMap Maps the fact to its end
+     * @return never null
+     */
+    public static <A, PointType_ extends Temporal & Comparable<PointType_>>
+            UniConstraintCollector<A, ?, ConnectedRangeChain<A, PointType_, Duration>>
+            toConnectedRangesByTime(Function<A, PointType_> startMap, Function<A, PointType_> endMap) {
```
triceo, line 2011:
```suggestion
            toConnectedTemporalRanges(Function<A, PointType_> startMap, Function<A, PointType_> endMap) {
```

Reads better to me.

triceo:
Also, these two I would still fix:
https://sonarcloud.io/project/issues?pullRequest=800&open=AY7zJHLnJu5nr6ccd4B4&id=ai.timefold%3Atimefold-solver
https://sonarcloud.io/project/issues?pullRequest=800&open=AY7zJHMEJu5nr6ccd4CG&id=ai.timefold%3Atimefold-solver

triceo:
Are we talking about the same method?

    Iterable<ConnectedRangeImpl<Interval_, Point_, Difference_>> removeInterval(Interval<Interval_, Point_> interval) {
        return IntervalClusterIterator::new;
    }

The method doesn't use the argument. It's also not part of the public API, so it doesn't need to keep the argument just in case we decide to use it in the future. I honestly don't see how this method removes anything.

Christopher-Chianelli:
> Are we talking about the same method?
> 
> ```
> Iterable<ConnectedRangeImpl<Interval_, Point_, Difference_>> removeInterval(Interval<Interval_, Point_> interval) {
>     return IntervalClusterIterator::new;
> }
> ```
> 
> The method doesn't use the argument. It's also not part of the public API, so it doesn't need to keep the argument just in case we decide to use it in the future. I honestly don't see how this method removes anything.

Yes, that method iterator is what removes the interval, think of it as a `flatMap`, it can return 0 interval clusters (if the removed interval is the last interval of the cluster), 1 interval cluster (if it did not split the cluster) or many clusters (if it was an interval that serves as a bridge between multiple otherwise disconnected interval clusters).

triceo:
And the `interval` argument to the method... how is that used?
I look at the code, and I see:

        return IntervalClusterIterator::new;

In other words:

       return new Iterable() {

            Iterator iterator() {
                return new IntervalClusterIterator();
            }

       }

The `interval` argument is nowhere.

Christopher-Chianelli:
It currently unused, yes, but IMO, removing it makes the calling code less clear.

Does `recalculateAfterRemoval()` or  `recalculateAfterRemoval(interval)` look more natural?

triceo:
To be honest, the code is very unclear as is - you're relying on side effects of other things; there is no obvious reason why the simple act of creating an instance of an iterator should cause the removal of anything. To make this situation a little bit less severe, you are adding an unused argument to the surrounding method, which only raises more questions.

In my opinion, when you need to do strange things to make some other part of the code look better, you should rethink the code as a whole.

Christopher-Chianelli:
The iterator did not remove anything, rather it any new interval clusters caused by the removal. Removed the method, kept the iterator.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeChainImpl.java
```diff
@@ -235,20 +239,39 @@ void removeInterval(Interval<Interval_, Point_> interval) {
...
+    }
+
     @Override
     public String toString() {
         return "Clusters {" +
-                "intervalClusters=" + getIntervalClusters() +
-                ", breaks=" + getBreaks() +
+                "intervalClusters=" + getConnectedRanges() +
+                ", breaks=" + getGaps() +
                 '}';
```
triceo, line 275: Please fix naming. Possibly in the entire file; speaks of clusters, but there no longer is any such class.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+    ConnectedRangeImpl(NavigableSet<IntervalSplitPoint<Interval_, Point_>> splitPointSet,
+            BiFunction<? super Point_, ? super Point_, ? extends Difference_> differenceFunction,
+            IntervalSplitPoint<Interval_, Point_> start) {
+        if (start == null) {
+            throw new IllegalArgumentException("start (" + start + ") is null");
+        }
+        if (differenceFunction == null) {
+            throw new IllegalArgumentException("differenceFunction (" + differenceFunction + ") is null");
+        }
```
triceo, line 31: For the sake of bravity, I'd prefer

    this.startSplitPoint = Objects.requireNonNull(start, "start");

triceo (approved):
LGTM after minor comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/RangeGapImpl.java
```diff
@@ -3,24 +3,24 @@
 import ai.timefold.solver.core.api.score.stream.common.ConnectedRange;
 import ai.timefold.solver.core.api.score.stream.common.RangeGap;
 
-final class RangeGapImpl<Interval_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
+final class RangeGapImpl<Range_, Point_ extends Comparable<Point_>, Difference_ extends Comparable<Difference_>>
```
triceo, line 6: Still clusters throughout this file.

PR #808, opened 2024-04-22
ci: don't run on MacOS

triceo:
It's too slow, and OpenRewrite is too flaky.


rsynek (approved):
Have you considered splitting tests to run at least the reliable subset of it on Mac as well?

triceo:
I did, but in the end I settled for Quickstarts, which are tested on Mac anyway.

PR #821, opened 2024-05-02
fix: forbid pinning uninitialized entities when unassigned not allowed

triceo:
Interestingly, plenty of tests were actually intentionally testing the situation where allowsUnassigned=false and yet value pinned to null, making sure the solver didn't fail in that case.

That was probably an oversight; if I had thought about it more at the time, I would've convinced myself that failing in this situation is the correct behavior. But it raises questions of backwards compatibility.

In my opinion, changing this is not a problem, because before this fix it would result in negative init score, and that would've made local search fail anyway.


Christopher-Chianelli (approved):
LGTM

Christopher-Chianelli:
> In my opinion, changing this is not a problem, because before this fix it would result in negative init score, and that would've made local search fail anyway.

Local search did not fail when it started with a negative init score; it continued normally.

triceo:
@Christopher-Chianelli You're right, the code _used to be_ there, but it must have been removed during some of the earlier work. I still believe the reported behavior is a bug and shouldn't have been allowed; I'll merge this.

PR #846, opened 2024-05-24
docs: fix ConstraintCollectors.average() example

winklerm:
Fixed the `ConstraintCollectors.average()` example referring to `ConstraintCollectors.sum()` instead.


triceo (approved):
Thank you!

PR #875, opened 2024-06-11
fix: Use custom thread factor class in SolverManager

Christopher-Chianelli:


core/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
```diff
@@ -1010,4 +1011,36 @@ void terminateScheduledSolverJobEarly_returnsInputProblem() throws ExecutionExce
...
+
+        @Override
+        public Thread newThread(Runnable runnable) {
+            return new Thread(runnable, CUSTOM_THREAD_NAME);
+        }
+    }
+
+    @Test
+    @Timeout(60)
+    void threadFactoryIsUsed() throws ExecutionException, InterruptedException {
```
zepfred, line 1026: Nice!

zepfred (approved):

triceo:
Why is this being done? Do we have a request somewhere, a justification?
This is highly unusual.

triceo:
If this is something for the Python solver, I'd prefer if we had a code branch specifically for the Python solver, and leave the original behavior of the Java solver unchanged.

Christopher-Chianelli:
Two reasons:

- It a genuine bug: `threadFactoryClass` on `SolverManager` literally **did nothing** before when used on `SolverManager`
- Due to how Python/JVM interacts, Python Solver requires daemon threads to terminate when used with Control-C/interrupts.

Christopher-Chianelli:
(If thread factory unset in the `SolverManagerConfig`, original behaviour is unchanged; `Executors.newFixedThreadPool(parallelSolverCount, Executors.defaultThreadFactory()) == Executors.newFixedThreadPool(parallelSolverCount)`

zepfred:
@triceo, this change does not affect the default behavior, which already sets the thread factory to `Executors.defaultThreadFactory()` when calling `Executors.newFixedThreadPool(parallelSolverCount);`.

PR #898, opened 2024-06-20
fix: fail fast when local search starts from an uninitialized solution

triceo:


core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
```diff
@@ -184,28 +180,29 @@ public void removePhaseLifecycleListener(PhaseLifecycleListener<Solution_> phase
...
-                        }
-                    }
-                    throw new IllegalStateException(getPhaseTypeString() + " phase (" + phaseIndex
-                            + ") needs to start from an initialized solution, but the planning variable (" + variableRef
-                            + ") is uninitialized for the entity (" + entity + ").\n"
-                            + "Maybe there is no Construction Heuristic configured before this phase to initialize the solution.\n"
-                            + "Or maybe the getter/setters of your planning variables in your domain classes aren't implemented correctly.");
-                }
-            });
+            var scoreDirector = phaseScope.getScoreDirector();
```
zepfred, line 183: The logic looks good, but we should include a test case that covers this behavior.

zepfred (approved):

PR #900, opened 2024-06-20
ci: Add downstream Python CI

Christopher-Chianelli:


triceo (commented):
Looks good, but please introduce Enterprise Python as well. (The Java solver does the same.)

triceo (approved):
Let's see this work in practice.

PR #902, opened 2024-06-21
ci: Add Python quickstarts CI

Christopher-Chianelli:


triceo (approved):
Let's see this work in practice.

PR #915, opened 2024-06-24
feat: Allow to consume the initialized solution

zepfred:
By definition, the initialized solution is always generated by the phase immediately prior to the first Local Search phase.

This pull request adds a new consumer, allowing the solver to inject the initialized solution when a valid initialization phase is identified.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
@@ -72,7 +72,8 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
 
     /**
      * Sets the initialized solution consumer, which is called before starting the first
-     * {@link ai.timefold.solver.core.impl.localsearch.LocalSearchPhase} phase.
+     * {@link ai.timefold.solver.core.impl.localsearch.LocalSearchPhase} or
+     * {@link ai.timefold.solver.core.impl.phase.custom.CustomPhase} phases.
      *
      * @param initializedSolutionConsumer never null, called only once before starting the first Local Search phase
      * @return this, never null
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
@@ -71,15 +71,15 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
             withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
 
     /**
-     * Sets the initialized solution consumer, which is called before starting the first
-     * {@link ai.timefold.solver.core.impl.localsearch.LocalSearchPhase} or
-     * {@link ai.timefold.solver.core.impl.phase.custom.CustomPhase} phases.
+     * Sets the consumer of the first initialized solution. First initialized solution is the solution at the end of
+     * the last phase that immediately precedes the first local search phase. This solution marks the beginning of actual
+     * optimization process.
      *
-     * @param initializedSolutionConsumer never null, called only once before starting the first Local Search phase
+     * @param firstInitializedSolutionConsumer never null, called only once before starting the first Local Search phase
      * @return this, never null
      */
     SolverJobBuilder<Solution_, ProblemId_>
-            withInitializedSolutionConsumer(Consumer<? super Solution_> initializedSolutionConsumer);
+            withFirstInitializedSolutionConsumer(Consumer<? super Solution_> firstInitializedSolutionConsumer);
 
     /**
      * Sets the custom exception handler.
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
@@ -71,6 +71,16 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
             withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
 
     /**
+     * Sets the initialized solution consumer, which is called before starting the first
+     * {@link ai.timefold.solver.core.impl.localsearch.LocalSearchPhase} phase.
+     *
+     * @param initializedSolutionConsumer never null, called only once before starting the first Local Search phase
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_>
+            withInitializedSolutionConsumer(Consumer<? super Solution_> initializedSolutionConsumer);
+
+    /**
      * Sets the custom exception handler.
      *
      * @param exceptionHandler never null, called if an exception or error occurs. If null it defaults to logging the
--- a/core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
@@ -229,7 +229,7 @@ public abstract class AbstractPhase<Solution_> implements Phase<Solution_> {
         }
 
         protected Builder(int phaseIndex, boolean triggerFirstInitializedSolutionEvent, String logIndentation,
-                          Termination<Solution_> phaseTermination) {
+                Termination<Solution_> phaseTermination) {
             this.phaseIndex = phaseIndex;
             this.triggerFirstInitializedSolutionEvent = triggerFirstInitializedSolutionEvent;
             this.logIndentation = logIndentation;
--- a/core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
@@ -32,7 +32,7 @@ public abstract class AbstractPhase<Solution_> implements Phase<Solution_> {
     protected final boolean assertStepScoreFromScratch;
     protected final boolean assertExpectedStepScore;
     protected final boolean assertShadowVariablesAreNotStaleAfterStep;
-    protected final boolean initializationPhase;
+    protected final boolean triggerFirstInitializedSolutionEvent;
 
     /** Used for {@link #addPhaseLifecycleListener(PhaseLifecycleListener)}. */
     protected PhaseLifecycleSupport<Solution_> phaseLifecycleSupport = new PhaseLifecycleSupport<>();
@@ -46,7 +46,7 @@ public abstract class AbstractPhase<Solution_> implements Phase<Solution_> {
         assertStepScoreFromScratch = builder.assertStepScoreFromScratch;
         assertExpectedStepScore = builder.assertExpectedStepScore;
         assertShadowVariablesAreNotStaleAfterStep = builder.assertShadowVariablesAreNotStaleAfterStep;
-        initializationPhase = builder.initializationPhase;
+        triggerFirstInitializedSolutionEvent = builder.triggerFirstInitializedSolutionEvent;
     }
 
     public int getPhaseIndex() {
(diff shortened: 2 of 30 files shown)


zepfred:
The job `Downstream - Timefold Solver Enterprise for Python` is failing because it is not fetching the correct branch version. The fix is included in this pull request, but it needs to be merged to take effect.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -70,6 +70,17 @@ default SolverJobBuilder<Solution_, ProblemId_> withProblem(Solution_ problem) {
     SolverJobBuilder<Solution_, ProblemId_>
             withFinalBestSolutionConsumer(Consumer<? super Solution_> finalBestSolutionConsumer);
 
+    /**
+     * Sets the initialized solution consumer, which is called before starting the first
+     * {@link ai.timefold.solver.core.impl.localsearch.LocalSearchPhase} or
+     * {@link ai.timefold.solver.core.impl.phase.custom.CustomPhase} phases.
```
triceo, line 76: This is not what was agreed, is it?
First local search, yes. First custom phase, no. Custom phase may still be initializing.

Also, arguably, the definition deserves a better explanation:

"Sets the consumer of the first initialized solution. First initialized solution is the solution at the end of the last phase that immediately precedes the first local search phase. This solution marks the beginning of actual optimization."

Also, I'd consider changing the name of the concept from "initialized solution" to "first initialized solution", to clearly explain what is the significance of it.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJob.java
```diff
@@ -275,4 +281,23 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+
+        private final ConsumerSupport<Solution_, ProblemId_> consumerSupport;
+
+        public InitializedSolutionPhaseListener(ConsumerSupport<Solution_, ProblemId_> consumerSupport) {
+            this.consumerSupport = consumerSupport;
+        }
+
+        @Override
+        public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
+            if (phaseScope.isInitializationPhase()) {
```
triceo, line 298: This seems to proliferate both to the phase, and to the phase scope. Is there any chance how we can only keep it on one of the two? It feels to me like this change touches more places than it needs to; if the information is available on the phase, or if it is available on the scope, why would it also need to be available on the other?

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -42,6 +44,27 @@ void consumeIntermediateBestSolution(Solution_ bestSolution, BooleanSupplier isE
         }
     }
 
+    // Called on the Solver thread.
+    void consumeInitializedSolution(Solution_ initializedSolution) {
+        if (initializedSolutionConsumer != null) {
+            // TODO - Do we need to sync it?
```
triceo, line 50: Why would we need to?

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -42,6 +44,27 @@ void consumeIntermediateBestSolution(Solution_ bestSolution, BooleanSupplier isE
...
+                throw new IllegalStateException("Interrupted when waiting for the final best solution consumption.");
+            }
+            try {
+                initializedSolutionConsumer.accept(initializedSolution);
+            } catch (Throwable throwable) {
+                exceptionHandler.accept(problemId, throwable);
+            } finally {
+                activeConsumption.release();
+            }
+        }
```
triceo, line 65: Isn't this a bit over-complicated? Why all the locking? We're calling a simple listener.

I'm not versed in this particular aspect of the code; perhaps @rsynek could explain?

core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
```diff
@@ -77,6 +79,11 @@ public boolean isAssertShadowVariablesAreNotStaleAfterStep() {
 
     public abstract String getPhaseTypeString();
 
+    @Override
+    public boolean isInitializationPhase() {
```
triceo, line 83: Arguably this needs a better name. What does it mean to be an "initialization phase"?

I'd be more literal here. I'd call this "triggersFirstInitializedSolutionEvent()". Javadoc can then reference the definition of what that event is.

triceo (changes requested):
Good start! Please see comments inline.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
zepfred: I don't think we need to sync it, unless I'm missing something.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
zepfred: I thought we had agreed on the first custom phase as well, which could be a custom local search.

image

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJob.java
zepfred: Okay. Let me check it.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
triceo: I believe that the part that you're sharing here was only a by-product of the conversation.
At the bottom of the document, you see proposal C saying that we only use the first Local Search.

We can not use a custom phase, because custom phase _could_ be some sort of local search, but it could also be a custom construction heuristic. The only place where we can be _sure_ is before the first actual local search.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
zepfred: I agree regarding the complexity. The initial goal was to bring this up for discussion because I don't think there is a need for a synchronous call.

I have simplified the logic and will now wait for comments from @rsynek.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
zepfred: Okay. It makes sense.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
rsynek: Generally speaking, if the consumption happens on the solver thread and it's a one-time event, I don't see any need of locking.

On the other hand, all the BSEs are being consumed on a consumer thread that is paired with the solver thread.

The reason is that consuming the solution is typically an IO operation that can delay the Solver. That being said, I leave it up to you to decide whether we keep such behavior (for a single solution, it might not be such a big deal, after all).

If you decide to move the consumption on the consumer thread, then locking might be necessary. Take a look at the final BSE consumption; the principle should be very similar.

rsynek: The Init event should in theory be always the first one, so the consumer will always grab it. However, in case of real-time planning, there might be a pending BSE waiting for consumption, that's why acquiring the `activeConsumption` lock might be necessary.

After the previous BSE is consumed, the Init event should have the right of way.

zepfred: Very useful information indeed. That makes sense, and I'll address that.

triceo:
@rsynek Can you please review the async-related stuff?

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -58,6 +69,12 @@ void consumeFinalBestSolution(Solution_ finalBestSolution) {
         if (bestSolutionConsumer != null) {
             scheduleIntermediateBestSolutionConsumption();
         }
+        // Make sure the first initialized solution is consumed before the final best solution consumer shutdown the executor.
```
zepfred, line 72: @rsynek, some tests indicate that the first initialized consumer may not run because the final solution starts and shuts down the executor before its execution.

I believe that may not occur in production because the LS wouldn't finish before the first initialized solution is consumed. Let me know if you think I should update the tests instead.

https://github.com/TimefoldAI/timefold-solver/actions/runs/9668030108/job/26671216009?pr=915
https://github.com/TimefoldAI/timefold-solver/actions/runs/9668030300/job/26671219202?pr=915

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -42,12 +47,29 @@ void consumeIntermediateBestSolution(Solution_ bestSolution, BooleanSupplier isE
...
+    }
+
     // Called on the Solver thread after Solver#solve() returns.
     void consumeFinalBestSolution(Solution_ finalBestSolution) {
         try {
             // Wait for the previous consumption to complete.
             // As the solver has already finished, holding the solver thread is not an issue.
             activeConsumption.acquire();
+            // Wait for the first solution consumption to complete
+            firstSolutionConsumption.acquire();
```
rsynek, line 72: Can the following (in theory) happen?

1. here, we acquire the `firstSolutionConsumption` lock
2. now if the init event comes, it waits on https://github.com/TimefoldAI/timefold-solver/pull/915/files#diff-07ca1817df09fb6ca99e48c7070b6e09d9992a9d31e7b02bd821bf62ecf52dedR55
3. the consumer finishes the final best solution and disposes the consumer thread
4. ` scheduleFirstInitializedSolutionConsumption` then fails due to the executor not accepting further tasks.

The important question is whether the final BSE can come before the init event and intuitively, I would say no.

rsynek (approved):
Nice job, @zepfred ! 

I described one hypothetical problematic situation, but the question is whether it can even happen.

core/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
```diff
@@ -263,6 +264,173 @@ void solveGenerics() throws ExecutionException, InterruptedException {
         solverJob.getFinalBestSolution();
     }
 
+    @Test
+    @Timeout(60)
+    void solveWithFirstInitializedSolutionConsumer() throws ExecutionException, InterruptedException {
```
triceo, line 269: This test is actually several independent tests in one; only CH, CH-LS, etc. Let's split them.

triceo (approved):
LGTM after comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
zepfred: Yes, I agree with you and couldn't imagine a scenario where BSE could occur before the init event. The init event will always be triggered at the end of the custom or CH phase immediately before an LS phase. So, the `consumeFirstInitializedSolution` will acquire the lock before the LS executes, as only one phase will try to lock  `firstSolutionConsumption`.

triceo: Devil's advocate here: the final CH (the same phase which triggers the FISE) also triggers a BSE. (FIS == BS most likely) Does this change anything?

zepfred: This situation is impossible. A final CH won't try to consume the FISE because it must immediately precede a LS, and its [flag](https://github.com/TimefoldAI/timefold-solver/pull/915/files#diff-a5c58876daf3e8f4f7fc44827a6c525a6a55e36d9491bbeb7a566b70c81317b4R305) wouldn't be enabled.

PR #934, opened 2024-07-04
fix: Do not throw an exception when the solution or entity classes are interfaces

Christopher-Chianelli:
The superclass of an interface is `null`. This caused this line of code to throw an exception for solution classes that are interface:

```java
var superclass = bottomClass.getSuperclass();
lineageClassList.addAll(getAllAnnotatedLineageClasses(superclass, annotation));
```

Since `getAllAnnotatedLineageClasses` expected superclass to not be `null`. `getAllAnnotatedLineageClasses` now returns an empty list for `null` arguments.


triceo:
I'm wondering if solutions as interfaces _should_ be supported. What would be the use case? What would be the downsides?

Considering that in all the years we have not seen anyone ask us for this, maybe we don't need this.

triceo (approved):

PR #935, opened 2024-07-05
fix: Continue the benchmark warm-up when one configuration fails

zepfred:
This pull request addresses a benchmark failure that occurs due to a bad configuration used during the warm-up phase.

image


triceo:
Can you please elaborate what was the problem and why this is the fix?

triceo:
What is the bad configuration? Why now, and not before?

triceo (approved):

PR #936, opened 2024-07-05
fix: Change the default moves for multi-variable domain models

zepfred:
This pull request updates the list of default moves for multi-variable domain models.


triceo (approved):

triceo:
I'm not entirely sure why the test coverage for this is in the Enterprise repo.
This doesn't need enterprise, does it? If not, the coverage should be here.

zepfred:
> I'm not entirely sure why the test coverage for this is in the Enterprise repo. This doesn't need enterprise, does it? If not, the coverage should be here.

That's correct, and it doesn't require the enterprise module. I could move it to the solver repo, but we won't test the scenario where the user enables nearby auto-configuration on a multi-variable domain model.

triceo:
Personally, I don't think we need to test it in this way at all.
Just use the factory and test that it returns the correct defaults.

triceo (approved):

PR #945, opened 2024-07-10
chore: deprecate constraint package

triceo:
Constraint packages are still used internally, as there is no way to remove them without breaking backwards compatibility. 
All user-facing references to them have been deprecated though.

### DIFF
--- a/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreStatisticPoint.java
+++ b/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreStatisticPoint.java
@@ -23,12 +23,8 @@ public class ConstraintMatchTotalBestScoreStatisticPoint extends StatisticPoint 
         return timeMillisSpent;
     }
 
-    public String getConstraintPackage() {
-        return constraintRef.packageName();
-    }
-
-    public String getConstraintName() {
-        return constraintRef.constraintName();
+    public ConstraintRef getConstraintRef() {
+        return constraintRef;
     }
 
     public int getConstraintMatchCount() {
@@ -39,13 +35,9 @@ public class ConstraintMatchTotalBestScoreStatisticPoint extends StatisticPoint 
         return scoreTotal;
     }
 
-    public String getConstraintId() {
-        return constraintRef.constraintId();
-    }
-
     @Override
     public String toCsvLine() {
-        return buildCsvLineWithStrings(timeMillisSpent, getConstraintPackage(), getConstraintName(),
+        return buildCsvLineWithStrings(timeMillisSpent, constraintRef.packageName(), constraintRef.constraintName(),
                 Integer.toString(constraintMatchCount), scoreTotal.toString());
     }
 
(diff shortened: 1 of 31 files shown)


benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreStatisticPoint.java
```diff
@@ -23,12 +23,8 @@ public long getTimeMillisSpent() {
         return timeMillisSpent;
     }
 
-    public String getConstraintPackage() {
```
zepfred, line 26: Should we first deprecate it before removing it?

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreStatisticPoint.java
```diff
@@ -23,12 +23,8 @@ public long getTimeMillisSpent() {
...
-        return constraintRef.packageName();
-    }
-
-    public String getConstraintName() {
```
zepfred, line 30: Same as above

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/constraintmatchtotalbestscore/ConstraintMatchTotalBestScoreStatisticPoint.java
zepfred: I'm assuming this is an 'internal' class and won't break the user's code. If so, please ignore my previous comment.

triceo: Yes, this class is internal, so I just went ahead and removed it. The information is there in `ConstraintRef` anyway.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/common/bi/AbstractBiConstraintStreamTest.java
```diff
@@ -942,16 +942,16 @@ public void groupBy_1Mapping1Collector() {
         // From scratch
         scoreDirector.setWorkingSolution(solution);
         assertScore(scoreDirector,
-                assertMatchWithScore(-1, TEST_CONSTRAINT_NAME, solution.getFirstEntity().toString(), 6),
-                assertMatchWithScore(-1, TEST_CONSTRAINT_NAME, solution.getEntityList().get(1).toString(), 5));
+                assertMatchWithScore(-1, null, TEST_CONSTRAINT_NAME, solution.getFirstEntity().toString(), 6),
```
zepfred, line 945: Are we still using the constraint package elsewhere? If not, why not remove it instead of using a `null` value?

triceo: I needed to be able to test what happens when there is no package. 
Which means I need to have a way to provide the package. (Which, until now, didn't exist in the test.)

zepfred (approved):
Looks good!

PR #946, opened 2024-07-11
feat: constraint weight override

triceo:
Planning solution gets an optional field of type `Map` (or more likely a custom map), with keys being the `constraintName` from `asConstraint(constraintName)`, and the value being the constraint weight. 
Using the map is optional. Constraint weights not present in the map use their default constraint weights as defined by `penalize(...)`. Constraints present in the map but not present in the `ConstraintProvider` trigger a fail-fast.

`asConstraintDescribed(constraintName, constraintDescription)` is introduced to keep constraint documentation where it belongs, with the constraints.

`ConstraintWeight`, `ConstraintConfiguration` and `penalizeConfigurable(...)` are deprecated. They continue working, but the new system doesn't use them.


triceo:
@lee-carlon This PR will continue evolving, but the docs are likely already stable. As always, feedback is welcome.

docs/src/modules/ROOT/pages/constraints-and-score/constraint-configuration.adoc
```diff
@@ -1,21 +1,139 @@
...
+    )
+);
+
+var solution = new VehicleRoutePlan();
+solution.setConstraintWeightOverrides(constraintWeightOverrides);
+
+...
+----
+
+The `Vehicle capacity` constraint in this planning solution will get a weight of `2hard`,
```
lee-carlon, line 118:
```suggestion
The `Vehicle capacity` constraint in this planning solution has a weight of `2hard`,
```

docs/src/modules/ROOT/pages/constraints-and-score/constraint-configuration.adoc
```diff
@@ -1,21 +1,139 @@
...
+as opposed to its original `1hard`.
+The `Service finished after max end time` constraint will have a weight of `0hard`,
```
lee-carlon, line 120:
```suggestion
The `Service finished after max end time` constraint has a weight of `0hard`,
```

docs/src/modules/ROOT/pages/constraints-and-score/constraint-configuration.adoc
```diff
@@ -1,21 +1,139 @@
...
+and therefore will be disabled entirely.
+
+This way, you can solve the same problem
+while applying different constraint weights to each instance.
```
lee-carlon, line 124:
```suggestion
In this way, you can solve the same problem
by applying different constraint weights to each instance.
```

lee-carlon (approved):
@triceo looks good. Just a few minor suggestions.

core/src/main/java/ai/timefold/solver/core/api/score/stream/tri/TriConstraintStream.java
```diff
@@ -1892,10 +1915,10 @@ default Constraint penalizeBigDecimal(String constraintPackage, String constrain
      * For non-int {@link Score} types use {@link #penalizeConfigurableLong(String, ToLongTriFunction)} or
      * {@link #penalizeConfigurableBigDecimal(String, TriFunction)} instead.
      *
-     * @deprecated Prefer {@link #penalizeConfigurable(ToIntTriFunction)}.
      * @param constraintName never null, shows up in {@link ConstraintMatchTotal} during score justification
      * @param matchWeigher never null, the result of this function (matchWeight) is multiplied by the constraintWeight
      * @return never null
+     * @deprecated Prefer {@link #penalize(Score, ToIntTriFunction)} and {@link ConstraintWeightOverrides}.
      */
     @Deprecated(forRemoval = true)
```
zepfred, line 1923: Should we include version `1.13.0` and update other stream classes?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/ConstraintWeightSupplier.java
```diff
@@ -0,0 +1,53 @@
...
+import ai.timefold.solver.core.api.domain.common.DomainAccessType;
+import ai.timefold.solver.core.api.domain.constraintweight.ConstraintConfiguration;
+import ai.timefold.solver.core.api.domain.solution.ConstraintWeightOverrides;
+import ai.timefold.solver.core.api.score.Score;
+import ai.timefold.solver.core.api.score.constraint.ConstraintRef;
+import ai.timefold.solver.core.api.score.stream.ConstraintProvider;
+import ai.timefold.solver.core.impl.domain.common.accessor.MemberAccessorFactory;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+public sealed interface ConstraintWeightSupplier<Solution_, Score_ extends Score<Score_>>
```
zepfred, line 14: Nice refactoring!

core/src/main/java/ai/timefold/solver/core/api/score/stream/tri/TriConstraintStream.java
triceo: Not sure here. I left it here like this, because it was already deprecated before. Saying it's "since 1.13.0" wouldn't be true, because it was deprecated years ago already.

migration/src/main/java/ai/timefold/solver/migration/v8/AsConstraintRecipe.java
```diff
@@ -237,9 +238,8 @@ public Expression visitExpression(Expression expression, ExecutionContext execut
                         } else if (select.getType().isAssignableFrom(quadConstraintStreamPattern)) {
                             templateCode = "#{any(ai.timefold.solver.core.api.score.stream.quad.QuadConstraintStream)}\n";
                         } else {
-                            LOGGER.warn("Cannot refactor to asConstraint() method" +
-                                    " for deprecated method called in expression (" + e + ").");
-                            return e;
+                            LOGGER.warn("Cannot refactor to asConstraint() method for deprecated method (" + method + ").");
```
zepfred, line 241:
```suggestion
                            LOGGER.warn("Cannot refactor to asConstraint() method for deprecated method ({}).", method);
```

zepfred (approved):
Great job. The new configuration mechanism looks much better!

zepfred:
It looks like the Python module is broken.

triceo:
Python needs a follow-up PR. I didn't provide one yet, because we're in the middle of moving repositories.

triceo:
@Christopher-Chianelli I've made several Python changes:

- Removed several deprecated methods.
- Added equivalent to `asConstraintDescribed(...)`.

I need your review on those.

I also attempted (but not finished) to redefine `ConstraintWeightOverrides` in Python. (Using the native Java one didn't work, as it kept complaining of incompatible generics on Java `HashMap`.) Please finish this work, so that I could learn from it.

python/python-core/src/main/python/domain/_types.py
```diff
@@ -0,0 +1,18 @@
+from ..score import Score
+from typing import Dict, Generic, Type, TypeVar, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ai.timefold.solver.core.api.domain.solution import ConstraintWeightOverrides as _JavaConstraintWeightOverrides
+
+Score_ = TypeVar('Score_', bound='Score')
+
+class ConstraintWeightOverrides(Generic[Score_]):
```
triceo, line 9: Needs finishing.

python/python-core/tests/test_solver_configuration.py
```diff
@@ -5,8 +5,7 @@
 
 from dataclasses import dataclass, field
 from typing import Annotated, List
-
-
+"""
```
triceo, line 8: Needs enabling once overrides work.

triceo:
Thanks, @Christopher-Chianelli!

PR #954, opened 2024-07-16
feat: add new Shadow variable listener

zepfred:
This pull request introduces a new shadow variable that automatically updates a list of elements in a planning list variable. It sources inverse and previous elements shadow variables and cascades updates to the next elements of a given domain entity.

```
public class Visit implements LocationAware {
    ...
    @InverseRelationShadowVariable(sourceVariableName = "visits")
    private Vehicle vehicle;
    @PreviousElementShadowVariable(sourceVariableName = "visits")
    private Visit previousVisit;
    @NextElementShadowVariable(sourceVariableName = "visits")
    private Visit nextVisit;
    @CascadeUpdateElementShadowVariable(sourceMethodName = "updateArrivalTime")
    private LocalDateTime arrivalTime;
    ...
    public void updateArrivalTime() {
        if (previousVisit == null && vehicle == null) {
            arrivalTime = null;
            return;
        }
        LocalDateTime departureTime = previousVisit == null ? vehicle.getDepartureTime() : previousVisit.getDepartureTime();
        arrivalTime = departureTime != null ? departureTime.plusSeconds(getDrivingTimeSecondsFromPreviousStandstill()) : null;
    }
```

In short, the new listener calls `updateArrivalTime` whenever the `vehicle` or `previousVisit` changes. It automatically cascades the update to `nextVisit` and stops when the `arrivalTime` value does not change after calling `updateArrivalTime` or when it reaches the tail.


core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadeUpdateElementShadowVariable.java
```diff
@@ -0,0 +1,32 @@
...
+ * <p>
+ * Important: it must only change the shadow variable(s) for which it's configured!
+ * It can be applied to multiple fields to modify different shadow variables.
+ * It should never change a genuine variable or a problem fact.
+ * It can change its shadow variable(s) on multiple entity instances
+ * (for example: an arrivalTime change affects all trailing entities too).
+ */
+@Target({ FIELD })
+@Retention(RUNTIME)
+public @interface CascadeUpdateElementShadowVariable {
```
triceo, line 24: TODO: Naming.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadeUpdateElementShadowVariable.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * A listener automatically sourced on a {@link PreviousElementShadowVariable} and {@link InverseRelationShadowVariable}.
+ * <p>
+ * Change shadow variables when the previous or inverse relation elements changes.
+ * <p>
+ * Automatically cascades change events to {@link NextElementShadowVariable} of a {@link PlanningListVariable}.
+ * <p>
+ * Important: it must only change the shadow variable(s) for which it's configured!
```
triceo, line 16:
```suggestion
 * Important: it must only change the shadow variable(s) for which it's configured.
```

We generally don't shout at people. :-)

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionMethodMemberAccessor.java
```diff
@@ -19,6 +19,10 @@ public final class ReflectionMethodMemberAccessor extends AbstractMemberAccessor
     private final MethodHandle methodHandle;
 
     public ReflectionMethodMemberAccessor(Method readMethod) {
+        this(readMethod, true);
+    }
+
+    public ReflectionMethodMemberAccessor(Method readMethod, boolean requiredReturnType) {
```
triceo, line 25: I think `returnTypeRequired` makes it more obvious that it's a `boolean`.
Please apply consistently throughout the codebase.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorImplementor.java
```diff
@@ -161,12 +162,17 @@ private static void createConstructor(ClassCreator classCreator, GizmoMemberInfo
...
                     methodCreator.newArray(Class.class, 0));
-            ResultHandle type =
-                    methodCreator.invokeVirtualMethod(
-                            MethodDescriptor.ofMethod(Method.class, "getGenericReturnType", Type.class),
-                            method);
-            methodCreator.writeInstanceField(FieldDescriptor.of(classCreator.getClassName(), GENERIC_TYPE_FIELD, Type.class),
-                    thisObj, type);
+            if (memberInfo.isRequiredReturnType()) {
+                // We create a field to store the result, only if the called method has a return type; otherwise, we
+                // will only execute it
```
triceo, line 167:
```suggestion
                // We create a field to store the result, only if the called method has a return type. 
                // Otherwise, we will only execute it.
```

It reads better if the lines are split by sentence, if possible.
(The IntelliJ Grazie Pro plugin helps with writing a lot.)

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorImplementor.java
```diff
@@ -246,9 +254,8 @@ private static void createGetName(ClassCreator classCreator, GizmoMemberInfo mem
 
         // If it is a method, assert that it has the required
         // properties
-        memberInfo.getDescriptor().whenIsMethod(method -> {
-            assertIsGoodMethod(method, memberInfo.getAnnotationClass());
-        });
+        memberInfo.getDescriptor().whenIsMethod(
+                method -> assertIsGoodMethod(method, memberInfo.isRequiredReturnType(), memberInfo.getAnnotationClass()));
```
triceo, line 258: It's a race to the finish line now. :-) We're going to have conflicts, you and I.
Nevermind, we'll figure it out when the time comes.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
```diff
@@ -0,0 +1,135 @@
...
+                .findFirst()
+                .orElse(null);
+        if (inverseRelationShadowDescriptor == null) {
+            throw new IllegalArgumentException(
+                    """
+                            The entityClass (%s) has an @%s annotated property (%s), but has no @InverseRelationShadowVariable shadow variable defined.
+                            Maybe add a new shadow variable @InverseRelationShadowVariable to the entity %s."""
+                            .formatted(entityDescriptor.getEntityClass(),
+                                    CascadeUpdateElementShadowVariable.class.getSimpleName(), variableMemberAccessor.getName(),
+                                    entityDescriptor.getEntityClass()));
```
triceo, line 57: I wonder if this is necessary. If the inverse is not defined, can we not use a supply? List variable works even without defining these, so the code to support that is already in there.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
```diff
@@ -0,0 +1,135 @@
...
+        var previousElementShadowDescriptor = entityDescriptor.getShadowVariableDescriptors().stream()
+                .filter(variableDescriptor -> PreviousElementShadowVariableDescriptor.class
+                        .isAssignableFrom(variableDescriptor.getClass()))
+                .findFirst()
+                .orElse(null);
+        if (previousElementShadowDescriptor == null) {
+            throw new IllegalArgumentException(
+                    """
+                            The entityClass (%s) has an @%s annotated property (%s), but has no @PreviousElementShadowVariable shadow variable defined.
+                            Maybe add a new shadow variable @PreviousElementShadowVariable to the entity %s."""
```
triceo, line 70: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
```diff
@@ -0,0 +1,135 @@
...
+        nextElementShadowVariableDescriptor = entityDescriptor.getShadowVariableDescriptors().stream()
+                .filter(variableDescriptor -> NextElementShadowVariableDescriptor.class
+                        .isAssignableFrom(variableDescriptor.getClass()))
+                .findFirst()
+                .orElse(null);
+        if (nextElementShadowVariableDescriptor == null) {
+            throw new IllegalArgumentException(
+                    """
+                            The entityClass (%s) has an @%s annotated property (%s), but has no @NextElementShadowVariable shadow variable defined.
+                            Maybe add a new shadow variable @NextElementShadowVariable to the entity %s."""
```
triceo, line 86: Dtto.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateVariableListenerTest.java
```diff
@@ -0,0 +1,168 @@
...
+                        "The entityClass (class ai.timefold.solver.core.impl.testdata.domain.shadow.wrong_cascade.TestdataCascadeWrongMethod)")
+                .withMessageContaining(
+                        "has an @CascadeUpdateElementShadowVariable annotated property (badUpdateCascadeValueWithReturnType)")
+                .withMessageContaining(
+                        "with sourceMethodName (class ai.timefold.solver.core.impl.testdata.domain.shadow.wrong_cascade.TestdataCascadeWrongMethod), ")
+                .withMessageContaining("but the method has not been found in the entityClass");
+    }
+
+    @Test
+    void updateAllNextValues() {
```
triceo, line 56: Maybe we should also test the case where two listeners point to the same method?

triceo (commented):
First review done, left some comments.
Overall, I like it.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1367,6 +1367,91 @@ public class Customer {
...
+@PlanningEntity
+public class Customer {
+
+    @InverseRelationShadowVariable(sourceVariableName = "customers")
+    private Vehicle vehicle;
+    @PreviousElementShadowVariable(sourceVariableName = "customers")
+    private Customer previousCustomer;
+    @NextElementShadowVariable(sourceVariableName = "customers")
+    private Customer nextCustomer;
+    @CascadeUpdateElementShadowVariable(sourceMethodName = "updateWeightAndArrivalTime")
```
triceo, line 1447: Isn't it rather the `targetMethodName`?
The relationship is inversed here.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1367,6 +1367,91 @@ public class Customer {
     public void setNextCustomer(Customer nextCustomer) {...}
 ----
 
+[#cascadeUpdateVariable]
+==== Cascade element update shadow variable
```
triceo, line 1371: This documentation is correct, but I've been wondering...
Should we rather (or maybe in addition) write a short section on "updating tail chains"? 
And maybe we don't hide it deep in a subsection on variable listeners.

Because as a user, that's what I will look for - not "cascade element update shadow variable".

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorImplementor.java
zepfred: Yeah, I think so :-)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
zepfred: I see your point. The main reason for adding these validations is to define the [dependencies](https://github.com/TimefoldAI/timefold-solver/pull/954/files#diff-4ed0ff66350dc1ce12d0b8dd2a56f04e6410eaaeef73cc15b1a0254c08d05adfR134) that will impact the global listeners sort.

I'm not sure if it's possible to do that without the related fields.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateVariableListenerTest.java
zepfred: Done!

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
zepfred: Yes, it makes sense.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateVariableListener.java
```diff
@@ -33,12 +35,25 @@ public void beforeVariableChanged(ScoreDirector<Solution_> scoreDirector, Object
...
+                scoreDirector.afterVariableChanged(currentEntity, sourceShadowVariableDescriptor.getVariableName());
+                newValueList.add(sourceShadowVariableDescriptor.getValue(currentEntity));
+            }
+            boolean hasChange = false;
+            for (int i = 0; i < oldValueList.size(); i++) {
+                if (!Objects.equals(oldValueList.get(i), newValueList.get(i))) {
+                    hasChange = true;
+                    break;
+                }
+            }
```
triceo, line 55: Arguably you can make this check when you're reading the new value, completely avoiding this loop and improving performance.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateVariableListener.java
```diff
@@ -33,12 +35,25 @@ public void beforeVariableChanged(ScoreDirector<Solution_> scoreDirector, Object
     public void afterVariableChanged(ScoreDirector<Solution_> scoreDirector, Object entity) {
         var currentEntity = entity;
         while (currentEntity != null) {
-            scoreDirector.beforeVariableChanged(currentEntity, sourceShadowVariableDescriptor.getVariableName());
-            Object oldValue = sourceShadowVariableDescriptor.getValue(currentEntity);
+            List<Object> oldValueList = new ArrayList<>(sourceShadowVariableDescriptorList.size());
```
triceo, line 38: Personally, I'd introduce two methods for this - one which only has 1 var, and another which handles 2+ vars. The 1-var path will be significantly faster, avoiding the list and the iterations. It is also the most frequent use case, so it makes sense to have a path handling it.

triceo (commented):
Reviewing the latest code.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
triceo: Don't you already know which fields this depends on?

Typically, a var listener is defined with a source variable.
In this case, the listener is _on_ the source variable, and it defines a target method.
You can think of the cascade shadow var as:

    @Cascade...(source = "var", target="run")
    Object var = ...

Therefore you should be depending on `var`, right? Or am I misunderstanding?

triceo: I've been wondering... how can you _know_ which vars this will depend on?
You are requiring the prev/next/etc. to exist, and you're basing it on that.
But is that correct? Why not let the user decide the source?

    `@Cascade...(targetMethodName = "...", source = "...")`

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
zepfred: I've moved the contento into a separate section and renamed the title. I believe it should be placed after `List variable shadow variables` because it relies on some shadow variables discussed in that section.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
zepfred: New approach implemented

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadeUpdateElementShadowVariableDescriptor.java
zepfred: There is a validation per source now.

triceo (changes requested):
Leaving some comments.
Also, now that we're basing it on supplies, I think it's worth it to run those benchmarks again and see what the performance is now.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1367,6 +1367,121 @@ public class Customer {
...
+    @InverseRelationShadowVariable(sourceVariableName = "customers")
+    private Vehicle vehicle;
+    @PreviousElementShadowVariable(sourceVariableName = "customers")
+    private Customer previousCustomer;
+    @NextElementShadowVariable(sourceVariableName = "customers")
+    private Customer nextCustomer;
+    @CascadingUpdateListener(targetMethodName = "updateArrivalTime", sourceVariableName = "vehicle")
+    @CascadingUpdateListener(targetMethodName = "updateArrivalTime", sourceVariableName = "previousCustomer")
+    private LocalDateTime arrivalTime;
+
```
triceo, line 1409: The very first example shouldn't be this complicated.
Reduce it to the bare minimum code. You can explain the extra features later.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberInfo.java
```diff
@@ -5,16 +5,23 @@
 public final class GizmoMemberInfo {
```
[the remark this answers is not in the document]
zepfred: I did not rebase yet.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateVariableListener.java
```diff
@@ -0,0 +1,116 @@
...
+            if (!execute(scoreDirector, currentEntity, targetVariableDescriptorList, targetMethod)) {
+                break;
+            }
+            currentEntity = getNextElement(currentEntity);
+        }
+    }
+
+    private Object getNextElement(Object entity) {
+        // The primary choice is to select the user-defined next element shadow variable.
+        return nextElementShadowVariableDescriptor != null ? nextElementShadowVariableDescriptor.getValue(entity)
```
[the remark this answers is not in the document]
zepfred: The implementation has changed after adding new base classes.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/nextprev/ExternalizedNextElementVariableSupply.java
```diff
@@ -0,0 +1,134 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.SingletonInverseVariableSupply;
+import ai.timefold.solver.core.impl.domain.variable.listener.SourcedVariableListener;
+import ai.timefold.solver.core.impl.util.IdentityElementAwareList;
+
+/**
+ * Alternative to {@link NextElementVariableListener}.
+ */
+public class ExternalizedNextElementVariableSupply<Solution_> implements
```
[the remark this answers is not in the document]
zepfred: I'm unsure if I understand that. Could you please provide more details?

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityElementAwareList.java
```diff
@@ -0,0 +1,93 @@
...
+ * This data structure differs from {@link ElementAwareList} because it provides a find operation in O(H).
+ * 
+ * The implementation relies on two different data structures: {@link java.util.IdentityHashMap} and {@link ElementAwareList}.
+ *
+ * The element comparison works in the same way as {@link java.util.IdentityHashMap},
+ * and two keys k1 and k2 are considered equal if and only if (k1==k2).
+ *
+ * @param <T> The element type.
+ */
+public class IdentityElementAwareList<T> implements Iterable<T> {
```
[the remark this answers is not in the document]
zepfred: I've renamed the class to `IndexedElementAwareList`, and I've been thinking about that. The new list class will offer constant time for find, add, and remove operations.

Everything would work fine with a `List`, except for the operation `ExternalizedNextElementVariableSupply#getNext`, which will be O(N), and the new implementation is O(H).

I suggest maintaining this class as there will be performance benefits.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/nextprev/ExternalizedNextElementVariableSupply.java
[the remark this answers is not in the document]
triceo: Now, `ExternalizedNextElementVariableSupply` is a standalone class. I'm wondering if instead it should be part of the externalized list variable state supply.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/DotNames.java
```diff
@@ -67,6 +68,8 @@ public final class DotNames {
     static final DotName PIGGYBACK_SHADOW_VARIABLE = DotName.createSimple(PiggybackShadowVariable.class.getName());
     static final DotName PREVIOUS_ELEMENT_SHADOW_VARIABLE = DotName.createSimple(PreviousElementShadowVariable.class.getName());
     static final DotName SHADOW_VARIABLE = DotName.createSimple(ShadowVariable.class.getName());
+    static final DotName CASCADE_UPDATE_ELEMENT_SHADOW_VARIABLE =
```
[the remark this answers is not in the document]
zepfred: Yes, no need for filst, because IndexView can identify multiple annotation definitions on the field.

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityElementAwareList.java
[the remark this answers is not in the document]
triceo: Doesn't a `Map` also give a constant time? I'm not sure why we're inventing a new type here.

zepfred:
@triceo, the tests are failing due to the rebase. I'll fix that, but the logic is ready for your review again.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
```diff
@@ -0,0 +1,206 @@
...
+
+        var targetMethodName = getTargetMethodName();
+        var sourceMethodMember = ConfigUtils.getDeclaredMembers(entityDescriptor.getEntityClass())
+                .stream()
+                .filter(member -> member.getName().equals(targetMethodName))
+                .findFirst()
+                .orElse(null);
+        if (sourceMethodMember == null) {
+            throw new IllegalArgumentException(
+                    "The entityClass (%s) has an @%s annotated property (%s), but the method \"%s\" is not found."
```
triceo, line 128:
```suggestion
                    "The entityClass (%s) has an @%s annotated property (%s), but the method \"%s\" cannot be found."
```

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
```diff
@@ -0,0 +1,206 @@
...
+        targetMethod = descriptorPolicy.getMemberAccessorFactory().buildAndCacheMemberAccessor(sourceMethodMember,
+                MemberAccessorFactory.MemberAccessorType.REGULAR_METHOD, null, descriptorPolicy.getDomainAccessType());
+    }
+
+    public void linkVariableDescriptorToSource(CascadingUpdateShadowVariable listener) {
+        var sourceDescriptor = entityDescriptor.getShadowVariableDescriptor(listener.sourceVariableName());
+        if (sourceDescriptor == null) {
+            throw new IllegalArgumentException(
+                    """
+                            The entityClass (%s) has an @%s annotated property (%s), but the shadow variable "%s" is not found.
```
triceo, line 143:
```suggestion
                            The entityClass (%s) has an @%s annotated property (%s), but the shadow variable "%s" cannot be found.
```

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/cascade/CollectionCascadingUpdateShadowVariableListenerTest.java
```diff
@@ -0,0 +1,171 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.impl.testdata.domain.cascade.multiple_var.supply.TestdataMultipleCascadingWithSupplyEntity;
+import ai.timefold.solver.core.impl.testdata.domain.cascade.multiple_var.supply.TestdataMultipleCascadingWithSupplySolution;
+import ai.timefold.solver.core.impl.testdata.domain.cascade.multiple_var.supply.TestdataMultipleCascadingWithSupplyValue;
+import ai.timefold.solver.core.impl.testdata.util.PlannerTestUtils;
+
+import org.junit.jupiter.api.Test;
+
+class CollectionCascadingUpdateShadowVariableListenerTest {
```
triceo, line 17: Arguably, if the test can be parameterized, we should be able to do this with 1 test class, not 4, right? The test classes are nearly identical.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1367,6 +1367,135 @@ public class Customer {
...
+    @CascadingUpdateShadowVariable(targetMethodName = "updateArrivalTime", sourceVariableName = "previousCustomer")
+    private LocalDateTime arrivalTime;
+
+    ...
+
+    public void updateArrivalTime() {...}
+----
+
+The listener behaves as described previously.
+It is called `updateArrivalTime` whenever a `previousCustomer` or `vehicle` changes.
```
triceo, line 1464:
```suggestion
It calls `updateArrivalTime` whenever a `previousCustomer` or `vehicle` changes.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1367,6 +1367,135 @@ public class Customer {
...
+It then propagates the changes to the subsequent elements,
+stopping when no change is identified, or it reaches the tail.
```
triceo, line 1466:
```suggestion
stopping when the method results in no change to the variable, or when it reaches the tail.
```

triceo (commented):
Looks good, with some comments. I still have to review the list situation; not entirely sure why we need it.

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityElementAwareList.java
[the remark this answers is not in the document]
zepfred: Lukas, after double-checking the class, I believe it makes complete sense to maintain the index, previous, and next elements in the 'ExternalizedNextElementVariableSupply'.

If `LocationInList` also includes the previous and next elements, the class `IndexedElementAwareList` will not be necessary.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/nextprev/ExternalizedNextElementVariableSupply.java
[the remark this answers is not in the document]
zepfred: Lukas, after double-checking the class, I believe it makes complete sense to maintain the index, previous, and next elements in the 'ExternalizedNextElementVariableSupply'.

If `LocationInList` also includes the previous and next elements, the class `IndexedElementAwareList` will not be necessary.

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityElementAwareList.java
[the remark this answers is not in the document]
zepfred: Looking at the `ExternalizedNextElementVariableSupply` logic, I realized that we have implemented a similar feature, but I've used the `ElementAwareListEntry` instead of `LocationInList` and delegated the responsibility to a separate data structure class.

I wonder if we should move this logic to `ExternalizedListVariableStateSupply`. Otherwise, I will create a new "LocationInList" class with the next element reference and use a similar approach as `ExternalizedListVariableStateSupply`. What do you think?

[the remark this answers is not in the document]
triceo: That's part of the reason why I suggested that `ExternalizedNextElementVariableSupply` and `ExternalizedListVariableStateSupply` shouldn't be separate. :-) It's the same logic, it should be in the same place.

zepfred:
@triceo, after considering adding the next element to `ExternalizedListVariableStateSupply`, I've decided to keep them in a separate externalized list.

`ExternalizedListVariableStateSupply` is always created when you have a planning list variable, but `ExternalizedNextElementVariableSupply` is only needed at this point for `@CascadingUpdateShadowVariable`. So, I believe we should not add it to `ExternalizedListVariableStateSupply` as it won't be necessary in general cases and could unnecessarily increase resource consumption.

PR #955, opened 2024-07-16
chore: Migrate Python repo

Christopher-Chianelli:


pom.xml
```diff
@@ -96,6 +96,22 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>non-essential</id>
```
triceo, line 100: Can you please explain what purpose this serves exactly?
Since the Python stuff apparently doesn't re-build anything, I don't understand why this profile needs to exist.
Plus I really don't like having entire modules marked as "non essential".

python/README.md
```diff
@@ -0,0 +1,186 @@
+![Timefold Logo](https://raw.githubusercontent.com/TimefoldAI/timefold-solver/main/docs/src/modules/ROOT/images/shared/timefold-logo.png)
+
+# Timefold Solver for Python
```
triceo, line 3: If the TOML stuff etc. needs to be in the root of the repo, why does the README need to be in a subfolder? And wouldn't it make sense to have one unified README for the entire repo? 

I doubt that people will click to the "python" subdir to see a Python-specific README; they'll stay at the top level, and onl\ see Java-specific stuff.

create-stubs.py
triceo, line 1: Once again I want to ask if this can be in the `python` subfolder.

python/jpyinterpreter/examples/.gitignore
triceo, line 1: Do we need the examples? Especially if we don't actually intend to promote the interpreter and are thinking of removing it?

python/jpyinterpreter/mvnw
triceo, line 1: This is duplication with elsewhere in the repo.
I'd rather reuse in CI than duplicate in Git.

python/jpyinterpreter/pyproject.toml
triceo, line 1: Suddenly this file doesn't need to be in the root of the repo?

triceo (commented):
I have reviewed the relevant stuff. I am assuming the actual Java and Python code stays the same, and therefore doesn't need reviewed.

Please check the release notes template and update it to include instructions for Python as well. As it stands right now, it is very Java-centric. 

But beware - JReleaser bugs will remove most of formatting, so don't expect tabs or spaces at the beginning of lines to be preserved in the final output. (That's why I don't actually include excerpts of `pom.xml` in there; it would come out entirely unformatted.)

pom.xml
Christopher-Chianelli: It does rebuild it, and doesn't require  the quarkus/spring/other stuff, which would only serve to pollute the Python source distribution.

python/jpyinterpreter/examples/.gitignore
Christopher-Chianelli: Can be removed

python/pom.xml
triceo, line 1: I'm wonedering... can we integrate the POM into the overall build? Would it make sense to simply build all things Java all at once, and then run the extra Python-specific bits?

python/jpyinterpreter/mvnw
Christopher-Chianelli: Cannot move; python build process cannot access parent directory.

python/python-core/.gitignore
triceo, line 1: Arguably this should be in the global `.gitignore`.

python/jpyinterpreter/pyproject.toml
Christopher-Chianelli: This is for jpyinterpreter, which doesn't rebuild timefold-solver (and is not distributed/only for testing)

pom.xml
triceo: Maybe it doesn't need to rebuild it anymore? It's the same repo, let's start treating it like that. Python will not be built separately, as an extra thing. They will be built together.

python/jpyinterpreter/examples/.gitignore
triceo: Please do that.

python/jpyinterpreter/pyproject.toml
triceo: Same question as above; now that they're not independent repos, maybe we can make it all part of the same build?

python/jpyinterpreter/mvnw
triceo: I am motivated by the fact that `mvnw` requires regular updates, as Maven is updated. I'd rather only update one place.

triceo:
Arguably the build process has become complicated enough now that it should be documented, at least in broad strokes. See [CONTRIBUTING](https://github.com/TimefoldAI/timefold-solver/blob/main/CONTRIBUTING.adoc).

python/README.md
Christopher-Chianelli: This is for the `pip` package description.

Christopher-Chianelli: (which cannot be `.adoc`)

triceo: Unfortunate. This means we probably need to have some duplication in the top-level README, for the sake of not totally scaring away Python people.

pom.xml
Christopher-Chianelli: This allows :

- `python -m build` / `pip install` workflows will ALWAYS build latest sources without requiring a previous `mvn clean install`

Previously, this wasn't possible, because the sources lived in different repositories.

Christopher-Chianelli: (or put another way, if you change a Java file, you can just run `tox` without needing to remember to do `mvn clean install` before)

triceo: I understand, but... why would I even _want_ a rebuild if I've just built it?
`mvn clean install` in the top level, that needs to be the goal, for our own dev experience.

python/jpyinterpreter/mvnw
Christopher-Chianelli: As I said before: `build` copies files to a new directory and build exclusively from that directory, and cannot copy files from parent or below.

Christopher-Chianelli: (or in other words, working directory is not `timefold-solver/python/...`, but `/tmp/build/...`)

python/jpyinterpreter/pyproject.toml
Christopher-Chianelli: That is done for `timefold` python, but cannot be done for `jpyinterpreter` since it cannot access parent directories.

pom.xml
Christopher-Chianelli: If you are working on Python, you never need to do `mvn clean install` or any familiarity of Java build tools at all.
This just makes all the tools in the Python ecosystem (`pip install`, `tox`, `build`, ...) just work with the Java code, without requiring any work.
This means if a user wanted to build the latest sources, this will work even if they don't have `mvn` installed and haven't built Timefold Solver locally:

```bash
pip install git+https://github.com/TimefoldAI/timefold-solver.git
```
 
And how are we suppose to determine if it is just built and not a stale copy?

python/pom.xml
Christopher-Chianelli: What do you mean? It is already part of the overall build.

pom.xml
triceo: > If you are working on Python, you never need to do `mvn clean install` or any familiarity of Java build tools at all.

There is no "working on Python". Every feature we had since the Python solver was introduced ended up requiring Python code as well. I'm not ruling out some changes will be exclusive to Python, but there will be a relatively small amount of them. IMO the development workflow needs to be tailored for what happens most often, not for corner cases. And from my current experience, the dominant workflow is working on Java _and_ Python.

> And how are we suppose to determine if it is just built and not a stale copy?

We aren't. Or rather the build system isn't supposed to. It's the developer's responsibility - just as it is currently with Java. If I run tests on stale Java code, I'll run into issues.

Christopher-Chianelli: That what I mean! You don't need to run `mvn clean install` to run Python tests! How are you running Python tests. This simplifies things, not complicates it.

triceo:
Good that Sonar works already! The results are bad though. :-)
I wonder what's different, the old repo was green.

pom.xml
triceo: I could argue it simplifies Python things while complicating the Java POM. 

But I see your point. Let's come up with a better name than `non-essential` and I'll shut up. Perhaps something that describes purpose of the profile.

Christopher-Chianelli:
> Good that Sonar works already! The results are bad though. :-) I wonder what's different, the old repo was green.

Seems Python tests aren't included; might be because I needed to move the properties to the `python` parent (since `sonar` fails if the sources/tests directory aren't defined in every single module (and Java/Python have different tests folders)

pom.xml
Christopher-Chianelli: `java-tools-and-integrations`?

triceo:
The tests are one thing, and the reliability rating is another.
Maybe we need to disable some rules that were disabled on the old repo?

Christopher-Chianelli:
> The tests are one thing, and the reliability rating is another. Maybe we need to disable some rules that were disabled on the old repo?

One of the most annoying ones is the name style rule, since the code in `jpyinterpreter` need to follow python conventions and begin special names with `$`. There are also a lot of false positives involving `Optional.isPresent()` checks. Also it says finite loops are infinite (thinks summing a known negative value increases increases the index instead of decreases the index). "BigDecimal.valueOf" warnings in places where we want the exact floating point representation in decimal to match Python's behaviour, Too many parents to match Python's parent structure, and Accessibility bypasses when the type of a class is unknown.

triceo:
This is not wrong IMO:
https://sonarcloud.io/project/security_hotspots?id=ai.timefold%3Atimefold-solver&pullRequest=955

I don't much care about Low/Medium. I'm assuming it's the High issues that affect the "E" rating:
https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG&pullRequest=955&id=ai.timefold%3Atimefold-solver

Christopher-Chianelli:
> This is not wrong IMO: https://sonarcloud.io/project/security_hotspots?id=ai.timefold%3Atimefold-solver&pullRequest=955
> 
Yes, and the high/medium ones are the one that are wrong (high ones are the incorrect "infinite loops", medium ones arethe incorrect "Optional.get without Optional.isPresent"; the issue linked is a low issue.

triceo:
The JVM/OS CI is failing. May be worth it to split the Java part from the Python part, to different jobs that can run in parallel.

Christopher-Chianelli:
> The JVM/OS CI is failing. May be worth it to split the Java part from the Python part, to different jobs that can run in parallel.

It failing on the Java part; for some reason, spotless formatting pass on Ubuntu and fails on Windows.

triceo:
The `\n` at the end of every line in the error output... smells fishy. Maybe some encoding/locale/separator property isn't propagated to Java from the parent process?

Christopher-Chianelli:
> The `\n` at the end of every line in the error output... smells fishy. Maybe some encoding/locale/separator property isn't propagated to Java from the parent process? Please look into it, we cannot merge if it fails the build.

It only happens in the `jpyinterpreter` python build; I changed it to build quickly like `timefold` python; the normal `mvn` build should catch formatting errors.

triceo:
Different error now.

python/jpyinterpreter/setup.py
```diff
@@ -15,9 +15,9 @@ class FetchDependencies(build_py):
     """
     def create_stubs(self, project_root, command):
         subprocess.run([str((project_root / command).absolute()), 'dependency:copy-dependencies'],
-                       cwd=project_root, check=True)
+                       '-Dpython', '-Dquickly', '-Dexec.skip', cwd=project_root, check=True)
```
triceo, line 18: This could go into the special Maven profile, no? That way, the config would be shared without having to repeat it.

Christopher-Chianelli:
Only enterprise + coverage tests are failing now; However, enterprise checks run on main workflow file and not the PR's workflow file, so we have no way to test it until this PR is merged.

Christopher-Chianelli:
From the look of it, `Sonar` is also running from `main`, so it doesn't run the Python tests and hence the poor coverage in `jpyinterpreter` + python code (which are covered mainly from python tests).

triceo (approved):

PR #981, opened 2024-07-18
fix: Adjust weights properly and don't throw NPEs 

triceo:


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/OverridesBasedConstraintWeightSupplier.java
```diff
@@ -102,6 +106,9 @@ The constraint (%s) is not in the default package (%s).
                     Constraint packages are deprecated, check your constraint implementation."""
                     .formatted(constraintRef, getDefaultConstraintPackage()));
         }
+        if (workingSolution == null) { // ConstraintVerifier is known to cause null here.
```
zepfred, line 109: I recommend adding a test case as the existing tests did not catch that.

zepfred (approved):

PR #990, opened 2024-07-23
fix: check duplicated class members for `@CascadingUpdateShadowVariable`

zepfred:
This pull request addresses an issue where the class has multiple members using the same name defined in `targetMethodName` for `@CascadingUpdateShadowVariable`.


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
```diff
@@ -118,20 +119,24 @@ public void linkVariableDescriptors(DescriptorPolicy descriptorPolicy) {
...
-                .findFirst()
-                .orElse(null);
-        if (sourceMethodMember == null) {
+                .toList();
+        if (sourceMethodMembers.size() > 1) {
+            throw new IllegalArgumentException("""
+                    The entity class (%s) has multiple members named as "%s".
+                    Maybe rename the method "%s" with a unique name.""".formatted(entityDescriptor.getEntityClass(),
+                    targetMethodName, targetMethodName));
+        } else if (sourceMethodMembers.stream().noneMatch(m -> Method.class.isAssignableFrom(m.getClass()))) {
```
Christopher-Chianelli, line 131: I would honestly do it like this:

```java
 var sourceMethodMembers = ConfigUtils.getDeclaredMembers(entityDescriptor.getEntityClass())
                .stream()
                .filter(member -> member instanceof Method method && 
                        method.getName().equals(targetMethod) && 
                        method.getParameterCount() == 0)
                .toList();
```

This:

- Select only 0-arg methods with the given name, which are the only valid choice

Since only 0-arg methods would be selected, it would be impossible to choose a field accidentally, and we can allow having a field with the same name as a method.

This also makes the `sourceMethodMembers.stream().noneMatch(m -> Method.class.isAssignableFrom(m.getClass()))` never trigger. In normal classes, `sourceMethodMembers.size() > 1` would also never trigger, but generated classes allow multiple 0-arg methods with the same name (provided they have a different return type).

Christopher-Chianelli (commented):
I honestly don't see a reason to outright prohibit having a field with the same name as the method, if it only makes sense for the target method to be selected as a member.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
zepfred: I see your point, but in that case, we won't fail if there is a field and a method using the same name.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
```diff
@@ -119,24 +119,28 @@ public void linkVariableDescriptors(DescriptorPolicy descriptorPolicy) {
                 .orElse(null);
 
         var targetMethodName = getTargetMethodName();
-        var sourceMethodMembers = ConfigUtils.getDeclaredMembers(entityDescriptor.getEntityClass())
+        var allSourceMethodMembers = ConfigUtils.getDeclaredMembers(entityDescriptor.getEntityClass())
                 .stream()
                 .filter(member -> member.getName().equals(targetMethodName))
```
Christopher-Chianelli, line 124: This line should be
```
member -> member instanceof Method method && member.getName().equals(targetMethodName) && method.getParameterCount() == 0
```

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
```diff
@@ -119,24 +119,28 @@ public void linkVariableDescriptors(DescriptorPolicy descriptorPolicy) {
...
                 .toList();
-        if (sourceMethodMembers.size() > 1) {
+        if (allSourceMethodMembers.size() > 1) {
             throw new IllegalArgumentException("""
                     The entity class (%s) has multiple members named as "%s".
                     Maybe rename the method "%s" with a unique name.""".formatted(entityDescriptor.getEntityClass(),
                     targetMethodName, targetMethodName));
-        } else if (sourceMethodMembers.stream().noneMatch(m -> Method.class.isAssignableFrom(m.getClass()))) {
+        }
+        var validSourceMethodMembers = allSourceMethodMembers.stream()
```
Christopher-Chianelli, line 132: Should be done on `allSourceMethodMembers`

Christopher-Chianelli (changes requested):
Merge the filter of `allSourceMethodMembers` and `validSourceMethodMembers` (the "allSourceMethodMembers.size() > 1" should not consider fields). Consider adding tests for

- Having a field and method with the target name works
- Having multiple methods with the target name works (i.e. overloads)
- Having only a field with the target name fails

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/cascade/CascadingUpdateShadowVariableDescriptor.java
Christopher-Chianelli: And why should we fail? Unlike variables (which are either fields or a getter/setter pair, and thus ambiguity arises if both a field and its getter are annotated with PlanningVariable), this is always a method and never a field (it even has method in the name), so having a field and method with the same name does not create ambiguity.

Christopher-Chianelli (approved):
LGTM, good job!

PR #993, opened 2024-07-25
chore: link the Github Release page from within docs

oemebamo:
Communication about new versions of the Timefold Solver happens via https://github.com/TimefoldAI/timefold-solver/releases

This commit links to that page from within the Antora docs navigation. So it becomes easier to find for people exploring the docs.


triceo (approved):
Thank you, @oemebamo!

PR #998, opened 2024-07-30
ci: external contributions require a label before CI can run

triceo:
This PR makes it so that pull requests that could steal secrets now require a specific label.
It also makes it so that a workflow runs automatically to grant that label to PRs by trusted users.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -8,8 +8,19 @@ on:
       - opened
       - reopened
       - synchronize
+      - labeled
 jobs:
+  safe_for_ci:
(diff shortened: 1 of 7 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -8,8 +8,19 @@ on:
       - opened
       - reopened
       - synchronize
+      - labeled
 jobs:
+  safe_for_ci:
+    name: "Ensure that PR is safe for CI"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail of not safe
```
winklerm, line 17: Nitpick: Typo?

winklerm (approved):

PR #1008, opened 2024-07-30
chore: do not require team during auth

winklerm:
Do not require a team when checking if the PR author is a member of the organisation in GH workflows.


triceo (approved):
Let's see what happens after we merge this.

PR #1010, opened 2024-07-31
feat: introduce a Ruin And Recreate move

triceo:
Introduces a Ruin and Recreate move for both basic and list variable. (Not chained.)

Also introduces a largely refactored "Optimization algorithms" section of the documentation.
It moves away from "everything on a single page" approach, and breaks the content down by topic - just as we already do for many other topics.


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -165,19 +175,24 @@ public void solvingError(SolverScope<Solution_> solverScope, Exception exception
         decider.solvingError(solverScope, exception);
     }
 
-    public static class Builder<Solution_> extends AbstractPhase.Builder<Solution_> {
+    public static class DefaultConstructionHeuristicPhaseBuilder<Solution_>
```
zepfred, line 178: The name `Builder` is used by other builders, e.g., `DefaultLocalSearchPhase`. Is there a reason for changing this one?

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseFactory.java
```diff
@@ -44,53 +47,65 @@ public DefaultConstructionHeuristicPhaseFactory(ConstructionHeuristicPhaseConfig
         super(phaseConfig);
     }
 
-    @Override
-    public ConstructionHeuristicPhase<Solution_> buildPhase(int phaseIndex, boolean triggerFirstInitializedSolutionEvent,
-            HeuristicConfigPolicy<Solution_> solverConfigPolicy, BestSolutionRecaller<Solution_> bestSolutionRecaller,
-            Termination<Solution_> solverTermination) {
-        ConstructionHeuristicType constructionHeuristicType_ = Objects.requireNonNullElse(
-                phaseConfig.getConstructionHeuristicType(),
+    protected DefaultConstructionHeuristicPhaseBuilder<Solution_> getBaseBuilder(int phaseIndex,
```
zepfred, line 50: I understand using the `DefaultConstructionHeuristicPhaseFactory` class to construct both phases, but I wonder if we can use a different approach. Here are some ideas:

Do we want to allow using `RuinRecreateConstructionHeuristicPhase` as an alternative construction phase to `DefaultConstructionHeuristicPhase`?

If so, wouldn't having a separate config class instead of using `ConstructionHeuristicPhaseConfig` make sense?

My point is the use of separate factory and phase classes, which would keep the changes located only in the respective classes, improving the concern separation.

However, if we want to use R&R as an LS method only, I don't see a reason to change the logic in `DefaultConstructionHeuristicPhaseFactory::buildPhase`.

Also, it seems unnecessary to add `isNested` to `DefaultConstructionHeuristicPhase` if the class `RuinRecreateConstructionHeuristicPhase` overrides the methods `doStep` and `phaseEnded`.

In short, can we maintain the default construction factory and phase classes and extend their behavior by creating separate R&R classes?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractUndoMove.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public final boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
+        return true; // Undo moves are always doable; the parent move was already done.
+    }
+
+    @Override
+    public final Move<Solution_> doMove(ScoreDirector<Solution_> scoreDirector) {
+        throw new UnsupportedOperationException("""
```
zepfred, line 32: Do we need a text block here?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveIterator.java
```diff
@@ -0,0 +1,76 @@
...
+    private final SolverScope<Solution_> solverScope;
+    private final ListVariableStateSupply<Solution_> listVariableStateSupply;
+    private final long minimumRuinedCount;
+    private final long maximumRuinedCount;
+    private final Random workingRandom;
+
+    public ListRuinRecreateMoveIterator(EntityIndependentValueSelector<Solution_> valueSelector,
+            RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> constructionHeuristicPhaseBuilder,
+            SolverScope<Solution_> solverScope,
+            ListVariableStateSupply<Solution_> listVariableStateSupply, long minimumRuinedCount, long maximumRuinedCount,
```
zepfred, line 29: Should we use an int instead of a long type?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelectorFactory.java
```diff
@@ -0,0 +1,52 @@
...
+        var valueSelector =
+                (EntityIndependentValueSelector<Solution_>) ValueSelectorFactory.<Solution_> create(new ValueSelectorConfig())
+                        .buildValueSelector(configPolicy, entityDescriptor, minimumCacheType, SelectionOrder.RANDOM,
+                                false, ValueSelectorFactory.ListValueFilteringType.ACCEPT_ASSIGNED);
+        var entityPlacerConfig = DefaultConstructionHeuristicPhaseFactory.buildListVariableQueuedValuePlacerConfig(configPolicy,
+                listVariableDescriptor);
+
+        var constructionHeuristicConfig = new ConstructionHeuristicPhaseConfig();
+        constructionHeuristicConfig.setEntityPlacerConfig(entityPlacerConfig);
+        var constructionHeuristicPhaseFactory =
```
zepfred, line 46: It looks like we could use a distinct factory class here instead of `DefaultConstructionHeuristicPhaseFactory`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateUndoMove.java
```diff
@@ -0,0 +1,65 @@
...
+        changedEntities.addAll(entityToOriginalPositionMap.keySet());
+        changedEntities.addAll(entityToNewPositionMap.keySet());
+
+        for (var entity : changedEntities) {
+            innerScoreDirector.beforeListVariableChanged(listVariableDescriptor, entity,
+                    listVariableDescriptor.getFirstUnpinnedIndex(entity),
+                    listVariableDescriptor.getListSize(entity));
+        }
+
+        for (var entry : entityToOriginalPositionMap.entrySet()) {
```
zepfred, line 42: Should we remove them from `entityToNewPositionMap` and add them back using `entityToOriginalPositionMap`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
```diff
@@ -0,0 +1,115 @@
...
+        this.ruinedValueList = ruinedValueList;
+        this.affectedEntitySet = affectedEntitySet;
+        this.entityToOriginalPositionMap = CollectionUtils.newIdentityHashMap(affectedEntitySet.size());
+        this.entityToNewPositionMap = CollectionUtils.newIdentityHashMap(affectedEntitySet.size());
+    }
+
+    @Override
+    protected Move<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
+        return new ListRuinRecreateUndoMove<>(this, listVariableStateSupply.getSourceVariableDescriptor(),
+                entityToNewPositionMap, entityToOriginalPositionMap);
```
zepfred, line 45:
```suggestion
                entityToOriginalPositionMap, entityToNewPositionMap);
```

zepfred, line 45: I suggest updating the logic from `ListRuinRecreateUndoMove` to make it consistent with the previous comment.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhase.java
```diff
@@ -0,0 +1,60 @@
...
+            extends DefaultConstructionHeuristicPhaseBuilder<Solution_> {
+
+        private List<Object> elementsToRecreate;
+
+        public RuinRecreateConstructionHeuristicPhaseBuilder(Termination<Solution_> phaseTermination,
+                EntityPlacer<Solution_> entityPlacer, ConstructionHeuristicDecider<Solution_> decider) {
+            super(0, false, "", phaseTermination, entityPlacer, decider);
+        }
+
+        public RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> setElementsToRecreate(List<Object> elements) {
```
zepfred, line 34:
```suggestion
public RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> withElementsToRecreate(List<Object> elements) {
```

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveSelectorFactory.java
```diff
@@ -0,0 +1,49 @@
...
+        var genuineVariableDescriptorList = entitySelector.getEntityDescriptor().getGenuineVariableDescriptorList();
+        if (genuineVariableDescriptorList.size() != 1) {
+            throw new UnsupportedOperationException(
+                    "Ruin and Recreate move selector currently only supports 1 planning variable.");
+        }
+        var variableDescriptor = genuineVariableDescriptorList.get(0);
+
+        var constructionHeuristicConfig = new ConstructionHeuristicPhaseConfig();
+        var constructionHeuristicPhaseFactory =
+                new DefaultConstructionHeuristicPhaseFactory<Solution_>(constructionHeuristicConfig);
```
zepfred, line 44: Same comment as the planning list version.

docs/src/modules/ROOT/pages/optimization-algorithms/move-selector-reference.adoc
```diff
@@ -0,0 +1,789 @@
...
+[#ruinRecreateMoveSelector]
+=== `RuinRecreateMoveSelector`
+
+The `RuinRecreateMove` selects a subset of entities and sets their values to null,
+effectively unassigning them.
+Then it runs a construction heuristic to assign them again.
+If xref:using-timefold-solver/modeling-planning-problems.adoc#planningVariableAllowingUnassigned[unassigned values] are allowed,
+it may leave them unassigned.
+
+This coarse-grained move is useful to help get the solver out of a local optimum.
```
zepfred, line 340:
```suggestion
This coarse-grained move is useful to help the solver to escape from a local optimum.
```

docs/src/modules/ROOT/pages/optimization-algorithms/move-selector-reference.adoc
```diff
@@ -0,0 +1,789 @@
...
+
+[#listRuinRecreateMoveSelector]
+=== `ListRuinRecreateMoveSelector`
+
+The `ListRuinRecreateMove` selects a subset of values, and removes them from their list variables.
+Then it runs a construction heuristic to assign them again.
+If xref:using-timefold-solver/modeling-planning-problems.adoc#planningListVariableAllowingUnassigned[unassigned values] are allowed,
+it may leave them unassigned.
+
+This coarse-grained move is useful to help get the solver out of a local optimum.
```
zepfred, line 569: ´´´suggestion
This coarse-grained move is useful to help the solver to escape from a local optimum.
´´´

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -0,0 +1,2134 @@
+[#optimizationAlgorithmsOverview]
```
zepfred, line 1: The new chapter organization looks better!

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
triceo: This is exactly the reason to change the name. :-) When looking for the builder in IntelliJ, I type "Builder" and get several different classes with exactly the same name. This is not practical.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractUndoMove.java
triceo: Personally, I think it's better if long exception messages are split in multiple lines,
for readability both in code and in sysout.
And to do that, I think a text block is more natural than `\n`.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseFactory.java
triceo: Yeah, I don't like the current (lack of) separation of concerns either. I'll see what can be done.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseFactory.java
```diff
@@ -162,59 +165,47 @@ public static EntityPlacerConfig buildListVariableQueuedValuePlacerConfig(
...
-        EnvironmentMode environmentMode = configPolicy.getEnvironmentMode();
-        ConstructionHeuristicDecider<Solution_> decider;
-        Integer moveThreadCount = configPolicy.getMoveThreadCount();
-        if (moveThreadCount == null) {
-            decider = new ConstructionHeuristicDecider<>(configPolicy.getLogIndentation(), termination, forager);
-        } else {
-            decider = TimefoldSolverEnterpriseService.loadOrFail(TimefoldSolverEnterpriseService.Feature.MULTITHREADED_SOLVING)
-                    .buildConstructionHeuristic(moveThreadCount, termination, forager, environmentMode, configPolicy);
-        }
-        if (environmentMode.isNonIntrusiveFullAsserted()) {
```
zepfred, line 180: I couldn't find any other place that updates the decider properties. Please confirm if setting them is still necessary.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import ai.timefold.solver.core.impl.constructionheuristic.DefaultConstructionHeuristicPhase.DefaultConstructionHeuristicPhaseBuilder;
+import ai.timefold.solver.core.impl.constructionheuristic.DefaultConstructionHeuristicPhaseFactory;
+import ai.timefold.solver.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider;
+import ai.timefold.solver.core.impl.constructionheuristic.placer.EntityPlacer;
+import ai.timefold.solver.core.impl.heuristic.HeuristicConfigPolicy;
+import ai.timefold.solver.core.impl.solver.termination.BasicPlumbingTermination;
+import ai.timefold.solver.core.impl.solver.termination.PhaseToSolverTerminationBridge;
+import ai.timefold.solver.core.impl.solver.termination.Termination;
+
+final class RuinRecreateConstructionHeuristicPhaseFactory<Solution_>
```
zepfred, line 13: Nice!

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -95,6 +98,10 @@ private void doStep(ConstructionHeuristicStepScope<Solution_> stepScope) {
         var step = stepScope.getStep();
         step.doMoveOnly(stepScope.getScoreDirector());
         predictWorkingStepScore(stepScope, step);
+        processWorkingSolutionDuringStep(stepScope);
+    }
+
+    protected void processWorkingSolutionDuringStep(ConstructionHeuristicStepScope<Solution_> stepScope) {
```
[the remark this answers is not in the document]
triceo: It is used in this class, and overridden in the R&R phase.

zepfred (approved):
The refactoring looks excellent. Thanks for addressing that!

Please double-check the logic still works with the new changes before merging it.

PR #1020, opened 2024-08-08
feat: Improve @CascadingUpdateShadowVariable logic

zepfred:
This pull request enhances the logic of the shadow variable @CascadingUpdateShadowVariable. The new approach eliminates the need to define sources and automatically runs the listener at the end of the events lifecycle.


core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadingUpdateShadowVariable.java
```diff
@@ -3,54 +3,24 @@
...
 import java.lang.annotation.Target;
 
-import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
-
 /**
- * Specifies that field may be updated by the target method when one or more source variables change.
+ * Specifies that field may be updated by the target method when a dependency changes.
  * <p>
- * Automatically cascades change events to {@link NextElementShadowVariable} of a {@link PlanningListVariable}.
+ * Automatically cascades change events to the next elements of a {@link PlanningListVariable}.
```
triceo, line 12:
```suggestion
 * Automatically cascades change events to the subsequent elements of a {@link PlanningListVariable}.
```

core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadingUpdateShadowVariable.java
```diff
@@ -3,54 +3,24 @@
...
 
-import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
-import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
-
 /**
- * Specifies that field may be updated by the target method when one or more source variables change.
+ * Specifies that field may be updated by the target method when a dependency changes.
```
triceo, line 10:
```suggestion
 * Specifies that a field may be updated by the target method when any of its variables change, genuine or shadow.
```

Is this correct?

core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadingUpdateShadowVariable.java
```diff
@@ -3,54 +3,24 @@
...
- * It is only possible to define either {@code sourceVariableName} or {@code sourceVariableNames}.
  * It can be applied to multiple fields to modify different shadow variables.
  * It should never change a genuine variable or a problem fact.
  * It can change its shadow variable(s) on multiple entity instances
  * (for example: an arrivalTime change affects all trailing entities too).
  */
 @Target({ FIELD })
 @Retention(RUNTIME)
-@Repeatable(CascadingUpdateShadowVariable.List.class)
 public @interface CascadingUpdateShadowVariable {
```
triceo, line 22: The Javadoc should also discuss the differences between this and a shadow variable.
Specifically, that nothing is allowed to depend on it, and that it will always be processed after all other shadow variables have been updated.

Maybe it should discuss what happens if multiple cascading vars on the same entity instance depend on each other. (Because they share fields.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/custom/CustomShadowVariableDescriptor.java
```diff
@@ -49,56 +49,62 @@ public void linkVariableDescriptors(DescriptorPolicy descriptorPolicy) {
...
+                            %s"""
+                            .formatted(entityDescriptor.getEntityClass(), ShadowVariable.class.getSimpleName(),
+                                    variableMemberAccessor.getName(), sourceVariableName,
+                                    sourceEntityDescriptor.getEntityClass(),
+                                    sourceEntityDescriptor.buildInvalidVariableNameExceptionMessage(sourceVariableName)));
         }
-        Class<? extends AbstractVariableListener> variableListenerClass = shadowVariable.variableListenerClass();
+        if (!sourceVariableDescriptor.canBeUsedAsSource()) {
+            throw new IllegalArgumentException(
+                    "The entityClass (%s) has a @%s annotated property (%s) with sourceVariableName (%s) which cannot be used as source."
```
triceo, line 82: Maybe these new exceptions should explain _why_ that's not the case. What did the user do wrong? From the message, it's not at all clear.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -33,12 +36,26 @@ public static <Solution_> VariableListenerSupport<Solution_> create(InnerScoreDi
...
     private boolean notificationQueuesAreEmpty = true;
     private int nextGlobalOrder = 0;
 
     VariableListenerSupport(InnerScoreDirector<Solution_, ?> scoreDirector, NotifiableRegistry<Solution_> notifiableRegistry) {
         this.scoreDirector = scoreDirector;
         this.notifiableRegistry = notifiableRegistry;
+        this.cascadingUpdateShadowVarDescriptorList = scoreDirector.getSolutionDescriptor().getEntityDescriptors().stream()
+                .flatMap(e -> e.getDeclaredCascadingUpdateShadowVariableDescriptors().stream())
+                .toList();
+        this.listVariableDescriptor = scoreDirector.getSolutionDescriptor().getEntityDescriptors().stream()
```
triceo, line 52: Afaik `SolutionDescriptor` has a method which can already retrieve the one and only list variable descriptor.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -33,12 +36,26 @@ public static <Solution_> VariableListenerSupport<Solution_> create(InnerScoreDi
...
+                .flatMap(e -> e.getGenuineVariableDescriptorList().stream())
+                .filter(VariableDescriptor::isListVariable)
+                .map(d -> (ListVariableDescriptor<Solution_>) d)
+                .findFirst()
+                .orElse(null);
+        this.listVariableEventList = new LinkedList<>();
```
triceo, line 58: Please don't use `LinkedList`.
https://kjellkod.wordpress.com/2012/02/25/why-you-should-never-ever-ever-use-linked-list-in-your-code-again/

I know that it has theoretical benefits in computer science, but in computer reality, they are no longer valid.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -195,34 +212,69 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
+        if (listVariableEventList.isEmpty() || cascadingUpdateShadowVarDescriptorList.isEmpty()) {
+            return;
+        }
+        for (var cascadingUpdateShadowVariableDescriptor : cascadingUpdateShadowVarDescriptorList) {
+            for (var event : listVariableEventList) {
+                var values = listVariableDescriptor.getValue(event.entity());
+                // Evaluate all elements inside the range
+                evaluateFromIndex(values, event.fromIndex(), event.toIndex(), true, cascadingUpdateShadowVariableDescriptor);
+                // Evaluate later elements, but stops when there is no change
+                evaluateFromIndex(values, event.toIndex(), values.size(), false, cascadingUpdateShadowVariableDescriptor);
```
triceo, line 250: I find it surprising that this is faster than the alternatives discussed.
If there's two events to say `process the entire list`, this will do it twice, even though it's unnecessary.

triceo (commented):
Glad to see that we're reaching completion!
Some comments.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/CascadingUpdateShadowVariable.java
zepfred: Yes, that's correct.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
zepfred: Right. However, the default moves do not generate two events when the source and destination entities are the same and we were executing unnecessary logic because it is not a valid state for the built-in moves.

It may be possible for the custom moves to generate two events, but it still seems incorrect to me to duplicate events as it would require additional work to create before/after events and undo the state.

triceo: Technically `CompositeMove` _could_ run any number of moves at the same time. But you're right - this is a corner case and the mechanism needs to work well for the likely cases.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1427,7 +1427,13 @@ A user-defined logic can only change shadow variables.
 It must never change a genuine planning variable or a problem fact.
 ====
 
-==== Multiple shadow variables
+[WARNING]
```
triceo, line 1430:
```suggestion
[NOTE]
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1462,6 +1468,11 @@ It is not allowed to specify the property `shadowEntityClass` in `@PiggybackShad
 when the source shadow variable is `@CascadingUpdateShadowVariable`.
 ====
 
+[WARNING]
+====
+The user's logic is responsible for defining the order in which each variable is updated.
```
triceo, line 1473: I don't understand what this is trying to say. Maybe it needs a bit of context?

Not sure if it's a warning. If it has more text explaining why this is bad, then I say it could be a warning.

zepfred: I believe it caused more confusion and I will remove it.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -195,34 +207,69 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
-        Solution_ workingSolution = scoreDirector.getWorkingSolution();
-        ShadowVariablesAssert snapshot =
+        var workingSolution = scoreDirector.getWorkingSolution();
+        var snapshot =
                 ShadowVariablesAssert.takeSnapshot(scoreDirector.getSolutionDescriptor(), workingSolution);
 
         forceTriggerAllVariableListeners(workingSolution);
 
-        final int SHADOW_VARIABLE_VIOLATION_DISPLAY_LIMIT = 3;
+        final var SHADOW_VARIABLE_VIOLATION_DISPLAY_LIMIT = 3;
```
triceo, line 272: We don't use `final` for local variables.

Also, arguably this should be standard `camelCase`, unless you want to make it a class-level constant. (Which it probably should be.)

triceo (dismissed):
LGTM with one minor final comment.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1412,98 +1414,23 @@ public class Customer {
...
-Timefold Solver triggers a single listener to run the user-defined logic whenever `previousCustomer` changes.
-It stops when both `arrivalTime` and `weightAtVisit` values do not change or when it reaches the end.
-
-[WARNING]
+[NOTE]
 ====
-It is not allowed to specify the property `shadowEntityClass` in `@PiggybackShadowVariable`
-when the source shadow variable is `@CascadingUpdateShadowVariable`.
+When distinct target methods are used by separate `@CascadingUpdateShadowVariable` variables in the same model,
+there will be no guarantees regarding which target method is executed first.
```
triceo, line 1433:
```suggestion
the order of their execution is undefined.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1412,98 +1414,23 @@ public class Customer {
...
-
-The listener behaves as described previously.
-It calls `updateArrivalTime` whenever a `previousCustomer` or `vehicle` changes.
-It then propagates the changes to the subsequent elements,
-stopping when the method results in no change to the variable, or when it reaches the tail.
-
 [WARNING]
 ====
 A user-defined logic can only change shadow variables.
 It must never change a genuine planning variable or a problem fact.
```
triceo, line 1427:
```suggestion
Changing a genuine planning variable or a problem fact will result in score corruption.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1371,8 +1371,10 @@ public class Customer {
 === Updating tail chains
 
 The annotation `@CascadingUpdateShadowVariable` provides a built-in listener that updates a set of connected elements.
-Timefold Solver triggers a user-defined logic whenever the a defined source variable changes.
-Moreover, it automatically propagates changes to the next elements when the value of the related shadow variable changes.
+Timefold Solver triggers a user-defined logic after all events are processed.
+Hence, the related listener is the final one executed during the event lifecycle.
+Moreover,
+it automatically propagates changes to the next elements when the value of the related shadow variable changes.
```
triceo, line 1377:
```suggestion
it automatically propagates changes to the subsequent elements in the list when the value of the related shadow variable changes.
```

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/wrong_cascade/TestdataCascadingPiggyback.java
```diff
@@ -4,33 +4,36 @@
...
     @PreviousElementShadowVariable(sourceVariableName = "valueList")
-    private TestdataCascadingNoSources previous;
+    private TestdataCascadingPiggyback previous;
     @NextElementShadowVariable(sourceVariableName = "valueList")
-    private TestdataCascadingNoSources next;
+    private TestdataCascadingPiggyback next;
     @CascadingUpdateShadowVariable(targetMethodName = "updateCascadeValue")
     private Integer cascadeValue;
+    @PiggybackShadowVariable(shadowVariableName = "cascadeValue")
+    private Integer cascadeValue2;
```
triceo, line 33: This should be no longer possible, right?

triceo (dismissed):

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/wrong_cascade/TestdataCascadingPiggyback.java
zepfred: Yes, there is a new test to cover that. Let me rename the class to make it clear.

triceo (dismissed):

PR #1030, opened 2024-08-13
perf: create less BigDecimal in load balancing

triceo:


zepfred (approved):
Looks good. I'm assuming the existing tests will cover this change.

PR #1031, opened 2024-08-13
ci: merge workflows requiring secrets to a single workflow

triceo:
- For external PRs from forks, this will only require one approval, as opposed to four.
- It also allows centralizing the logic for checking privileges.
- Overall, this reduces the number of workflow files.


winklerm (approved):
Nice, thanks!

PR #1037, opened 2024-08-13
feat: improve JVM exceptions

zepfred:
This PR improves the compatibility JVM error messages.

### DIFF
--- a/python/python-core/src/main/python/_timefold_java_interop.py
+++ b/python/python-core/src/main/python/_timefold_java_interop.py
@@ -76,7 +76,7 @@ def init(*args, path: list[str] = None, include_timefold_jars: bool = True) -> N
         If the Timefold dependencies should be added to `path`.
         Defaults to True.
     """
-    from _jpyinterpreter import init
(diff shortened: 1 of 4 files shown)


python/python-core/src/main/python/_timefold_java_interop.py
```diff
@@ -80,20 +80,30 @@ def init(*args, path: list[str] = None, include_timefold_jars: bool = True) -> N
...
-        args = [jpype.getDefaultJVMPath()]
+        args = [get_default_jvm_path()]
     init(*args, path=path, include_translator_jars=False)
 
     from ai.timefold.solver.python.logging import PythonDelegateAppender
     PythonDelegateAppender.setLogEventConsumer(PythonConsumer(forward_logging_events))
     update_log_level()
 
 
+def get_default_jvm_path():
```
Christopher-Chianelli, line 98: Can use `path_supplier=jpype.getDefaultJVMPath` as a parameter to make it unit testable, and do `path_supplier()` in the `try` block.

PR #1039, opened 2024-08-14
feat: update the best solution for custom construction heuristics

zepfred:
This pull request modifies the logic of `DefaultCustomPhase` to update the best solution when the returned solution is not better than the starting phase one.

When the planning list variable accepts unassigned list values, the custom CH does not update the best solution because it may be possible that it does not improve the starting score (`0hard/0soft`).


core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultCustomPhase.java
```diff
@@ -74,6 +78,10 @@ public void stepEnded(CustomStepScope<Solution_> stepScope) {
 
     public void phaseEnded(CustomPhaseScope<Solution_> phaseScope) {
         super.phaseEnded(phaseScope);
+        // Only update the best solution if the custom phase first initializes the solution and made any change.
+        if (triggersFirstInitializedSolutionEvent() && !phaseScope.getStartingScore().equals(phaseScope.getBestScore())) {
```
triceo, line 82: If the score is improved, shouldn't it _always_ give the better solution?
What would be a situation where the best solution should be ignored?

What happens if the custom phase is in between two local search phases? Is this behavior still correct?

zepfred:
@triceo, to resolve the problem, we can create a constraint that rewards assigned values. I am closing this issue because it is not currently relevant.

PR #1056, opened 2024-08-26
fix: Use inserted range when retracting range in `toConnectedRanges`

### LINKED ISSUE #953: ConstraintCollectors.toConnectedRanges: "this.splitPoint" is null
**Describe the bug**
By using the new `ConstraintCollectors.toConnectedRanges` we always run into the exception where the `splitPoint == null`which crashes the entire application. To reproduce I recreated the example from the [documentation](https://docs.timefold.ai/timefold-solver/latest/constraints-and-score/score-calculation#collectorsConnectedRanges), which has the same issue.

**Expected behavior**
Runs without exception.

**Actual behavior**
Crashes with:
```log
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.lang.Comparable.compareTo(Object)" because "this.splitPoint" is null
	at ai.timefold.solver.core.impl.score.stream.collector.connected_ranges.RangeSplitPoint.compareTo(RangeSplitPoint.java:109)
	at ai.timefold.solver.core.impl.score.stream.collector.connected_ranges.RangeSplitPoint.compareTo(RangeSplitPoint.java:9)
	at java.base/java.util.TreeMap.compare(TreeMap.java:1604)
	at java.base/java.util.TreeMap.getFloorEntry(TreeMap.java:463)
	at java.base/java.util.TreeMap.floorKey(TreeMap.java:1045)
	at java.base/java.util.TreeSet.floor(TreeSet.java:426)
	at ai.timefold.solver.core.impl.score.stream.collector.connected_ranges.ConnectedRangeTracker.remove(ConnectedRangeTracker.java:80)
	at ai.timefold.solver.core.impl.score.stream.collector.ConnectedRangesCalculator.retract(ConnectedRangesCalculator.java:30)
	at ai.timefold.solver.core.impl.score.stream.collector.bi.ObjectCalculatorBiCollector.lambda$accumulator$0(ObjectCalculatorBiCollector.java:26)
	at ai.timefold.solver.core.impl.score.stream.bavet.common.AbstractGroupNode.retract(AbstractGroupNode.java:261)
	at ai.timefold.solver.core.impl.score.stream.bavet.common.StaticPropagationQueue.propagate(StaticPropagationQueue.java:93)
	at ai.timefold.solver.core.impl.score.stream.bavet.common.StaticPropagationQueue.propagateRetracts(StaticPropagationQueue.java:83)
	at ai.timefold.solver.core.impl.score.stream.bavet.common.Propagator.propagateEverything(Propagator.java:64)
	at ai.timefold.solver.core.impl.score.stream.bavet.BavetConstraintSession.calculateScoreInLayer(BavetConstraintSession.java:92)
	at ai.timefold.solver.core.impl.score.stream.bavet.BavetConstraintSession.calculateScore(BavetConstraintSession.java:83)
	at ai.timefold.solver.core.impl.score.director.stream.BavetConstraintStreamScoreDirector.calculateScore(BavetConstraintStreamScoreDirector.java:49)
	at ai.timefold.solver.core.impl.score.director.AbstractScoreDirector.assertExpectedUndoMoveScore(AbstractScoreDirector.java:642)
	at ai.timefold.solver.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.doMove(ConstructionHeuristicDecider.java:138)
	at ai.timefold.solver.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.decideNextStep(ConstructionHeuristicDecider.java:107)
	at ai.timefold.solver.core.impl.constructionheuristic.DefaultConstructionHeuristicPhase.solve(DefaultConstructionHeuristicPhase.java:62)
	at ai.timefold.solver.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:82)
	at ai.timefold.solver.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:200)
	at <mypackage>.ConnectedRangeDemo.main(ConnectedRangeDemo.java:22)
```

**To Reproduce**
```java
public class Equipment {
    private int id;
    private int capacity;

    public Equipment(int id, int capacity) {
        this.id = id;
        this.capacity = capacity;
    }

    public int getId() { return id; }
    public int getCapacity() { return capacity; }
}
```

```java
@PlanningEntity
public class Job {
    private int id;
    private int requiredEquipmentId;
    private Integer start;

    public Job() {}
    public Job(int id, int requiredEquipmentId) {
        this.id = id;
        this.requiredEquipmentId = requiredEquipmentId;
    }

    @PlanningId
    public int getId() { return id; }

    public int getRequiredEquipmentId() { return requiredEquipmentId; }

    @PlanningVariable
    public Integer getStart() { return start; }
    public void setStart(Integer start) { this.start = start; }

    public Integer getEnd() { return start == null ? null : start + 10; }
}
```

```java
@PlanningSolution
public class Planner {
    @PlanningScore
    private HardMediumSoftScore score;

    private final List<Equipment> equipments = new ArrayList<>();

    @PlanningEntityCollectionProperty
    private final List<Job> jobs = new ArrayList<>();

    @ValueRangeProvider
    public CountableValueRange<Integer> getStartOffsetRange() {
        return ValueRangeFactory.createIntValueRange(0, 100);
    }

    public Planner() {}
    public Planner(List<Equipment> equipments, List<Job> jobs) {
        this.equipments.addAll(equipments);
        this.jobs.addAll(jobs);
    }
}
```

```java
public class MyConstraintProvider implements ConstraintProvider {
    public MyConstraintProvider() {}

    @Override
    public Constraint[] defineConstraints(ConstraintFactory constraintFactory) {
        return new Constraint[]{doNotOverAssignEquipment(constraintFactory)};
    }

    public Constraint doNotOverAssignEquipment(ConstraintFactory constraintFactory) {
        return constraintFactory.forEach(Equipment.class)
                .join(Job.class, Joiners.equal(Equipment::getId, Job::getRequiredEquipmentId))
                .groupBy((equipment, job) -> equipment, ConstraintCollectors.toConnectedRanges((equipment, job) -> job,
                        Job::getStart,
                        Job::getEnd,
                        (a, b) -> b - a))
                .flattenLast(ConnectedRangeChain::getConnectedRanges)
                .filter((equipment, connectedRange) -> connectedRange.getMaximumOverlap() > equipment.getCapacity())
                .penalize(HardMediumSoftScore.ONE_HARD)
                .asConstraint("Concurrent equipment usage over capacity");
    }
}
```

```java
public class ConnectedRangeDemo {
    public static void main(String[] args) {
        var e1 = new Equipment(1, 1);
        var j1 = new Job(1, e1.getId());
        var j2 = new Job(2, e1.getId());
        var problem = new Planner(List.of(e1), List.of(j1, j2));

        var config = new SolverConfig()
                .withSolutionClass(Planner.class)
                .withEntityClasses(Job.class)
                .withConstraintProviderClass(MyConstraintProvider.class)
                .withEnvironmentMode(EnvironmentMode.FULL_ASSERT);
        var solver = SolverFactory.create(config).buildSolver();
        solver.solve(problem);
    }
}
```

## Environment
Tested on `1.12.0` and `1.10.0`

```
openjdk version "21.0.1" 2023-10-17 LTS
OpenJDK Runtime Environment Temurin-21.0.1+12 (build 21.0.1+12-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.1+12 (build 21.0.1+12-LTS, mixed mode)
```

Christopher-Chianelli:
The `ConnectedRangesCalculator` incorrectly assumes the range for each value is constant. The fix is to save the range on insert, and then pass the saved range on retract. The other calculators do not have this problem, because either:

- Their input should be immutable (ex: number)
- They used a map in their internal data structure that saved the original value

Fixes #953


core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ObjectCalculatorBiCollector.java
```diff
@@ -21,9 +21,9 @@ public ObjectCalculatorBiCollector(BiFunction<? super A, ? super B, ? extends In
     @Override
     public TriFunction<Calculator_, A, B, Runnable> accumulator() {
         return (calculator, a, b) -> {
-            final Input_ mapped = mapper.apply(a, b);
-            calculator.insert(mapped);
-            return () -> calculator.retract(mapped);
+            final var mapped = mapper.apply(a, b);
+            final var saved = calculator.insert(mapped);
```
triceo, line 25: `final` shouldn't be necessary here. The variable is _effectively final_ without it too.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/quad/ObjectCalculatorQuadCollector.java
```diff
@@ -23,9 +23,9 @@ public ObjectCalculatorQuadCollector(QuadFunction<? super A, ? super B, ? super
     @Override
     public PentaFunction<Calculator_, A, B, C, D, Runnable> accumulator() {
         return (calculator, a, b, c, d) -> {
-            final Input_ mapped = mapper.apply(a, b, c, d);
-            calculator.insert(mapped);
-            return () -> calculator.retract(mapped);
+            final var mapped = mapper.apply(a, b, c, d);
+            final var saved = calculator.insert(mapped);
```
triceo, line 27: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/tri/ObjectCalculatorTriCollector.java
```diff
@@ -21,9 +21,9 @@ public ObjectCalculatorTriCollector(TriFunction<? super A, ? super B, ? super C,
     @Override
     public QuadFunction<Calculator_, A, B, C, Runnable> accumulator() {
         return (calculator, a, b, c) -> {
-            final Input_ mapped = mapper.apply(a, b, c);
-            calculator.insert(mapped);
-            return () -> calculator.retract(mapped);
+            final var mapped = mapper.apply(a, b, c);
+            final var saved = calculator.insert(mapped);
```
triceo, line 25: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/uni/ObjectCalculatorUniCollector.java
```diff
@@ -21,9 +21,9 @@ public ObjectCalculatorUniCollector(Function<? super A, ? extends Input_> mapper
     @Override
     public BiFunction<Calculator_, A, Runnable> accumulator() {
         return (calculator, a) -> {
-            final Input_ mapped = mapper.apply(a);
-            calculator.insert(mapped);
-            return () -> calculator.retract(mapped);
+            final var mapped = mapper.apply(a);
+            final var saved = calculator.insert(mapped);
```
triceo, line 25: Dtto.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeSolvingTest.java
```diff
@@ -0,0 +1,137 @@
...
+        var config = new SolverConfig()
+                .withSolutionClass(Planner.class)
+                .withEntityClasses(Job.class)
+                .withConstraintProviderClass(MyConstraintProvider.class)
+                .withEnvironmentMode(EnvironmentMode.FULL_ASSERT)
+                .withTerminationConfig(
+                        new TerminationConfig()
+                                .withScoreCalculationCountLimit(1_000L));
+        var solver = SolverFactory.create(config).buildSolver();
+        solver.solve(problem);
```
triceo, line 135: Please write a test which tests the collector directly.

I think we have been overdoing it with these integration tests lately. There are several reasons why tests that run the solver should be the exception, rather than the norm:

- They take a relatively long time and if overused, our test suite will take impractically long to run.
- They do not actually test anything. Yes, the test otherwise would've failed without this fix, but that's not a guarantee of anything into the future. If the solver changes how it iterates through things, because for example the RNG impl in the JDK changed, the test may no longer be testing the same conditions and therefore no longer catch the issue.

These kinds of tests should only be used in cases where it is not possible or practical to reproduce a particular behavior on a lower level. There are cases like that, but this is not one of them - we know how to test collectors directly, and we should take the time to reproduce the actual problematic behavior.

triceo (changes requested):
We need a different test.
Otherwise LGTM.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ObjectCalculatorBiCollector.java
Christopher-Chianelli: I prefer final to explicitly mark it as final/as something that shouldn't be changed.

triceo (approved):

PR #1060, opened 2024-08-28
fix: allow @PlanningId to be defined on interfaces

Christopher-Chianelli:
Previously, `getAllMembers` excluded members defined in interfaces. Presumably, this was because methods would have implementations in the implementing class.
However, there is a flaw in that logic: implementations of a method do not inherit the annotations of the declared method. As a result, when LookupManager looks up the accessor of a class whose @PlanningId is defined in an interface,
it finds none.

`getAllMembers` is changed to add all interface methods to the member list, thus allowing interface members to be used in lookups.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
@@ -369,22 +369,19 @@ public class ConfigUtils {
         Stream<Member> memberStream = Stream.empty();
         while (clazz != null) {
             var fieldStream = Stream.of(clazz.getDeclaredFields())
-                    .filter(field -> field.isAnnotationPresent(annotationClass) && !field.isSynthetic())
-                    .sorted(alphabeticMemberComparator);
-            var methodStream = Stream.of(clazz.getDeclaredMethods())
-                    .filter(method -> method.isAnnotationPresent(annotationClass) && !method.isSynthetic());
+                    .filter(field -> field.isAnnotationPresent(annotationClass) && !field.isSynthetic());
 
-            for (var implementedInterface : clazz.getInterfaces()) {
-                methodStream = Stream.concat(
-                        methodStream,
-                        Arrays.stream(implementedInterface.getMethods())
-                                .filter(method -> method.isAnnotationPresent(annotationClass) && !method.isSynthetic()));
-            }
-            methodStream = methodStream.sorted(alphabeticMemberComparator);
+            // Implementations of an interface method do not inherit the annotations of the declared method in
+            // the interface, so we need to also add the interface's methods to the stream.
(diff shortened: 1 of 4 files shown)


core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -372,12 +372,19 @@ public static List<Member> getAllMembers(Class<?> baseClass, Class<? extends Ann
                     .filter(field -> field.isAnnotationPresent(annotationClass) && !field.isSynthetic())
                     .sorted(alphabeticMemberComparator);
             var methodStream = Stream.of(clazz.getDeclaredMethods())
-                    .filter(method -> method.isAnnotationPresent(annotationClass) && !method.isSynthetic())
-                    .sorted(alphabeticMemberComparator);
+                    .filter(method -> method.isAnnotationPresent(annotationClass) && !method.isSynthetic());
```
triceo, line 375: You can filter the entire concated stream at the end, avoiding this duplication.

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -372,12 +372,19 @@ public static List<Member> getAllMembers(Class<?> baseClass, Class<? extends Ann
...
+
+            for (var implementedInterface : clazz.getInterfaces()) {
+                methodStream = Stream.concat(
+                        methodStream,
+                        Arrays.stream(implementedInterface.getMethods())
+                                .filter(method -> method.isAnnotationPresent(annotationClass) && !method.isSynthetic()));
+            }
```
triceo, line 382: Since we're already using stream concatenation, no need to mix normal iteration with it too. Use `flatMap` to achieve the same result.

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -372,12 +372,19 @@ public static List<Member> getAllMembers(Class<?> baseClass, Class<? extends Ann
...
+            methodStream = methodStream.sorted(alphabeticMemberComparator);
             memberStream = Stream.concat(memberStream, Stream.concat(fieldStream, methodStream));
             clazz = clazz.getSuperclass();
         }
-        return memberStream.collect(Collectors.toList());
+        return memberStream.distinct().collect(Collectors.toList());
```
triceo, line 387: Wouldn't it be better to first distinct and then sort? Either way, both can be done at this final stream, and not on the intermediate streams.

triceo (dismissed):
LGTM, with small comments.
An explanation of why this was necessary would be useful.

triceo (approved):

PR #1066, opened 2024-09-02
Update modeling-planning-problems.adoc

RutgerLubbers:
InverseRelationShadowVariable is on a collection of the holder of the relationship.  For instance, see the `Consumer` to `Facility` in the `facility-location` quickstart project.


triceo (approved):
Thank you for the PR, well spotted!

PR #1067, opened 2024-09-03
fix: initialize nested selectors correctly

triceo:
When a selector is created within another selector, it has to be added to the selector's lifecycle. If that doesn't happen, the selector won't be properly initialized as its solver/phase/step events won't be called.

Thankfully, this issue didn't show until now, because none of the inner selectors actually required their state. In my recent experiments though, I managed to create situations where state would be required, leading to NPEs.


core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -824,15 +823,33 @@ public int countReinitializableVariables(Object entity) {
...
+                (effectiveMovableEntityFilter == null
+                        || effectiveMovableEntityFilter.test(workingSolution, entity));
     }
 
     @Override
     public String toString() {
         return getClass().getSimpleName() + "(" + entityClass.getName() + ")";
     }
 
+    @FunctionalInterface
```
zepfred, line 837: This class appears to be hidden here, and I think it should be moved to a separate package (e.g., ai.timefold.solver.core.impl.heuristic.selector.common.decorator), and it also deserves better documentation.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/PinEntityFilter.java
```diff
@@ -22,7 +21,7 @@ public PinEntityFilter(MemberAccessor memberAccessor) {
     }
 
     @Override
-    public boolean accept(ScoreDirector<Solution_> scoreDirector, Object entity) {
+    public boolean test(Solution_ solution, Object entity) {
```
zepfred, line 24: I understand your point about using BiPredicate, but I'm wondering if MovableFilter should have a contract consistent with SelectionFilter. For example, it could have a method named accept instead of test.

zepfred: Also, it seems that accepting the score director instead of the solution class will require fewer changes in other places.

triceo: The point of this _is_ to accept the solution. Score director is a much heavier class which is not available at all times, and it shouldn't be required when the code doesn't need it.

In fact, this was the original change that triggered the entire PR. For move streams, I need this method to operate without a score director. This change is 100 % intentional.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/KOptListMoveSelector.java
```diff
@@ -26,51 +28,51 @@ final class KOptListMoveSelector<Solution_> extends GenericMoveSelector<Solution
     private final int[] pickedKDistribution;
 
     private ListVariableStateSupply<Solution_> listVariableStateSupply;
-    private EntityIndependentValueSelector<Solution_> effectiveOriginSelector;
-    private EntityIndependentValueSelector<Solution_> effectiveValueSelector;
+
+    @Override
+    public void solvingEnded(SolverScope<Solution_> solverScope) {
```
zepfred, line 33: Let's move this method below to the constructor.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListSwapMoveSelector.java
```diff
@@ -41,41 +43,36 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
-        movableLeftValueSelector = filterPinnedListPlanningVariableValuesWithIndex(leftValueSelector, listVariableStateSupply);
-        movableRightValueSelector =
-                filterPinnedListPlanningVariableValuesWithIndex(rightValueSelector, listVariableStateSupply);
     }
 
     @Override
     public void solvingEnded(SolverScope<Solution_> solverScope) {
         super.solvingEnded(solverScope);
         listVariableStateSupply = null;
-        movableLeftValueSelector = null;
```
zepfred, line 53: I liked the simplification made with the "effective" selectors, and I noted a pattern where we recreate the selectors when the solver starts. You might have already considered this before making the changes, but would it be possible for the selectors to be inconsistent if we don't recreate them when the solver starts?

triceo: Where have you noticed me recreating selectors during solvingStarted()?
If there is still a place like that, it's a bug and it needs to be removed.

zepfred: I did not check any other selector outside this PR. The patterns I noted are the ones I shared before.

zepfred: I checked all the selectors but couldn't find any that recreates the selectors at the `solvingStarted` method. Regardless, the question still remains: Could not recreating the selectors lead to an inconsistent state?

triceo: No. Selectors don't really have state, until solvingStarted() is called on them. The way we recreate selectors is by wrapping them - the parent always delegates the lifecycle calls to the child. Since the recreation now happens in the selector constructor, none of the selectors could have had any lifecycle methods called on them yet. Therefore this is wrapping selectors without state with other selectors without state. This is safe, unlike what was there previously.

zepfred (approved):
LGTM!

PR #1068, opened 2024-09-03
chore: replace Chad with Carl

triceo:
It's Carl. It's always been Carl. Chad is dead, long live Carl.


lee-carlon (approved):
Looks good.

PR #1072, opened 2024-09-04
feat: add new move count metric

zepfred:
This pull request includes a new metric for counting the number of moves, which differs from the score count when using R&R moves.


benchmark/src/main/java/ai/timefold/solver/benchmark/impl/SolverBenchmarkFactory.java
```diff
@@ -91,6 +91,8 @@ protected List<SolverMetric> getSolverMetrics(ProblemBenchmarksConfig config) {
                 .orElseGet(ProblemStatisticType::defaultList)) {
             if (problemStatisticType == ProblemStatisticType.SCORE_CALCULATION_SPEED) {
                 out.add(SolverMetric.SCORE_CALCULATION_COUNT);
+            } else if (problemStatisticType == ProblemStatisticType.MOVE_CALCULATION_SPEED) {
```
triceo, line 94: Question: does "move calculation" make sense? Move's aren't calculated.
Arguably "move execution" would be more correct here.
Any other ideas?

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/movecalculationspeed/MoveCalculationSpeedProblemStatistic.java
```diff
@@ -0,0 +1,56 @@
...
+    public SubSingleStatistic<?, ?> createSubSingleStatistic(SubSingleBenchmarkResult subSingleBenchmarkResult) {
+        return new MoveCalculationSpeedSubSingleStatistic<>(subSingleBenchmarkResult);
+    }
+
+    /**
+     * @return never null
+     */
+    @Override
+    protected List<LineChart<Long, Long>> generateCharts(BenchmarkReport benchmarkReport) {
+        LineChart.Builder<Long, Long> builder = new LineChart.Builder<>();
```
triceo, line 36: For new code, I'd generally go with `var`.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/movecalculationspeed/MoveCalculationSpeedSubSingleStatistic.java
```diff
@@ -0,0 +1,91 @@
...
+                if (timeMillisSpent >= nextTimeMillisThreshold) {
+                    registry.getGaugeValue(SolverMetric.MOVE_CALCULATION_COUNT, runTag, moveCalculationCountNumber -> {
+                        long moveCalculationCount = moveCalculationCountNumber.longValue();
+                        long calculationCountInterval = moveCalculationCount - lastMoveCalculationCount.get();
+                        long timeMillisSpentInterval = timeMillisSpent - lastTimeMillisSpent;
+                        if (timeMillisSpentInterval == 0L) {
+                            // Avoid divide by zero exception on a fast CPU
+                            timeMillisSpentInterval = 1L;
+                        }
+                        long moveCalculationSpeed = calculationCountInterval * 1000L / timeMillisSpentInterval;
```
triceo, line 61: We already have this code elsewhere. It's a bit of a fishy logic, we should extract it to a method and reuse.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractMetricMove.java
```diff
@@ -0,0 +1,22 @@
+package ai.timefold.solver.core.impl.heuristic.move;
+
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+
+/**
+ * Abstract superclass for {@link Move} which includes the ability to enable or disable the move metric collection.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+public abstract class AbstractMetricMove<Solution_> implements Move<Solution_> {
```
triceo, line 10: Please think of a different way of doing this. `Move` is something that people extend; we don't want them to think about metrics as they do so. Moves should not have to understand if they're being measured or not.

triceo (commented):
I'm leaving a review before I leave for PTO.
It's OK if merged once my comments are resolved, provided that someone else sees the final PR as well.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/SolverBenchmarkFactory.java
zepfred: What about `MOVE_COUNT_SPEED`?

zepfred: image
image
image

zepfred:
@triceo, I see some possible points of discussion beside the ones you may find during the review:

>Optionally, the metric also has sub-metrics per move type. (Some other metrics can already do that, for inspiration.)

I couldn't find any useful sub-metric for the move count. We already have the move count per step.

>Solver logging additionally includes move speed (move count over time), along with score speed which it already does.

I initially included it but then removed it. My point is that it may confuse the user as it is the same number when not using R&R moves. I believe that the benchmark should be used to compare move speeds, and the score calculation speed should still be the primary metric.

>There is a score count related type of Termination. We need one of those for move count as well.

I considered adding `unimprovedMoveCount`, but I'm unsure if it's worth it. This would require changing some base classes to store the best move count and the last accepted step move count.

>Does this become the new metric by which we measure solver performance, and therefore is it enabled by default? If so, documentation needs to change, replacing uses of score calculation speed by move speed.

I believe the primary metric remains score calculation speed, and we should delegate the comparison of move speed to the benchmark module. With that in mind, the performance page will not change significantly except for adding a reference to the new metric. It doesn't make sense to me to compare a configuration using R&R with one that doesn't. Hence, when comparing two configurations using R&R, the score calculation still makes sense for evaluating performance.

Christopher-Chianelli (commented):
I am okay with the move evaluation speed name, but I am concern that one of the tests, where Ruin moves are not used, have different counts for score calculation count and move evaluation counts (which smells like a bug). Otherwise LGTM.

core/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
```diff
@@ -515,13 +515,15 @@ void testScoreCalculationCountForFinishedJob() throws ExecutionException, Interr
 
         solverJob.getFinalBestSolution();
         assertThat(solverJob.getScoreCalculationCount()).isEqualTo(5L);
+        assertThat(solverJob.getMoveEvaluationCount()).isEqualTo(4L);
```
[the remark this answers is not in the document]
zepfred: That's because we calculate the score without applying any moves during the solve start phase:

1. ai.timefold.solver.core.impl.solver.DefaultSolver#solvingStarted
2. ai.timefold.solver.core.impl.solver.AbstractSolver#solvingStarted
3. ai.timefold.solver.core.impl.solver.recaller.BestSolutionRecaller#solvingStarted (line 50)

triceo:
> I couldn't find any useful sub-metric for the move count. We already have the move count per step.


Some other metrics can track their values per-move. So, for example, you could have "move execution count for change moves" etc. This is what I mean. There is precedent.
 
> > Solver logging additionally includes move speed (move count over time), along with score speed which it already does.
> 
> I initially included it but then removed it. My point is that it may confuse the user as it is the same number when not using R&R moves. I believe that the benchmark should be used to compare move speeds, and the score calculation speed should still be the primary metric.

Good point on it being the same number in many cases.
Not sure on what the default should be; arguably, score calc speed is now the confusing one, because based on different moves, it changes. Move execution speed is the one that's "stable".
 
> > There is a score count related type of Termination. We need one of those for move count as well.
> 
> I considered adding `unimprovedMoveCount`, but I'm unsure if it's worth it. This would require changing some base classes to store the best move count and the last accepted step move count.

I do think it is worth the effort, for the same reason as I said above. Score calculation count is not stable, its definition changes based on moves used. Move execution count is stable and always means the same thing.

core/src/test/java/ai/timefold/solver/core/api/solver/SolverManagerTest.java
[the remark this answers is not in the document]
triceo: A comment added to the test would help avoiding this question in the future.

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -596,6 +596,27 @@ Useful for comparing different score calculators and/or constraint implementatio
 Also useful to measure the scalability cost of an extra constraint.
 
 
+[#benchmarkReportMoveEvaluationSpeedSummary]
+=== Move evaluation speed summary (graph And table)
```
triceo, line 600:
```suggestion
=== Move evaluation speed summary (graph and table)
```

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -33,6 +33,18 @@ and still compare it with the original's calculation speed.
...
 ====
 
+The solver usually calculates the score for each move it evaluates.
+This means a direct relationship exists between the score calculation and the evaluated move, typically a `1:1` ratio.
+However, some moves,
+such as xref:optimization-algorithms/move-selector-reference.adoc#ruinRecreateMoveSelector[Ruin and Recreate],
+involve processing the score multiple times within the same move.
+
+The use of moves like Ruin and Recreate may increase the number of score calculations,
+which makes the relationship `1:1` invalid.
```
triceo, line 43:
```suggestion
making this ratio `N:1`.
```

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -33,6 +33,18 @@ and still compare it with the original's calculation speed.
...
+In such cases,
+the metric xref:using-timefold-solver/benchmarking-and-tweaking.adoc#benchmarkReportMoveEvaluationSpeedSummary[__move evaluation speed per second__] makes more sense,
+as it will count the number of moves individually.
```
triceo, line 46:
```suggestion
as the move count will stay stable no matter how many score calculations the moves perform.
```

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -596,6 +596,27 @@ Useful for comparing different score calculators and/or constraint implementatio
...
+[IMPORTANT]
+====
+When improving your move evaluation,
+it's important to note that comparing a configuration
+that uses xref:optimization-algorithms/move-selector-reference.adoc#ruinRecreateMoveSelector[Ruin and Recreate] moves with one that doesn't wouldn't be fair.
+This is because the configuration using Ruin and Recreate will likely execute fewer moves than one that doesn't.
+On the other hand, the Ruin and Recreate moves will likely calculate the score more times, improving the __score calculation speed__.
+The recreate step runs a construction heuristic,
+which uses greedy logic to find a better location to assign each one of the entities removed with the ruin step.
+====
```
triceo, line 618: This note arguably applies to score calculation count, but is irrelevant in move count, no? Move counts are the same regardless of what moves are used.

(It is true that RnR is slow and therefore will decrease move count, but you can say that about any slow move. This is not specific to RnR.)

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -753,6 +774,34 @@ After those few seconds of initialization, the calculation speed is relatively s
...
+image::using-timefold-solver/benchmarking-and-tweaking/moveEvaluationSpeedStatistic.png[align="center"]
+
+
+[NOTE]
+====
+The initial high calculation speed is typical during solution initialization:
+it's far easier to calculate the score of a solution if only a handful planning entities have been initialized,
+than when all the planning entities are initialized.
+
+After those few seconds of initialization, the evaluation speed is relatively stable,
```
triceo, line 800:
```suggestion
After the construction heuristic phase, the evaluation speed is relatively stable,
```

Let's not promise seconds. :-) We've seen it can be minutes, hours.

triceo (approved):
LGTM after the suggested changes are applied.

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
zepfred: Makes sense.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseBuilder.java
```diff
@@ -58,6 +58,7 @@ public EntityPlacer<Solution_> getEntityPlacer() {
 
     @Override
     public DefaultConstructionHeuristicPhase<Solution_> build() {
+        this.setEnableCollectMetrics(false);
```
triceo, line 61: Arguably, if this is only used in one place and the default is true, this should rather be called `disableMetricCollection()`, inverting the logic.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhase.java
```diff
@@ -130,7 +130,9 @@ public void stepStarted(LocalSearchStepScope<Solution_> stepScope) {
     public void stepEnded(LocalSearchStepScope<Solution_> stepScope) {
         super.stepEnded(stepScope);
         decider.stepEnded(stepScope);
-        collectMetrics(stepScope);
+        if (stepScope.isPhaseEnableCollectMetrics()) {
+            collectMetrics(stepScope);
+        }
```
triceo, line 135: Why would metrics collection ever be disabled in Local Search? What changed that the metric collection is now conditional?

Also, let's change the method. Arguably, this belongs on the phase scope, and then it can be called `isMetricCollectionEnabled()`. If we need it, that is.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractStepScope.java
```diff
@@ -44,10 +46,23 @@ public void setBestScoreImproved(Boolean bestScoreImproved) {
...
+        }
+    }
+
     // ************************************************************************
     // Calculated methods
     // ************************************************************************
 
+    public boolean isPhaseEnableCollectMetrics() {
+        return getPhaseScope().isEnableCollectMetrics();
+    }
```
triceo, line 64: Ah, I see, it _is_ on the phase, you just make it available on the step like this.
I wouldn't - it increases amount of code in one place (here) to decrease amount of code later (at call time). But arguably, it adds more code than it removes.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -355,4 +381,17 @@ public void destroyYielding() {
         }
     }
 
+    public void incrementMoveEvaluationCountPerType(Move<?> move) {
+        addMoveEvaluationCountPerType(move.getSimpleMoveTypeDescription(), 1L);
+    }
+
+    public void addMoveEvaluationCountPerType(String moveType, long count) {
```
triceo, line 388: Why is this method public, if we already have the one above?

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -62,6 +67,11 @@ public class SolverScope<Solution_> {
      */
     private final Map<Tags, List<AtomicReference<Number>>> stepScoreMap = new ConcurrentHashMap<>();
 
+    /**
+     * Used for tracking move count per move type
+     */
+    private final Map<String, AtomicLong> moveEvaluationCountPerTypeMap = new ConcurrentHashMap<>();
```
triceo, line 73: If the map is already concurrent, do you need the `AtomicLong`?
Wouldn't locking the entire write operation be a much better guarantee of atomicity?

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -4,35 +4,44 @@
 [#scoreCalculationPerformanceTricksOverview]
 == Overview
 
-The `Solver` will normally spend most of its execution time running score calculation,
+The `Solver` will normally spend most of its execution time evaluating moves and running score calculation,
 which is called in its deepest loops.
-Faster score calculation will return the same solution in less time with the same algorithm,
+Faster move evaluation will return the same solution in less time with the same algorithm,
```
triceo, line 9: In this case, we are arguably talking about faster score calculation.
That is the deciding factor for speed, if everything else is equal. Move evaluation speed is _determined_ by score calculation speed.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -4,35 +4,44 @@
...
+This is a good measurement of Move evaluation performance,
 despite that it is affected by non-score calculation execution time.
 It depends on the problem scale of the problem dataset.
 Normally, even for large scale problems, it is higher than ``1000``,
 except if you are using xref:constraints-and-score/score-calculation.adoc#easyScoreCalculation[``EasyScoreCalculator``].
 
 [IMPORTANT]
 ====
-When improving your score calculation, focus on maximizing the score calculation speed,
+When improving your move evaluation, focus on maximizing the move evaluation speed,
```
triceo, line 25: See, I don't agree with this statement.

To improve your performance, you have to increase your score calculation speed - if you focus on move evaluation speed, it means you start removing slow moves, and that is counter-productive. You need to increase the score calculation speed while _keeping_ the existing moves.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -4,35 +4,44 @@
...
+The `Solver` will normally spend most of its execution time evaluating moves and running score calculation,
 which is called in its deepest loops.
-Faster score calculation will return the same solution in less time with the same algorithm,
+Faster move evaluation will return the same solution in less time with the same algorithm,
 which normally means a better solution in equal time.
 
 
-[#scoreCalculationSpeed]
-== Score calculation speed
+[#moveEvaluationSpeed]
```
triceo, line 13: Have you checked all the references to this?
Antora has a habit of failing when refs don't exist.

Also, I'd call this section "Move evaluation and score calculation speed".

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -458,5 +467,7 @@ For example, move multiple items from the same container to another container.
 == `stepLimit` benchmark
 
 Not all score constraints have the same performance cost.
-Sometimes one score constraint can kill the score calculation performance outright.
-Use the xref:using-timefold-solver/benchmarking-and-tweaking.adoc#benchmarker[Benchmarker] to do a one minute run and check what happens to the score calculation speed if you comment out all but one of the score constraints.
\ No newline at end of file
+Sometimes one score constraint can kill the move evaluation performance outright.
```
triceo, line 470:
```suggestion
Sometimes one score constraint can performance outright.
```

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -595,6 +595,27 @@ Useful for comparing different score calculators and/or constraint implementatio
 (presuming that the solver configurations do not differ otherwise).
 Also useful to measure the scalability cost of an extra constraint.
 
+[IMPORTANT]
+====
+When improving your score speed,
+it's important to note that comparing a configuration
+that uses xref:optimization-algorithms/move-selector-reference.adoc#ruinRecreateMoveSelector[Ruin and Recreate] moves
```
triceo, line 602:
```suggestion
When improving your score calculation speed,
comparing a configuration that uses 
xref:optimization-algorithms/move-selector-reference.adoc#ruinRecreateMoveSelector[Ruin and Recreate] moves
```

docs/src/modules/ROOT/pages/using-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -753,6 +774,50 @@ After those few seconds of initialization, the calculation speed is relatively s
...
+    </problemBenchmarks>
+----
+
+.Move Evaluation Speed Statistic
+image::using-timefold-solver/benchmarking-and-tweaking/moveEvaluationSpeedStatistic.png[align="center"]
+
+
+[NOTE]
+====
+The initial high calculation speed is typical during solution initialization:
```
triceo, line 796:
```suggestion
The initial high evaluation speed is typical during solution initialization:
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -287,7 +287,7 @@ but not for slow stepping algorithms (such as Tabu Search).
 
 Both cause congestion in xref:enterprise-edition/enterprise-edition.adoc#multithreadedSolving[multi-threaded solving] with most appenders, see below.
 
-In Eclipse, `debug` logging to the console tends to cause congestion with a score calculation speeds above 10 000 per second.
+In Eclipse, `debug` logging to the console tends to cause congestion with a move evaluation speeds above 10 000 per second.
```
triceo, line 290:
```suggestion
In Eclipse, `debug` logging to the console tends to cause congestion with move evaluation speeds above 10 000 per second.
```

triceo (approved):
LGTM after comments are resolved.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhase.java
zepfred: > Why would metrics collection ever be disabled in Local Search? What changed that the metric collection is now conditional?

Do we want to account for the number of step moves in the R&R construction heuristic phase? If so, then we can remove the if block. However, I used the same idea from the move speed, viewing the R&R as a single move.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
zepfred: We use it in the MT synchronization process.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
zepfred: > If the map is already concurrent, do you need the AtomicLong?

Considering no metric collects this information over time and the map is not being updated concurrently, we can use the Long instead.

> Wouldn't locking the entire write operation be a much better guarantee of atomicity?

Since we are using ConcurrentHashMap and atomicity is no longer necessary, I don't see a reason to do that.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
zepfred: Yes, I did check the references.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhase.java
triceo: The RnR should be viewed as a single move, that is definitely correct. But the RnR doesn't run Local Search, so why would this be affected? LS is _outside_ of the RnR move, it reports to the CH, so I'm not sure why those steps are reported here.

zepfred: Good point; they are not reported on the LS. Let me remove the if block.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractStepScope.java
```diff
@@ -44,10 +46,18 @@ public void setBestScoreImproved(Boolean bestScoreImproved) {
         this.bestScoreImproved = bestScoreImproved;
     }
 
+    public void incrementMoveEvaluationCount(Move<?> move) {
+        if (getPhaseScope().isEnableCollectMetrics()) {
```
triceo, line 50: Let's change the name; 'isMetricCollectionEnabled()', better English.
Also, it would be nice if the method were documented - why do we ever need to disable metrics?

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -93,7 +93,12 @@ public void setBestSolutionStepIndex(int bestSolutionStepIndex) {
 
     public abstract AbstractStepScope<Solution_> getLastCompletedStepScope();
 
-    public boolean isEnableCollectMetrics() {
+    /**
+     * @return true, if the metrics collection, such as
+     *         {@link ai.timefold.solver.core.config.solver.monitoring.SolverMetric#MOVE_COUNT_PER_TYPE MOVE_COUNT_PER_TYPE},
+     *         is enabled; false, if the collection is disabled for the phase.
```
triceo, line 99:
```suggestion
     * @return true, if the metrics collection, such as
     *         {@link ai.timefold.solver.core.config.solver.monitoring.SolverMetric#MOVE_COUNT_PER_TYPE MOVE_COUNT_PER_TYPE},
     *         is enabled.
     *.        This is disabled for nested phases, such as Construction heuristics in Ruin and Recreate.
```

triceo:
There was something still bugging me about this PR. So I opened the IDE and looked at it. I know what it was now. Disabling metric collection is a very specific thing, it only ever exists for RnR. Yet, it proliferates everywhere - to every phase, to every step, everywhere. 

I fixed it by only dealing with disabling metric collection when already in RnR. If we ever need it in more phases than here, we can gentrify it - but right now, the correct decision was to not introduce this anywhere but in RnR.

Please take a look.

zepfred:
> There was something still bugging me about this PR. So I opened the IDE and looked at it. I know what it was now. Disabling metric collection is a very specific thing, it only ever exists for RnR. Yet, it proliferates everywhere - to every phase, to every step, everywhere.
> 
> I fixed it by only dealing with disabling metric collection when already in RnR. If we ever need it in more phases than here, we can generify it - but right now, the correct decision was to not introduce this anywhere but in RnR.
> 
> I have similarly updated the enterprise PR, which no longer needs to check whether metrics are enabled. As long as RnR doesn't trigger any metrics, nobody else (not LS, not PS, ...) needs to know anything about metrics being enabled or disabled.
> 
> Please take a look and check my thinking.

Looks good to me!

PR #1075, opened 2024-09-05
ci: use ARM64 runners

triceo:


triceo:
- Solver CI w/ Java: 30 % slower.
- Solver CI w/ Python: 35 % slower.
- Native solver CI: 18 % slower.

All wall clock times, start to finish, as reported by Github Actions.

Conclusion: Even though these machines are cheaper, they are also much slower. The overall cost will be higher using these machines.

PR #1084, opened 2024-09-12
feat: add solver job start event

zepfred:
This pull request adds a new job configuration that allows an action to be run when the solver starts the solving process. The current approach uses a `PhaseLifecycleListenerAdapter` to trigger the event because the API `SolverEventListener` would need to be changed in order to add a new solver event.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
@@ -84,10 +84,10 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
     /**
      * Sets the consumer for when the solver starts its solving process.
      *
-     * @param startSolverJobConsumer never null, called only once when the solver is starting the solving process
+     * @param solverJobStartedConsumer never null, called only once when the solver is starting the solving process
      * @return this, never null
      */
-    SolverJobBuilder<Solution_, ProblemId_> withStartSolverJobConsumer(Consumer<? super Solution_> startSolverJobConsumer);
+    SolverJobBuilder<Solution_, ProblemId_> withSolverJobStartedConsumer(Consumer<? super Solution_> solverJobStartedConsumer);
 
     /**
      * Sets the custom exception handler.
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
@@ -82,6 +82,14 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
             withFirstInitializedSolutionConsumer(Consumer<? super Solution_> firstInitializedSolutionConsumer);
 
     /**
+     * Sets the runnable action for when the solver starts its solving process.
+     *
+     * @param startSolverJobHandler never null, called only once when the solver is starting the solving process
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> withStartSolverJobHandler(Runnable startSolverJobHandler);
+
+    /**
      * Sets the custom exception handler.
      *
      * @param exceptionHandler never null, called if an exception or error occurs. If null it defaults to logging the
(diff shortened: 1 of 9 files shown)


triceo:
We've already added events before, haven't we? The solution initialization event comes to mind.
What makes this one different, why are we suddenly afraid of changing the API for it?

zepfred:
> We've already added events before, haven't we? The solution initialization event comes to mind. 

We used the `PhaseLifecycleListenerAdapter` for this specific event, which was the appropriate choice as it relates to the phase itself.

> What makes this one different, why are we suddenly afraid of changing the API for it?

In this case, it is a solver event and should be called once during `DefaultSolver#solvingStarted`, but it can also be solved using the phase event lifecycle as proposed by this PR. My goal was to keep this pull request simple because changing the contract `SolverEventListener` would impact many other areas of the solver.

We may consider creating a separate issue to refactor `SolverEventListener` to make the solver events system more extensible.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -62,6 +65,11 @@ void consumeFirstInitializedSolution(Solution_ firstInitializedSolution) {
         scheduleFirstInitializedSolutionConsumption();
     }
 
+    // Called on the solver thread when it starts
+    void triggerStartSolverJob() {
```
triceo, line 69: Shouldn't this run on a different thread? 
The solver thread cannot be blocked by user code. Have you seen what platform does in the listeners? It's heavy.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -81,6 +81,14 @@ default SolverJobBuilder<Solution_, ProblemId_> withProblem(Solution_ problem) {
     SolverJobBuilder<Solution_, ProblemId_>
             withFirstInitializedSolutionConsumer(Consumer<? super Solution_> firstInitializedSolutionConsumer);
 
+    /**
+     * Sets the runnable action for when the solver starts its solving process.
+     *
+     * @param startSolverJobHandler never null, called only once when the solver is starting the solving process
+     * @return this, never null
+     */
+    SolverJobBuilder<Solution_, ProblemId_> withStartSolverJobHandler(Runnable startSolverJobHandler);
```
triceo, line 90: I'm wondering... how do I find out which job is starting? Maybe we want to replace `Runnable` with something that can give some context? Not sure.

triceo (commented):
Looks good so far.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverJob.java
```diff
@@ -306,4 +311,24 @@ public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
             }
         }
     }
+
+    /**
+     * A listener that is triggered once when the solver starts the solving process.
+     */
+    private final class StartSolverJobPhaseLifecycleListener extends PhaseLifecycleListenerAdapter<Solution_> {
```
triceo, line 318: `SolverLifecycleListenerAdapter` can listen to `solvingStarted` events. This is what you should listen to, instead of reading the phases.

triceo (commented):
Please switch to listen to the proper event. After that, LGTM.

triceo (dismissed):

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverManager.java
```diff
@@ -170,10 +175,12 @@ public void terminateEarly(ProblemId_ problemId) {
     public void close() {
         solverThreadPool.shutdownNow();
         problemIdToSolverJobMap.values().forEach(DefaultSolverJob::close);
+        resourcesToRelease.forEach(DefaultSolverJob::close);
     }
 
     void unregisterSolverJob(ProblemId_ problemId) {
-        problemIdToSolverJobMap.remove(getProblemIdOrThrow(problemId));
+        var job = problemIdToSolverJobMap.remove(getProblemIdOrThrow(problemId));
+        resourcesToRelease.add(job);
```
triceo, line 183: This creates a bit of a memory leak for as long as the solver manager is alive.
If we're unregistering the job, why not close it immediately?
We should only remember the jobs that are never unregistered.

triceo: Also, arguably, if the job requires closing, then the job should implement `Closable` or `AutoClosable`.

triceo: Also, Sonar reports a minor issue.

triceo (approved):

PR #1091, opened 2024-09-16
fix: prevent excessive logging of constraint weights

triceo:
Contains two commits:

- First fixes the actual issue.
- Second refactors some of the code to decrease complexity, as recommended by Sonar. The logic of the code did not change, only who calls what from where.

Merge individually, do not squash.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
@@ -248,16 +248,11 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
     }
 
     @Override
-    public Solution_ cloneWorkingSolution() {
-        return cloneSolution(workingSolution);
-    }
-
-    @Override
     public Solution_ cloneSolution(Solution_ originalSolution) {
         SolutionDescriptor<Solution_> solutionDescriptor = getSolutionDescriptor();
-        Score_ originalScore = (Score_) solutionDescriptor.getScore(originalSolution);
+        Score_ originalScore = solutionDescriptor.getScore(originalSolution);
         Solution_ cloneSolution = solutionDescriptor.getSolutionCloner().cloneSolution(originalSolution);
-        Score_ cloneScore = (Score_) solutionDescriptor.getScore(cloneSolution);
+        Score_ cloneScore = solutionDescriptor.getScore(cloneSolution);
         if (scoreDirectorFactory.isAssertClonedSolution()) {
             if (!Objects.equals(originalScore, cloneScore)) {
                 throw new IllegalStateException("Cloning corruption: "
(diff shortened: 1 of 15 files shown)


core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -297,33 +292,29 @@ protected void setCalculatedScore(Score_ score) {
...
     @Override
     public AbstractScoreDirector<Solution_, Score_, Factory_> clone() {
-        // Breaks incremental score calculation.
-        // Subclasses should overwrite this method to avoid breaking it if possible.
-        AbstractScoreDirector<Solution_, Score_, Factory_> clone =
-                (AbstractScoreDirector<Solution_, Score_, Factory_>) scoreDirectorFactory
-                        .buildScoreDirector(lookUpEnabled, constraintMatchEnabledPreference);
-        clone.setWorkingSolution(cloneWorkingSolution());
-        return clone;
+        throw new UnsupportedOperationException("Cloning score directors is not supported.");
```
zepfred, line 301: Should we update the `upgrade-to-latest-version.adoc` file to include information about this breaking change?

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -228,17 +230,6 @@ default Score_ doAndProcessMove(Move<Solution_> move, boolean assertMoveScoreFro
      */
     SupplyManager getSupplyManager();
 
-    /**
```
zepfred, line 231: Should we deprecate it before removing it?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintSessionFactory.java
```diff
@@ -81,18 +98,32 @@ public BavetConstraintSession<Score_> buildSession(Solution_ workingSolution, bo
                 castConstraint.collectActiveConstraintStreams(constraintStreamSet);
                 constraintWeightMap.put(constraint, constraintWeight);
             } else {
```
zepfred, line 100:
```suggestion
} else if (constraintWeightLoggingEnabled) {
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
triceo: `AbstractScoreDirector` is not public API and as such, I could have removed the method if I wanted to and didn't need to tell anyone. :-) I decided to keep it, but insert the exception, just to verify my assumption that it is not being used anywhere in our code.

If any users actually used this method, they deserve to suffer. (This is not public API.)

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
triceo: Same answer as above. I can afford to remove it without deprecating, because this is internal code. Nobody uses this method, as evidenced by the code compiling.

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintSessionFactory.java
triceo: Normally I would apply this change, but considering that this PR is split in two commits and I won't be squashing them, I decided to skip this one. It's mostly an aesthetic change anyway.

PR #1095, opened 2024-09-19
feat: multi-threaded solving support for Ruin and Recreate moves

triceo:


core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import ai.timefold.solver.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+import ai.timefold.solver.core.impl.testdata.domain.TestdataEntity;
+import ai.timefold.solver.core.impl.testdata.domain.TestdataSolution;
+import ai.timefold.solver.core.impl.testdata.domain.TestdataValue;
+
+import org.junit.jupiter.api.Test;
+
+class RuinRecreateMoveTest {
```
zepfred, line 21: Should we add multithreading tests using the R&R moves?

zepfred (approved):
Looks good! I have a minor comment about adding a test that reproduces the bug.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveTest.java
triceo: I was thinking about it, but the only way to do that would be to add (another) test that runs the entire solver. I don't want to do that - those tests don't actually test anything specific and they take too long. We should have less of those, not more.

PR #1107, opened 2024-09-23
feat: Bring Move into public API

triceo:
- Introduces a new public `Move` interface, largely similar to the old non-public one. One major difference is that this interface does not require or even support undo moves, they are now automatically created by the solver.
- This new interface requires the introduction of a whole bunch of meta-model interfaces, which are public API replacements for the internal solution, entity and variable descriptors.
- Converts CH and LS to use the new move interface. For backwards compatibility though, the selectors still produce moves using the old interface, and therefore `LegacyMoveAdapter` converts between the two.
- Removes the undo move tabu functionality as there is no way to implement it now, and it wasn't used for years anyway.
- Deprecates all undo-related functionality in the old move selector framework and removes all undo move implementations from the codebase. This allowed to remove a whole bunch of test coverage which is no longer meaningful.
- Updates documentation where undo moves are discussed.


core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/PlanningEntityMetaModel.java
```diff
@@ -0,0 +1,110 @@
...
+     * @throws IllegalArgumentException if the variable does not exist on the entity
+     */
+    @SuppressWarnings("unchecked")
+    default <Value_> VariableMetaModel<Solution_, Entity_, Value_> variable(String variableName) {
+        for (var variableMetaModel : variables()) {
+            if (variableMetaModel.name().equals(variableName)) {
+                return (VariableMetaModel<Solution_, Entity_, Value_>) variableMetaModel;
+            }
+        }
+        throw new IllegalArgumentException(
```
zepfred, line 74: Should we use the "formatted" instead?

core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/PlanningVariableMetaModel.java
```diff
@@ -0,0 +1,51 @@
...
+    default boolean isGenuine() {
+        return true;
+    }
+
+    /**
+     * Returns whether the planning variable allows null values.
+     *
+     * @return {@code true} if the planning variable allows null values, {@code false} otherwise.
+     */
+    boolean allowsUnassigned();
```
zepfred, line 42: I wonder if `allowsUnassigned` should be part of `VariableMetaModel`. If we need to cast variable models to check this, perhaps we could avoid that by adding it to the parent class.

core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/PlanningVariableMetaModel.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Returns whether the planning variable is chained.
+     *
+     * @return {@code true} if the planning variable is chained, {@code false} otherwise.
+     */
+    boolean isChained();
```
zepfred, line 49: Same as above

core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/ShadowVariableMetaModel.java
```diff
@@ -0,0 +1,37 @@
+package ai.timefold.solver.core.api.domain.metamodel;
+
+/**
+ * A {@link VariableMetaModel} that represents a shadow planning variable.
```
zepfred, line 4: Custom shadow variables can be modified by user code

core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/DefaultUnassignedLocation.java
```diff
@@ -0,0 +1,11 @@
+package ai.timefold.solver.core.api.domain.metamodel;
+
+record DefaultUnassignedLocation() implements UnassignedLocation {
```
zepfred, line 3: I understand your strategy, but I suggest a different approach. What if `DefaultUnassignedLocation` extends `DefaultLocationInList` and throws exceptions for `DefaultUnassignedLocation::entity()` or `DefaultUnassignedLocation::index()`?

We still fail fast, and there will be no need to cast values in the internal classes, which I didn't like when using `LocationInList` previously. Additionally, we could add a method `ElementLocation::isAssigned()` and override it in `DefaultUnassignedLocation`.

core/src/main/java/ai/timefold/solver/core/api/move/Move.java
```diff
@@ -0,0 +1,138 @@
...
+     * <p>
+     * This method is only called after {@link #execute(MutableSolutionState)}, which might affect the return values.
+     * <p>
+     * Duplicate entries in the returned {@link Collection} are best avoided.
+     * The returned {@link Collection} is recommended to be in a stable order.
+     * For example, use {@link List} or {@link LinkedHashSet}, but not {@link HashSet}.
+     *
+     * @return never null
+     */
+    default Collection<?> extractPlanningEntities() {
```
zepfred, line 97: Are there any move implementations that do not provide these methods? Does it make sense to have them for any move? Depending on the answers, why not transfer these tabu-related operations to a specific tabu-move interface, such as `TabuMove extends Move`?

Next to that, the TB logic should use `TabuMove` instead of
`Move`. Does it make sense?

core/src/main/java/ai/timefold/solver/core/api/move/MutableSolutionState.java
```diff
@@ -0,0 +1,80 @@
...
+ * although we will strive to avoid this as much as possible.
+ * <p>
+ * We encourage you to try the API and give us feedback on your experience with it,
+ * before we finalize the API.
+ * Please direct your feedback to
+ * <a href="https://github.com/TimefoldAI/timefold-solver/discussions">Timefold Solver Github</a>.
+ * 
+ * @param <Solution_>
+ */
+public interface MutableSolutionState<Solution_> extends SolutionState<Solution_> {
```
zepfred, line 27: What about the unassign operation?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningEntityMetaModel.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public List<VariableMetaModel<Solution_, Entity_, ?>> variables() {
+        return Collections.unmodifiableList(variables);
+    }
+
+    void addVariable(VariableMetaModel<Solution_, Entity_, ?> variable) {
+        if (variable.entity() != this) {
+            throw new IllegalArgumentException("The variable (" + variable + ") is not part of this entity (" + this + ").");
```
zepfred, line 50: Let's use "formatted"

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningSolutionMetaModel.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public List<PlanningEntityMetaModel<Solution_, ?>> entities() {
+        return Collections.unmodifiableList(entities);
+    }
+
+    void addEntity(PlanningEntityMetaModel<Solution_, ?> planningEntityMetaModel) {
+        if (planningEntityMetaModel.solution() != this) {
+            throw new IllegalArgumentException("The entityMetaModel (" + planningEntityMetaModel
```
zepfred, line 38: Let's use "formatted"

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultShadowVariableMetaModel.java
```diff
@@ -0,0 +1,36 @@
...
+        return (Class<Value_>) variableDescriptor.getVariablePropertyType();
+    }
+
+    @Override
+    public String name() {
+        return variableDescriptor.getVariableName();
+    }
+
+    @Override
+    public ShadowVariableDescriptor<Solution_> variableDescriptor() {
```
zepfred, line 26: Do we need this method?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/InnerVariableMetaModel.java
```diff
@@ -0,0 +1,10 @@
+package ai.timefold.solver.core.impl.domain.solution.descriptor;
+
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+
+public sealed interface InnerVariableMetaModel<Solution_>
```
zepfred, line 5: Why not move this interface to the public API and return a generic type? Alternatively, we could add the method to `VariableMetaModel`, as it appears to be a common method for all its child classes.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedListVariableStateSupply.java
```diff
@@ -4,16 +4,16 @@
...
 import java.util.Objects;
 
+import ai.timefold.solver.core.api.domain.metamodel.ElementLocation;
+import ai.timefold.solver.core.api.domain.metamodel.LocationInList;
 import ai.timefold.solver.core.api.score.director.ScoreDirector;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
-import ai.timefold.solver.core.impl.heuristic.selector.list.ElementLocation;
-import ai.timefold.solver.core.impl.heuristic.selector.list.LocationInList;
 
 final class ExternalizedListVariableStateSupply<Solution_>
```
zepfred, line 12: Here is one of the places we could enhance the experience by modifying the default contract so that `ExternalizedListVariableStateSupply::getLocationInList` returns `LocationInList`. I'm uncertain whether calling `ensureAssigned()` would disrupt existing code, but if not, it would be logical to apply this validation when retrieving an element
`getLocationInList`.

My point is to return something that can be used without casting or converting with `ensureAssigned`. On the other hand, I understand that changing the contract could impact many other areas.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractMove.java
```diff
@@ -7,26 +7,24 @@
...
-    public final Move<Solution_> doMove(ScoreDirector<Solution_> scoreDirector) {
-        var undoMove = createUndoMove(scoreDirector);
-        doMoveOnly(scoreDirector);
-        return undoMove;
-    }
-
     @Override
     public final void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
-        doMoveOnGenuineVariables(scoreDirector);
+        // LegacyMoveAdapter does not wrap the score director in a VariableChangeRecordingScoreDirector
```
zepfred, line 22: I'm unsure if the comment makes sense. It looks like we will always use `VariableChangeRecordingScoreDirector`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractSimplifiedMove.java
```diff
@@ -10,20 +11,20 @@
...
-        var recordingScoreDirector = new VariableChangeRecordingScoreDirector<>(scoreDirector);
-        doMoveOnly(recordingScoreDirector);
-        return new RecordedUndoMove<>(this, recordingScoreDirector.getVariableChanges());
-    }
-
     @Override
     public final void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
-        doMoveOnGenuineVariables(scoreDirector);
-        scoreDirector.triggerVariableListeners();
+        // LegacyMoveAdapter does not wrap the score director in a VariableChangeRecordingScoreDirector
```
zepfred, line 21: ditto

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/factory/LegacyIteratorAdapter.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+        return moveIterator.hasNext();
+    }
+
+    @Override
+    public Move<Solution_> next() {
+        return new ai.timefold.solver.core.impl.heuristic.move.LegacyMoveAdapter<>(moveIterator.next());
```
zepfred, line 22: Is there a reason for not importing the class?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/SubChainChangeMove.java
```diff
@@ -68,19 +68,13 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
     protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
-        Object firstEntity = subChain.getFirstEntity();
-        Object lastEntity = subChain.getLastEntity();
-        Object oldFirstValue = variableDescriptor.getValue(firstEntity);
+        var firstEntity = subChain.getFirstEntity();
+        var lastEntity = subChain.getLastEntity();
+        var oldFirstValue = variableDescriptor.getValue(firstEntity);
         // Close the old chain
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 77: Let's make it consistent and rename the var to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/SubChainReversingSwapMove.java
```diff
@@ -67,34 +69,27 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
+        var leftFirstValue = variableDescriptor.getValue(leftFirstEntity);
+        var leftLastEntity = leftSubChain.getLastEntity();
+        var rightFirstEntity = rightSubChain.getFirstEntity();
+        var rightFirstValue = variableDescriptor.getValue(rightFirstEntity);
+        var rightLastEntity = rightSubChain.getLastEntity();
+        var leftLastEntityValue = variableDescriptor.getValue(leftLastEntity);
+        var rightLastEntityValue = variableDescriptor.getValue(rightLastEntity);
         // Change the entities
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 83: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/EntityOrderInfo.java
```diff
@@ -86,9 +87,12 @@ public <Node_> Node_ predecessor(Node_ object, ListVariableStateSupply<?> listVa
     }
 
     public <Node_> boolean between(Node_ start, Node_ middle, Node_ end, ListVariableStateSupply<?> listVariableStateSupply) {
-        var startElementLocation = (LocationInList) listVariableStateSupply.getLocationInList(start);
-        var middleElementLocation = (LocationInList) listVariableStateSupply.getLocationInList(middle);
-        var endElementLocation = (LocationInList) listVariableStateSupply.getLocationInList(end);
+        var startElementLocation = listVariableStateSupply.getLocationInList(start)
```
zepfred, line 90: ditto

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/KOptListMoveIterator.java
```diff
@@ -376,8 +377,10 @@ int getShortestCycleIdentifier(EntityOrderInfo entityOrderInfo, Object[] removeE
 
     private int getSegmentSize(EntityOrderInfo entityOrderInfo, Object from, Object to) {
         var entityToEntityIndex = entityOrderInfo.entityToEntityIndex();
-        var startElementLocation = (LocationInList) listVariableStateSupply.getLocationInList(from);
-        var endElementLocation = (LocationInList) listVariableStateSupply.getLocationInList(to);
+        var startElementLocation = listVariableStateSupply.getLocationInList(from)
```
zepfred, line 380: Let's apply `var` to all possible places.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/MultipleDelegateList.java
```diff
@@ -70,7 +70,7 @@ public void applyChangesFromCopy(MultipleDelegateList<?> copy) {
 
     public int getIndexOfValue(ListVariableStateSupply<?> listVariableStateSupply, Object value) {
         var elementLocation = listVariableStateSupply.getLocationInList(value);
-        if (elementLocation instanceof LocationInList elementLocationInList) {
+        if (elementLocation instanceof LocationInList<?> elementLocationInList) {
```
zepfred, line 73: This is kind of casting operation we can avoid by changing the `ListVariableStateSupply` contract.

See my suggestions related to `LocationInList`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/TwoOptListMove.java
```diff
@@ -120,11 +109,11 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
     }
 
     private void doTailSwap(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
-        List<Object> firstListVariable = variableDescriptor.getValue(firstEntity);
-        List<Object> secondListVariable = variableDescriptor.getValue(secondEntity);
-        int firstOriginalSize = firstListVariable.size();
-        int secondOriginalSize = secondListVariable.size();
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 112: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/TwoOptListMove.java
```diff
@@ -153,8 +142,8 @@ private void doTailSwap(ScoreDirector<Solution_> scoreDirector) {
     }
 
     private void doSublistReversal(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
-        List<Object> listVariable = variableDescriptor.getValue(firstEntity);
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 145: ditto

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/TwoOptListMove.java
```diff
@@ -133,12 +122,12 @@ private void doTailSwap(ScoreDirector<Solution_> scoreDirector) {
                 secondEdgeEndpoint,
                 secondOriginalSize);
 
-        List<Object> firstListVariableTail = firstListVariable.subList(firstEdgeEndpoint, firstOriginalSize);
-        List<Object> secondListVariableTail = secondListVariable.subList(secondEdgeEndpoint, secondOriginalSize);
+        var firstListVariableTail = firstListVariable.subList(firstEdgeEndpoint, firstOriginalSize);
+        var secondListVariableTail = secondListVariable.subList(secondEdgeEndpoint, secondOriginalSize);
 
         int tailSizeDifference = secondListVariableTail.size() - firstListVariableTail.size();
```
zepfred, line 128: Let's apply `var` to all possible places.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
```diff
@@ -84,11 +81,23 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
...
-            var location = (LocationInList) listVariableStateSupply.getLocationInList(ruinedValue);
+            var location = listVariableStateSupply.getLocationInList(ruinedValue)
+                    .ensureAssigned();
             entityToNewPositionMap.computeIfAbsent(location.entity(), ignored -> new TreeSet<>())
                     .add(new RuinedLocation(ruinedValue, location.index()));
+            entityToInsertedValuesMap.computeIfAbsent(location.entity(), ignored -> new ArrayList<>()).add(ruinedValue);
+        }
+
+        for (var entry : entityToInsertedValuesMap.entrySet()) {
+            recordingScoreDirector.recordListAssignment(listVariableDescriptor, entry.getKey(), entry.getValue());
```
zepfred, line 98: Is the call to `recordingScoreDirector.afterListVariableChanged` no longer necessary?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListSwapMove.java
```diff
@@ -119,20 +119,15 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
-    public ListSwapMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
-        return new ListSwapMove<>(variableDescriptor, rightEntity, rightIndex, leftEntity, leftIndex);
-    }
-
     @Override
     protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
-        Object leftElement = variableDescriptor.getElement(leftEntity, leftIndex);
-        Object rightElement = variableDescriptor.getElement(rightEntity, rightIndex);
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 124: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/SubListChangeMove.java
```diff
@@ -88,23 +87,17 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
-    @Override
-    protected AbstractMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
-        return new SubListChangeMove<>(variableDescriptor, destinationEntity, destinationIndex, length, sourceEntity,
-                sourceIndex, reversing);
-    }
-
     @Override
     protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 92: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/ChangeMove.java
```diff
@@ -48,15 +48,9 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
-    @Override
-    public ChangeMove<Solution_> createUndoMove(ScoreDirector<Solution_> scoreDirector) {
-        Object oldValue = variableDescriptor.getValue(entity);
-        return new ChangeMove<>(variableDescriptor, entity, oldValue);
-    }
-
     @Override
     protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 53: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
```diff
@@ -81,10 +76,10 @@ public SwapMove<Solution_> rebase(ScoreDirector<Solution_> destinationScoreDirec
 
     @Override
     protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector) {
-        InnerScoreDirector<Solution_, ?> innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
-        for (GenuineVariableDescriptor<Solution_> variableDescriptor : variableDescriptorList) {
-            Object oldLeftValue = variableDescriptor.getValue(leftEntity);
-            Object oldRightValue = variableDescriptor.getValue(rightEntity);
+        var innerScoreDirector = (VariableDescriptorAwareScoreDirector<Solution_>) scoreDirector;
```
zepfred, line 79: Let's rename it to `castScoreDirector`

core/src/main/java/ai/timefold/solver/core/impl/move/director/EphemeralMoveDirector.java
```diff
@@ -0,0 +1,48 @@
...
+import ai.timefold.solver.core.api.move.Move;
+import ai.timefold.solver.core.impl.score.director.VariableDescriptorAwareScoreDirector;
+
+/**
+ * The only move director that supports undoing moves.
+ * Moves are undone when the director is {@link #close() closed}.
+ * 
+ * @param <Solution_>
+ */
+public final class EphemeralMoveDirector<Solution_> extends MoveDirector<Solution_>
```
zepfred, line 14: Cool!

core/src/main/java/ai/timefold/solver/core/impl/move/director/ListVariableAfterUnassignmentAction.java
```diff
@@ -0,0 +1,20 @@
...
+import ai.timefold.solver.core.api.move.Rebaser;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.VariableDescriptorAwareScoreDirector;
+
+record ListVariableAfterUnassignmentAction<Solution_>(Object element,
+        ListVariableDescriptor<Solution_> variableDescriptor) implements ChangeAction<Solution_> {
+
+    @Override
+    public void undo(VariableDescriptorAwareScoreDirector<Solution_> scoreDirector) {
+        scoreDirector.beforeListVariableElementAssigned(variableDescriptor, element);
```
zepfred, line 12: No actions here?

core/src/main/java/ai/timefold/solver/core/impl/move/director/ListVariableBeforeAssignmentAction.java
```diff
@@ -0,0 +1,20 @@
...
+import ai.timefold.solver.core.api.move.Rebaser;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.VariableDescriptorAwareScoreDirector;
+
+record ListVariableBeforeAssignmentAction<Solution_>(Object element,
+        ListVariableDescriptor<Solution_> variableDescriptor) implements ChangeAction<Solution_> {
+
+    @Override
+    public void undo(VariableDescriptorAwareScoreDirector<Solution_> scoreDirector) {
+        scoreDirector.afterListVariableElementUnassigned(variableDescriptor, element);
```
zepfred, line 12: ditto

core/src/main/java/ai/timefold/solver/core/impl/move/director/MoveDirector.java
```diff
@@ -0,0 +1,133 @@
...
+import ai.timefold.solver.core.api.move.Rebaser;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.DefaultPlanningListVariableMetaModel;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.DefaultPlanningVariableMetaModel;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.BasicVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.move.InnerMutableSolutionState;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.impl.score.director.VariableDescriptorAwareScoreDirector;
+
+public sealed class MoveDirector<Solution_>
```
zepfred, line 17: Cool!

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/ChainedChangeMoveTest.java
```diff
@@ -11,176 +11,145 @@
...
 
-        SingletonInverseVariableSupply inverseVariableSupply = SelectorTestUtils.mockSingletonInverseVariableSupply(
+        var inverseVariableSupply = SelectorTestUtils.mockSingletonInverseVariableSupply(
                 new TestdataChainedEntity[] { a1, a2, a3, b1 });
 
-        ChainedChangeMove<TestdataChainedSolution> move =
-                new ChainedChangeMove<>(variableDescriptor, a3, b1, inverseVariableSupply);
-        assertThat(move.isMoveDoable(scoreDirector)).isTrue();
-        ChainedChangeMove<TestdataChainedSolution> undoMove = move.createUndoMove(scoreDirector);
-        move.doMove(scoreDirector);
```
zepfred, line 46: This comment applies to all places where we are removing the undo validation. I understand the move logic has changed, but I am still concerned about removing the undo validation entirely.

I wonder how much effort it would take to use the ephemeral director while maintaining the undo validations.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -1842,20 +1842,7 @@ A move that is currently not doable can become doable when the working solution
 otherwise we probably shouldn't have created it in the first place.
 
 Each move has an __undo move__: a move (normally of the same type) which does the exact opposite.
```
zepfred, line 1844: Is the phrase "normally of the same type" still applicable?

docs/src/modules/ROOT/pages/upgrading-timefold-solver/backwards-compatibility.adoc
```diff
@@ -30,7 +30,23 @@ Those documented `impl` classes are reliable and safe to use (unless explicitly
...
 
-[NOTE]
-====
-The Python Solver is currently in beta and its API is subject to change.
-====
\ No newline at end of file
+[#previewFeatures]
+= Preview features
+
+Timefold Solver includes several components which are only available as preview features.
```
zepfred, line 36:
```suggestion
Timefold Solver may include components that are only available as preview features. Currently available are:
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/MoveTabuAcceptor.java
```diff
@@ -3,22 +3,16 @@
...
 import java.util.Collections;
 
-import ai.timefold.solver.core.impl.heuristic.move.Move;
+import ai.timefold.solver.core.api.move.Move;
 import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchMoveScope;
 import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchStepScope;
 
 public class MoveTabuAcceptor<Solution_> extends AbstractTabuAcceptor<Solution_> {
 
-    protected boolean useUndoMoveAsTabuMove = true;
```
zepfred, line 12: Will removing the TS undo moves affect any existing code?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/InnerVariableMetaModel.java
triceo: The public API has only the methods it needs, and no more. Variable descriptors will not be exposed as public API - they do far too much, and expose far too many internals of the solver. In fact, the reason for metamodel's existence is so that I can hide the descriptors.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedListVariableStateSupply.java
triceo: This change to require casting is very deliberate and it is intended to _prevent user error_ - it is a way to prevent NPEs. If you know something is supposed to be assigned, you better say it out loud in the code, and make sure that you handle the error case.

You'll see this pattern much more in new Java code. With sealed classes and the advent of pattern matching, this is going to be the new way of writing Java code. If you haven't yet, watch this talk to understand what is already possible in Java, what is soon going to be possible, and how that's going to make Java code better and less buggy:

https://www.youtube.com/watch?v=LgquOtnjMJ8

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/factory/LegacyIteratorAdapter.java
triceo: We now have two `Move`s - one public, the other not. Unfortunately, I cannot import both. So I chose to import the public one, and the private one uses FQNs.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/TailChainSwapMove.java
```diff
@@ -188,68 +188,43 @@ public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
...
+                reverseChain(castScoreDirector, lastEntityInChain, leftAnchor, rightEntity);
+                castScoreDirector.changeVariableFacade(variableDescriptor, leftNextEntity, rightEntity);
             }
         }
     }
 
-    protected void reverseChain(InnerScoreDirector scoreDirector, Object fromValue, Object fromEntity, Object toEntity) {
-        Object entity = fromValue;
-        Object newValue = fromEntity;
+    protected void reverseChain(VariableDescriptorAwareScoreDirector<Solution_> scoreDirector, Object fromValue,
```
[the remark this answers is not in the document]
triceo: It sure looks like it, but it isn't. :-)
Reusing it caused test failures.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/MultipleDelegateList.java
triceo: Answered above.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
triceo: No. At this point, the only thing left to do is to record the assignments, because the CH has already made all the changes. The assignments do need to be recorded though, because we need to be able to undo them later.

core/src/main/java/ai/timefold/solver/core/impl/move/director/ListVariableAfterUnassignmentAction.java
triceo: An action is a command to undo something which already happened.
This is an after-unassigned action, so the inverse of that is before-assigned.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/ChainedChangeMoveTest.java
triceo: In my opinion, there is no need. All undo operations are now centralized on the recording score director and the actions, so all testing belongs there. Doing it on the moves would be duplication.

In some tests, I still do it, because those tests rely on the state resulting from the undo moves. So there, the undo is necessary for the test to be correct.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/MoveTabuAcceptor.java
triceo: Yes and no. The undo tabu no longer does anything - so in that sense, the user code will be affected. But all the public APIs are unaffected - so user code will still compile and run without any issues.

Considering I didn't even _know_ we had undo move tabu before I started looking into Move Streams, I think it's safe to say that it's an obscure feature which won't be missed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/factory/LegacyIteratorAdapter.java
zepfred: Do we have two `LegacyMoveAdapter` classes?

triceo: No. Because the legacy move is there to convert the old move to the new move.
The reverse operation is not necessary.

core/src/main/java/ai/timefold/solver/core/api/domain/metamodel/PlanningVariableMetaModel.java
triceo: This is intentional, and is related to the pattern matching that I mentioned in another comment. List variable cannot be chained, and therefore shouldn't confuse people with that option.

That's why we have inheritance - more specific objects carry more specific methods.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultShadowVariableMetaModel.java
triceo: The method returns a more specific type, so I'd say yes. It prevents user from needing to cast.

core/src/main/java/ai/timefold/solver/core/api/move/MutableSolutionState.java
triceo: This class technically doesn't need even the assign operation. :-) They are unused, prepared for the next phase. I'll leave this there, the other operations will show up in other PRs.

core/src/main/java/ai/timefold/solver/core/api/move/Move.java
triceo: Interesting idea. Either is good, IMO. In order to not complicate the type hierarchy, I don't see a problem with keeping it like this.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/factory/LegacyIteratorAdapter.java
zepfred: I was referring to importing the class `ai.timefold.solver.core.impl.heuristic.move.LegacyMoveAdapter`.

zepfred (commented):
LGTM! 

The new API looks great, and eliminating the undo move will be a valuable enhancement.

PR #1111, opened 2024-09-26
fix: Python bytecode generation overrides, pop off result of setattr and delattr, fix enum

### LINKED ISSUE #959: ArrivalTimeUpdatingVariableListener is not working and not updating arrival time
**Describe the bug**
ArrivalTimeUpdatingVariableListener is not working and not updating arrival time in tests folder

**Expected behavior**
arrival_time variable should not be None

**Actual behavior**
arrival_time variable is None in the output when running tests code to check the output

**To Reproduce**
run test_vehicle_routing.py in tests folder (https://github.com/TimefoldAI/timefold-solver-python/blob/main/tests/test_vehicle_routing.py)

## Environment

**Timefold Solver Version or Git ref**: https://github.com/TimefoldAI/timefold-solver-python/blob/main/tests/test_vehicle_routing.py

**Python version used:** 3.10.14

**Output of `uname -a` or `ver`:** x86_64 GNU/Linux

### LINKED ISSUE #1074: Python: fix bug with shadow variable listeners

Christopher-Chianelli:
- Python bytecode generation now generates overrides for all signatures, meaning the `type(...).method` workaround can be removed.

- The bug in the variable listener was due to the generic case of setattr and delattr (when field infomation is not available due to cyclical dependencies), which call the object's `__setattr__` and `__delattr__` respectively, but did not pop off the result, causing an inconsistent stack size when an object attribute is assigned or deleted inside a loop.

- Enum values are now stored in the CPythonInterpreter's shared instance map, meaning objects will get the same Java instance for the same Python value regardless of the time of conversion.

### DIFF
--- a/python/jpyinterpreter/src/main/java/ai/timefold/jpyinterpreter/PythonClassTranslator.java
+++ b/python/jpyinterpreter/src/main/java/ai/timefold/jpyinterpreter/PythonClassTranslator.java
@@ -1,9 +1,8 @@
 package ai.timefold.jpyinterpreter;
 
-import static ai.timefold.jpyinterpreter.PythonBytecodeToJavaBytecodeTranslator.ARGUMENT_SPEC_INSTANCE_FIELD_NAME;
-
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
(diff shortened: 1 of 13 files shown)


triceo (dismissed):

triceo (approved):
Fixed the formatting.

PR #1114, opened 2024-09-26
chore: Improve SolverManager error message in Python

### LINKED ISSUE #1097: Feat: Better error log to showcase why solution failed for python.
**Is your feature request related to a problem? Please describe.**
It's really difficult to understand why solver failed as there is only solution failed text not the reason or exception

**Describe the solution you'd like**
It will be better if we get to see the exception or error causing solving failed

Christopher-Chianelli:
- Traceback, original error message and exception type is now included in the error log in Python


triceo (dismissed):
LGTM when the Sonar issues are gone.

PR #1124, opened 2024-10-01
perf: decrease lambda creation on the hot path during tuple update

triceo:
Tuple updates are done very frequently.
As I was benchmarking a different issue, I noticed that we create gigabytes of large capturing lambdas during tuple update, and therefore on the hot path.
This refactoring intends to decrease the number of lambdas created, as well as make them capture less and contain less conditional logic, helping with branch predictions.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractJoinNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractJoinNode.java
@@ -28,11 +28,11 @@ public abstract class AbstractJoinNode<LeftTuple_ extends AbstractTuple, Right_,
     protected final int inputStoreIndexRightOutTupleList;
     private final int outputStoreIndexLeftOutEntry;
     private final int outputStoreIndexRightOutEntry;
-    private final StaticPropagationQueue<OutTuple_> propagationQueue;
 
-    protected final TupleConsumer<LeftTuple_, Right_> tupleInserter;
-    protected final TupleConsumer<LeftTuple_, Right_> innerLeftUpdater;
-    protected final TupleConsumer<LeftTuple_, Right_> innerRightUpdater;
+    protected final TupleConsumer<LeftTuple_, Right_> leftTupleUpdater;
+    protected final TupleConsumer<LeftTuple_, Right_> rightTupleUpdater;
+    protected final TupleConsumer<LeftTuple_, Right_> outTupleInserter;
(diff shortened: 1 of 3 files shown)


core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractJoinNode.java
```diff
@@ -61,28 +73,6 @@ protected final void insertOutTuple(LeftTuple_ leftTuple, UniTuple<Right_> right
...
-    protected final void innerUpdateLeft(LeftTuple_ leftTuple, Consumer<Consumer<UniTuple<Right_>>> rightTupleConsumer) {
-        // Prefer an update over retract-insert if possible
-        ElementAwareList<OutTuple_> outTupleListLeft = leftTuple.getStore(inputStoreIndexLeftOutTupleList);
-        // Propagate the update for downstream filters, matchWeighers, ...
-        if (!isFiltering) {
-            for (var outTuple : outTupleListLeft) {
-                updateOutTupleLeft(outTuple, leftTuple);
-            }
-        } else {
-            rightTupleConsumer.accept(rightTuple -> {
```
triceo, line 79: This lambda will no longer be created. In my benchmarks, this took gigabytes of heap and therefore also GC cycles.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractJoinNode.java
```diff
@@ -61,28 +73,6 @@ protected final void insertOutTuple(LeftTuple_ leftTuple, UniTuple<Right_> right
...
-        if (!isFiltering || testFiltering(leftTuple, rightTuple)) {
-            insertOutTuple(leftTuple, rightTuple);
-        }
-    }
-
-    protected final void innerUpdateLeft(LeftTuple_ leftTuple, Consumer<Consumer<UniTuple<Right_>>> rightTupleConsumer) {
-        // Prefer an update over retract-insert if possible
-        ElementAwareList<OutTuple_> outTupleListLeft = leftTuple.getStore(inputStoreIndexLeftOutTupleList);
-        // Propagate the update for downstream filters, matchWeighers, ...
-        if (!isFiltering) {
```
triceo, line 74: This condition is now outside of the hot loop, which is fine, because it is a one-time check that can be externalized.

zepfred (approved):

PR #1128, opened 2024-10-03
perf: lookup indexers in HashMap

triceo:


triceo:
Closing. The improvement was a false positive.

PR #1136, opened 2024-10-07
feat: support assigned values for Recommendations as well

triceo:


core/src/main/java/ai/timefold/solver/core/api/solver/SolutionManager.java
```diff
@@ -137,10 +137,122 @@ default ScoreAnalysis<Score_> analyze(Solution_ solution, ScoreAnalysisFetchPoli
...
+     * As defined by {@link #recommendAssignment(Object, Object, Function, ScoreAnalysisFetchPolicy)},
+     * with {@link ScoreAnalysisFetchPolicy#FETCH_ALL}.
+     */
+    default <EntityOrElement_, Proposition_> List<RecommendedAssignment<Proposition_, Score_>> recommendAssignment(
+            Solution_ solution, EntityOrElement_ evaluatedEntityOrElement,
+            Function<EntityOrElement_, Proposition_> propositionFunction) {
+        return recommendAssignment(solution, evaluatedEntityOrElement, propositionFunction, FETCH_ALL);
+    }
+
+    /**
```
zepfred, line 150: Very informative!

core/src/main/java/ai/timefold/solver/core/api/solver/SolutionManager.java
```diff
@@ -137,10 +137,122 @@ default ScoreAnalysis<Score_> analyze(Solution_ solution, ScoreAnalysisFetchPoli
...
+     * This is because they shared state,
+     * and the original state of the solution was for Shift to be unassigned.
+     * </li>
+     * </ol>
+     *
+     * If instead the proposition function returned Ann and Bob directly, the immutable planning variables,
+     * this problem would have been avoided.
+     * Alternatively, the proposition function could have returned a defensive copy of the Shift.
+     *
+     * @param solution never null; must be fully initialized except for one entity or element
```
zepfred, line 231: This param explanation is unclear to me. Can we explicitly state the difference between an initialized and uninitialized solution?

core/src/main/java/ai/timefold/solver/core/impl/solver/Assigner.java
```diff
@@ -5,45 +5,47 @@
...
+    public List<Recommendation_> apply(InnerScoreDirector<Solution_, Score_> scoreDirector) {
         var solutionDescriptor = scoreDirector.getSolutionDescriptor();
         var initializationStatistics = solutionDescriptor.computeInitializationStatistics(originalSolution);
         var uninitializedCount =
                 initializationStatistics.uninitializedEntityCount() + initializationStatistics.unassignedValueCount();
         if (uninitializedCount > 1) {
             throw new IllegalStateException("""
                     Solution (%s) has (%d) uninitialized elements.
-                    Fit Recommendation API requires at most one uninitialized element in the solution."""
+                    Assignment Recommendation API requires at most one uninitialized element in the solution."""
```
zepfred, line 41: Should we also mention the fully initialized solution?

core/src/main/java/ai/timefold/solver/core/impl/solver/AssignmentProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public List<Recommendation_> apply(InnerScoreDirector<Solution_, Score_> scoreDirector) {
+        // The cloned element may already be assigned.
+        // If it is, we need to unassign it before we can run the construction heuristic.
+        var supplyManager = scoreDirector.getSupplyManager();
+        var solutionDescriptor = solverFactory.getSolutionDescriptor();
+        var listVariableDescriptor = solutionDescriptor.getListVariableDescriptor();
+        if (listVariableDescriptor != null) {
```
zepfred, line 54: I'm worried that this action will not be undone. We can unassign the element, but I can't find the corresponding undo action. Could this lead to an inconsistent state?

triceo: The action will not be undone. It doesn't need to be.
The solution at this point is already cloned - so there is no external impact.
And we do not need to undo this change for internal reasons either - the CH needs to always start from this new state.

core/src/main/java/ai/timefold/solver/core/impl/solver/AssignmentProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    private Recommendation_ execute(InnerScoreDirector<Solution_, Score_> scoreDirector, Move<Solution_> move, long moveIndex,
+            In_ clonedElement, Function<In_, Out_> propositionFunction) {
+        var undo = move.doMove(scoreDirector);
+        var newScoreAnalysis = scoreDirector.buildScoreAnalysis(fetchPolicy == ScoreAnalysisFetchPolicy.FETCH_ALL);
+        var newScoreDifference = newScoreAnalysis.diff(originalScoreAnalysis);
+        var result = propositionFunction.apply(clonedElement);
+        var recommendation = recommendationConstructor.apply(moveIndex, result, newScoreDifference);
```
zepfred, line 153: I understand what the logic is doing here and have some questions:

1. Wouldn't this create a capturing lambda?
2. We will return `DefaultRecommendedAssignment` or `DefaultRecommendedFit`. What is the advantage of using the lambda instead of just instantiating the default implementations?

triceo: Wrt. 1 - no. The lambda is already created and passed as an argument. Also, all of the variables are coming as arguments, nothing is captured externally. 

Wrt. 2 - assignment and fit have no common ancestor. Therefore it would be impossible to have the same class handle both assignments and fits. Doing it like this allows me to reuse 100 % of the code.

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -338,7 +338,7 @@ void analyzeWithUninitializedSolution(SolutionManagerSource SolutionManagerSourc
 
     @ParameterizedTest
     @EnumSource(SolutionManagerSource.class)
-    void recommendFit(SolutionManagerSource SolutionManagerSource) {
+    void recommendAssignment(SolutionManagerSource SolutionManagerSource) {
```
zepfred, line 341: Maybe we should create a separate test for that. Even though `recommendFit` is deprecated now, we still should test it to ensure no regressions are introduced.

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -386,22 +387,71 @@ void recommendFit(SolutionManagerSource SolutionManagerSource) {
...
-        assertThatThrownBy(() -> solutionManager.recommendFit(uninitializedSolution, uninitializedEntity,
+        assertThatThrownBy(() -> solutionManager.recommendAssignment(uninitializedSolution, uninitializedEntity,
                 TestdataShadowedEntity::getValue))
                 .hasMessageContaining("Solution (Generated Solution 0) has (3) uninitialized elements.");
     }
 
     @ParameterizedTest
     @EnumSource(SolutionManagerSource.class)
-    void recommendFitWithUnassigned(SolutionManagerSource SolutionManagerSource) {
+    void recommendAssignmentWithUnassigned(SolutionManagerSource SolutionManagerSource) {
```
zepfred, line 454: ditto

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -475,7 +581,7 @@ void recommendFitUninitializedSolutionWithUnassigned(SolutionManagerSource Solut
 
     @ParameterizedTest
     @EnumSource(SolutionManagerSource.class)
-    void recommendFitMultivar(SolutionManagerSource SolutionManagerSource) {
+    void recommendAssignmentMultivar(SolutionManagerSource SolutionManagerSource) {
```
zepfred, line 584: ditto

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -578,7 +778,7 @@ void recommendFitUninitializedSolutionWithMultivar(SolutionManagerSource Solutio
         solution.setMultiVarEntityList(List.of(uninitializedEntity, secondUninitializedEntity));
 
         var solutionManager = SolutionManagerSource.createSolutionManager(SOLVER_FACTORY_MULTIVAR);
-        assertThatThrownBy(() -> solutionManager.recommendFit(solution, uninitializedEntity,
+        assertThatThrownBy(() -> solutionManager.recommendAssignment(solution, uninitializedEntity,
```
zepfred, line 781: ditto

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -604,7 +804,8 @@ void recommendFitChained(SolutionManagerSource SolutionManagerSource) {
 
         var solutionManager = SolutionManagerSource.createSolutionManager(SOLVER_FACTORY_CHAINED);
         var recommendationList =
-                solutionManager.recommendFit(solution, uninitializedEntity, TestdataShadowingChainedEntity::getChainedObject);
+                solutionManager.recommendAssignment(solution, uninitializedEntity,
```
zepfred, line 807: ditto

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -672,14 +940,14 @@ void recommendFitTwoUninitializedEntityWithChained(SolutionManagerSource Solutio
         uninitializedSolution.setChainedEntityList(Arrays.asList(b1, c1, c2, uninitializedEntity, uninitializedEntity2));
 
         var solutionManager = SolutionManagerSource.createSolutionManager(SOLVER_FACTORY_CHAINED);
-        assertThatThrownBy(() -> solutionManager.recommendFit(uninitializedSolution, uninitializedEntity,
+        assertThatThrownBy(() -> solutionManager.recommendAssignment(uninitializedSolution, uninitializedEntity,
```
zepfred, line 943: The suggestion to continue testing the `recommendFit` test applies to all instances where we are changing it to `recommendAssignment`.

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -386,22 +387,71 @@ void recommendFit(SolutionManagerSource SolutionManagerSource) {
...
+                    .score()).isEqualTo(SimpleScore.of(-2));
+        });
+        var thirdRecommendation = recommendationList.get(2);
+        assertSoftly(softly -> {
+            softly.assertThat(thirdRecommendation.proposition()).isEqualTo(solution.getValueList().get(1));
+            softly.assertThat(thirdRecommendation.scoreAnalysisDiff()
+                    .score()).isEqualTo(SimpleScore.of(-2));
+        });
+        // Ensure the original solution is in its original state.
+        assertSoftly(softly -> {
```
zepfred, line 429: My comment in the `AssignmentProcessor` class relates to this validation. Please verify if I'm overlooking anything.

triceo: The original solution will be in its original state, because it is never touched. :-) Before anything happens, the solution will be cloned, and the recommendations only ever work with the clone. This test proves that.

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -459,14 +509,70 @@ void recommendFitWithUnassigned(SolutionManagerSource SolutionManagerSource) {
...
+        // At this point, entity 0 is unassigned.
+        // Entity 1 is assigned to value #1.
+        // But only entity2 should be processed for recommendations.
+        var solutionManager = SolutionManagerSource.createSolutionManager(SOLVER_FACTORY_UNASSIGNED);
+        assertThat(solutionManager).isNotNull();
+        var recommendationList =
+                solutionManager.recommendAssignment(solution, evaluatedEntity, TestdataAllowsUnassignedEntity::getValue);
+
+        // Three values means there need to be four recommendations, one extra for unassigned.
+        assertThat(recommendationList).hasSize(valueSize + 1);
```
zepfred, line 526: How will this unassigned recommendation work if we already unassign it at the beginning of the `AssignmentProcessor` logic?

triceo: The processor only unassigns the value if already assigned.
It has to, because otherwise CH wouldn't do anything. (= All values already assigned.)

This is an implementation detail. There were two options to choose from for implementing this feature:

- Either articially unassign a value to make CH work as usual.
- Or refactor CH to be able to "ignore" that all entities are already assign, and try to assign one element anyway.

I decided that the former is not only easier, but also more natural.
It has no impact on the solution on the outside, as all of this only ever happens on the internal, cloned solution.

The test proves that the code works as intended.

core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -764,13 +1101,33 @@ void recommendFitTwoUninitializedEntityWithList(SolutionManagerSource SolutionMa
...
+        var evaluatedValue = new TestdataListValueWithShadowHistory("c1");
+        var c = new TestdataListEntityWithShadowHistory("c", c0, evaluatedValue);
+        var d = new TestdataListEntityWithShadowHistory("d");
+        var solution = new TestdataListSolutionWithShadowHistory();
+        solution.setEntityList(Arrays.asList(a, b, c, d));
+        solution.setValueList(Arrays.asList(b0, c0, evaluatedValue));
+
+        var solutionManager = SolutionManagerSource.createSolutionManager(SOLVER_FACTORY_LIST);
+        var recommendationList =
+                solutionManager.recommendAssignment(solution, evaluatedValue, v -> new Pair<>(v.getEntity(), v.getIndex()));
```
zepfred, line 1124: Should this test fail as per the validation in `Assigner#apply`?

docs/src/modules/ROOT/pages/responding-to-change/responding-to-change.adoc
```diff
@@ -725,6 +725,7 @@ but do accept infeasible solutions too.
 
 
 [#recommendedFitAPI]
+[#assignmentRecommendationAPI]
```
zepfred, line 728: Let's make sure there are no broken references to this anchor.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolutionManager.java
```diff
@@ -113,12 +116,20 @@ public ScoreAnalysis<Score_> analyze(Solution_ solution, ScoreAnalysisFetchPolic
         return analysis;
     }
 
+    @Override
+    public <In_, Out_> List<RecommendedAssignment<Out_, Score_>> recommendAssignment(Solution_ solution,
```
zepfred, line 120: I believe we need to add a new overloaded method without `fetchPolicy` param as we are using such a method in the documentation.

triceo: It's a default method on `SolutionManager`.

persistence/jackson/src/test/java/ai/timefold/solver/jackson/api/TimefoldJacksonModuleTest.java
```diff
@@ -174,15 +174,15 @@ private static String getSerializedScoreAnalysis() {
     }
 
     @Test
-    void recommendationFit() throws JsonProcessingException {
+    void recommendedAssignment() throws JsonProcessingException {
```
zepfred, line 177: The same comment about continuing to test the deprecated stuff applies here.

zepfred (dismissed):
My main concern about `AssignmentProcessor` has already been addressed. Therefore, the changes look great to me. 
+1 after checking the comments related to the tests of deprecated methods.

PR #1139, opened 2024-10-07
chore: improve CI

zepfred:


.github/workflows/release.yml
```diff
@@ -118,4 +142,13 @@ jobs:
...
+        if: ${{ github.event.inputs.dryRun == 'false' }}
+
+      - name: Put back the 999-SNAPSHOT version on the release branch
+        run: |
+          git checkout -B $RELEASE_BRANCH_NAME
+          mvn -Dfull versions:set -DnewVersion=999-SNAPSHOT
+          sed -i "s/^timefold_solver_python_version.*=.*/timefold_solver_python_version = '999-dev0'/" setup.py
+          find . -name 'pom.xml' | xargs git add
+          git commit -am "build: move back to 999-SNAPSHOT"
+          git push origin $RELEASE_BRANCH_NAME
```
triceo, line 154: Does this work? You can't push directly to a .x branch, they're protected.

zepfred: At this point, we will push to the branch `__timefold_release_branch__`, but you are correct about the release pointing to this commit. Let's move it to the `finish_release` workflow.

triceo (approved):

.github/workflows/release.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           fetch-depth: 0
           ref: ${{ github.event.inputs.sourceBranch }}
 
-      - name: Init release branch
+      - name: Delete existing release branch
```
triceo, line 51:
```suggestion
      - name: Delete existing release branch (if exists)
```

PR #1140, opened 2024-10-10
chore: update the CI to compile and deploy Antora docs

zepfred:
The PR updates the secure PR pipeline and adds a new job for compiling and deploying the Antora docs.

1. The compilation step is necessary because certain errors are only detected during the release process, which could delay the solver's release.
2. The deployment step can help identify issues during the documentation development phase, such as incorrect title configuration, inconsistent menu items, etc.
3. The deployment step ensures that it is not used on the main branch and that the production remains unchanged.


.github/workflows/pull_request_secure.yml
```diff
@@ -259,6 +259,53 @@ jobs:
...
+
+      - name: Deploy Documentation (Preview Mode)
+        if: ${{ github.ref != 'refs/heads/main' }}
+        id: deploy
+        uses: cloudflare/wrangler-action@v3
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          workingDirectory: ./apps/docs
+          #          command: pages deploy ./public-serve --project-name=timefold-docs --branch=${{ github.ref }}
```
triceo, line 306: No deploy?

zepfred: I'm still testing it on my fork.

pieterdeschepper (changes requested):
The problem with this approach is that it uses another antora-playbook file, which is  duplicated from the frontend, and is prone to be wrong when we make updates in the frontend. I'd prefer if the one from frontend is used, but with cleared sources and only the solver added as source.

.github/workflows/pull_request_secure.yml
```diff
@@ -259,6 +259,61 @@ jobs:
...
+          GURL="$(git remote get-url origin)"
+          cd -
+          sed -i '/algolia/,+1d' apps/docs/antora-playbook.yml
+          sed -i '/timefold-orbit/,+3d' apps/docs/antora-playbook.yml
+          sed -i '/timefold-field/,+3d' apps/docs/antora-playbook.yml
+          sed -i '/timefold-models/,+3d' apps/docs/antora-playbook.yml
+          sed -i '/timefold-employee/,+3d' apps/docs/antora-playbook.yml
+          sed -i '/tags/,1d' apps/docs/antora-playbook.yml
+          sed -i "s#git@github.com:timefoldai/timefold-solver.git#$GURL#g" apps/docs/antora-playbook.yml
+          sed -i "s/branches:[ .-~]*/branches: $BRANCH_NAME/" apps/docs/antora-playbook.yml
```
pieterdeschepper, line 303: Instead of all these `sed` replaces, I'd prefer something like the below for a couple of reasons:
- It's independent of any changes made to the original file (right now, making changes in the sources of the original file, might break your sed-replace
- It doesn't give any insights into our private repo / setup
The following code just clears out the whole `sources` section and only puts the relevant solver source in there:

```awk -vGURL="${GURL}" -vBRANCH_NAME="${BRANCH_NAME}" 'BEGIN{ print_flag=1 }
{
    if( $0 ~ /  sources:/ )
    {
       print "  sources:"
       print "    - url: ", GURL
       print "      branches: [",BRANCH_NAME,"]"
       print "      start_path: docs/src"
       print_flag=0;
       next
    }
    if( $0 ~ /^$/ )
    {
        print_flag=1;
    }
    if ( print_flag == 1 )
        print $0

} ' antora-playbook.yml

docs/src/antora-playbook-template.yml
pieterdeschepper, line 1: I don't think we need this file anymore?

.github/workflows/pull_request_secure.yml
zepfred: I found an easier way to do this

docs/src/antora-playbook-template.yml
pieterdeschepper, line 1: Please remove this file

pieterdeschepper (changes requested):

triceo:
@pieterdeschepper Please review.

@zepfred Maybe we want to remove the `-Dfull` from the solver CI? It exists to build the docs, but now that we have a separate job for it, we can save some time.

pieterdeschepper (approved):

PR #1148, opened 2024-10-16
docs: bring back some old images

triceo:
These are not used in the docs, but Geoffrey uses them for his presentations.


ge0ffrey (approved):
conferenceSchedulingValueProposition.png/svg can be removed I think, that's really old

PR #1158, opened 2024-10-24
chore: Clean up TODOs after merging nullness markers

triceo:


greyhairredbear:
Sorry for missing those last todos, thanks for cleaning up - in case my review is sufficient here I can gladly approve

triceo:
@greyhairredbear No worries, I missed them too. :-) I'll leave the review with Fred, since your review wouldn't pass the check to allow me to merge the PR - but feel free to add your comments, if you see something fishy.

greyhairredbear:
> @greyhairredbear No worries, I missed them too. :-) I'll leave the review with Fred, since your review wouldn't pass the check to allow me to merge the PR - but feel free to add your comments, if you see something fishy.

Yeah, thought the approval would need to be by someone with the correct role for this repo. 

PR LGTM, `emptyMatchAnalysisIfNull` looks neat 👍 and I guess `buildScoreCorruptionAnalysis` would only be called in `FULL_ASSERT` mode, so potential performance impacts of allocating new empty lists are negligible, right?

triceo:
> `buildScoreCorruptionAnalysis` would only be called in `FULL_ASSERT` mode, so potential performance impacts of allocating new empty lists are negligible, right?

Correct. And besides, as far as I know, `Collections.emptyList()` doesn't actually create a new instance of anything - it returns a constant.

zepfred (approved):

PR #1160, opened 2024-10-24
feat: add nullness annotations to public API part 2

greyhairredbear:
As discussed, this is a continuation of #1009. Opening again as draft. 

#### Open items

- [ ] missing annotations in score.stream.common.ConstraintCollectors
- [x] score.stream.penta
- [x] score.stream.quad
- [ ] check impl related to score package
- [ ] check if all impl related to solver package has been annotated
- [ ] config
- [ ] timefold-solver-benchmark
- [ ] timefold-solver-test

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/score/stream/penta/PentaJoiner.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/stream/penta/PentaJoiner.java
@@ -2,6 +2,7 @@ package ai.timefold.solver.core.api.score.stream.penta;
 
 import ai.timefold.solver.core.api.score.stream.Joiners;
 import ai.timefold.solver.core.api.score.stream.quad.QuadConstraintStream;
+import org.jspecify.annotations.NonNull;
 
 /**
  * Created with {@link Joiners}.
@@ -11,6 +12,7 @@ import ai.timefold.solver.core.api.score.stream.quad.QuadConstraintStream;
  */
 public interface PentaJoiner<A, B, C, D, E> {
 
-    PentaJoiner<A, B, C, D, E> and(PentaJoiner<A, B, C, D, E> otherJoiner);
+    @NonNull
+    PentaJoiner<A, B, C, D, E> and(@NonNull PentaJoiner<A, B, C, D, E> otherJoiner);
 
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/score/stream/quad/QuadConstraintStream.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/stream/quad/QuadConstraintStream.java
@@ -36,6 +36,8 @@ import ai.timefold.solver.core.api.score.stream.tri.TriConstraintStream;
 import ai.timefold.solver.core.api.score.stream.uni.UniConstraintStream;
 import ai.timefold.solver.core.impl.util.ConstantLambdaUtils;
 
+import org.jspecify.annotations.NonNull;
+
 /**
  * A {@link ConstraintStream} that matches four facts.
  *
@@ -60,10 +62,9 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * (such as {@link Joiners#equal(TriFunction, Function)},
      * because the latter applies hashing and/or indexing, so it doesn't create every combination just to filter it out.
      *
-     * @param predicate never null
-     * @return never null
      */
-    QuadConstraintStream<A, B, C, D> filter(QuadPredicate<A, B, C, D> predicate);
+    @NonNull
+    QuadConstraintStream<A, B, C, D> filter(@NonNull QuadPredicate<A, B, C, D> predicate);
 
     // ************************************************************************
     // If (not) exists
@@ -80,13 +81,12 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * a different definition of exists applies.
      * (See {@link ConstraintFactory#from(Class)} Javadoc.)
      *
-     * @param otherClass never null
-     * @param joiner never null
      * @param <E> the type of the fifth matched fact
-     * @return never null, a stream that matches every tuple of A, B, C and D where E exists for which the
+     * @return a stream that matches every tuple of A, B, C and D where E exists for which the
      *         {@link PentaJoiner} is true
      */
-    default <E> QuadConstraintStream<A, B, C, D> ifExists(Class<E> otherClass, PentaJoiner<A, B, C, D, E> joiner) {
+    default <E> @NonNull QuadConstraintStream<A, B, C, D> ifExists(@NonNull Class<E> otherClass,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner) {
         return ifExists(otherClass, new PentaJoiner[] { joiner });
     }
 
@@ -94,15 +94,13 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * As defined by {@link #ifExists(Class, PentaJoiner)}.
      * For performance reasons, indexing joiners must be placed before filtering joiners.
      *
-     * @param otherClass never null
-     * @param joiner1 never null
-     * @param joiner2 never null
      * @param <E> the type of the fifth matched fact
-     * @return never null, a stream that matches every tuple of A, B, C and D where E exists for which the
+     * @return a stream that matches every tuple of A, B, C and D where E exists for which the
      *         {@link PentaJoiner}s are true
      */
-    default <E> QuadConstraintStream<A, B, C, D> ifExists(Class<E> otherClass, PentaJoiner<A, B, C, D, E> joiner1,
-            PentaJoiner<A, B, C, D, E> joiner2) {
+    default <E> @NonNull QuadConstraintStream<A, B, C, D> ifExists(@NonNull Class<E> otherClass,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner1,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner2) {
         return ifExists(otherClass, new PentaJoiner[] { joiner1, joiner2 });
     }
 
@@ -110,16 +108,13 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * As defined by {@link #ifExists(Class, PentaJoiner)}.
      * For performance reasons, indexing joiners must be placed before filtering joiners.
      *
-     * @param otherClass never null
-     * @param joiner1 never null
-     * @param joiner2 never null
-     * @param joiner3 never null
      * @param <E> the type of the fifth matched fact
-     * @return never null, a stream that matches every tuple of A, B, C and D where E exists for which the
+     * @return a stream that matches every tuple of A, B, C and D where E exists for which the
      *         {@link PentaJoiner}s are true
      */
-    default <E> QuadConstraintStream<A, B, C, D> ifExists(Class<E> otherClass, PentaJoiner<A, B, C, D, E> joiner1,
-            PentaJoiner<A, B, C, D, E> joiner2, PentaJoiner<A, B, C, D, E> joiner3) {
+    default <E> @NonNull QuadConstraintStream<A, B, C, D> ifExists(@NonNull Class<E> otherClass,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner1, @NonNull PentaJoiner<A, B, C, D, E> joiner2,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner3) {
         return ifExists(otherClass, new PentaJoiner[] { joiner1, joiner2, joiner3 });
     }
 
@@ -127,18 +122,13 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * As defined by {@link #ifExists(Class, PentaJoiner)}.
      * For performance reasons, indexing joiners must be placed before filtering joiners.
      *
-     * @param otherClass never null
-     * @param joiner1 never null
-     * @param joiner2 never null
-     * @param joiner3 never null
-     * @param joiner4 never null
      * @param <E> the type of the fifth matched fact
-     * @return never null, a stream that matches every tuple of A, B, C and D where E exists for which the
+     * @return a stream that matches every tuple of A, B, C and D where E exists for which the
      *         {@link PentaJoiner}s are true
      */
-    default <E> QuadConstraintStream<A, B, C, D> ifExists(Class<E> otherClass, PentaJoiner<A, B, C, D, E> joiner1,
-            PentaJoiner<A, B, C, D, E> joiner2, PentaJoiner<A, B, C, D, E> joiner3,
-            PentaJoiner<A, B, C, D, E> joiner4) {
+    default <E> @NonNull QuadConstraintStream<A, B, C, D> ifExists(@NonNull Class<E> otherClass,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner1, @NonNull PentaJoiner<A, B, C, D, E> joiner2,
+            @NonNull PentaJoiner<A, B, C, D, E> joiner3, @NonNull PentaJoiner<A, B, C, D, E> joiner4) {
         return ifExists(otherClass, new PentaJoiner[] { joiner1, joiner2, joiner3, joiner4 });
     }
 
@@ -150,13 +140,12 @@ public interface QuadConstraintStream<A, B, C, D> extends ConstraintStream {
      * but we can't fix it with a {@link SafeVarargs} annotation because it's an interface method.
      * Therefore, there are overloaded methods with up to 4 {@link PentaJoiner} parameters.
      *
-     * @param otherClass never null
-     * @param joiners never null
      * @param <E> the type of the fifth matched fact
-     * @return never null, a stream that matches every tuple of A, B, C and D where E exists for which the
+     * @return a stream that matches every tuple of A, B, C and D where E exists for which the
      *         {@link PentaJoiner}s are true
      */
-    <E> QuadConstraintStream<A, B, C, D> ifExists(Class<E> otherClass, PentaJoiner<A, B, C, D, E>... joiners);
+    <E> @NonNull QuadConstraintStream<A, B, C, D> ifExists(@NonNull Class<E> otherClass,
+            @NonNull PentaJoiner<A, B, C, D, E>... joiners);
 
     /**
      * Create a new {@link QuadConstraintStream} for every tuple of A, B, C and D where E exists for which the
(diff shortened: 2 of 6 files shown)


core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
```diff
@@ -219,226 +219,233 @@ public final class ConstraintCollectors {
...
+    public static <A> @NonNull UniConstraintCollector<A, ?, Long>
+            sumLong(@NonNull ToLongFunction<? super A> groupValueMapping) {
         return InnerUniConstraintCollectors.sum(groupValueMapping);
     }
 
     /**
      * As defined by {@link #sum(ToIntFunction)}.
      */
-    public static <A, Result> UniConstraintCollector<A, ?, Result> sum(Function<? super A, Result> groupValueMapping,
+    // TODO: nullability of further params (suppose all are @NonNull)
```
greyhairredbear, line 238: @triceo should `zero`, `adder` and `subtractor` all be `@NonNull` here?

greyhairredbear:
@triceo I have a general question here on `ConstraintCollectors`: There are many functions without any information on nullability in the Javadoc, but I guess for most of the parameters (`mapper`s and `valueSetFunction`s and similar likes), `@NonNull` should be appropriate, correct? Are there any instances in this file where a parameter or a return value might be `@Nullable`? I coudn't find any quickly searching/browsing through this file, but I want to make sure I'm correct before over-promising

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintCollectors.java
triceo: Yes. There are overloads for that method which do not require these arguments - so when an argument is actually present, it is required unless Javadoc says otherwise.

triceo:
None of these methods will ever return `null` - that is for sure.
When it comes to argument, I don't expect them to be nullable. If any of them are, I'd expect the Javadoc to say so.

greyhairredbear:
Small sidenote since I stumbled upon it: `toConnectedRanges`/`tonConnectedTemporalRanges` has some overloads that IntelliJ reports as unused, so I guess there are no tests for those either. Don't know how much value there would be of testing those specific cases, just wanted to flag this for the small chance it has gone unnoticed.

greyhairredbear:
@triceo As my next step is the `config` package: Lots of the members of a `config` are initialized with `null` and there is little information in the Javadoc. I guess annotating getters and parameters for setters `with...`-methods as `@Nullable`, do you agree?

greyhairredbear:
Hi @triceo, just wanted to keep you updated on my progress here, since I didn't manage to complete the work here by tonight. 

Currently, I'm still working on the `config` package, but I have some free resources tomorrow and Tuesday night, so I suppose I should be through with the changes by then. There are still some todos in the PR, so I'll add comments on those so we can track them easier. 

I hope this delay in schedule doesn't lead to too many inconveniences for you, I hope the time buffer until the release date on Nov 10th (iirc) is sufficient.

triceo:
Hey @greyhairredbear - thanks for letting me know! As far as we can still merge it this week, we should be good.

greyhairredbear:
@triceo I guess I'm through with the last commit, I'll comment and tag you on the open todos/questions I have tomorrow (probably before I start my workday). After completing that I suppose this should be good to merge

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/common/nearby/NearbySelectionConfig.java
```diff
@@ -236,7 +246,8 @@ public NearbySelectionConfig withBetaDistributionBeta(Double betaDistributionBet
     // Builder methods
     // ************************************************************************
 
-    public void validateNearby(SelectionCacheType resolvedCacheType, SelectionOrder resolvedSelectionOrder) {
+    // TODO clarify if @NonNull is ok for both params
+    public void validateNearby(@NonNull SelectionCacheType resolvedCacheType, @NonNull SelectionOrder resolvedSelectionOrder) {
```
greyhairredbear, line 250: @triceo

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
```diff
@@ -325,7 +329,8 @@ public String toString() {
         return getClass().getSimpleName() + "(" + entityClass + ")";
     }
 
-    public static <Solution_> boolean hasSorter(EntitySorterManner entitySorterManner,
+    // TODO: entityDescriptor @NonNull?
+    public static <Solution_> boolean hasSorter(@NonNull EntitySorterManner entitySorterManner,
             EntityDescriptor<Solution_> entityDescriptor) {
```
greyhairredbear, line 334: @triceo

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -234,28 +237,26 @@ public Config_ withFixedProbabilityWeight(Double fixedProbabilityWeight) {
      * Gather a list of all descendant {@link MoveSelectorConfig}s
      * except for {@link UnionMoveSelectorConfig} and {@link CartesianProductMoveSelectorConfig}.
      *
-     * @param leafMoveSelectorConfigList not null
      */
-    public void extractLeafMoveSelectorConfigsIntoList(List<MoveSelectorConfig> leafMoveSelectorConfigList) {
+    // TODO: list elements @NonNull?
+    public void extractLeafMoveSelectorConfigsIntoList(@NonNull List<@NonNull MoveSelectorConfig> leafMoveSelectorConfigList) {
```
greyhairredbear, line 242: @triceo There are multiple `List<T>` member in the `*Config`-classes. My current assumption was that `null` does not make a whole lot of sense here, so I annotated the type param as `@NonNull`. Is this assumption valid for you?

core/src/main/java/ai/timefold/solver/core/config/solver/monitoring/SolverMetric.java
```diff
@@ -90,12 +92,15 @@ public enum SolverMetric {
...
         return meterId;
     }
 
-    public static void registerScoreMetrics(SolverMetric metric, Tags tags, ScoreDefinition<?> scoreDefinition,
-            Map<Tags, List<AtomicReference<Number>>> tagToScoreLevels, Score<?> score) {
+    public static void registerScoreMetrics(@NonNull SolverMetric metric, @NonNull Tags tags,
+            // TODO: @NonNull Map<@NonNull Tags, @NonNull List<@NonNull AtomicReference<@NonNull Number>>> tagToScoreLevels?
+            @NonNull ScoreDefinition<?> scoreDefinition,
+            @NonNull Map<@NonNull Tags, List<AtomicReference<Number>>> tagToScoreLevels,
+            @NonNull Score<?> score) {
```
greyhairredbear, line 103: @triceo does it make sense to have `@NonNull Map<@NonNull Tags, @NonNull List<@NonNull AtomicReference<@NonNull Number>>>` as type for `tagToScoreLevels`?

benchmark/src/main/java/ai/timefold/solver/benchmark/config/PlannerBenchmarkConfig.java
```diff
@@ -543,168 +509,172 @@ public void setThreadFactoryClass(Class<? extends ThreadFactory> threadFactoryCl
...
         return warmUpHoursSpentLimit;
     }
 
-    public void setWarmUpHoursSpentLimit(Long warmUpHoursSpentLimit) {
+    public void setWarmUpHoursSpentLimit(@NonNull Long warmUpHoursSpentLimit) {
         this.warmUpHoursSpentLimit = warmUpHoursSpentLimit;
     }
 
-    public Long getWarmUpDaysSpentLimit() {
+    public @NonNull Long getWarmUpDaysSpentLimit() {
```
triceo, line 552: I'm afraid this pattern is not correct. :-(

- Getters in configs have to be `Nullable`, because the fields may hold `null`. `null` here means that the user chose not to specify. (Why else would we choose `Long` instead of `long`, avoiding reference types altogether?)
- Setters likewise have to be able to accept `null`, because the solver code needs to be able to express that something should explicitly be set to its default value.
- It is only the withers which should not accept null and never return it, because we understand withers to mean that the user is making an explicit choice, and therefore `null` no longer applies. 

This is an unfortunate miscommunication. We have discussed this topic in comments above - even after re-reading my answer, I still think I said the same thing as I am saying here.

Sadly, I can not accept this and the annotations in configs need to be reviewed with the above in mind. :-( Interestingly, it seems like not all configs use this wrong pattern - some configs properly assign nullability as we agreed.

Maybe this file was annotated before our agreement?

core/src/main/java/ai/timefold/solver/core/config/solver/monitoring/SolverMetric.java
triceo: Not our best API design here. :-(

You know Kotlin better than I do. Would these annotations on this crazy type actually help? I feel like they would, but I will defer to your judgment here.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/common/nearby/NearbySelectionConfig.java
triceo: Non-null OK for both.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
triceo: Non-null OK for both.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
triceo: Yes, it makes sense. I would not expect null inside of this collection.

triceo (changes requested):
I have hopefully answered all the comments.
Unfortunately, in at least one of the configs, the nullability doesn't follow the agreed pattern, and therefore I must request changes.

benchmark/src/main/java/ai/timefold/solver/benchmark/config/PlannerBenchmarkConfig.java
greyhairredbear: This is a mistake on my part. Seems like some error using multi-cursor and then pasting the wrong value. I'll carefully review getters and setters myself to make sure this didn't happen more often. Thanks for catching that. 

In general, I'm pretty sure I adhered to the pattern as discussed (getters/setters nullable, withers taking nonnull params)

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -295,7 +304,8 @@ public static int ceilDivide(int dividend, int divisor) {
         return (dividend / divisor) + correction;
     }
 
-    public static int resolvePoolSize(String propertyName, String value, String... magicValues) {
+    // TODO what to make of magicValues?
+    public static int resolvePoolSize(String propertyName, @NonNull String value, String... magicValues) {
```
greyhairredbear, line 308: @triceo `magicValues` only used for generating the message during exception handling in here - `Arrays.toString` does allow passing null, so `@Nullable` might be fine as an annotation for `magicValues`. Does that sound correct?

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -58,7 +61,8 @@ public class ConfigUtils {
      * @param <T> the new instance type
      * @return new instance of clazz
      */
-    public static <T> T newInstance(Object configBean, String propertyName, Class<T> clazz) {
+    // TODO: propertyName, clazz
+    public static <T> @NonNull T newInstance(@Nullable Object configBean, String propertyName, Class<T> clazz) {
```
greyhairredbear, line 65: @triceo are `propertyName` and `clazz` non-null?

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -87,8 +91,9 @@ public static <T> T newInstance(Supplier<String> ownerDescriptor, String propert
         }
     }
 
-    public static void applyCustomProperties(Object bean, String beanClassPropertyName,
-            Map<String, String> customProperties, String customPropertiesPropertyName) {
+    // TODO: param customPropertiesPropertyName
+    public static void applyCustomProperties(@NonNull Object bean, @NonNull String beanClassPropertyName,
+            @Nullable Map<@NonNull String, @NonNull String> customProperties, String customPropertiesPropertyName) {
```
greyhairredbear, line 96: @triceo `customPropertiesPropertyName` is also only used for exception messages, is this `@Nullable`?

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -384,6 +392,7 @@ public static List<Member> getAllMembers(Class<?> baseClass, Class<? extends Ann
         return memberStream.distinct().sorted(alphabeticMemberComparator).collect(Collectors.toList());
     }
 
+    // TODO
     @SafeVarargs
     public static Class<? extends Annotation> extractAnnotationClass(Member member,
             Class<? extends Annotation>... annotationClasses) {
```
greyhairredbear, line 398: @triceo From the function code, it seems to me the parameters here should all be `@NonNull`, is that right?

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -415,6 +425,7 @@ Maybe the member (%s) should return a parameterized %s."""
                                 memberName, type, memberName, type.getSimpleName())));
     }
 
+    // TODO
     public static Optional<Class<?>> extractGenericTypeParameter(String parentClassConcept, Class<?> parentClass, Class<?> type,
             Type genericType, Class<? extends Annotation> annotationClass, String memberName) {
```
greyhairredbear, line 430: @triceo I suppose the parameters all should be `@NonNull`, right?

greyhairredbear:
@triceo I fixed the wrongly annotated getters/setters in `PlannerBenchmarkConfig` and reviewed the other `Config` files, didn't find any other mistakes like this. I've also commented on the last remaining todos for this PR.

triceo (commented):
Answering more of the comments.

triceo:
The Sonar warnings, some of them important, are not your concern - don't worry about them. Something for me to look into after.

core/src/main/java/ai/timefold/solver/core/config/solver/monitoring/SolverMetric.java
greyhairredbear: I think it makes sense to use the annotation too - when unwrapping the mapped value, things might get messy: 

```
scoreLevels?.forEach { it?.get()?.times(1) }
```

Especially when one for whatever reason would have to handle the null-case here. That being said, the safe-calls wouldn't be mandatory here, since all the types here would be platform types. Still, it's better to be certain as a user, and with the correct annotations its explicitly stated in the code.

greyhairredbear: And I fear I've seen far worse API designs 😅

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
greyhairredbear: Just wanted to change this accordingly and found there is a null-check for `annotationClass`, so I'll make that param nullable

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -527,9 +518,9 @@ public SolverConfig withClassLoader(ClassLoader classLoader) {
      * As defined by {@link ScoreDirectorFactoryConfig#withEasyScoreCalculatorClass(Class)}, but returns this.
      *
      * @param easyScoreCalculatorClass sometimes null
-     * @return this, never null
      */
-    public SolverConfig withEasyScoreCalculatorClass(Class<? extends EasyScoreCalculator> easyScoreCalculatorClass) {
+    // TODO docu says param of a with-method may be null
+    public @NonNull SolverConfig withEasyScoreCalculatorClass(Class<? extends EasyScoreCalculator> easyScoreCalculatorClass) {
```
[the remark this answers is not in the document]
greyhairredbear: @triceo I guess this is the last open question. Should I just make these parameters nullable according to the existing documentation?

core/src/main/java/ai/timefold/solver/core/config/solver/monitoring/SolverMetric.java
```diff
@@ -90,10 +93,11 @@ public enum SolverMetric {
...
     }
 
-    public String getMeterId() {
+    public @NonNull String getMeterId() {
         return meterId;
     }
 
+    @NullMarked
     public static void registerScoreMetrics(SolverMetric metric, Tags tags, ScoreDefinition<?> scoreDefinition,
             Map<Tags, List<AtomicReference<Number>>> tagToScoreLevels, Score<?> score) {
```
greyhairredbear, line 102: @triceo schemagen failed when annotating the `List<AtomicReference<Number>>` param, so I used `@NullMarked` here. 

However, this introduced a new IDE warning (not sure if this is a compiler warning too) on line `116`, since `Metrics.gauge()` is annotated to return a nullable type. From reading the code of `gauge`, I understand this shouldn't be the case when passing a non-null reference in as the third param (which `new AtomicReference` certainly is). 

So I guess we could: 
1. leave it as it is
2. wrap `gauge` with `Objects.requireNonNull` 
3. wrap `gauge` with `Objects.requireNonNullOrElse` (but I guess this might be a bit confusing, since the else-branch should never be reached)

Any thoughts?

triceo: I'd leave it for now. 

I'll have some downtime before the release, I'll go back to my notes and review some of the places where maybe I want to think about it a bit more. This will be one of them.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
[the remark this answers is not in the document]
greyhairredbear: > The Github UX for this needs to improve.
That's certainly true, especially on the conversation view. In the "files changed" view, there is this "Conversations" dropdown which I used to keep track of unresolved comments, but clicking on the unresolved conversation and then being taken to the right comment seemed very unreliable for me on large diffs like these. 

I'll push the changes in a couple of minutes

triceo (approved):
Thank you, @greyhairredbear, for this work!
The solver and its users will be all the better for it.

I'll merge when the CI machinery approves. (Sonar won't, but I'll take care of that later.)

PR #1169, opened 2024-10-28
feat: change move support in Move Streams

triceo:
Doesn't do anything yet. WIP.


triceo:
Outdated, replaced by https://github.com/TimefoldAI/timefold-solver/pull/1349

PR #1185, opened 2024-11-07
feat: enable score analysis without justifications

triceo:
Introduces an option to disable constraint match justifications,
while still keeping constraint matching enabled.
In some cases, the dataset is so large that justifications are pointless, 
and impractical to send over the wire as well.
In this case, score analysis only provides constraint match count, and not the full match analysis.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
@@ -95,7 +95,7 @@ public abstract class AbstractScoreDirectorFactory<Solution_, Score_ extends Sco
 
     @Override
     public InnerScoreDirector<Solution_, Score_> buildScoreDirector() {
-        return buildScoreDirector(true, ConstraintMatchPolicy.ENABLED);
+        return buildScoreDirector(false, ConstraintMatchPolicy.DISABLED);
     }
 
     @Override
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
@@ -94,11 +94,6 @@ public abstract class AbstractScoreDirectorFactory<Solution_, Score_ extends Sco
     // ************************************************************************
 
     @Override
-    public InnerScoreDirector<Solution_, Score_> buildScoreDirector() {
-        return buildScoreDirector(false, ConstraintMatchPolicy.DISABLED);
-    }
-
-    @Override
     public void assertScoreFromScratch(Solution_ solution) {
         // Get the score before uncorruptedScoreDirector.calculateScore() modifies it
         Score_ score = getSolutionDescriptor().getScore(solution);
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
@@ -6,6 +6,7 @@ import ai.timefold.solver.core.api.score.director.ScoreDirector;
 import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.BasicVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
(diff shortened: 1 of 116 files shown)


core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
```diff
@@ -94,14 +95,14 @@ public void setTrackingWorkingSolution(boolean trackingWorkingSolution) {
 
     @Override
     public InnerScoreDirector<Solution_, Score_> buildScoreDirector() {
-        return buildScoreDirector(true, true);
+        return buildScoreDirector(false, ConstraintMatchPolicy.DISABLED);
```
zepfred, line 98: Should we set `ENABLED_WITHOUT_JUSTIFICATIONS` or `ENABLED` instead?

triceo: The method is only used in tests. I will remove it entirely.

zepfred: I understand the reason now after reading the comments in the `InnerScoreDirectorFactory`

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirectorFactory.java
```diff
@@ -29,56 +30,53 @@ public interface InnerScoreDirectorFactory<Solution_, Score_ extends Score<Score
     InnerScoreDirector<Solution_, Score_> buildScoreDirector();
 
     /**
-     * Like {@link #buildScoreDirector()}, but optionally disables {@link ConstraintMatch} tracking and look up
-     * for more performance (presuming the {@link ScoreDirector} implementation actually supports it to begin with).
+     * Like {@link #buildScoreDirector()}, but optionally enables {@link ConstraintMatch} tracking and look up
```
zepfred, line 33:
```suggestion
     * Like {@link #buildScoreDirector()}, but disables {@link ConstraintMatch} tracking and look up
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirectorFactory.java
```diff
@@ -29,56 +30,53 @@ public interface InnerScoreDirectorFactory<Solution_, Score_ extends Score<Score
     InnerScoreDirector<Solution_, Score_> buildScoreDirector();
```
zepfred, line 30: Is there documentation for the default value of `ConstraintMatchPolicy` used in `buildScoreDirector()`?

core/src/main/java/ai/timefold/solver/core/impl/score/director/easy/EasyScoreDirector.java
```diff
@@ -28,9 +29,9 @@ public final class EasyScoreDirector<Solution_, Score_ extends Score<Score_>>
     private final EasyScoreCalculator<Solution_, Score_> easyScoreCalculator;
 
     public EasyScoreDirector(EasyScoreDirectorFactory<Solution_, Score_> scoreDirectorFactory,
-            boolean lookUpEnabled, boolean constraintMatchEnabledPreference, boolean expectShadowVariablesInCorrectState,
+            boolean lookUpEnabled, boolean expectShadowVariablesInCorrectState,
```
zepfred, line 32: It seems that changing the `ConstraintMatchPolicy` is no longer possible. Should it be allowed?

zepfred (commented):
The changes look good to me. Will we need to provide the Python counterpart?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirectorFactory.java
zepfred: Sounds good

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirectorFactory.java
triceo: The method got removed.

core/src/main/java/ai/timefold/solver/core/impl/score/director/easy/EasyScoreDirector.java
triceo: Easy has no support for constraint matches, so there is really no point for setting that policy from the outside.

triceo:
Wrt. Python - I checked and I don't think so.

zepfred (approved):

PR #1191, opened 2024-11-08
fix: Add PlanningPinToIndex to Python

### LINKED ISSUE #1190: PlanningPinToIndex Python
**Describe the bug**
I wanted to do some experiments with PlanningPinToIndex, but I get the following error (Python):
```
NameError: name 'PlanningPinToIndex' is not defined
```
I might be very wrong, but I think it should be defined [here](https://github.com/TimefoldAI/timefold-solver/blob/20ca32204aad5339cce6ff38d83f69d9af7368f7/python/python-core/src/main/python/domain/_annotations.py), but it seems to be missing.

**To Reproduce**
```
from typing import Annotated
from timefold.solver.domain import *
from pydantic import BaseModel, Field

@planning_entity
class TF_Expert(BaseModel):
    id: Annotated[str, PlanningId]
    first_unpinned_index: Annotated[int, PlanningPinToIndex] = 0
    assignments: Annotated[list[str],
                      PlanningListVariable(),
                      Field(default_factory=list)]
```

## Environment

1.15.0b0 (Python 3.11.10)

openjdk 17.0.12 2024-07-16
OpenJDK Runtime Environment (build 17.0.12+7-Ubuntu-1ubuntu222.04)
OpenJDK 64-Bit Server VM (build 17.0.12+7-Ubuntu-1ubuntu222.04, mixed mode, sharing)

Linux host 6.8.0-48-generic #48~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct  7 11:24:13 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Christopher-Chianelli:
JPyInterpreter

- Allow annotations to specify a type to override the generated getter return type. When a type is specified, it goes through the generic PythonLikeObject to Java Object convertor.

Python

- Add PlanningPinToIndex
- Make PlanningPin use the new JPyInterpreter mechanism to remove code that converts a PlanningPin to Pinning filter

Build

- Fix Python 3.12 environment name in `tox.ini`


Christopher-Chianelli:
Fixes #1190

triceo (changes requested):
The field should not support `null` values - there is no reason for why the field should have a reference type. If the Javadoc says it, let's fix the Javadoc.

triceo:
> * `@PlanningPin` allows null/using Boolean (i.e. not allowing Integer will be inconsistent)

You could also say that we shouldn't repeat mistakes of the past. 

> * There is no performance benefit from not allowing `Integer`, since the field is automatically boxed by `MemberAccessor`

I agree. Performance is irrelevant here. I simply do not see the point of allowing this field to have a third state (`null`).

> * Users might have restrictions that prevent them from using primitive types.

This particular type has not supported `null` since the start, and there were zero complaints. That tells me `null` is not necessary there. (Besides the fact that there is no practical use for putting `null` here - zero is the perfect default for this use case.)

triceo (approved):

PR #1196, opened 2024-11-11
feat: add new termination config based on the best solution improvement

zepfred:
This pull request introduces a new termination condition that activates when the solver cannot improve the best solution within a specified time frame. This time frame is determined by the speed of move evaluations, the last time the solver successfully improved the solution, and the duration without any improvements.

image
image

### DIFF
--- a/.github/workflows/pull_request_secure.yml
+++ b/.github/workflows/pull_request_secure.yml
@@ -264,7 +264,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: approval_required
     name: Build Documentation
-    env:
+    environment:
+      name: "documentation (preview)"
+      url: ${{ steps.deploy.outputs.deployment-url }}
       BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
     steps:
       - name: Checkout frontend


benchmark/src/main/resources/benchmark.xsd
```diff
@@ -587,6 +587,9 @@
           <xs:element minOccurs="0" name="moveCountLimit" type="xs:long"/>
                               
           
+          <xs:element minOccurs="0" name="unimprovedBestSolutionLimit" type="xs:double"/>
```
triceo, line 590: We need a better name for this one - one that actually indicated what is going to happen.

- `automaticExhaustionDetection` (we already use "exhaust" for exhaustive search, so maybe not)
- `automaticDepletionDetection`
- ... ?

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedBestSolutionTermination.java
```diff
@@ -0,0 +1,178 @@
...
+import ai.timefold.solver.core.api.score.Score;
+import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
+import ai.timefold.solver.core.impl.phase.scope.AbstractStepScope;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+import ai.timefold.solver.core.impl.solver.thread.ChildThreadType;
+
+public final class UnimprovedBestSolutionTermination<Solution_> extends AbstractTermination<Solution_> {
+
+    // Minimal interval of time to avoid early conclusions
+    protected static final long MINIMAL_INTERVAL_TIME = 10L;
```
triceo, line 12: Needs a unit in the name.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedBestSolutionTermination.java
```diff
@@ -0,0 +1,178 @@
...
+    // The ratio specifies the minimum criteria to determine a flat line between two move count values.
+    // A value of 0.2 represents 20% of the execution time of the current growth curve.
+    // For example, the first best solution is found at 0 seconds,
+    // while the last best solution is found at 60 seconds.
+    // Given the total time of 60 seconds,
+    // we will identify a flat line between the last best solution and the discovered new best solution
+    // if the time difference exceeds 12 seconds.
+    private final double maxUnimprovedBestSolutionLimit;
```
triceo, line 20: Needs a better name, describing the functionality.
This also needs to show up in the docs and in the images, as it is one of the key parameters to understand the algorithm.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedBestSolutionTermination.java
```diff
@@ -0,0 +1,178 @@
...
+    // Similar to unimprovedBestSolutionLimit,
+    // this criterion is specifically used to identify flat lines among multiple curves before the termination.
+    // The goal is to adjust the stop criterion based on the latest curve found when there are several.
+    private final double minUnimprovedBestSolutionLimit;
```
triceo, line 24: Based on the current description, I do not understand what this does.
Also, same comment as above.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedBestSolutionTermination.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // ************************************************************************
+    // Terminated methods
+    // ************************************************************************
+
+    @Override
+    public boolean isSolverTerminated(SolverScope<Solution_> solverScope) {
+        throw new UnsupportedOperationException(
+                "%s can only be used for phase termination.".formatted(getClass().getSimpleName()));
```
triceo, line 89: I understand why this is, but as a future improvement, we should think about ways of making this termination the default without the user having to configure anything.
I want this to eventually be the only termination the user will ever need.

This will require the termination to _somehow_ survive the CH phase.

docs/src/modules/ROOT/images/optimization-algorithms/overview/terminationMultipleGrowthCurve.png
triceo, line 1: This is not a typical curve for medium-soft scores.
When a medium score improves, typically the best score becomes _worse_ than it was before.

Also, we should clearly explain that this image applies to situations where the score has multiple levels.

docs/src/modules/ROOT/images/optimization-algorithms/overview/terminationGrowthCurve.png
triceo, line 1: The visual of the chart does not fit within the rest of our docs. It looks entirely different, looks out of place. Not sure how to solve this one, though.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -565,6 +565,58 @@ Terminates as soon as a feasible solution has been discovered.
 
 This `Termination` is usually combined with other terminations.
 
+[#unimprovedBestSolutionTermination]
+==== `UnimprovedBestSolutionTermination`
+
+`UnimprovedBestSolutionTermination` terminates when the solver cannot improve the best solution in an estimated time interval.
+The estimation of the termination interval is based on the last time
+the solver improved the solution and how much time has passed without any improvements.
```
triceo, line 573: Are we using time? I thought we were using move counts.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -565,6 +565,58 @@ Terminates as soon as a feasible solution has been discovered.
...
+Let's consider a solution process based on the following image.
+
+image::optimization-algorithms/overview/terminationGrowthCurve.png[align="center"]
+
+In the previous image, the estimated unimproved time is represented by the flat red line.
+With a configuration for `unimprovedBestSolutionLimit` set to `0.5`,
+and considering that the last best score was achieved at `10 seconds`,
+the solver will terminate the process if no improvement is found within a maximum of 10 seconds.
+This `10-seconds` limit is calculated
+by multiplying the estimated total time (`20 seconds`) by the `unimprovedBestSolutionLimit` (`0.5`).
```
triceo, line 591: Isn't this a bit confusing?
Why not change the definition and instead of basing it on estimated total time, it would be based on the time it spent improving?

The limit would be `1.0` in that case, and the docs would say eg. "the last best score was achieved at 10 seconds, and the solver will terminate in another 10 seconds."

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -565,6 +565,58 @@ Terminates as soon as a feasible solution has been discovered.
...
+
+
+[NOTE]
+====
+This `Termination` configuration includes some built-in properties:
+
+* The `Termination` process requires the growth curve to last at least `10 seconds`,
+meaning that the solver will not finish until approximately `10 seconds` after the growth curve evaluation begins.
```
triceo, line 599: We should discuss this interval, and give it a name.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -565,6 +565,58 @@ Terminates as soon as a feasible solution has been discovered.
...
+====
+
+Now, let's discuss the problem-solving process involving multiple growth curves:
+
+image::optimization-algorithms/overview/terminationMultipleGrowthCurve.png[align="center"]
+
+The termination process will track the solver's progress to identify various growth curves.
+Each time a new curve is identified, the newly identified curve is used to trigger the termination process.
+This means
+that the previously explained interval calculation will only consider the last curve rather than the entire flow.
```
triceo, line 613: We keep talking about curves here, but maybe we should talk about the last point of improvement. (Which is also a concept which needs a good name, and needs to be shown in the graph.)

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -565,6 +565,58 @@ Terminates as soon as a feasible solution has been discovered.
...
+
+The previous chart displays a red line that is smaller than the size needed to trigger termination,
+but it is still significant enough to indicate the beginning of a new curve.
+The termination threshold is calculated as `80%` of the size of the flat line that would end the solving process.
+
+This `Termination` is usually combined with other terminations.
```
triceo, line 619: Please explain why, and with which.

triceo:
I left first initial comments. I think we need:

- Better naming overall.
- For all events on "the curve" to have a name.
- For all the possible scenarios to have an explanatory picture.

In the current state, I worry people wouldn't understand. (If I didn't already understand, I think the current explanation would leave me confused.)

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
zepfred: We estimate time using the move evaluation speed and move count. However, I am considering using time instead. This would solve a corner case I couldn't address while maintaining similar behavior to the current implementation.

docs/src/modules/ROOT/images/optimization-algorithms/overview/terminationMultipleGrowthCurve.png
zepfred: You're thinking about score levels while I am considering the solution score. If the medium score improves, the solution score improves wherever the soft levels are. The current implementation of the solver utilizes the lexicographic dominance relation.

https://en.wikipedia.org/wiki/Lexicographic_optimization

docs/src/modules/ROOT/images/optimization-algorithms/overview/terminationGrowthCurve.png
zepfred: That makes sense. I'm trying to determine which tool we used previously.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
zepfred: I observed this by running some experiments. However, I didn't run a proper statistical analysis. I have decided to set this as a termination setting to enable its configuration.

PR #1201, opened 2024-11-12
chore: update antora file

zepfred:


docs/src/antora.yml
```diff
@@ -1,19 +1,19 @@
 # This file is temporary, since for the time being, we need to build docs from the main branch.
 # Remove when 1.16.0 is out, otherwise this file quickly becomes stale.
```
triceo, line 2:
```suggestion
# Remove when 1.17.0 is out, otherwise this file quickly becomes stale.
```

triceo (approved):
LGTM, with the change to the comment.

PR #1226, opened 2024-11-26
chore: use MavenCentral in JReleaser

zepfred:
This PR updates the configuration of JReleaser to deploy the solver using Maven Central.

1. Uses two new env var for authentication: `JRELEASER_MAVENCENTRAL_USERNAME` and `JRELEASER_MAVENCENTRAL_PASSWORD`.
2. Updates the config according to the [documentation](https://jreleaser.org/guide/latest/reference/deploy/maven/maven-central.html).


triceo (approved):
Let's see what happens when we actually do a release.

PR #1228, opened 2024-11-27
chore: unify the handling of list variable shadows

triceo:
All shadow variables related to a list variable are now handled in one unified loop, as opposed to each having its own variable listener. 
This should technically be a performance improvement, but in fact a barely significant regression of 3 % was measured in some cases. (Other cases show minor improvements.)

This change also makes internalized previous/next shadows possible, and that in turn will help with the revamp of shadow variables that is currently in development.
Therefore we introduce this change even though it may bring a small performance regression.


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupply.java
```diff
@@ -1,18 +1,91 @@
...
+     * @return null if the element is the first element in the list
+     */
+    Object getPreviousElement(Object element);
+
+    /**
+     *
+     * @param element never null
+     * @return null if the element is the last element in the list
+     */
+    Object getNextElement(Object element);
```
triceo, line 89: These methods can now be used to access prev/next element, without having to mess with variable descriptors and externalization.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupply.java
```diff
@@ -1,18 +1,91 @@
...
+ * @see ExternalizedListVariableStateSupply The external representation of these shadow variables,
+ *      which doesn't care whether the variable is internal or externalized.
+ */
 public interface ListVariableStateSupply<Solution_> extends
         SourcedVariableListener<Solution_>,
         ListVariableListener<Solution_, Object, Object>,
         SingletonInverseVariableSupply,
-        IndexVariableSupply,
-        ListVariableElementStateSupply<Solution_> {
+        IndexVariableSupply {
```
triceo, line 42: This class handles index/inverse as well.
One-stop shop for all your list variable needs.

triceo:
@Christopher-Chianelli Since you're going to be using this later, would you mind reviewing?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -0,0 +1,291 @@
...
+            this.indexChanged = indexChanged;
+        }
+
+    }
+
+    public ElementLocation getLocationInList(Object planningValue) {
+        if (!canUseExternalizedLocation) {
+            return Objects.requireNonNullElse(elementLocationMap.get(planningValue), ElementLocation.unassigned());
+        } else {
+            var inverse = getInverseSingleton(planningValue);
```
Christopher-Chianelli, line 226: `getInverseSingleton` and `getIndex` might do a map lookup on the same value. Might be worth it to do a single map lookup here if requiresLocationMap is true.

Christopher-Chianelli (approved):
LGTM; possible performance improvements:

- Merge NextElementVariableProcessor and PrevElementVariableProcessor into a single, final class, and add a private final int field that +1 for NextElementProcessor, and -1 for PrevElementProcessor.
- Do not call `getInverseSingleton` and `getIndex` in `getLocationInList`, since they both do a map lookup on the same element.

PR #1232, opened 2024-11-27
fix: Documentation CI - attempt 4

zepfred:


.github/workflows/pull_request_secure.yml
```diff
@@ -287,13 +281,60 @@ jobs:
...
+      - name: Install yq
+        run: |
+          sudo wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq 
+          sudo chmod +x /usr/bin/yq          
+
+      - name: Configuring Antora version
+        working-directory: "./timefold-solver"
+        run: |
+          LATEST_VERSION="SNAPSHOT"
+          echo "LATEST_VERSION=$LATEST_VERSION" >> "$GITHUB_ENV"
```
triceo, line 301: I'm thinking maybe we do the `sed` right here, instead of first sending the information to env and then reading it from there?

It's a lot of boilerplate already, and I don't think the extra script makes it easier to read or maintain.

triceo (approved):

PR #1234, opened 2024-11-28
feat: add new termination config based on the best solution improvement

zepfred:
Reopens #1196

This pull request introduces a new termination condition that activates when the solver cannot improve the best solution within a specified time frame. This time frame is determined by the speed of move evaluations, the last time the solver successfully improved the solution, and the duration without any improvements.

image
image


triceo:
Closed in favor of https://github.com/TimefoldAI/timefold-solver/pull/1313.

PR #1247, opened 2024-12-03
fix: Use a stable recipe order for `NullableRecipe.getRecipeList()`

timtebeek:
Hi! Great to see the recipes here; I do have one request: can we make the order of the recipe list stable?

For context: we generate documentation for recipes, even when maintained externally. Before this change we would see a new recipe list order for every run. After this change the order should be fixed, such that we do not keep changing that documentation page every time.

Thanks in advance!

### DIFF
--- a/migration/src/main/java/ai/timefold/solver/migration/v8/NullableRecipe.java
+++ b/migration/src/main/java/ai/timefold/solver/migration/v8/NullableRecipe.java
@@ -25,7 +25,7 @@ public final class NullableRecipe extends AbstractRecipe {
     @Override
     public List<Recipe> getRecipeList() {
         var packageName = "ai.timefold.solver.core.api.score.stream";
-        var methodsToRename = Map.ofEntries(
+        var methodsToRename = List.of(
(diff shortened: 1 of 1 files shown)


triceo (approved):
LGTM. Thanks, @timtebeek!

While I have you, quick question - how do I run these tests in parallel? They take _a lot_ of time, and when I tried parallelizing them via JUnit 5, I got a whole bunch of exceptions. I searched the OpenRewrite docs and nothing came up. Surely, there has to be a way...

timtebeek:
> While I have you, quick question - how do I run tests for these recipes in parallel? They take _a lot_ of time, and when I tried [parallelizing them via JUnit 5](https://junit.org/junit5/docs/5.3.0-M1/user-guide/index.html#writing-tests-parallel-execution), I got a whole bunch of exceptions. I searched the OpenRewrite docs and nothing came up. Surely, there has to be a way...

Good question! I guess we've never explored that option, nor do I know what issues that would run into. I'd briefly on the migration module here added the same files as I had seen on `core/src/test/resources/junit-platform.properties`, but saw no issues or noticable improvement. What did you run into before?

PR #1252, opened 2024-12-04
test: run OpenRewrite recipe tests in parallel

triceo:


triceo:
@timtebeek For whatever reason, I am no longer seeing the failures I saw before. I will have the entire CI run, that tests on all platforms, maybe something shows up.

triceo:
OK, still no sign of any exception. Ran it a hundred times locally, ran the entire CI three times - I guess I'll attribute the exceptions I saw to [cosmic bit flips](https://www.scienceabc.com/innovation/what-are-bit-flips-and-how-are-spacecraft-protected-from-them.html). If it ever happens again, I'll write it down and report directly to OpenRewrite. Thanks @timtebeek!

timtebeek:
Looks like a great improvement indeed; locally I see the 160 tests in `AsConstraintRecipeTest` go down from ~35 seconds to just 10. Now I wonder if we can do the same. 😅

zepfred (approved):

PR #1258, opened 2024-12-06
fix: ensure the sort manner is applied to the default CH config

zepfred:
This pull request corrects a bug that occurs when an entity selector configuration is created without using the sorting settings, particularly when the related entity has a difficulty weight factory defined  `@PlanningEntity(difficultyWeightFactoryClass...`.

The bug only affects solver configurations that do not define a phase list, as the solver sets the entity selector configuration before inner operations update the heuristic policy configuration. The issue does not affect the value selector config because no config is created by default (`DefaultSolverFactory#buildPhaseList`).

### DIFF
--- a/core/src/test/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseTest.java
@@ -27,6 +27,9 @@ import ai.timefold.solver.core.impl.testdata.domain.TestdataValue;
 import ai.timefold.solver.core.impl.testdata.domain.allows_unassigned.TestdataAllowsUnassignedEasyScoreCalculator;
 import ai.timefold.solver.core.impl.testdata.domain.allows_unassigned.TestdataAllowsUnassignedEntity;
 import ai.timefold.solver.core.impl.testdata.domain.allows_unassigned.TestdataAllowsUnassignedSolution;
+import ai.timefold.solver.core.impl.testdata.domain.difficultyComparator.TestdataDifficultyWeightEntity;
+import ai.timefold.solver.core.impl.testdata.domain.difficultyComparator.TestdataDifficultyWeightSolution;
+import ai.timefold.solver.core.impl.testdata.domain.difficultyComparator.TestdataDifficultyWeightValue;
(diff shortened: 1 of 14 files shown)


core/src/test/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseTest.java
```diff
@@ -341,4 +344,36 @@ void constructionHeuristicAllocateToValueFromQueue() {
...
+        solverConfig = PlannerTestUtils.buildSolverConfig(TestdataDifficultyWeightSolution.class,
+                TestdataDifficultyWeightEntity.class)
+                .withPhases(new ConstructionHeuristicPhaseConfig());
+
+        solution = new TestdataDifficultyWeightSolution("s1");
+        solution.setValueList(
+                List.of(new TestdataDifficultyWeightValue("v1"),
+                        new TestdataDifficultyWeightValue("v2")));
+        solution.setEntityList(List.of(new TestdataDifficultyWeightEntity("e1")));
+        solution = PlannerTestUtils.solve(solverConfig, solution);
```
triceo, line 373: I'll be honest - I don't like this. We can't be testing everything by running the solver. It takes too much time, and the tests get bloated.

I'd verify that this was detected on the Entity Descriptor, and that's that.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/difficultyweight/TestdataDifficultyWeightEntity.java
```diff
@@ -0,0 +1,61 @@
...
+import ai.timefold.solver.core.api.domain.variable.PlanningVariable;
+import ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import ai.timefold.solver.core.impl.testdata.domain.TestdataObject;
+
+@PlanningEntity(difficultyWeightFactoryClass = TestdataDifficultyWeightFactory.class)
+public class TestdataDifficultyWeightEntity extends TestdataObject {
+
+    public static final String VALUE_FIELD = "value";
+    private boolean comparisonCalled = false;
```
triceo, line 13: This is no longer necessary, some methods possibly too. Please simplify.

triceo (approved):
LGTM after last comment resolved.

PR #1260, opened 2024-12-06
feat: Verify variable change calls on every move in TRACKED_FULL_ASSERT

Christopher-Chianelli:
Additional changes:

- Assigned/Unassigned values are now tracked.

Future changes:

- Ensuring there are no changes outside the affected range.


triceo:
I've been thinking. `VariableListenerSupport` already tracks these events - see `Notification`, its children and where they are created. Why not hook this code in there instead? Seems to me that we are duplicating this information, making the assert mode even slower where maybe we can reuse what we've already built anyway.

triceo:
Doesn't look as if we'll be merging this. 
Closing, please reopen if you actually want this merged, @Christopher-Chianelli.

PR #1277, opened 2024-12-16
perf: reduce the LocationInList allocation rate

triceo:
Although the actual move evaluation speed increase is small (~ 3 %), the allocation rate drops substantially.


rsynek (approved):

rsynek (commented):
Sonar complains about insufficient coverage of the new code. It's true that I haven't noticed a new tests being added for the `ListVariableSupply`.

PR #1300, opened 2025-01-05
perf: reduce GC pressure during joining

triceo:


core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/bi/IndexedIfExistsBiNode.java
```diff
@@ -31,7 +30,7 @@ public IndexedIfExistsBiNode(boolean shouldExist,
     }
 
     public IndexedIfExistsBiNode(boolean shouldExist,
-            BiFunction<A, B, IndexProperties> mappingAB, Function<C, IndexProperties> mappingC,
+            BiMapping<A, B> mappingAB, IndexerFactory.UniMapping<C> mappingC,
```
zepfred, line 33:
```suggestion
            BiMapping<A, B> mappingAB, UniMapping<C> mappingC,
```

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/ComparisonIndexer.java
```diff
@@ -43,8 +43,8 @@ public ComparisonIndexer(JoinerType comparisonJoinerType, int propertyIndex,
     }
 
     @Override
-    public ElementAwareListEntry<T> put(IndexProperties indexProperties, T tuple) {
-        Key_ indexKey = indexProperties.toKey(propertyIndex);
+    public ElementAwareListEntry<T> put(Object indexProperties, T tuple) {
```
zepfred, line 46: Can we guarantee that the type of `indexProperties` is always `Key_`?

It seems better to me that using a contract requiring the type `Key_` is preferable to allowing two possible types.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/IndexerFactory.java
```diff
@@ -526,87 +467,119 @@ public Function<Right_, ai.timefold.solver.core.impl.score.stream.bavet.common.i
...
+
+    private static <A, B, C, D> QuadMapping<A, B, C, D> wrap(QuadFunction<A, B, C, D, Object> mapping) {
+        return (a, b, c, d) -> {
+            var result = mapping.apply(a, b, c, d);
+            return Objects.requireNonNullElse(result, NullIndexProperties.INSTANCE);
+        };
+    }
+
+    @FunctionalInterface
+    public interface UniMapping<A> extends Function<A, Object> {
```
zepfred, line 570: The logic seems correct. I wonder if the result type could be something other than `Object`, which would clarify the contract for inner nodes.

triceo: That is unfortunately the one downside of this change - it can only be `Object`.

Previously, it was `IndexProperties`, because everything got mapped to a subtype of `IndexProperties`. But this is no longer the case.

If we have a single key, that key will be returned directly, as opposed to being wrapped in `SingleIndexProperties`. That is where the savings come from. So, unfortunately, this can only ever be `Object` - and that object will sometimes be `IndexProperties`, and sometimes just an arbitrary user type.

I tried to make the situation easier by using `UniMapping` instead of `Function<A, Object>`. This way, when you see `UniMapping`, you already know what the `Object` means, and why.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/ComparisonIndexer.java
triceo: Actually, `Key_` is here just so that the code is a bit shorter.
Technically, when generics is erases, it becomes `Comparable` - and that is as much as we know at this point. We never head the information about the type of the key - only the user knows that.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/IndexerFactory.java
```diff
@@ -430,61 +384,48 @@ public <A> Function<A, ai.timefold.solver.core.impl.score.stream.bavet.common.in
...
-    public Function<Right_, ai.timefold.solver.core.impl.score.stream.bavet.common.index.IndexProperties> buildRightMapping() {
+    public UniMapping<Right_> buildRightMapping() {
         var joinerCount = joiner.getJoinerCount();
         return switch (joinerCount) {
             case 0 -> a -> NoneIndexProperties.INSTANCE;
             case 1 -> {
                 var mapping = joiner.getRightMapping(0);
-                yield a -> new ai.timefold.solver.core.impl.score.stream.bavet.common.index.SingleIndexProperties<>(
-                        mapping.apply(a));
+                yield mapping::apply;
```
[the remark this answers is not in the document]
triceo: Yes, see previous comment on the PR. This is one of the key savings of the PR.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/IndexerFactory.java
zepfred: Okay. Saving about `30%` of allocation is a great improvement.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/IndexerFactory.java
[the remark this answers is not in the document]
zepfred: I meant, should this be `yield wrap(mapping);` instead?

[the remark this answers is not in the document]
triceo: Hmm, yes, well spotted.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractIndexedIfExistsNode.java
```diff
@@ -23,21 +22,19 @@ public abstract class AbstractIndexedIfExistsNode<LeftTuple_ extends AbstractTup
...
     private final int inputStoreIndexLeftProperties;
     private final int inputStoreIndexLeftCounterEntry;
     private final int inputStoreIndexRightProperties;
     private final int inputStoreIndexRightEntry;
     private final Indexer<ExistsCounter<LeftTuple_>> indexerLeft;
     private final Indexer<UniTuple<Right_>> indexerRight;
 
-    protected AbstractIndexedIfExistsNode(boolean shouldExist,
-            Function<Right_, IndexProperties> mappingRight,
+    protected AbstractIndexedIfExistsNode(boolean shouldExist, IndexerFactory.UniMapping<Right_> mappingRight,
```
zepfred, line 33:
```suggestion
    protected AbstractIndexedIfExistsNode(boolean shouldExist, UniMapping<Right_> mappingRight,
```

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractUnindexedIfExistsNode.java
```diff
@@ -77,9 +75,7 @@ public final void updateLeft(LeftTuple_ leftTuple) {
             ElementAwareList<FilteringTracker<LeftTuple_>> leftTrackerList = leftTuple.getStore(inputStoreIndexLeftTrackerList);
             leftTrackerList.forEach(FilteringTracker::remove);
             counter.countRight = 0;
-            for (UniTuple<Right_> tuple : rightTupleList) {
-                updateCounterFromLeft(leftTuple, tuple, counter, leftTrackerList);
-            }
+            rightTupleList.forEach(leftTuple, counter, leftTrackerList, this::updateCounterFromLeft);
```
zepfred, line 78: Same as above

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractUnindexedIfExistsNode.java
```diff
@@ -44,17 +44,15 @@ public final void insertLeft(LeftTuple_ leftTuple) {
...
 
         if (!isFiltering) {
             counter.countRight = rightTupleList.size();
         } else {
-            ElementAwareList<FilteringTracker<LeftTuple_>> leftTrackerList = new ElementAwareList<>();
-            for (UniTuple<Right_> tuple : rightTupleList) {
-                updateCounterFromLeft(leftTuple, tuple, counter, leftTrackerList);
-            }
+            var leftTrackerList = new ElementAwareList<FilteringTracker<LeftTuple_>>();
+            rightTupleList.forEach(leftTuple, counter, leftTrackerList, this::updateCounterFromLeft);
```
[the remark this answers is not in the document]
triceo: No, there's no way how it could - it's all type-safe. If I were passing the parameter in the wrong place, the compiler would scream at me.

(And it did, until I fixed all of those places.)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractUnindexedIfExistsNode.java
```diff
@@ -106,32 +102,27 @@ public final void insertRight(UniTuple<Right_> rightTuple) {
...
+        var rightEntry = rightTupleList.add(rightTuple);
         rightTuple.setStore(inputStoreIndexRightEntry, rightEntry);
         if (!isFiltering) {
             leftCounterList.forEach(this::incrementCounterRight);
         } else {
-            ElementAwareList<FilteringTracker<LeftTuple_>> rightTrackerList = new ElementAwareList<>();
-            for (ExistsCounter<LeftTuple_> tuple : leftCounterList) {
-                updateCounterFromRight(rightTuple, tuple, rightTrackerList);
-            }
+            var rightTrackerList = new ElementAwareList<FilteringTracker<LeftTuple_>>();
```
zepfred, line 110: Same as above

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractUnindexedIfExistsNode.java
```diff
@@ -106,32 +102,27 @@ public final void insertRight(UniTuple<Right_> rightTuple) {
...
             // No fail fast if null because we don't track which tuples made it through the filter predicate(s)
             insertRight(rightTuple);
             return;
         }
         if (isFiltering) {
-            ElementAwareList<FilteringTracker<LeftTuple_>> rightTrackerList = updateRightTrackerList(rightTuple);
-            for (ExistsCounter<LeftTuple_> tuple : leftCounterList) {
-                updateCounterFromRight(rightTuple, tuple, rightTrackerList);
-            }
+            var rightTrackerList = updateRightTrackerList(rightTuple);
```
zepfred, line 124: Same as above

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/AbstractUnindexedJoinNode.java
```diff
@@ -40,19 +40,16 @@ public final void insertLeft(LeftTuple_ leftTuple) {
...
-        ElementAwareListEntry<LeftTuple_> leftEntry = leftTupleList.add(leftTuple);
+        var leftEntry = leftTupleList.add(leftTuple);
         leftTuple.setStore(inputStoreIndexLeftEntry, leftEntry);
-        ElementAwareList<OutTuple_> outTupleListLeft = new ElementAwareList<>();
+        var outTupleListLeft = new ElementAwareList<OutTuple_>();
         leftTuple.setStore(inputStoreIndexLeftOutTupleList, outTupleListLeft);
-        for (UniTuple<Right_> tuple : rightTupleList) {
-            insertOutTupleFiltered(leftTuple, tuple);
-        }
+        rightTupleList.forEach(leftTuple, (rightTuple, leftTuple_) -> insertOutTupleFiltered(leftTuple_, rightTuple));
```
zepfred, line 47: That's what I meant in my comments in `AbstractUnindexedIfExistsNode`

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/quad/IndexedIfExistsQuadNode.java
```diff
@@ -31,7 +30,7 @@ public IndexedIfExistsQuadNode(boolean shouldExist,
     }
 
     public IndexedIfExistsQuadNode(boolean shouldExist,
-            QuadFunction<A, B, C, D, IndexProperties> mappingABCD, Function<E, IndexProperties> mappingE,
+            QuadMapping<A, B, C, D> mappingABCD, IndexerFactory.UniMapping<E> mappingE,
```
zepfred, line 33:
```suggestion
            QuadMapping<A, B, C, D> mappingABCD, UniMapping<E> mappingE,
```

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/tri/IndexedIfExistsTriNode.java
```diff
@@ -31,7 +30,7 @@ public IndexedIfExistsTriNode(boolean shouldExist,
     }
 
     public IndexedIfExistsTriNode(boolean shouldExist,
-            TriFunction<A, B, C, IndexProperties> mappingABC, Function<D, IndexProperties> mappingD,
+            TriMapping<A, B, C> mappingABC, IndexerFactory.UniMapping<D> mappingD,
```
zepfred, line 33:
```suggestion
            TriMapping<A, B, C> mappingABC, UniMapping<D> mappingD,
```

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareList.java
```diff
@@ -72,8 +93,13 @@ public void remove(ElementAwareListEntry<T> entry) {
         } else {
             entry.next.previous = entry.previous;
         }
+        entry.list = null;
         entry.previous = null;
         entry.next = null;
+        if (availableBlankEntry == null) { // Entry will be reused.
```
zepfred, line 99: Nice!

zepfred (approved):
The allocation improvement is a great achivement. The logic looks great to me!

One last comment. Do we have enough test coverage for `IndexerFactory`? I'm wondering why no tests caught the issue regarding value wrapping?

PR #1306, opened 2025-01-07
feat: add support for visualizing Bavet node graph

triceo:


core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/visual/NodeGraph.java
```diff
@@ -0,0 +1,179 @@
...
+            var stream = (BavetAbstractConstraintStream<?>) castConstraint.getScoringConstraintStream();
+            var node = buildHelper.findParentNode(stream);
+            sinkList.add(new GraphSink<>(node, castConstraint));
+        }
+        return new NodeGraph<>(solution, sourceList.stream().distinct().toList(),
+                edgeList.stream().distinct().toList(),
+                sinkList.stream().distinct().toList());
+    }
+
+    public String write() {
```
zepfred, line 59: I'm unsure about the method name. What about `parse`?

PR #1308, opened 2025-01-07
fix: Use Named Solver fields and method parameters to determine list of solver names in Quarkus

Christopher-Chianelli:
In order to know the names of Solver beans to generate, the SolverBuildTimeConfig must extend the SolverRuntimeConfig, since all build time properties of SolverBuildTimeConfig are optional (so a named Solver might only be defined by termination, which is a runtime property).

Quarkus does not provide a mechanism for defining additional produces/instances of a bean at runtime (it goes against doing all possible work at build time), and thus, all names of Solver beans MUST be known at build time.

You cannot read a RuntimeConfig object inside a processor; all its methods will return null. Thus, it is impossible to get a list of runtime properties that were defined at build time.

Having SolverBuildTimeConfig extend SolverRuntimeConfig causes an issue. Namely, Quarkus, will generate a warning if a SolverRuntimeConfig property changes at runtime, since, by definition, it is also a SolverBuildTimeConfig property.

Quarkus provides a handy build item for surpressing the warning (although its intended use is to hide sensitive data): SuppressNonRuntimeConfigChangedWarningBuildItem.
Unfortunately, we need the EXACT name of the property to hide (there was no programmatic method available for getting the name from a method or class). Thus, we use ConfigurationBuildItem to get all build time properties (which natuarally includes the SolverBuildTimeConfig properties), filter them to check only properties that start with TimefoldBuildTimeConfig prefix, and suppress warnings for any property that match the filter and ends with a runtime property suffix.

### DIFF
--- a/quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolversInvalidSolutionClassTest.java
+++ b/quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolversInvalidSolutionClassTest.java
@@ -4,6 +4,10 @@ package ai.timefold.solver.quarkus;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+import ai.timefold.solver.core.api.solver.SolverManager;
 import ai.timefold.solver.quarkus.testdata.chained.domain.TestdataChainedQuarkusSolution;
 import ai.timefold.solver.quarkus.testdata.normal.constraints.TestdataQuarkusConstraintProvider;
 import ai.timefold.solver.quarkus.testdata.normal.domain.TestdataQuarkusEntity;
@@ -77,6 +81,14 @@ class TimefoldProcessorMultipleSolversInvalidSolutionClassTest {
                     .isInstanceOf(IllegalStateException.class)
                     .hasMessageContaining(
                             "Unused classes ([ai.timefold.solver.quarkus.testdata.chained.domain.TestdataChainedQuarkusSolution]) found with a @PlanningSolution annotation."));
+
+    @Inject
+    @Named("solver1")
+    SolverManager<?, ?> solverManager1;
+
+    @Inject
+    @Named("solver2")
+    SolverManager<?, ?> solverManager2;
 
     @Test
     void test() {
--- a/quarkus-integration/quarkus-benchmark/deployment/src/test/java/ai/timefold/solver/benchmark/quarkus/TimefoldBenchmarkProcessorMultipleSolversConfigTest.java
+++ b/quarkus-integration/quarkus-benchmark/deployment/src/test/java/ai/timefold/solver/benchmark/quarkus/TimefoldBenchmarkProcessorMultipleSolversConfigTest.java
@@ -5,9 +5,13 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import java.util.concurrent.ExecutionException;
 
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
 import ai.timefold.solver.benchmark.quarkus.testdata.normal.constraints.TestdataQuarkusConstraintProvider;
 import ai.timefold.solver.benchmark.quarkus.testdata.normal.domain.TestdataQuarkusEntity;
 import ai.timefold.solver.benchmark.quarkus.testdata.normal.domain.TestdataQuarkusSolution;
+import ai.timefold.solver.core.api.solver.SolverManager;
 
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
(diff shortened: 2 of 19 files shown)


triceo:
> In order to know the names of Solver beans to generate, the SolverBuildTimeConfig must extend the SolverRuntimeConfig, since all build time properties of SolverBuildTimeConfig are optional (so a named Solver might only be defined by termination, which is a runtime property).

What you say is true, currently, but I wonder if it's wise to hardcode that assumption. ("all build-time properties are optional") If we make one extend the other, we are effectively making a decision that it will stay like this forever, and whoever decides to change it, they'll have to redo a lot of what we do here.

What if instead we made that coupling just a little bit looser? SolverBuildTimeConfig doesn't extend SolverRuntimeConfig - instead they share an interface which specifies the common properties? (Wouldn't that also solve the problem of Quarkus complaining about overriding build-time properties? The properties would no longer be overridden, because on the class level, they'd be different properties.

But maybe I'm missing something here.

Christopher-Chianelli:
> > In order to know the names of Solver beans to generate, the SolverBuildTimeConfig must extend the SolverRuntimeConfig, since all build time properties of SolverBuildTimeConfig are optional (so a named Solver might only be defined by termination, which is a runtime property).
> 
> What you say is true, currently, but I wonder if it's wise to hardcode that assumption. ("all build-time properties are optional") If we make one extend the other, we are effectively making a decision that it will stay like this forever, and whoever decides to change it, they'll have to redo a lot of what we do here.

Quarkus users don't use the config classes; the config classes are just a programmatic way to define properties (i.e. making SolverBuildTimeConfig not extend SolverRuntimeConfig is backwards compatiable, provided SolverRuntimeConfig still exists). Moreover, the deployment classes (which SolverBuildTimeConfig is) are basically private API, since users do not have access to them at runtime.

> What if instead we made that coupling just a little bit looser? SolverBuildTimeConfig doesn't extend SolverRuntimeConfig - instead they share an interface which specifies the common properties? (Wouldn't that also solve the problem of Quarkus complaining about overriding build-time properties? The properties would no longer be overridden, because on the class level, they'd be different properties. Quarkus doesn't know they share an interface - Quarkus only sees the final class.)
> 
> But maybe I'm missing something here.

To do that, the build time and runtime properties must have a different prefix. They currently have the same prefix of `quarkus.timefold.solver`. As a result, any shared property is automatically a build time property to Quarkus,. To change that is to break backwards compatibility.

Christopher-Chianelli:
If we want to break that assumption (all build time properties optional), we break backwards compatibility (since now all Quarkus users need to specify it). We also break backwards compatibility if we only make it required when multiple solvers are used (ex: `quarkus.timefold.solver.names`) (since now all Quarkus users that use named solvers need to specify it).

triceo (approved):
LGTM.

Before I merge though - do we have a test case to check if the overriding actually works? If not, let's add it please.

Christopher-Chianelli:
I already added a test case; although I think I have an idea that might be able to remove BuildConfig extending RuntimeConfig. Namely, we MIGHT be able to use the `@Named("...") Solver solver` fields to determine the list of names. Although I still need to check if what I am thinking of is even possible.

triceo:
I have seen the tests in this PR, and they do override the properties - but they do not check if the overrides were actually applied, in practice, in the solver config. 

> Although I still need to check if what I am thinking of is even possible.

If you have an idea to improve this still, go for it. We do not need to merge this immediately.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorMultipleSolversInvalidSolutionClassTest.java
```diff
@@ -78,6 +82,14 @@ class TimefoldProcessorMultipleSolversInvalidSolutionClassTest {
                     .hasMessageContaining(
                             "Unused classes ([ai.timefold.solver.quarkus.testdata.chained.domain.TestdataChainedQuarkusSolution]) found with a @PlanningSolution annotation."));
 
+    @Inject
+    @Named("solver1")
+    SolverManager<?, ?> solverManager1;
+
+    @Inject
+    @Named("solver2")
+    SolverManager<?, ?> solverManager2;
```
triceo, line 91: Why do we need to add these, if the test worked without them before?
Smells of a backwards incompatible change.

Christopher-Chianelli: If you are using named solver managers, you should have these annotations somewhere in your code base. I argue the old behaviour is a bug. And all bugs are incompatible changes. As noted in my commit:

"""
As a consequence, there is a behaviour change in the very contrived
case of "exactly one named solver in properties (no default), no Named annotations,
injected default solver". Since there was no Named annotations,
the (empty) default solver properties will be checked instead of the single
named solver properties. I consider the old behaviour to be a bug
(named properties should not affect default properties),
but the tests expect that behaviour.
"""

Christopher-Chianelli: (there is also a question if we should add a fail fast if there a named build time solver without a corresponding Named annotation (ex: misspelled the name by accident, "solve1" instead of "solver1"))

triceo: I think that adding the fail-fast makes sense. We can give a nice exception message, listing all the named solvers present in config which do not have the accompanying annotation.

PR #1309, opened 2025-01-08
fix: Ruin and Recreate move reversion

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/move/director/VariableChangeRecordingScoreDirector.java
```diff
@@ -203,4 +203,20 @@ public void recordListAssignment(ListVariableDescriptor<Solution_> variableDescr
...
+     * Used when moves with nested phases create a new solution but do not record events before the list changes,
+     * such as ruin and recreate.
+     * The before event must be manually recorded,
+     * or the original list will not be restored during the rollback action.
+     *
+     * @param variableDescriptor never null
+     * @param entity never null
+     * @param list never null and may be empty
+     */
+    public void recordBeforeListVariableChanged(ListVariableDescriptor<Solution_> variableDescriptor, Object entity,
```
triceo, line 218: Two questions:


1. If this is `recordBeforeListVariableChanged`, does it mean that `recordListAssignment()` should instead be `recordAfterListVariableChanged`?
2.  We typically call the usual before/after methods on the score director, except in this case, where we have a whole new bunch of methods. It is confusing. It looks as is these methods are glued to the recording score director "from the side". They look like they don't belong. I wonder... does the score director need refactoring in order for these methods to fit more naturally?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
```diff
@@ -95,6 +95,19 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
...
+                // meaning it is a new destination entity without a ListVariableBeforeChangeAction
+                // to restore the original elements.
+                // We need to ensure the before action is executed in order to restore the original elements.
+                var beforeActionElementList = new ArrayList<>(listVariableDescriptor.getValue(entry.getKey()));
+                for (var element : entry.getValue()) {
+                    var location = listVariableStateSupply.getLocationInList(element).ensureAssigned();
+                    beforeActionElementList.set(location.index(), null);
+                }
+                recordingScoreDirector.recordBeforeListVariableChanged(listVariableDescriptor, entry.getKey(),
+                        beforeActionElementList.stream().filter(Objects::nonNull).toList());
```
triceo, line 109: This is strange, and probably inefficient.
Why put the null entries into the list, only to remove them later?

zepfred: I'm still considering the approach. The logic confirmed that my investigation into the cause of the issue was accurate. The goal is to recompute the related entity's list after the change has been completed.

It is not ready yet.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
```diff
@@ -95,7 +98,32 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
...
         for (var entry : entityToInsertedValuesMap.entrySet()) {
-            recordingScoreDirector.recordListAssignment(listVariableDescriptor, entry.getKey(), entry.getValue());
+            if (!entityToOriginalPositionMap.containsKey(entry.getKey())) {
+                // The entity has not been evaluated while creating the entityToOriginalPositionMap,
+                // meaning it is a new destination entity without a ListVariableBeforeChangeAction
+                // to restore the original elements.
+                // We need to ensure the before action is executed in order to restore the original elements.
+                var beforeActionElementList =
+                        constructionHeuristicPhase.getMissingUpdatedElementsMap().get(entry.getKey());
+                recordingScoreDirector
```
triceo, line 108: I'm wondering if this recorder type isn't overkill. Effectively what you need to do is:
- Have before/after events that are delegated to the underlying score director.
- Have before/after events that are not delegated.

So why not create duplicates of the before/after methods, which clearly say they will not be delegated? That way, the complexity of the recorder and of the custom action can be removed. At least externally. You can still keep them internally, if you need them.

Alternatively, you can do something like 

    recordingScoreDirector.undelegated(scoreDirector -> ...);

And hide all of the undelegated operations inside this one lambda. That way, it is very clear what is and what is not delegated, and you still avoid the (external) complexity of the recorder and the custom action. (Both can remain internally, if that makes it easier to implement the functionality.)

triceo (commented):
I believe that the recorder situation can be simplified. The move should not really deal with what's recorded, and what's a custom change. The move should just say what it wants - it wants some changes to not be delegated.

zepfred:
Fixed by #1320

PR #1313, opened 2025-01-10
feat: add new adaptive termination type

### LINKED ISSUE #1311: Adaptive termination
Absolute terminations (such as time spent, unimproved time spent etc.) are not good enough, because they require a different configuration for every problem and every data set. (You need to know how long to solve for.) We want to provide a new kind of termination, which is only configured in relative terms - not in absolute time, or absolute score difference.

Please see [the proposal](https://github.com/user-attachments/files/18360255/Quick.sheets.1.pdf). We define several parameters:

- A time window G, a time interval. Default value proposed to be 30 seconds.
- A score difference D, a ratio of two the first and the last score. Default value proposed to be 0.01, or 1 %.

The termination would work like this:

1. The grace period starts and has a length of `G`. During the grace period, termination is not allowed.
2. At the end of the grace period, a softest score is read, and compared to the softest score at the beginning of the grace period. The difference of the two becomes `Y_1`. If `Y_1 == 0`, immediate termination.
3. In the next move, a time windows `T_N` is established, by subtracting `G` from the current time. The softest score at the end of `T_N` (= the current score) is compared to the softest score at the beginning of `T_N`, and the difference becomes `Y_N`. 
4. `Y = Y_N/Y_1`. If `D > Y`, terminate. Otherwise:
5. Back to 3, or back to 1 if the softest-but-last score level changed. If the score only has one level, back to 3. If the score has multiple levels, the grace period re-starts (back to 1) every time the second-to-last level improves. Example: For `HardSoftScore`, the grace period restarts every time the hard score improves.

Naming not decided yet.

Christopher-Chianelli:
The adaptive termination type has the concept of a grace period G and minimum improvement ratio R.

The adaptive termination type works as follows.

- When in the grace period, do not terminate.

- If a non-lowest score level changes at any point, reset the grace period (or enter a new one if the grace period is finished).

- After the duration G of the grace period passes, record the difference of final best score and the starting best score as Y_0. If this is 0, terminate.

- On each step, calculate the difference between the current best score and the best score G moments ago as Y_n. If Y_n / Y < R (i.e. the score did not improve by at least R% of the improvement in the grace window), terminate.

Fixes #1311


core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -0,0 +1,171 @@
...
+    boolean isGracePeriodActive;
+    private long gracePeriodStartTimeMillis;
+    private LevelScoreDiff gracePeriodScoreDiff;
+
+    private final NavigableMap<Long, Score_> scoresByTime;
+
+    public AdaptiveTermination(long gracePeriodMillis, double minimumImprovementRatio) {
+        this.gracePeriodMillis = gracePeriodMillis;
+        this.minimumImprovementRatio = minimumImprovementRatio;
+        this.scoresByTime = new TreeMap<>();
```
triceo, line 28: I was skeptical of this approach, as fast-stepping algorithms will easily put tens of thousands of entries to this map per second, and I was expecting that to be both CPU-intensive and memory-intensive.

However, in the experiments I ran, this did not materialize. Impact of the termination is barely visible in the data. Interesting.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private record LevelScoreDiff(boolean harderScoreChanged, double softestScoreDiff) {
```
triceo, line 31: Typically, I would be all for a carrier type like this. In this case, and on the hot path, I'd rather store the two variables individually and avoid allocating the instance of the record. 

In fact, this is the only thing that actually made a measurable impact in my experiments - `LevelScoreDiff` was ~10th most allocated object on the heap.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
triceo: I wonder though if the `TreeMap` is bringing any tangible benefits. Consider this:

- We don't actually need sorting, do we? The values are sequential, a `List` would do.
- We are forced to box `long` into `Long`, causing some unnecessary GC pressure.

I wonder if we could convert this into `List<Score_>`. The index would have to become relative to the last reset, so starting with zero.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -0,0 +1,171 @@
...
+            return false;
+        }
+
+        return scoreDiff.softestScoreDiff / gracePeriodScoreDiff.softestScoreDiff < minimumImprovementRatio;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean isSolverTerminated(SolverScope<Solution_> solverScope) {
+        return isTerminated(System.currentTimeMillis(), (Score_) solverScope.getBestScore());
```
triceo, line 117: I wonder if we should use nanotime instead, and truncate it to millis.
The precision of `currentTimeMillis` is questionable, see Javadoc:

> Note that while the unit of time of the return value is a millisecond, the granularity of the value depends on the underlying operating system and may be larger. For example, many operating systems measure time in units of tens of milliseconds.

triceo: Also... should this terminate the solver even in, for example, a CH phase? Arguably it should not, the CH phase generally doesn't improve.

The way how unimproved terminations get around it is by forcing the adaptive termination to be only possible in local search phase. Maybe we do something like that too?

Although I really don't like that, because it causes config pain. (You need to define LS phase, and put the termination inside of it.) Maybe we refactor the terminations a bit, so that it is possible for an unimproved termination to be defined on the solver-level, but never terminate a CH? Would need a discussion, I think.

(My ultimate goal is to have only one termination. This one. And this configuration quirk is preventing that from happening.)

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    public boolean isPhaseTerminated(AbstractPhaseScope<Solution_> phaseScope) {
+        return isTerminated(System.currentTimeMillis(), phaseScope.getBestScore());
+    }
+
+    @Override
+    public double calculateSolverTimeGradient(SolverScope<Solution_> solverScope) {
+        return -1.0;
```
triceo, line 127: I wonder if there's a way to estimate this based on how close we are to the minimum necessary improvement.

triceo (commented):
That was quick! Leaving suggestions for improvement.

Maybe we introduce some TRACE-level logging? This termination is doing smart things, and some visibility into the mechanism may be necessary. We should find a decent way of logging the decrease of the ratio - maybe log every time in halves?

Also required:
- Documentation. Preferrably with a visualization as above.
- Naming. (Already started a conversation on Slack.)
- Add this to quickstarts?

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.TreeMap;
+
+import ai.timefold.solver.core.api.score.Score;
+import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
+import ai.timefold.solver.core.impl.phase.scope.AbstractStepScope;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+import ai.timefold.solver.core.impl.solver.thread.ChildThreadType;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
```
triceo, line 13: You can use `@NullMarked` which means that everything is `@NonNull` unless marked as `@Nullable`. In my experience, this saves a lot of hassle.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -1,57 +1,71 @@
...
+            // optimization: since most of the time the score the same,
+            // we can use equals to avoid creating double arrays in the
+            // vast majority of comparisons
+            return 0.0;
+        }
+        var scoreDiffs = end.subtract(start).toLevelDoubles();
+        var softestLevel = scoreDiffs.length - 1;
+        for (int i = 0; i < softestLevel; i++) {
+            if (scoreDiffs[i] != 0.0) {
+                throw new HardLevelImprovedSignal();
```
triceo, line 65: May I suggest we return an impossible value instead of using exceptions for control flow? Perhaps [NaN](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Double.html#NaN)?

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
```diff
@@ -1,57 +1,71 @@
...
+    private static final class HardLevelImprovedSignal extends Exception {
+        @Override
+        public Throwable fillInStackTrace() {
+            return this;
+        }
+    }
+
+    private static <Score_ extends Score<Score_>> double softImprovement(@NonNull Score_ start,
+            @NonNull Score_ end) throws HardLevelImprovedSignal {
+        if (start.equals(end)) {
```
triceo, line 55: Can the score be _worse_ here? Arguably, it can - sometimes the solver picks a worse move to try to get out of a local optima.

Christopher-Chianelli: No, it uses best score, not step score. Best score should never decrease.
```
stepScope.getPhaseScope().getBestScore()
```

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
triceo: I understand your reasoning, but I disagree. In my opinion, what you're doing here is misusing exceptions - you want to terminate without a return, and exceptions are a _nice_ way of doing it. When you have a hammer, everything becomes a nail.

Exceptions are meant for _exceptional_ cases - this situation is not exceptional.
Exceptions are meant for reporting errors, not for control flow. This is not an error.
And finally, unless I'm mistaken, exceptions mess with JVM's optimizations, because they abort the flow.

I could go on, but I'd just be duplicating what was already said much better in [the Java bible](https://www.amazon.com/Effective-Java-Joshua-Bloch/dp/0134685997). :-)

triceo: Good, data is always better than no data. Doesn't this in fact prove that using the exception is the slower option, as I expected?

> that there is no way to enforce the caller to actually check it is NaN.

The argument would have value if this was public API where there was a risk that the user can misuse it. However, this is a private method inside of an internal implementation - external users don't enter the picture, and for our purposes we make sure it works through tests.

Christopher-Chianelli: > Good, data is always better than no data. Doesn't this in fact prove that using the exception is the slower option, as I expected?

Yes (although I wouldn't use prove; this is only a single benchmark run, on a single machine).

> > that there is no way to enforce the caller to actually check it is NaN.
> 
> The argument would have value if this was public API where there was a risk that the user can misuse it. However, this is a private method inside of an internal implementation - external users don't enter the picture, and for our own purposes we make sure it works through tests.

I will rename the method to something like `softImprovementOrNaNOnHardImprovement` to prevent headaches for future me.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveScoreRingBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+
+import java.util.Arrays;
+import java.util.Objects;
+
+import ai.timefold.solver.core.api.score.Score;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+final class AdaptiveScoreRingBuffer<Score_ extends Score<Score_>> {
```
triceo, line 11: Is there a way to refactor the ring buffer to make it a general-purpose collection?
I can see it being useful to the Solver for more than this.

Christopher-Chianelli: Yes, but then we need to box the longs (unless you want to keep it as a map with long keys?) . Additionally it not a typical ring buffer (which normally acts a queue); its method removes all elements up to the floored query (and keeps the query entry). It also relies on the fact the queries will always increase due to the behaviour of `System.nanoTime`.

If you can spec out what you want the RingBuffer to be able to do, I can refactor it to match.

triceo: Nevermind then - the generalization would hurt the primary purpose.
But the description of differences from a "typical" ring buffer would make for a nice Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AdaptiveTermination.java
Christopher-Chianelli: It cannot terminate in the CH phase, because the CH ALWAYS improve (namely, it improves the init score every step, which resets the grace period).

triceo: Not necessarily true - list var with allows unassigned has init score always zero.
But doesn't matter - this termination, for now, is only allowed on the LS phase.

Christopher-Chianelli: CH in unassigned does not decrease the best score; it only can increase it (it can leave the value unassigned (no increase), or assign a value (which more than likely will increase the medium score).

triceo: Those are assumptions - as long as the thing is technically possible, we need to have an answer. In this case, this (along with the unimproved terminations) will be exempt from terminating the CH, once it's even possible to configure them on the top level. For now, until we implement that, it's not a problem at all.

In fact, I am surprised why this termination even cares about init score. How does it receive events from CH, if it needs to be configured on the LS only? That smells like a bug.

Christopher-Chianelli: It was from `solvingStarted` (where the best score is typically uninitialized). All terminations gets the `solvingStarted` event even if configured on a phase level.

Christopher-Chianelli: (in short, the termination was incorrectly using `solvingStarted` instead of `phaseStarted` to initialize).

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
```diff
@@ -0,0 +1,224 @@
...
+            System.arraycopy(nanoTimeRingBuffer, 0, newNanoTimeRingBuffer, firstLength, secondLength);
+            System.arraycopy(scoreRingBuffer, 0, newScoreRingBuffer, firstLength, secondLength);
+            readIndex = 0;
+            writeIndex = totalLength;
+        }
+        nanoTimeRingBuffer = newNanoTimeRingBuffer;
+        scoreRingBuffer = newScoreRingBuffer;
+    }
+
+    public void clear() {
```
triceo, line 119: I wonder if:

- Throwing away the array and re-allocating it would be faster than clearing it. (Maybe not.)
- We should allocate the array using the original size, not using whatever size it is now.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
```diff
@@ -0,0 +1,224 @@
...
+        Arrays.fill(scoreRingBuffer, null);
+    }
+
+    /**
+     * Returns the first element of the score ring buffer.
+     * Does not remove the element.
+     *
+     * @return the first element of the score ring buffer
+     */
+    public @NonNull Score_ peekFirst() {
```
triceo, line 132: Since the array starts as empty, this should not be annotated as `@NonNull`, because `null` is entirely possible.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Returns the latest score prior or at the given time, and removes
+     * all time-score pairings prior to it.
+     *
+     * @param nanoTime the queried time in nanoseconds.
+     * @return the latest score prior to the given time.
+     */
+    public @NonNull Score_ pollLatestScoreBeforeTimeAndClearPrior(long nanoTime) {
+        if (readIndex == writeIndex && nanoTimeRingBuffer[writeIndex] == 0L) {
+            throw new IllegalStateException("Ring buffer is empty");
```
triceo, line 185: Do you want to treat this as an exception? I see two options here:

- Either this is an exceptional state which must not happen, because the algorithm would not be correct. Then the message should start with "Impossible state: ", and I'm going to argue it should be thrown by the code calling this method, not by this method directly.
- Or this is a totally possible state, which we just want to signal to the calling code. In that case, this exception should be replaced by a return on `null` - see our previous conversation on exceptions and control flow.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
```diff
@@ -0,0 +1,179 @@
...
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+import ai.timefold.solver.core.impl.solver.thread.ChildThreadType;
+
+import org.jspecify.annotations.NonNull;
+
+public final class DiminishedReturnsTermination<Solution_, Score_ extends Score<Score_>>
+        extends AbstractTermination<Solution_> {
+    static final long NANOS_PER_MILLISECOND = 1_000_000;
+
+    private final long slidingWindowNanos;
```
triceo, line 15: If the length of the window was a `Duration` (see my comment on the termination config), then you could avoid this and just call `slidingWindowDuration.toMillis()`.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
```diff
@@ -0,0 +1,179 @@
...
+            var timeElapsedNanos = currentTime - gracePeriodStartTimeNanos;
+            if (timeElapsedNanos >= slidingWindowNanos) {
+                // grace period over, record the reference diff
+                isGracePeriodActive = false;
+                gracePeriodSoftestImprovementDouble = endpointDiff;
+                if (endpointDiff < 0.0) {
+                    // Should be impossible; the only cases where the best score improves
+                    // and have a lower softest level are if either a harder level or init score
+                    // improves, but if that happens, we reset the grace period.
+                    throw new IllegalStateException("The score deteriorated from (%s) to (%s) during the grace period."
```
triceo, line 119:
```suggestion
                    throw new IllegalStateException("Impossible state: The score deteriorated from (%s) to (%s) during the grace period."
```

triceo: This is a convention we generally honor in other parts of the codebase.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/TerminationFactory.java
```diff
@@ -122,6 +122,19 @@ The termination with bestScoreFeasible (%s) can only be used with a score type \
...
 
+        if (terminationConfig.getDiminishedReturnsConfig() != null) {
+            var adaptiveTerminationConfig = terminationConfig.getDiminishedReturnsConfig();
+            var gracePeriodMillis = adaptiveTerminationConfig.calculateSlidingWindowMilliseconds();
+            if (gracePeriodMillis == null) {
+                gracePeriodMillis = 30_000L;
+            }
+            var minimumImprovementRatio = adaptiveTerminationConfig.getMinimumImprovementRatio();
+            if (minimumImprovementRatio == null) {
+                minimumImprovementRatio = 0.01;
```
triceo, line 133: See Slack thread on the "correct" default value.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorOverridePropertiesAtRuntimeTest.java
```diff
@@ -48,6 +48,9 @@ private static String getRequiredProperty(String name) {
             .overrideConfigKey("quarkus.timefold.solver.termination.best-score-limit", "0")
             .overrideConfigKey("quarkus.timefold.solver.move-thread-count", "4")
             .overrideConfigKey("quarkus.timefold.solver-manager.parallel-solver-count", "1")
+            .overrideConfigKey("quarkus.timefold.solver.termination.diminished-returns.enabled", "false")
```
triceo, line 51: Arguably this is not necessary, if either of the other two properties is present?

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorXMLDiminishedReturnsTest.java
```diff
@@ -0,0 +1,24 @@
+package ai.timefold.solver.quarkus;
+
+class TimefoldProcessorXMLDiminishedReturnsTest {
```
triceo, line 3: Delete or finish?

quarkus-integration/quarkus/runtime/src/main/java/ai/timefold/solver/quarkus/config/DiminishedReturnsRuntimeConfig.java
```diff
@@ -0,0 +1,44 @@
...
+     * For example, if the {@link #slidingWindowDuration} is "30s",
+     * the {@link #minimumImprovementRatio} is 0.25, and the
+     * score improves by 100soft during the first 30 seconds of local search,
+     * then the local search phase will terminate when the difference between
+     * the current best score and the best score from 30 seconds ago is less than
+     * 25soft (= 0.25 * 100soft).
+     * <br/>
+     * Defaults to 0.1.
+     */
+    Optional<Double> minimumImprovementRatio();
```
triceo, line 43: `OptionalDouble` exists.
(For some reason, `OptionalBoolean` does not.)

quarkus-integration/quarkus/runtime/src/main/java/ai/timefold/solver/quarkus/config/DiminishedReturnsRuntimeConfig.java
```diff
@@ -0,0 +1,44 @@
...
+     * initial improvement. Must be positive.
+     * <br/>
+     * For example, if the {@link #slidingWindowDuration} is "30s",
+     * the {@link #minimumImprovementRatio} is 0.25, and the
+     * score improves by 100soft during the first 30 seconds of local search,
+     * then the local search phase will terminate when the difference between
+     * the current best score and the best score from 30 seconds ago is less than
+     * 25soft (= 0.25 * 100soft).
+     * <br/>
+     * Defaults to 0.1.
```
triceo, line 41: Please remember to update.

triceo (commented):
Looking good!

Still need to add documentation. Since this termination is non-trivial, we'll need an illustration (akin to what I provided above) together with an explanation of the concepts, and broad strokes of the algorithm.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorOverridePropertiesAtRuntimeTest.java
Christopher-Chianelli: The other two properties are optional, which make it hard to use the defaults unless there a third property.
Additionally, if you want to use diminishing returns by default, but have the option to disable it, you need to somehow delete the defined properties if there is no enabled property.

triceo: I see the reason for the property existing - I'm just arguing you don't need to specify it, if you specify any of the others.

Does it get confusing as you switch from one to the other? Probably. But how often does that actually happen?

core/src/main/java/ai/timefold/solver/core/config/solver/termination/DiminishedReturnsTerminationConfig.java
```diff
@@ -0,0 +1,210 @@
...
+                    + ") or slidingWindowDays (" + slidingWindowDays + ").");
+        }
+        long slidingWindowMillis = 0L
+                + requireNonNegative(slidingWindowMilliseconds, "slidingWindowMilliseconds")
+                + requireNonNegative(slidingWindowSeconds, "slidingWindowSeconds") * 1000L
+                + requireNonNegative(slidingWindowMinutes, "slidingWindowMinutes") * 60_000L
+                + requireNonNegative(slidingWindowHours, "slidingWindowHours") * 3_600_000L
+                + requireNonNegative(slidingWindowDays, "slidingWindowDays") * 86_400_000L;
+        return slidingWindowMillis;
+    }
```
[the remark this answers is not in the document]
Christopher-Chianelli: This is how the other terminations do this.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
```diff
@@ -0,0 +1,224 @@
...
+ * {@link DiminishedReturnsScoreRingBuffer} automatically clear entries prior
+ * to the queried time when polled (if the queried time is not in the
+ * {@link DiminishedReturnsScoreRingBuffer}, then the largest entry prior
+ * to the queried time is kept and returned).
+ * The ring buffers will automatically resize when filled past their capacity.
+ * 
+ * @param <Score_> The score type
+ */
+final class DiminishedReturnsScoreRingBuffer<Score_ extends Score<Score_>> {
+    private static final int DEFAULT_CAPACITY = 4096;
```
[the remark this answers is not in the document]
Christopher-Chianelli: Arbitrary but commonly used; 4096 is the page size in the linux kernel, the typical size of a file system sector, etc. It is also a common buffer size used in Java libraries. Want to use a power of two for alignment + probably faster modulo.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
Christopher-Chianelli: We should keep the current size; the size of this buffer is based on score calculation speed (the higher the speed, the more entries in the buffer). Since score calculation speed stays roughly the same in LS, the buffer has an ideal size that it will always regrow to if was shrunk.

Creating a new array each time creates additional memory pressure (although clearing should be rare in most cases) and may invalidate the CPU cache.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
Christopher-Chianelli: Exceptional state which must not happen; correctness assumes there at least one element in the buffer (which is inserted on start & whenever grace is reset).

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsScoreRingBuffer.java
Christopher-Chianelli: The buffer starts with the score at the start of a phase, and polling will always keep one element in the buffer. Thus, it is an impossible state to have the buffer empty when it's called.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
Christopher-Chianelli: It was done like this to match how other terminations do it.

quarkus-integration/quarkus/runtime/src/main/java/ai/timefold/solver/quarkus/config/DiminishedReturnsRuntimeConfig.java
Christopher-Chianelli: As for why `OptionalBoolean` does not exist: https://mail.openjdk.org/pipermail/lambda-dev/2012-November/006750.html

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
 
 Switching xref:using-timefold-solver/running-the-solver.adoc#environmentMode[EnvironmentMode] can heavily impact when this termination ends.
 
+[#diminishedReturnsTermination]
+==== `DiminishedReturnsTermination`
+
+Terminates when the rate of improvement is below a percentage of the initial rate of improvement.
+The rate of improvement is the score difference between the final best score and the initial best score in a window, divided by the duration of the window.
```
triceo, line 644: I think we should talk about the softest score here.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
...
+The first window is called a grace period, which cannot trigger a termination.
+
+image::optimization-algorithms/overview/diminishedReturnsTermination.png[align="center"]
+
+[NOTE]
+====
+If the hard score improves, the grace period is reset and the initial rate of improvement is recalculated.
+====
```
triceo, line 652: Since this is a core part of the algorithm, I don't think it should be a NOTE. Rather, a normal text like everything else.

Also, the above comment about the softest score will pay off here, where we talk about the hard score all of a sudden.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
...
+[source,xml,options="nowrap"]
+----
+  <localSearch>
+    <termination>
+      <diminishedReturns>
+        <slidingWindowSeconds>500</slidingWindowSeconds>
+      </diminishedReturns>
+    </termination>
+  </localSearch>
+----
```
triceo, line 704: I consider this section too verbose. It's IMO enough to show the duration and the seconds, and then mention we also support the others.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
...
+  <localSearch>
+    <termination>
+      <diminishedReturns>
+        <minimumImprovementRatio>0.2</minimumImprovementRatio>
+      </diminishedReturns>
+    </termination>
+  </localSearch>
+----
+
+This `Termination` can only be used for a `Phase` (such as ``<localSearch>``), not for the `Solver` itself.
```
triceo, line 759: I'd invert this statement - this termination cannot be used globally on the solver, only on a phase, and not on a CH phase.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
...
 
 [#combiningMultipleTerminations]
```
triceo, line 761: I'd mention that the termination can run for a very long time, and add a recommendation to combine the termination with a maximum time spent, to ensure it doesn't run for longer than acceptable.

Do we have a test for this? When I was toying with the termination, I saw some weird behavior where a composite termination wasn't working properly.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -637,6 +637,126 @@ This is useful for benchmarking.
 
 Switching xref:using-timefold-solver/running-the-solver.adoc#environmentMode[EnvironmentMode] can heavily impact when this termination ends.
 
+[#diminishedReturnsTermination]
+==== `DiminishedReturnsTermination`
```
triceo, line 641: I'd start with an introductory paragraph explaining why we've done this. Absolute terminations are hard to configure, because they either depend on hardware (time-based) or on the problem (score-based). The relative termination doesn't have this problem.

triceo (approved):
LGTM after comments resolved.
Looking forward to using this one!

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
Christopher-Chianelli: Copied from unimproved time spent termination, which list them all: https://docs.timefold.ai/timefold-solver/latest/optimization-algorithms/overview#unimprovedTimeMillisSpentTermination

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
Christopher-Chianelli: Copied from step count termination https://docs.timefold.ai/timefold-solver/latest/optimization-algorithms/overview#stepCountTermination

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
Christopher-Chianelli: [We have AND/OR composition tests](https://github.com/TimefoldAI/timefold-solver/blob/8ca431c4a6a2edf4b74f903d8aa40a45f5b0355d/core/src/test/java/ai/timefold/solver/core/impl/solver/termination/AndCompositeTerminationTest.java#L15). When I tried composing it with a time spent termination, it terminated as expected (did one with a short time spent, long grace, and another with a long time spent, short grace; both terminated quickly).

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
Christopher-Chianelli: And it can be used on a CH phase (no terminations are exclusive to LS; all phase exclusive terminations will work on any phase). We are also inconsistent in how the unimproved terminations handle it; unimproved time spent is allowed to be configured on a solver level (and just ignores the CH: https://github.com/TimefoldAI/timefold-solver/blob/4b446075bea81eea393edd887a32960470e2970b/core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedTimeMillisSpentScoreDifferenceThresholdTermination.java#L61-L71), whereas unimproved step count must be configured on phases (https://github.com/TimefoldAI/timefold-solver/blob/4b446075bea81eea393edd887a32960470e2970b/core/src/main/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedStepCountTermination.java#L25-L26).

triceo: This is interesting - the inconsistency is surprising me. I thought the fail-fast was present in all the unimproved terminations. We'll get to fix that eventually.

PR #1320, opened 2025-01-13
fix: ensure entity placer is not shared in ruin and recreate moves

zepfred:
Fixes TimefoldAI/timefold-employee-scheduling#305

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/director/VariableChangeRecordingScoreDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/director/VariableChangeRecordingScoreDirector.java
@@ -4,6 +4,7 @@ import java.util.IdentityHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import ai.timefold.solver.core.api.score.director.ScoreDirector;
 import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
@@ -17,9 +18,8 @@ import ai.timefold.solver.core.impl.score.director.VariableDescriptorCache;
 public final class VariableChangeRecordingScoreDirector<Solution_>
         implements RevertableScoreDirector<Solution_> {
 
-    private final InnerScoreDirector<Solution_, ?> delegate;
+    private final InnerScoreDirector<Solution_, ?> backingScoreDirector;
     private final List<ChangeAction<Solution_>> variableChanges;
-    private boolean triggerDelegateActions = true;
 
     /*
      * The fromIndex of afterListVariableChanged must match the fromIndex of its beforeListVariableChanged call.
@@ -40,24 +40,23 @@ public final class VariableChangeRecordingScoreDirector<Solution_>
      */
     private final Map<Object, Integer> cache;
 
-    public VariableChangeRecordingScoreDirector(ScoreDirector<Solution_> delegate) {
-        this(delegate, true);
+    public VariableChangeRecordingScoreDirector(ScoreDirector<Solution_> backingScoreDirector) {
+        this(backingScoreDirector, true);
     }
 
-    public VariableChangeRecordingScoreDirector(ScoreDirector<Solution_> delegate, boolean requiresIndexCache) {
-        this.delegate = (InnerScoreDirector<Solution_, ?>) delegate;
+    public VariableChangeRecordingScoreDirector(ScoreDirector<Solution_> backingScoreDirector, boolean requiresIndexCache) {
+        this.backingScoreDirector = (InnerScoreDirector<Solution_, ?>) backingScoreDirector;
         this.cache = requiresIndexCache ? new IdentityHashMap<>() : null;
         // Intentional LinkedList; fast clear, no allocations upfront,
         // will most often only carry a small number of items.
         this.variableChanges = new LinkedList<>();
     }
 
-    private VariableChangeRecordingScoreDirector(InnerScoreDirector<Solution_, ?> delegate,
-            List<ChangeAction<Solution_>> variableChanges, Map<Object, Integer> cache, boolean triggerDelegateActions) {
-        this.delegate = delegate;
+    private VariableChangeRecordingScoreDirector(InnerScoreDirector<Solution_, ?> backingScoreDirector,
+            List<ChangeAction<Solution_>> variableChanges, Map<Object, Integer> cache) {
+        this.backingScoreDirector = backingScoreDirector;
         this.variableChanges = variableChanges;
         this.cache = cache;
-        this.triggerDelegateActions = triggerDelegateActions;
     }
 
     @Override
@@ -75,9 +74,9 @@ public final class VariableChangeRecordingScoreDirector<Solution_>
         var listIterator = variableChanges.listIterator(changeCount);
         while (listIterator.hasPrevious()) { // Iterate in reverse.
             var changeAction = listIterator.previous();
-            changeAction.undo(delegate);
+            changeAction.undo(backingScoreDirector);
         }
-        delegate.triggerVariableListeners();
+        Objects.requireNonNull(backingScoreDirector).triggerVariableListeners();
         variableChanges.clear();
         if (cache != null) {
             cache.clear();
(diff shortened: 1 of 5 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhase.java
```diff
@@ -28,4 +40,26 @@ public String getPhaseTypeString() {
...
+        if (elementsToRuinSet != null) {
+            var listVariableDescriptor = stepScope.getPhaseScope().getSolverScope().getSolutionDescriptor()
+                    .getListVariableDescriptor();
+            var entity = stepScope.getStep().extractPlanningEntities().iterator().next();
+            if (!elementsToRuinSet.contains(entity)) {
+                // Sometimes, the list of elements to be ruined does not include new destinations select by the CH.
+                // In these cases,
+                // we need to record the element list before making any move changes
+                // so that it can be referenced
+                // to restore the solution to its original state in the inner nodes.
```
triceo, line 54:
```suggestion
                // Sometimes, the list of elements to be ruined does not include new destinations selected by the CH.
                // In these cases, we need to record the element list before making any move changes
                // so that it can be referenced to restore the solution to its original state in the inner nodes.
```

triceo: What nodes are we talking about here? Perhaps a different word is in order? "Node" is a deep CS terminology.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseBuilder.java
```diff
@@ -24,29 +26,73 @@ public static <Solution_> RuinRecreateConstructionHeuristicPhaseBuilder<Solution
             HeuristicConfigPolicy<Solution_> solverConfigPolicy, ConstructionHeuristicPhaseConfig constructionHeuristicConfig) {
         var constructionHeuristicPhaseFactory =
                 new RuinRecreateConstructionHeuristicPhaseFactory<Solution_>(constructionHeuristicConfig);
-        return (RuinRecreateConstructionHeuristicPhaseBuilder<Solution_>) constructionHeuristicPhaseFactory.getBuilder(0, false,
+        var builder = (RuinRecreateConstructionHeuristicPhaseBuilder<Solution_>) constructionHeuristicPhaseFactory.getBuilder(0,
+                false,
                 solverConfigPolicy, TerminationFactory.<Solution_> create(new TerminationConfig())
                         .buildTermination(solverConfigPolicy));
+        if (solverConfigPolicy.getMoveThreadCount() != null && solverConfigPolicy.getMoveThreadCount() >= 1) {
+            builder.multiThread = true;
```
triceo, line 34: `multithreaded`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseBuilder.java
```diff
@@ -24,29 +26,73 @@ public static <Solution_> RuinRecreateConstructionHeuristicPhaseBuilder<Solution
...
         this.elementsToRecreate = elements;
         return this;
     }
 
+    public RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> withElementsToRuin(Set<Object> elements) {
+        this.elementsToRuin = elements;
+        return this;
+    }
+
+    public RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> withCopyEntityPlacer(boolean multiThread) {
```
triceo, line 84: Shouldn't this be called `withMultithreaded`?

triceo (commented):
So what was the problem after all? I only see the changes we talked about - yet that version didn't work yet. So, what changed?

Let's discuss the custom actions on Slack.

Left a few more comments, but otherwise LGTM.

zepfred:
> So what was the problem after all? I only see the changes we talked about - yet that version didn't work yet. So, what changed?

`constructionHeuristicPhaseBuilder` is shared among all threads. Therefore, every time a thread builds the R&R phase, it updates the reference `elementsToRecreate`, which filters out non-ruined elements. Additionally, the entity placer and decider are both shared. When a main thread executes a chosen move, it utilizes the same decider, leading to the CH phase in move threads ending prematurely and not reassigning the ruined elements.

In short, `constructionHeuristicPhaseBuilder` was causing race conditions, and `constructionHeuristicPhaseBuilder#ensureThreadSafe()` fixes that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhase.java
zepfred: I meant the score director restoring the original state.

core/src/main/java/ai/timefold/solver/core/impl/move/director/VariableChangeRecordingScoreDirector.java
```diff
@@ -143,10 +173,23 @@ public SolutionDescriptor<Solution_> getSolutionDescriptor() {
         return delegate.getSolutionDescriptor();
     }
 
+    /**
+     * Returns the score director to which events are delegated.
+     */
     public InnerScoreDirector<Solution_, ?> getDelegate() {
         return delegate;
     }
```
triceo, line 181:
```suggestion
    public InnerScoreDirector<Solution_, ?> getBacking() {
        return backingScoreDirector;
    }
```

core/src/main/java/ai/timefold/solver/core/impl/move/director/VariableChangeRecordingScoreDirector.java
```diff
@@ -143,10 +173,23 @@ public SolutionDescriptor<Solution_> getSolutionDescriptor() {
...
 
+    /**
+     * The {@code VariableChangeRecordingScoreDirector} score director includes two main tasks:
+     * tracking any variable change and firing events to a delegated score director.
+     * This method returns a copy of the score director
+     * that only tracks variable changes without firing any delegated score director events.
+     */
+    public VariableChangeRecordingScoreDirector<Solution_> getNonDelegating() {
+        return new VariableChangeRecordingScoreDirector<>(delegate, variableChanges, cache, false);
```
triceo, line 190: For safety, shouldn't `delegate` be `null` here instead?
That way, you can possibly also avoid the boolean - if the delegate is null, do not delegate.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseBuilderTest.java
```diff
@@ -0,0 +1,48 @@
...
+                        InitializingScoreTrendLevel.ANY, InitializingScoreTrendLevel.ANY, InitializingScoreTrendLevel.ANY }))
+                .build();
+        var constructionHeuristicConfig = mock(ConstructionHeuristicPhaseConfig.class);
+        var builder = RuinRecreateConstructionHeuristicPhaseBuilder.create(solverConfigPolicy, constructionHeuristicConfig);
+        var phase = builder.build();
+        assertThat(phase.getEntityPlacer()).isSameAs(builder.getEntityPlacer());
+    }
+
+    @Test
+    void buildMultiThread() {
```
triceo, line 32:
```suggestion
    void buildMultiThreaded() {
```

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateConstructionHeuristicPhaseBuilderTest.java
```diff
@@ -0,0 +1,48 @@
...
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.impl.score.trend.InitializingScoreTrend;
+import ai.timefold.solver.core.impl.testdata.domain.TestdataSolution;
+
+import org.junit.jupiter.api.Test;
+
+class RuinRecreateConstructionHeuristicPhaseBuilderTest {
+
+    @Test
+    void buildSingleThread() {
```
triceo, line 19:
```suggestion
    void buildSingleThreaded() {
```

triceo (approved):
LGTM after comments resolved and CI green.

PR #1324, opened 2025-01-15
perf: improve performance of Bavet indexing

triceo:
Refactors indexing in Bavet so that:

- IndexKeys (formerly IndexProperties) become bimorphic as opposed to megamorphic, enabling all kinds of JVM optimizations.
- Many instances of IndexKeys are no longer created as they were unnecessary. (Case in point: SingleIndexProperties is gone, without replacement.)
- There is less indirection in the nodes.
- Naming of keys, functions and mappings is standardized across the board.

In the score director benchmarks, these changes show throughput improvements of up to 10 %, with GC pressure reduced by up to 40 %.


triceo:
Please review after we've done with the release, @zepfred.
You have already seen most of these changes before, in the commit that was originally reverted.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/ComparisonIndexer.java
```diff
@@ -14,19 +14,37 @@
...
+    private final KeyRetriever<Key_> keyRetriever;
     private final Supplier<Indexer<T>> downstreamIndexerSupplier;
     private final Comparator<Key_> keyComparator;
     private final boolean hasOrEquals;
     private final NavigableMap<Key_, Indexer<T>> comparisonMap;
 
-    public ComparisonIndexer(JoinerType comparisonJoinerType, Supplier<Indexer<T>> downstreamIndexerSupplier) {
-        this(comparisonJoinerType, 0, downstreamIndexerSupplier);
+    /**
+     * Construct an {@link ComparisonIndexer} which immediately ends in a {@link NoneIndexer}.
```
zepfred, line 24: The name `NoneIndexer` seems off to me. Maybe should choose a more meaningful name, such as `RawDataIndexer` or something similar.

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/index/ComparisonIndexer.java
triceo: I agree it looks weird, but it matches with `NoneJoiner`. In the absence of better names that would work for both, I'll leave it as is. It's a private API anyway - if a better idea comes along, we can change it any time we like.

PR #1332, opened 2025-01-16
chore: make the pull request CI consistent with the release CI

zepfred:


.github/workflows/pull_request.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           cache: 'maven'
 
       - name: Build and test timefold-solver
-        run: mvn -B verify
+        run: mvn -Dfull -B verify
```
triceo, line 42: The reason we don't do this is because it makes the jobs considerably longer.
Arguably, we want to do this once, not in every combination.

PR #1335, opened 2025-01-19
perf: further improve Bavet join performance

triceo:


triceo:
Although this reduces memory use massively, it also - in one benchmark - reduces performance by 50 %. There are no obvious culprits for that reduction, and no more time to experiment. Closing.

PR #1341, opened 2025-01-20
perf: remove certain polymorphic call sites

triceo:
- Slightly improves performance.
- Makes profiling easier by replacing calls to lambda instances by actual methods.


core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/DynamicPropagationQueue.java
```diff
@@ -19,15 +19,16 @@
 final class DynamicPropagationQueue<Tuple_ extends AbstractTuple, Carrier_ extends AbstractPropagationMetadataCarrier<Tuple_>>
         implements PropagationQueue<Carrier_> {
 
+    private static final int INITIAL_CAPACITY = 1000; // Selected arbitrarily.
```
zepfred, line 22: Would it make sense to use the entity count to estimate capacity?

zepfred (approved):
The changes look good!

I wonder about the benefit of replacing well-defined child classes with static build methods like the ones from `TupleLifecycle`. IMO, the child classes provide a more elegant solution.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/DynamicPropagationQueue.java
triceo: Not really. The tuples could be anything - in case of a groupBy, these will be the groups and therefore the number will likely be very low. In case of a join, the tuples will the the out tuples, and there could be thousands, tens of thousands, even more. IMO there is no good number here.

zepfred:
@triceo, please take a look at the Sonar rating.

PR #1345, opened 2025-01-22
chore: move Bavet implementation to a different package

triceo:
With Move Streams, Bavet will not just be used by Constraint Streams.

This PR only moves things between packages.
It touches no public API and has no impact on the user.


Christopher-Chianelli (approved):
I would personally prefer for ConstraintCollector implementations (which do not use the details of the Bavet network) to stay in the `ai.timefold.solver.core.impl.score.stream.collector` package (which maps nicely from `ai.timefold.solver.core.api.score.stream`, which contains ConstraintCollectors (and is their access point). However, this is not a block for me. LGTM.
`

PR #1349, opened 2025-01-23
Move Streams Work In Progress

triceo:


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
```diff
@@ -0,0 +1,222 @@
...
+    @SuppressWarnings("unchecked")
+    public <Tuple_ extends AbstractTuple> TupleLifecycle<Tuple_>
+            getAggregatedTupleLifecycle(List<? extends Stream_> streamList) {
+        var tupleLifecycles = streamList.stream()
+                .filter(this::isStreamActive)
+                .map(s -> getTupleLifecycle(s, tupleLifecycleMap))
+                .toArray(TupleLifecycle[]::new);
+        return switch (tupleLifecycles.length) {
+            case 0 ->
+                throw new IllegalStateException("Impossible state: None of the streamList (" + streamList + ") are active.");
```
zepfred, line 86: Please apply the previous comment to all similar occurrences

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
```diff
@@ -0,0 +1,222 @@
...
+                throw new IllegalStateException("Impossible state: the tupleSourceStream (" + k
+                        + ") is reserving a store after it has been extracted.");
+            } else {
+                return index + 1;
+            }
+        });
+    }
+
+    public int extractTupleStoreSize(Stream_ tupleSourceStream) {
+        Integer lastIndex = storeIndexMap.put(tupleSourceStream, Integer.MIN_VALUE);
```
zepfred, line 116: Should it be zero instead of Integer.MIN_VALUE?

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachExcludingUnassignedUniNode.java
```diff
@@ -26,22 +26,23 @@ public void insert(A a) {
...
+            updateExisting(a, tuple);
+        } else { // Tuple no longer passes the filter.
+            retract(a);
         }
     }
 
     @Override
     public void retract(A a) {
-        if (!filter.test(a)) { // The tuple was never inserted because it did not pass the filter.
+        var tuple = tupleMap.remove(a);
```
zepfred, line 41: Why we don't need to filter before removing?

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachExcludingUnassignedUniNode.java
```diff
@@ -26,22 +26,23 @@ public void insert(A a) {
...
+        if (tuple == null) { // The tuple was never inserted because it did not pass the filter.
```
zepfred, line 42: The comment is inconsistent

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachUniNode.java
```diff
@@ -3,20 +3,21 @@
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.UniTuple;
 
-public final class ForEachIncludingUnassignedUniNode<A> extends AbstractForEachUniNode<A> {
+public sealed class ForEachUniNode<A>
```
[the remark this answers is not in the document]
zepfred: The original name indicates that we will iterate through all nodes, including those that are unassigned. The new name does not.

core/src/main/java/ai/timefold/solver/core/impl/bavet/NodeNetwork.java
```diff
@@ -29,23 +30,27 @@ public int layerCount() {
...
                 .stream()
                 .filter(entry -> entry.getKey().isAssignableFrom(factClass))
                 .map(Map.Entry::getValue)
-                .flatMap(List::stream)
-                .toArray(AbstractForEachUniNode[]::new);
+                .flatMap(List::stream);
     }
 
-    public void propagate() {
+    public void settle() {
```
zepfred, line 47: Propagate seems to be a common term used in CS. I'm unsure if the method name change is good idea.

core/src/main/java/ai/timefold/solver/core/impl/move/director/MoveDirector.java
```diff
@@ -134,4 +166,8 @@ public VariableDescriptorAwareScoreDirector<Solution_> getScoreDirector() {
         return scoreDirector;
     }
 
+    public void resetWorkingSolution(Solution_ workingSolution) {
```
zepfred, line 169: Please add a comment about why we are adding an empty method

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/common/bridge/AftBridgeUniDataStream.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Objects;
+
+import ai.timefold.solver.core.impl.bavet.common.TupleSource;
+import ai.timefold.solver.core.impl.move.streams.dataset.AbstractDataStream;
+import ai.timefold.solver.core.impl.move.streams.dataset.AbstractUniDataStream;
+import ai.timefold.solver.core.impl.move.streams.dataset.DefaultDataStreamFactory;
+import ai.timefold.solver.core.impl.move.streams.dataset.common.DataNodeBuildHelper;
+
+public final class AftBridgeUniDataStream<Solution_, A>
```
zepfred, line 11: What is the meaning of Aft?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/common/bridge/AftBridgeUniDataStream.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        AftBridgeUniDataStream<?, ?> that = (AftBridgeUniDataStream<?, ?>) o;
```
zepfred, line 32:
```suggestion
        var that = (AftBridgeUniDataStream<?, ?>) o;
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DatasetSession.java
```diff
@@ -0,0 +1,76 @@
...
+        if (effectiveClassToFromSolutionNodeArray == null) {
+            return;
+        }
+        for (var node : effectiveClassToFromSolutionNodeArray) {
+            node.read(solutionView, solution);
+        }
+    }
+
+    @Override
+    public void settle() {
```
zepfred, line 72: Why are we overriding it?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
triceo: Actually, no. This just indicates "no value".
Also, this code did not change from what was there before.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachExcludingUnassignedUniNode.java
triceo: Performance optimization.
Removing from a map will return null if the thing was never inserted; therefore I can avoid calling the filter, and still get the same result.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachExcludingUnassignedUniNode.java
triceo: It is consistent. The tuple didn't pass the filter during insertion, and therefore it was not inserted.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachUniNode.java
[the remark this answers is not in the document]
triceo: You're right that CS is lacking comments around nodes etc. It is not practical though, nor the point of this PR, to fix CS now. Documenting of, and education about, CS is something we definitely need to do at some point though.

core/src/main/java/ai/timefold/solver/core/impl/bavet/NodeNetwork.java
triceo: The term was actually conflating two different things:
- Propagation, which is what happens to tuples in nodes.
- Settling, which is a process in which a series of nodes have their propagations triggered.

The current situation is clearer now. Once all nodes have propagated, the node network is considered settled.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/common/bridge/AftBridgeUniDataStream.java
triceo: "Aft" and "fore" are english words, opposites of each other. "Fore" means in front of something, and "aft" means behind something.

triceo:
I have changed exception formatting wherever practical.
It is not in scope of this PR to refactor every class that had to be touched.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DatasetSessionFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public DatasetSessionFactory(DefaultDataStreamFactory<Solution_> dataStreamFactory) {
+        this.dataStreamFactory = dataStreamFactory;
+    }
+
+    public DatasetSession<Solution_> buildSession() {
+        var activeDataStreamSet = new LinkedHashSet<AbstractDataStream<Solution_>>();
+        var datasets = dataStreamFactory.getDatasets();
+        for (var dataset : datasets) {
+            dataset.collectActiveDataStreams(activeDataStreamSet);
```
zepfred, line 27: I wonder if we should change the logic to return the list of active datasets. This way, we wouldn't need to pass activeDatasetStreamSet as a parameter.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DefaultDataStreamFactory.java
```diff
@@ -0,0 +1,147 @@
...
+         * 3/ FromType super FactType; querying for interface where declared type is its implementation.
+         */
+        boolean hasMatchingType = problemFactOrEntityClassSet.stream()
+                .anyMatch(factType -> fromType.isAssignableFrom(factType) || factType.isAssignableFrom(fromType));
+        if (!hasMatchingType) {
+            List<String> canonicalClassNameList = problemFactOrEntityClassSet.stream()
+                    .map(Class::getCanonicalName)
+                    .sorted()
+                    .toList();
+            throw new IllegalArgumentException("Cannot use class (" + fromType.getCanonicalName()
```
zepfred, line 80: Let's use `String::format`

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/FilterUniDataStream.java
```diff
@@ -0,0 +1,52 @@
...
+
+final class FilterUniDataStream<Solution_, A>
+        extends AbstractUniDataStream<Solution_, A> {
+
+    private final Predicate<A> predicate;
+
+    public FilterUniDataStream(DefaultDataStreamFactory<Solution_> dataStreamFactory,
+            AbstractUniDataStream<Solution_, A> parent, Predicate<A> predicate) {
+        super(dataStreamFactory, parent);
+        this.predicate = predicate;
```
zepfred, line 18:
```suggestion
        this.predicate = Objects.requireNonNull(predicate, "The predicate (null) cannot be null.");
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/ForEachDataStream.java
```diff
@@ -0,0 +1,115 @@
...
+    private final Collection<A> source;
+
+    public ForEachDataStream(DefaultDataStreamFactory<Solution_> dataStreamFactory, Class<A> forEachClass) {
+        this(dataStreamFactory, forEachClass, (Predicate<A>) null);
+    }
+
+    public ForEachDataStream(DefaultDataStreamFactory<Solution_> dataStreamFactory, Class<A> forEachClass,
+            Predicate<A> filter) {
+        super(dataStreamFactory, null);
+        this.forEachClass = forEachClass;
```
zepfred, line 35: We are using `Objects::requireNonNull` in some places and in others not.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/IfExistsUniDataStream.java
```diff
@@ -0,0 +1,112 @@
...
+        parentA.collectActiveDataStreams(dataStreamSet);
+        parentBridgeB.collectActiveDataStreams(dataStreamSet);
+        dataStreamSet.add(this);
+    }
+
+    @Override
+    public void buildNode(DataNodeBuildHelper<Solution_> buildHelper) {
+        TupleLifecycle<UniTuple<A>> downstream = buildHelper.getAggregatedTupleLifecycle(childStreamList);
+        var indexerFactory = new IndexerFactory<>(joiner);
+        var node = indexerFactory.hasJoiners()
```
zepfred: I would use an if block instead of a ternary operator

core/src/main/java/ai/timefold/solver/core/impl/move/streams/generic/ListChangeMove.java
```diff
@@ -0,0 +1,178 @@
...
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.api.domain.variable.PlanningListVariable;
+import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningListVariableMetaModel;
+import ai.timefold.solver.core.preview.api.domain.metamodel.VariableMetaModel;
+import ai.timefold.solver.core.preview.api.move.MutableSolutionView;
+import ai.timefold.solver.core.preview.api.move.Rebaser;
+
+import org.jspecify.annotations.NonNull;
+
+/**
```
zepfred, line 17: The class is very well-documented. I missed these types of comments in the other move classes and data stream classes.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/BavetConcatConstraintStream.java
```diff
@@ -0,0 +1,8 @@
+package ai.timefold.solver.core.impl.score.stream.bavet.common;
```
zepfred, line 1: I wonder if these stream classes should be part of the package ` ai.timefold.solver.core.impl.bavet.common`

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -71,7 +71,8 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
      * {@link BavetAbstractConstraintStream} implement equals/hashcode ignoring child streams.
      * <p>
      * {@link BavetConstraintSessionFactory#buildSession(Object, ConstraintMatchPolicy, boolean, Consumer)} needs this to happen
-     * for all streams.
+     * for all
```
zepfred, line 74: Please adjust the comment

core/src/test/java/ai/timefold/solver/core/impl/move/streams/dataset/UniDatasetStreamTest.java
```diff
@@ -0,0 +1,107 @@
...
+import ai.timefold.solver.core.preview.api.move.SolutionView;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+import org.junit.jupiter.api.Test;
+
+class UniDatasetStreamTest {
+
+    @Test
+    void forEach() {
```
zepfred, line 21: Nice!

zepfred (approved):
LGTM!

I understand that this is still in the early stages, but I want to bring this to your attention. The new stream classes and nearly all the move classes lack comments that would help explain the code in the future once Move Streams is completed.

triceo:
> I understand that this is still in the early stages, but I want to bring this to your attention. The new stream classes and nearly all the move classes lack comments that would help explain the code in the future once Move Streams is completed.

I see the same issue. At the moment, it is not practical to document these surfaces, as all are likely to change a thousand times before the feature is final. But when we are getting ready to publish the feature, I will write all sorts of docs.

Also, compared to constraint streams, I won't be the only one working on this. Once the groundwork is laid, we'll each implement some of the required features. The situation of constraint streams, where I'm still pretty much the only person who understands all of it, will not repeat here - that was a unique circumstance, and we have ways of avoiding it now.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DatasetSessionFactory.java
triceo: It is a sort of a visitor pattern. Personally, I don't mind, and making this change would introduce yet another completely unrelated change to this already large PR. 

It's been like this forever, I'm keeping it.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/BavetConcatConstraintStream.java
triceo: I'm not making it too generic - when we need concat in data streams, we'll do it. I doubt we will need it there.

core/src/test/java/ai/timefold/solver/core/impl/move/streams/dataset/UniDatasetStreamTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        var solutionDescriptor = dataStreamFactory.getSolutionDescriptor();
+        solutionDescriptor.visitAll(solution, datasetSession::insert);
+
+        datasetSession.settle();
+        return datasetSession;
+    }
+
+    @NullMarked
+    private record TestdataListSolutionView(TestdataListSolution solution) implements SolutionView<TestdataListSolution> {
```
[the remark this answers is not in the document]
triceo: Totally possible. When that happens, I'll do it.
I am trying to do the minimal possible thing, not to over-build this from the start.
The code will evolve where it needs to.

PR #1352, opened 2025-01-24
feat: add Local Search reconfiguration process

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/ReconfigurableAcceptor.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        var accepted = evaluate(moveScope);
+        var improved = moveScope.getScore().compareTo(moveScope.getStepScope().getPhaseScope().getBestScore()) > 0;
+        if (improved) {
+            restartStrategy.reset(moveScope);
+        }
+        return accepted;
+    }
+
+    protected abstract boolean evaluate(LocalSearchMoveScope<Solution_> moveScope);
```
triceo, line 67: Please rename this method to describe what it will actually do.
In general, avoid words like "evaluate", "do" etc., because they do not say anything.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/ReconfigurableAcceptor.java
```diff
@@ -0,0 +1,70 @@
...
+
+    protected abstract <Score_ extends Score<Score_>> void reset(Score_ score);
```
triceo, line 69: Reset what? Why?
Arguably this method needs a better name, and/or Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+        this.clock = clock;
+        this.scalingFactor = scalingFactor;
+    }
+
+    @Override
+    public void phaseStarted(LocalSearchPhaseScope<Solution_> phaseScope) {
+        restartTriggered = false;
+        if (gracePeriodMillis == 0) {
+            // 10 seconds of grace period
+            gracePeriodMillis = clock.millis() + 10_000;
```
triceo, line 43: The [Javadoc of Clock](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html#millis--) recommends to not use `millis()`. Arguably, it'd be simpler if this code operated on `Instant` anyway - `long` brings no semantics, `Instant` does.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+
+    protected boolean isGracePeriodFinished() {
+        return gracePeriodFinished;
+    }
+
+    protected void disableTriggerFlag() {
+        restartTriggered = false;
+    }
+
+    abstract boolean process(LocalSearchMoveScope<Solution_> moveScope);
```
triceo, line 87: `process` is another name that doesn't actually describe anything specific.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/UnimprovedMoveCountRestartStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public boolean process(LocalSearchMoveScope<Solution_> moveScope) {
+        var currentMoveCount = moveScope.getStepScope().getPhaseScope().getSolverScope().getMoveEvaluationCount();
+        if (lastImprovementMoveCount == 0) {
+            lastImprovementMoveCount = currentMoveCount;
+            return false;
+        }
+        if (currentMoveCount - lastImprovementMoveCount >= nextRestart) {
+            logger.debug("Restart triggered with geometric factor {}, scaling factor of {}, move count ({})",
```
triceo, line 62: Let's make this `trace`.

The solver's logging is designed poorly - it uses various solving levels to map to how "deep" you are in the solving process. INFO is phases, DEBUG is steps, TRACE is moves. This is unfortunate, but we can not really change this.

Let's put these things on TRACE, which is the lowest possible level - if people enable this one, they will get everything.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/RestartStrategy.java
```diff
@@ -0,0 +1,11 @@
+package ai.timefold.solver.core.impl.localsearch.decider.acceptor.restart;
+
+import ai.timefold.solver.core.impl.localsearch.event.LocalSearchPhaseLifecycleListener;
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchMoveScope;
+
+public interface RestartStrategy<Solution_> extends LocalSearchPhaseLifecycleListener<Solution_> {
```
triceo, line 6: Needs to be Javadoc'd.
Also arguably better naming - what does it mean to be "triggered"? What does "reset" do?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+    protected double currentGeometricGrowFactor;
+
+    protected AbstractGeometricRestartStrategy(Clock clock, double scalingFactor) {
+        this.clock = clock;
+        this.scalingFactor = scalingFactor;
+    }
+
+    @Override
+    public void phaseStarted(LocalSearchPhaseScope<Solution_> phaseScope) {
+        restartTriggered = false;
```
triceo, line 40: Arguably you want to call the method you have designed for this.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+            }
+        }
+        return restartTriggered;
+    }
+
+    protected boolean isGracePeriodFinished() {
+        return gracePeriodFinished;
+    }
+
+    protected void disableTriggerFlag() {
```
triceo, line 83: Why would I want to disable this?

Looks like a very low-level method is being exposed. Aren't we exposing implementation here?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
triceo: Also, the magic constant should become an actual constant, and be Javadoc'd. Why is the value what it is?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/ReconfigurationStrategy.java
```diff
@@ -0,0 +1,16 @@
...
+import ai.timefold.solver.core.impl.phase.scope.AbstractStepScope;
+
+public sealed interface ReconfigurationStrategy<Solution_> extends PhaseLifecycleListener<Solution_>
+        permits NoOpReconfigurationStrategy, RestoreBestSolutionReconfigurationStrategy {
+
+    <Score_ extends Score<Score_>> Score_ apply(AbstractStepScope<Solution_> stepScope);
+
+    default boolean isTriggered(AbstractStepScope<Solution_> stepScope) {
+        return stepScope.getPhaseScope().isReconfigurationTriggered();
+    }
```
triceo, line 14: Arguably we need more descriptive names, and/or Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestoreBestSolutionReconfigurationStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+    public RestoreBestSolutionReconfigurationStrategy(MoveSelector<Solution_> moveSelector, Acceptor<Solution_> acceptor) {
+        this.moveSelector = moveSelector;
+        this.acceptor = acceptor;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <Score_ extends Score<Score_>> Score_ apply(AbstractStepScope<Solution_> stepScope) {
+        var solverScope = stepScope.getPhaseScope().getSolverScope();
+        logger.debug("Resetting working solution, score ({})", solverScope.getBestScore());
```
triceo, line 31: Trace.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestoreBestSolutionReconfigurationStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+        solverScope.setWorkingSolutionFromBestSolution();
+        // Changing the working solution requires reinitializing the move selector and acceptor
+        // 1 - The move selector will reset all cached lists using old solution entity references
+        moveSelector.phaseStarted(stepScope.getPhaseScope());
+        // 2 - The acceptor will restart its search from the updated working solution (last best solution)
+        acceptor.phaseStarted((LocalSearchPhaseScope<Solution_>) stepScope.getPhaseScope());
```
triceo, line 37: Looks like this code should be shared with something somewhere else. (Sorry for not being more specific, I don't remember 100 % of our codebase.)

If something _has to happen_ when we are restoring the working solution, then arguably every piece of code that restores the working solution should share this code. This is how you prevent bugs.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestoreBestSolutionReconfigurationStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void stepEnded(AbstractStepScope<Solution_> stepScope) {
+        // Do nothing
+    }
+
+    @Override
+    public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        Objects.requireNonNull(moveSelector);
+        Objects.requireNonNull(acceptor);
```
triceo, line 56: Fail earlier.

core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AcceptorFactoryTest.java
```diff
@@ -1,12 +1,8 @@
 package ai.timefold.solver.core.impl.localsearch.decider.acceptor;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
-import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
+import static org.assertj.core.api.Assertions.*;
```
triceo, line 3: Please configure your IDE never to use asterisk imports.

triceo (commented):
First round of review is in.

My most general comment is: we need a different name for "reconfiguration". I know that the papers use this word, but we cannot. We already have "configuration", and using "reconfiguration" would be confusing - the "configuration" in "reconfiguration" has nothing to do with solver config.

Reheating? Restart? Refresh? Let's discuss.

triceo:
Another concern: we only support LA (and DLAS). But shouldn't we support simulated annealing too? In SA, the process is actually called reheating, and is well researched as far as I know.

For the record: I don't much care for SA, because it is hard to configure and therefore not much useful. But if we don't support reheating the SA, it should be a conscious decision, and not just something we forgot to do.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchMoveScope;
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchPhaseScope;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Restart strategy, which exponentially increases the metric that triggers the restart process.
+ * The first restart occurs after the {@code grace period + 1 * scalingFactor} metric.
```
triceo, line 14: I wonder about the use of the term "grace period" here. We already use it in the new termination.

On one hand, it is the same thing, so it should use the same name.
On the other hand, it is used in a different context, so maybe it shouldn't?
I'm abvialent.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final double GEOMETRIC_FACTOR = 1.4; // Value extracted from the cited paper
+    protected final Clock clock;
+    protected final Logger logger = LoggerFactory.getLogger(AbstractGeometricRestartStrategy.class);
+    protected final double scalingFactor;
+
+    private boolean gracePeriodFinished;
+    private boolean restartTriggered;
+    private long gracePeriodMillis;
+    protected long nextRestart;
+    protected double currentGeometricGrowFactor;
```
triceo, line 31: grow_th_Factor?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public boolean isTriggered(LocalSearchMoveScope<Solution_> moveScope) {
+        if (!restartTriggered && (gracePeriodFinished || clock.millis() >= gracePeriodMillis)) {
+            gracePeriodFinished = true;
+            var triggered = process(moveScope);
+            if (triggered) {
+                currentGeometricGrowFactor = Math.ceil(currentGeometricGrowFactor * GEOMETRIC_FACTOR);
+                nextRestart = (long) Math.ceil(currentGeometricGrowFactor * scalingFactor);
```
triceo, line 72: Reuse this computation - it is important enough not to be duplicated.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
triceo: Also, I don't particularly like that we are exposing mutable fields to implementations.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/LocalSearchDecider.java
```diff
@@ -94,6 +102,15 @@ public void decideNextStep(LocalSearchStepScope<Solution_> stepScope) {
         InnerScoreDirector<Solution_, ?> scoreDirector = stepScope.getScoreDirector();
         scoreDirector.setAllChangesWillBeUndoneBeforeStepEnds(true);
         int moveIndex = 0;
+        if (reconfigurationStrategy.isTriggered(stepScope)) {
```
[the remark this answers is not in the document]
zepfred: Yes, I confirmed that moving it to the `stepEnd` event works.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/ReconfigurableAcceptor.java
zepfred: I realized we don't need this method.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
zepfred: The goal is to avoid restarting too early, and I used the term to maintain consistence with the new termination.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricRestartStrategy.java
zepfred: I'm unsure about this suggestion. Can you please elaborate on that?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestoreBestSolutionReconfigurationStrategy.java
zepfred: The logic has been moved to `LocalSearchDecider`.

core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/UnimprovedMoveCountStuckCriterionTest.java
```diff
@@ -15,106 +17,108 @@
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
-class UnimprovedMoveCountRestartStrategyTest {
+class UnimprovedMoveCountStuckCriterionTest {
 
     @Test
-    void isTriggered() {
+    void isSolverStuck() {
         var clock = mock(Clock.class);
```
triceo, line 24: Instead of mocking the Clock, you can actually get a new instance which you can control to do literally anything you like.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/RestartableAcceptor.java
```diff
@@ -11,60 +10,54 @@
...
-public abstract class ReconfigurableAcceptor<Solution_> extends AbstractAcceptor<Solution_> {
+public abstract class RestartableAcceptor<Solution_> extends AbstractAcceptor<Solution_> {
 
-    private final RestartStrategy<Solution_> restartStrategy;
+    private final StuckCriterion<Solution_> stuckCriterionDetection;
+    protected boolean restartTriggered;
 
-    protected ReconfigurableAcceptor(RestartStrategy<Solution_> restartStrategy) {
-        this.restartStrategy = restartStrategy;
+    protected RestartableAcceptor(StuckCriterion<Solution_> stuckCriterionDetection) {
```
triceo, line 18: IMO it's enough to call the variable `stuckCriterion`.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/lateacceptance/DiversifiedLateAcceptanceAcceptor.java
```diff
@@ -53,7 +53,7 @@ private void validate() {
 
     @Override
     @SuppressWarnings({ "rawtypes", "unchecked" })
-    protected boolean evaluate(LocalSearchMoveScope<Solution_> moveScope) {
+    protected boolean applyAcceptanceCriteria(LocalSearchMoveScope<Solution_> moveScope) {
```
triceo, line 56: IMO since this is an acceptor, a simple `accept()` would suffice.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricStuckCriterion.java
```diff
@@ -18,29 +18,28 @@
...
  * @param <Solution_> the solution type
  */
-public abstract class AbstractGeometricRestartStrategy<Solution_> implements RestartStrategy<Solution_> {
+public abstract class AbstractGeometricStuckCriterion<Solution_> implements StuckCriterion<Solution_> {
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractGeometricStuckCriterion.class);
     private static final double GEOMETRIC_FACTOR = 1.4; // Value extracted from the cited paper
-    protected final Clock clock;
-    protected final Logger logger = LoggerFactory.getLogger(AbstractGeometricRestartStrategy.class);
-    protected final double scalingFactor;
+    private static final long GRACE_PERIOD_MILLIS = 30_000; // Same value as DiminishedReturnsTermination
```
triceo, line 24: Questionable. If the diminished termination is enabled, don't we want the solver to be able to restart before the termination triggers? Then arguably this should be _less_ than whatever the termination is configured to be. But I'm not sure if we even want that.

Should we detect the termination in the solver config? Should we depend on whatever values the user has configured there? I'm not sure either.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricStuckCriterion.java
```diff
@@ -18,29 +18,28 @@
...
     }
 
     @Override
     public void phaseStarted(LocalSearchPhaseScope<Solution_> phaseScope) {
-        restartTriggered = false;
         if (gracePeriodMillis == 0) {
-            // 10 seconds of grace period
-            gracePeriodMillis = clock.millis() + 10_000;
+            // 30 seconds of grace period
+            gracePeriodMillis = clock.instant().toEpochMilli() + GRACE_PERIOD_MILLIS;
```
triceo, line 42: My earlier comment to not use `clock.millis()` was intending to suggest that `gracePeriod` should be an `Instant` (so something like `gracePeriodEnd`), and you can use `Instant.plusMillis(...)` to modify it.

`Instant` has semantics, `long` does not.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/StuckCriterion.java
```diff
@@ -0,0 +1,22 @@
+package ai.timefold.solver.core.impl.localsearch.decider.acceptor.restart;
+
+import ai.timefold.solver.core.impl.localsearch.event.LocalSearchPhaseLifecycleListener;
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchMoveScope;
+
+/**
+ * Base contract
+ * for defining strategies that identify when the {@link ai.timefold.solver.core.api.solver.Solver solver} is stuck.
```
triceo, line 8: Remove the FQN, shorten the comment to a one-liner. :-)

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/StuckCriterion.java
```diff
@@ -0,0 +1,22 @@
...
+ * 
+ * @param <Solution_>
+ */
+public interface StuckCriterion<Solution_> extends LocalSearchPhaseLifecycleListener<Solution_> {
+
+    /**
+     * Main logic that applies a specific metric to determine if a solver is stuck and cannot find better solutions
+     * in the current solving flow.
```
triceo, line 16:
```suggestion
     * Main logic that applies a specific metric to determine if a solver is stuck in a local optimum.
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/UnimprovedMoveCountStuckCriterion.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Criterion based on the unimproved move count.
+ * It exponentially increases the unimproved move count values that trigger the
+ * {@link RestartStrategy restart} process.
+ */
+public class UnimprovedMoveCountStuckCriterion<Solution_> extends AbstractGeometricStuckCriterion<Solution_> {
+
+    // Multiplier defined through experiments
+    protected static final long UNIMPROVED_MOVE_COUNT_MULTIPLIER = 300_000;
```
triceo, line 19: For ESS or FSR, this will be equivalent to 30-60 seconds.
For smaller problems though, this may be barely a few seconds.
IMO this highlights the need for some sort of a relative criterion.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/AdaptedSolverScope.java
```diff
@@ -0,0 +1,317 @@
...
+import ai.timefold.solver.core.impl.score.definition.ScoreDefinition;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.impl.solver.AbstractSolver;
+import ai.timefold.solver.core.impl.solver.change.DefaultProblemChangeDirector;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+import ai.timefold.solver.core.impl.solver.thread.ChildThreadType;
+
+import io.micrometer.core.instrument.Tags;
+
+public class AdaptedSolverScope<Solution_> extends SolverScope<Solution_> {
```
triceo, line 25: This is _a lot_ of code just to pass in one extra parameter. It also creates uncertainty as to what type exactly is the solver scope, and who set that type there. 

Since solver scope is mutable and is already a "god class" used for storing whatever context we have, why not make the decider just another field on the original solver scope? Arguably though, it should not be on the solver scope, because it's the `localSearchDecider` - it should be on local search's phase scope.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestartStrategy.java
```diff
@@ -0,0 +1,38 @@
+package ai.timefold.solver.core.impl.localsearch.decider.reconfiguration;
+
+import ai.timefold.solver.core.impl.localsearch.decider.acceptor.restart.StuckCriterion;
+import ai.timefold.solver.core.impl.phase.event.PhaseLifecycleListener;
+import ai.timefold.solver.core.impl.phase.scope.AbstractStepScope;
+
+/**
+ * Base contract for defining restart strategies.
+ * The restart process is initiated
+ * when the {@link ai.timefold.solver.core.impl.localsearch.decider.LocalSearchDecider decider} identifies
```
triceo, line 10: Where possible, please use imports. FQNs arguably don't provide much benefit, but they hinder readability of those comments.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/reconfiguration/RestoreBestSolutionRestartStrategy.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
+        // Do nothing
+    }
+
+    @Override
+    public void solvingStarted(SolverScope<Solution_> solverScope) {
+        var castSolverScope = (AdaptedSolverScope<Solution_>) solverScope;
```
triceo, line 50: I think if you move this to phaseStarted, you'll be able to put the decided on local search phase scope, where it belongs.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -260,16 +260,16 @@ public int getNextStepIndex() {
...
 
-    public boolean isReconfigurationTriggered() {
-        return this.reconfigurationTriggered;
+    public boolean isSolverStuck() {
+        return this.solverStuck;
     }
 
-    public void triggerReconfiguration() {
-        this.reconfigurationTriggered = true;
+    public void triggerSolverStuck() {
```
triceo, line 267: Arguably, `setSolverStuck(boolean ...)` gets rid of two methods, and it is in line with how other fields on the class are modified.

triceo (commented):
Next round of comments.

core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/UnimprovedMoveCountStuckCriterionTest.java
zepfred: I need to create a mock and ensure we can test the logic after the grace period.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricStuckCriterion.java
zepfred: I understand your viewpoint and struggled to select an appropriate default value. That's why I opted for the same value as the diminished termination.

However, I believe we should not rely on that termination at all. This should stand as an independent value.

core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/UnimprovedMoveCountStuckCriterionTest.java
triceo: But isn't `Clock` designed for that?

You can create `Clock` instances based on any `Instant` you want. You can store the clock with that instant in the class, and the class will immediately use that time value. 

Or you can configure a clock that ticks per a certain value, and have time move every 5 minutes or something like that. IMO mocking the Clock is not how the class is supposed to be used.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/restart/AbstractGeometricStuckCriterion.java
```diff
@@ -0,0 +1,107 @@
...
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchPhaseScope;
+import ai.timefold.solver.core.impl.solver.scope.SolverScope;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Restart strategy, which exponentially increases the metric that triggers the restart process.
+ * The first restart occurs after the {@code grace period + 1 * scalingFactor} metric.
+ * Following that, the metric increases exponentially: 1, 2, 3, 5, 7, 10, 14...
```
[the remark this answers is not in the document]
Christopher-Chianelli: I would write the formula down; either

`r_new = r_old * GEOMETRIC_FACTOR`

or

`r = (gracePeriod + scalingFactor) * GEOMETRIC_FACTOR^restartCount`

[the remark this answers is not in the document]
zepfred: > That sequence looks more quadratic than exponential. This is probably because the default GEOMETRIC_FACTOR is between 1 and 2 (and thus the massive growth expected of an exponential takes longer to ramp up).

Yes, you are right. The article states that better results can be achieved with values between 1 and 2. I liked the second formula.

PR #1370, opened 2025-01-31
fix: prevent possible concurrent modification of a collection

triceo:
We have seen reports of the TreeMap getting concurrently accessed and throwing exceptions. Unfortunately, we have never seen any code to reproduce these failures, as concurrency issues are notoriously hard to reproduce. Therefore we refactor the mechanism to make any such issues impossible.

For the same reason as there is no reproducer, there is also no new test coverage. But existing test coverage still passes.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/change/ProblemChange.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/change/ProblemChange.java
@@ -2,8 +2,11 @@ package ai.timefold.solver.core.api.solver.change;
 
 import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
 import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.api.domain.variable.VariableListener;
 import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.solver.Solver;
+import ai.timefold.solver.core.api.solver.event.BestSolutionChangedEvent;
+import ai.timefold.solver.core.impl.heuristic.move.Move;
 
 import org.jspecify.annotations.NonNull;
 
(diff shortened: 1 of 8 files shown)


triceo:
I have put this together with @rsynek, who is the original author of this code.
I also need a review from @zepfred, who touched this code most recently.

core/src/main/java/ai/timefold/solver/core/impl/solver/BestSolutionHolder.java
```diff
@@ -3,54 +3,104 @@
...
         }
-        SortedMap<BigInteger, List<CompletableFuture<Void>>> containedProblemChangesPerVersion =
-                problemChangesPerVersion.headMap(versionedBestSolution.getVersion().add(BigInteger.ONE));
 
-        List<CompletableFuture<Void>> containedProblemChanges = new ArrayList<>();
-        for (Map.Entry<BigInteger, List<CompletableFuture<Void>>> entry : containedProblemChangesPerVersion.entrySet()) {
-            containedProblemChanges.addAll(entry.getValue());
-            problemChangesPerVersion.remove(entry.getKey());
+        var boundaryVersion = versionedBestSolution.version().add(BigInteger.ONE);
+        if (lastProcessedVersion.getAndSet(boundaryVersion).compareTo(boundaryVersion) >= 0) {
```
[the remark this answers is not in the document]
zepfred: This also makes me think. If we don't set it in this block, where should we set it?

core/src/main/java/ai/timefold/solver/core/impl/solver/BestSolutionHolder.java
```diff
@@ -3,54 +3,104 @@
...
+                problemChangesPerVersionRef.getAndUpdate(map -> createNewProblemChangesMap(map.tailMap(boundaryVersion)));
+        // At this point, the old map is not accessible to any other thread.
+        // We also do not need to clear it, because this being the only reference, 
+        // garbage collector will do it for us.
+        var containedProblemChanges = oldProblemChangesPerVersion.headMap(boundaryVersion)
+                .values()
+                .stream()
+                .flatMap(Collection::stream)
+                .toList();
+        return new BestSolutionContainingProblemChanges<>(versionedBestSolution.bestSolution(), containedProblemChanges);
```
zepfred, line 103: Maybe we can update `lastProcessedVersion` before returning the changes.

core/src/main/java/ai/timefold/solver/core/impl/solver/BestSolutionHolder.java
```diff
@@ -61,77 +113,56 @@ BestSolutionContainingProblemChanges<Solution_> take() {
...
-            problemChangesPerVersion.clear();
-        } finally {
-            problemChangesLock.unlock();
-        }
+        // The map is an atomic reference. 
+        // We first replace the reference with a new map atomically, avoiding synchronization issues.
+        // Then we process the old map, which is safe because no one can access it anymore.
+        // We do not need to clear it, because this being the only reference,
+        // the garbage collector will do it for us.
+        problemChangesPerVersionRef.getAndSet(createNewProblemChangesMap())
```
zepfred, line 158: Nice!

zepfred (approved):
The changes have simplified the class's logic and look great to me!

core/src/main/java/ai/timefold/solver/core/api/solver/change/ProblemChange.java
```diff
@@ -2,30 +2,42 @@
...
  * <li>calculates the score and makes the {@link PlanningSolution updated working solution}
- * the new {@link PlanningSolution best solution}; note that this {@link PlanningSolution solution} is not published
- * via the {@link ai.timefold.solver.core.api.solver.event.BestSolutionChangedEvent}, as it hasn't been initialized yet</li>
+ * the new {@link PlanningSolution best solution};
+ * note that this {@link PlanningSolution solution} is not published via the {@link BestSolutionChangedEvent},
+ * as it hasn't been initialized yet</li>
  * <li>restarts solving to fill potential uninitialized {@link PlanningEntity planning entities}</li>
  * </ol>
  * <p>
+ * From the above, it follows that the solver will require some time
```
rsynek, line 31: Valuable info for users, thanks!

rsynek (approved):
I like that a part of locking has been replaced by immutability, that's nice.

PR #1391, opened 2025-02-16
 docs: hotfix a minor error in the python quickstarts tutorial. 

phamtrongthang123:
In section **8. Create the application**, there is a line: 
```python
    return Timetable(demo_data.name, timeslots, rooms, lessons)
```
This will cause `NameError: name 'demo_data' is not defined` because `demo_data` is not defined anywhere in the `generate_demo_data()` function.  A possible hotfix is to just directly assign a name:
```python
    return Timetable("DEMO_DATA", timeslots, rooms, lessons)
```


triceo (approved):

triceo:
Thank you!

PR #1395, opened 2025-02-17
fix: fix statistics for list variables allowing unassigned values

zepfred:
This PR updates the statistical calculation method to estimate value counts similarly to how we handle basic variables.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -1111,7 +1111,7 @@ public class SolutionDescriptor<Solution_> {
      */
     public double getProblemScale(Solution_ solution) {
         long logBase = getMaximumValueRangeSize(solution);
-        if (logBase == 1 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
+        if (logBase > 0 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
             // List variable allowing unassigned values must be increased; otherwise, the scale will be zero.
             logBase++;
         }
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -1110,11 +1110,7 @@ public class SolutionDescriptor<Solution_> {
      * @return {@code >= 0}
      */
     public double getProblemScale(Solution_ solution) {
-        long logBase = getMaximumValueRangeSize(solution);
-        if (logBase > 0 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
-            // List variable allowing unassigned values must be increased; otherwise, the scale will be zero.
-            logBase++;
-        }
+        var logBase = Math.max(2, getMaximumValueRangeSize(solution));
         ProblemScaleTracker problemScaleTracker = new ProblemScaleTracker(logBase);
         visitAllEntities(solution, entity -> {
             var entityDescriptor = findEntityDescriptorOrFail(entity.getClass());
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -1111,6 +1111,10 @@ public class SolutionDescriptor<Solution_> {
      */
     public double getProblemScale(Solution_ solution) {
         long logBase = getMaximumValueRangeSize(solution);
+        if (logBase == 1 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
+            // List variable allowing unassigned values must be increased; otherwise, the scale will be zero.
+            logBase++;
+        }
         ProblemScaleTracker problemScaleTracker = new ProblemScaleTracker(logBase);
         visitAllEntities(solution, entity -> {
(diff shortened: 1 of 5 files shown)


zepfred:
The commit f3b830c2435599ae436aa9aea4234a13ae7a14c0 caused the issue

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1111,6 +1111,10 @@ public long getMaximumValueRangeSize(Solution_ solution) {
      */
     public double getProblemScale(Solution_ solution) {
         long logBase = getMaximumValueRangeSize(solution);
+        if (logBase == 1 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
+            // List variable allowing unassigned values must be increased; otherwise, the scale will be zero.
```
Christopher-Chianelli, line 1115: Comment incorrect; scale would be NaN, since `log(1) = 0`; as for what `logBase` actually does, it is the base of the logarithm used in calculations (which is converted back to base 10 after all the calculations are done).
It defaults to the maximum value range size, since for basic planning variables, usually each entity increases the size by `maxValueRangeSize`, and` log_{maxValueRangeSize}(maxValueRangeSize) = 1`, increasing accuracy. It appears there an inconsistency in value range size calculation, since basic variables does add 1 if allow unassigned is true: https://github.com/TimefoldAI/timefold-solver/blob/e8302c2484c531f135508d19e00994a05ae4f265/core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/CompositeValueRangeDescriptor.java#L82 (inconsitency is introduced by this check that is only done for basic but not list: https://github.com/TimefoldAI/timefold-solver/blob/e8302c2484c531f135508d19e00994a05ae4f265/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/GenuineVariableDescriptor.java#L77-L80 )

triceo: I don't mind fixing the source data - just be careful, because other places depend on values returned by the descriptors and ranges, so adding 1 there somewhere could possibly change behavior of code in other places. Therefore making that change would require a more careful investigation of the rest of the code using this data.

That said, it may be worthwhile, because this inconsistency between basic and list var is unfortunate, and something you don't want to have to remember.

zepfred: > Comment incorrect; scale would be NaN, since log(1) = 0

`MathUtils.getLogInBase(logBase, 10d)` returns Infinity because `log(1) = 0` and dividing non-zero by zero returns Infinity. Next, the result divided by infinity is zero: ` (result / (double) MathUtils.LOG_PRECISION) / MathUtils.getLogInBase(logBase, 10d)`.

> (inconsistency is introduced by this check that is only done for basic but not list:

It makes sense. My first commit updated that section of the code, but some tests started to fail, and I reverted precisely because of Lukas' comment.

Christopher-Chianelli: I would change the code to
`var logBase = Math.max(2, getMaximumValueRangeSize(solution))`; simpler, and prevents the bug in other cases (the bug being, using a log base below 2 for calculations).

zepfred: Should we consider this for variables that do not accept unassigned values?

Christopher-Chianelli: Yes; (hence remove the `if` and replace it by a `max`)

triceo (approved):
LGTM when green.
(Which currently it is not.)

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1110,7 +1110,11 @@ public long getMaximumValueRangeSize(Solution_ solution) {
      * @return {@code >= 0}
      */
     public double getProblemScale(Solution_ solution) {
-        long logBase = getMaximumValueRangeSize(solution);
+        long logBase = Math.max(2, getMaximumValueRangeSize(solution));
+        if (logBase == 1 && getListVariableDescriptor() != null && getListVariableDescriptor().allowsUnassignedValues()) {
+            // List variable allowing unassigned values must be increased; otherwise, the scale will be zero.
+            logBase++;
+        }
```
Christopher-Chianelli, line 1117: This `if` should be removed; it will never trigger (logBase is at least 2).
```suggestion
```

Christopher-Chianelli (changes requested):

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
zepfred: Using the minimum of 2 in any situation will result in failure if the value count is zero. Your suggestion makes it simpler, but we still need the if block.

Christopher-Chianelli: Add this line to `ProblemScaleTracker.addBasicProblemScale`:
```
    public void addBasicProblemScale(long count) {
        if (count == 0) {
            // Log(0) = -infinity; also an invalid problem (since there are no
            // valid values for the variable, including null)
            return;
        }
        basicProblemScaleLog += MathUtils.getScaledApproximateLog(MathUtils.LOG_PRECISION, logBase, count);
    }
```

Christopher-Chianelli (changes requested):
See my comment; add a change to ProblemScaleTracker to make the min work

Christopher-Chianelli (approved):

PR #1404, opened 2025-02-20
docs: Use correct quarkus endpoint in quickstart guide

nathancolgate:
The sample quarkus application only uses a `/timetables` endpoint, not a `/timetables/solve` endpoint.

### DIFF
--- a/docs/src/modules/ROOT/pages/quickstart/shared/try-the-application.adoc
+++ b/docs/src/modules/ROOT/pages/quickstart/shared/try-the-application.adoc
@@ -6,7 +6,7 @@ The following example uses the Linux command `curl` to send a POST request:
 
 [source,shell]
 ----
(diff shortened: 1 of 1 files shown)


TomCools:
Hi @nathancolgate, thank you so much for all your edits. I am going through them now.
This one I am going to solve the other way round to be in line with the other examples (where we do have /timetable/solve).

TomCools:
Thank you for this @nathancolgate, I am going through your PR's now. 

There seems to be an inconsistency between how we handle this in our different examples. I think we'll align all examples on /timetables (without the /solve) so we will accept this PR and fix the other stuff in other PRs ourselves.

TomCools (approved):

nathancolgate:
Thanks @TomCools! Overall it was a great experience, even for a non-java guy like me! Hurray for better docs!

PR #1407, opened 2025-02-20
docs: Lesson constructor for tests

nathancolgate:
My tests were failing because of this error:

```
/src/test/java/org/acme/schooltimetabling/solver/TimetableConstraintProviderTest.java:[28,30] no suitable constructor found for Lesson(java.lang.String,java.lang.String,java.lang.String,java.lang.String,org.acme.schooltimetabling.domain.Timeslot,org.acme.schooltimetabling.domain.Room)
```

So i borrowed a second constructor from some of the OptaPlanner docs.

### DIFF
--- a/docs/src/modules/ROOT/pages/quickstart/shared/school-timetabling/school-timetabling-model.adoc
+++ b/docs/src/modules/ROOT/pages/quickstart/shared/school-timetabling/school-timetabling-model.adoc
@@ -269,6 +269,12 @@ public class Lesson {
         this.studentGroup = studentGroup;
     }
 
+    public Lesson(String id, String subject, String teacher, String studentGroup, Timeslot timeslot, Room room) {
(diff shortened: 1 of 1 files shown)


docs/src/modules/ROOT/pages/quickstart/shared/school-timetabling/school-timetabling-model.adoc
```diff
@@ -269,6 +269,12 @@ public class Lesson {
         this.studentGroup = studentGroup;
     }
 
+    public Lesson(String id, String subject, String teacher, String studentGroup, Timeslot timeslot, Room room) {
```
TomCools, line 272: I would add a comment this constructor is only for tests and make the constructor "package private" (no access modifier).

TomCools (approved):

PR #1421, opened 2025-02-24
feat: Declarative Shadow Variables

Christopher-Chianelli:
VariableListeners are notoriously difficult to write, especially when you have multiple variable depending on each others.
Declarative shadow variables fixes this by providing a declarative way to define the dependencies of your shadow variables and how to calculate them.

First, annotate your variables as `@ProvidedShadowVariable`:

```java
@PlanningEntity
public class Visit {
    String id;

    @InvalidityMarker
    boolean isInvalid;   // If true, the visit is in a "loop" and some of its provided shadow variables will be null

    @ProvidedShadowVariable(MyShadowVariableProvider.class)
    LocalDateTime serviceReadyTime;
    
    // ...
}
```

Next, create a `ShadowVariableProvider` to define your variables:

```java
public class MyShadowVariableProvider implements ShadowVariableProvider {
    @Override
    public void defineVariables(ShadowVariableFactory variableFactory) {
        // The previous visit; you don't need to use @PreviousElementShadowVariable
        var previousVisit = variableFactory.entity(Visit.class).previous();
        // The vehicle the visit is assignedf to; you don't need to use @InverseRelationShadowVariable
        var vehicle = variableFactory.entity(Visit.class).inverse(Vehicle.class);
        
        // The service ready time; when the technician will arrive
        var serviceReadyTime = variableFactory.newShadow(Visit.class)
               // If a visit has a previous visit, the ready time 
                .compute(previousVisit.fact(Location.class, Visit::getLocation),
                         previousVisit.variable(LocalDateTime.class, "serviceFinishTime"),
                         (visit, previousLocation, previousEndTime) -> previousEndTime.plus(visit.getLocation().travelTimeFrom(previousLocation))
                )
                // If a visit has no previous visit, use the vehicle's start time
                .orCompute(vehicle.fact(Location.class, Vehicle::getLocation),
                        vehicle.fact(LocalDateTime.class, Vehicle::getStartTime),
                        (visit, startLocation, startTime) -> startTime.plus(visit.getLocation().travelTimeFrom(startLocation))
                )
                // If a visit is unassigned, its serviceReadyTime is null
                .as("serviceReadyTime");
        
         // The service start time; when service can start
         // If the visit is a part of a group, service cannot start until all technicians arrive
         // Otherwise, it the same as the service start time
         var visitGroup = variableFactory
                        .entity(Visit.class)
                        .group(Visit.class, Visit::getVisitGroup);
         var serviceStartTime = variableFactory.newShadow(Visit.class)
                // If the visit has a visit group, its start time is the max of the ready times of the visit in its visit group
                .compute(visitGroup.variables(LocalDateTime.class, "serviceReadyTime"),
                        (visit, groupReadyTimes) -> groupReadyTimes.isEmpty() ? null : Collections.max(groupReadyTimes))
                // Otherwise, it the service ready time of this visit
                .orCompute(serviceReadyTime, (visit, readyTime) -> readyTime)
                .as("serviceStartTime");

         // Each service ends 30 minutes after it started
         var serviceFinishTime = variableFactory.newShadow(Visit.class)
                .compute(serviceStartTime, (visit, startTime) -> startTime.plusMinutes(30))
                .as("serviceFinishTime");
    }
}
```

and finally, write tests to ensure your shadow variables work correctly:

```java
@Test
public void shadowVariables() {
      var sessionFactory = ShadowVariableSessionFactory.create(
              SolutionDescriptor.buildSolutionDescriptor(RoutePlan.class,
                      Vehicle.class, Visit.class),
              new MyShadowVariableProvider());

      var vehicle = Vehicle("v1");
      var visit1 = Visit("c1");
      var visit2 = Visit("c2");
      var visit3 = Visit("c3");

      var session = sessionFactory.forEntities(vehicle, visit1, visit2, visit3);
      session.setInverse(visit1, vehicle);
      session.setPrevious(visit2, visit1);
      session.setPrevious(visit3, visit2);
      session.updateVariables();

      assertThat(visit1.getServiceReadyTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME);
      assertThat(visit1.getServiceStartTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME);
      assertThat(visit1.getServiceFinishTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(30L));
      assertThat(visit1.isInvalid()).isFalse();

      assertThat(visit2.getServiceReadyTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(60L));
      assertThat(visit2.getServiceStartTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(60L));
      assertThat(visit2.getServiceFinishTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(90L));
      assertThat(visit2.isInvalid()).isFalse();

      assertThat(visit3.getServiceReadyTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(120L));
      assertThat(visit3.getServiceStartTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(120L));
      assertThat(visit3.getServiceFinishTime()).isEqualTo(TestShadowVariableProvider.BASE_START_TIME.plusMinutes(150L));
      assertThat(visit3.isInvalid()).isFalse();
}
```

Behind the scenes, Timefold Solver calculates a valid topological order for each of your shadow variables.
This allows Timefold Solver to:

1. Only recalculate changed shadow variables. A shadow variable is recalculated if any of its declared inputs has changed. Using an undeclared variable input will lead to score corruptions.
2. Determine what variables are in a "loop". Variables are in a "loop" if:

- They directly or indirectly depend on each other.
- They depend on looped variables.

Currently a WIP:

- [ ] Write javadoc
- [ ] Write documentation
- [ ] Write additional tests
- [ ] Write additional use cases
- [ ] Finalize API


triceo:
Before I do a proper review, I suggest we do some house-keeping first. (Otherwise comments would get lost when files are being renamed and moved around.)

- Are we happy with the package? Variables generally go into `api.domain.variable`, so in this case `preview.api.domain.variable.declarative`?
- Preview features need to be declared in `PreviewFeature`; features not enabled through this enum need to fail fast when used. (In this case, probably when the new annotations are being processed.)

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -637,10 +663,21 @@ public boolean hasVariableDescriptor(String variableName) {
...
-    public VariableDescriptor<Solution_> getVariableDescriptor(String variableName) {
+    public @Nullable VariableDescriptor<Solution_> getVariableDescriptor(String variableName) {
         return effectiveVariableDescriptorMap.get(variableName);
     }
 
+    public @NonNull VariableDescriptor<Solution_> getVariableDescriptorOrFail(String variableName) {
+        var variableDescriptor = effectiveVariableDescriptorMap.get(variableName);
+        if (variableDescriptor == null) {
+            throw new IllegalArgumentException(
+                    "Entity class %s does not hava a \"%s\" genuine or shadow variable. Maybe you meant one of %s?"
```
triceo, line 674: Using raw strings (`"""`) will let you remove the escaping.
Also, the maybe should be on a new line, which raw strings also make easier.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/AbstractVariableReference.java
```diff
@@ -0,0 +1,23 @@
+package ai.timefold.solver.core.impl.domain.variable.provided;
+
+import ai.timefold.solver.core.preview.api.variable.provided.VariableReference;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+public abstract sealed class AbstractVariableReference<Solution_, Entity_, Value_> implements VariableReference<Entity_, Value_>
```
triceo, line 8: It says class, but it doesn't actually have any concrete methods, which suggests it's an interface instead. :-) We typically call these interfaces `InnerX`, where `X` is the public thing that's being extended.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/ChangedVariableNotifier.java
```diff
@@ -0,0 +1,39 @@
...
+    private final BiConsumer<VariableDescriptor<Solution_>, Object> beforeVariableChanged;
+    private final BiConsumer<VariableDescriptor<Solution_>, Object> afterVariableChanged;
+
+    public ChangedVariableNotifier(BiConsumer<VariableDescriptor<Solution_>, Object> beforeVariableChanged,
+            BiConsumer<VariableDescriptor<Solution_>, Object> afterVariableChanged) {
+        this.beforeVariableChanged = beforeVariableChanged;
+        this.afterVariableChanged = afterVariableChanged;
+    }
+
+    public static <Solution_> ChangedVariableNotifier<Solution_> empty() {
```
triceo, line 18: Arguably, this can be a singleton.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/ChangedVariableNotifier.java
```diff
@@ -0,0 +1,39 @@
...
+        return new ChangedVariableNotifier<>(
+                (vd, entity) -> {
+                },
+                (vd, entity) -> {
+                });
+    }
+
+    public static <Solution_> ChangedVariableNotifier<Solution_> of(InnerScoreDirector<Solution_, ?> scoreDirector) {
```
triceo, line 26: And this can be stored on a score director, meaning there would be a 1:1 mapping between these instances and score directors.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/DefaultGroupVariableReference.java
```diff
@@ -0,0 +1,244 @@
...
+import ai.timefold.solver.core.impl.domain.variable.supply.SupplyManager;
+import ai.timefold.solver.core.preview.api.variable.provided.GroupVariableReference;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+public final class DefaultGroupVariableReference<Solution_, Entity_, ParentValue_, Value_>
+        extends AbstractVariableReference<Solution_, Entity_, List<Value_>>
+        implements GroupVariableReference<Entity_, Value_> {
+    @NonNull
```
triceo, line 18: Let's start using a class-level `@NullMarked` - that eliminates the need to use `@NonNull`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/DefaultShadowCalculationBuilder.java
```diff
@@ -0,0 +1,114 @@
...
+import ai.timefold.solver.core.api.function.PentaFunction;
+import ai.timefold.solver.core.api.function.QuadFunction;
+import ai.timefold.solver.core.api.function.TriFunction;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.supply.SupplyManager;
+import ai.timefold.solver.core.preview.api.variable.provided.ShadowCalculationBuilder;
+import ai.timefold.solver.core.preview.api.variable.provided.SingleVariableReference;
+import ai.timefold.solver.core.preview.api.variable.provided.VariableReference;
+
+public class DefaultShadowCalculationBuilder<Solution_, Entity_, Value_> implements ShadowCalculationBuilder<Entity_, Value_> {
```
triceo, line 14: Also, you seem to be using the nullability markers somewhat selectively.
Why are they used in one place, and not in other places?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/InvalidityMarkerVariableDescriptor.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public void processAnnotations(DescriptorPolicy descriptorPolicy) {
+
+    }
+
+    @Override
+    public List<VariableDescriptor<Solution_>> getSourceVariableDescriptorList() {
+        return List.of();
```
triceo, line 30: IMO `Collections.emptyList()` better describes what will happen here.

core/src/main/java/ai/timefold/solver/core/preview/api/variable/provided/GroupVariableReference.java
```diff
@@ -0,0 +1,19 @@
+package ai.timefold.solver.core.preview.api.variable.provided;
+
+import java.util.List;
+import java.util.function.Function;
+
+public interface GroupVariableReference<Entity_, Value_> extends VariableReference<Entity_, List<Value_>> {
```
triceo, line 6: Arguably, the public APIs need the nullness markers the most.

core/src/main/java/ai/timefold/solver/core/preview/api/variable/provided/ShadowVariableSession.java
```diff
@@ -0,0 +1,13 @@
+package ai.timefold.solver.core.preview.api.variable.provided;
+
+public interface ShadowVariableSession {
```
triceo, line 3: My understanding of this API is still limited, but do we actually need this public API anywhere in the solver? IMO this is internal in core, and only exposed in the tests; therefore I'd make this public in the tests, not in the solver itself.

core/src/main/java/ai/timefold/solver/core/preview/api/variable/provided/ShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,13 @@
+package ai.timefold.solver.core.preview.api.variable.provided;
+
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.provided.MockShadowVariableSessionFactory;
+
+public interface ShadowVariableSessionFactory {
```
triceo, line 6: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/provided/DefaultShadowCalculationBuilder.java
Christopher-Chianelli: Mostly, they are used in classes where nullability is mixed (i.e. some fields/methods can be null, others cannot).

core/src/main/java/ai/timefold/solver/core/preview/api/variable/provided/ShadowVariableSession.java
Christopher-Chianelli: It can be moved; it is here for convenience when testing the prototype. (core internally depends on the default implementation which has before/after methods).

triceo:
Hypothetical scenario:

- I have two different shadow providers.
- But they end up declaring/depending on the same shadow variables.

Do they all become part of the same graph? In other words - is the graph limited to one shadow provider, or is it built from all shadow providers combined, and then split into individual independent components?

Christopher-Chianelli:
> Hypothetical scenario:
> 
>     * I have two different shadow providers.
> 
>     * But they end up declaring/depending on the same shadow variables.
> 
> 
> Do they all become part of the same graph? In other words - is the graph limited to one shadow provider, or is it built from all shadow providers combined, and then split into individual independent components?

Although untested, it is built from all shadow providers combined, and then split into individual independent components. Defining the same shadow variable twice is an error though; (you can use it multiple time, but cannot declare it multiple times).

triceo:
Please add this to the list of things that need to be tested.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/InvalidityMarker.java
```diff
@@ -0,0 +1,13 @@
...
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target({ METHOD, FIELD })
+@Retention(RUNTIME)
+public @interface InvalidityMarker {
```
ge0ffrey, line 12: Will this mark any type of invalidity or only loops?

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableSession.java
```diff
@@ -0,0 +1,17 @@
+package ai.timefold.solver.core.preview.api.domain.variable.declarative;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public interface ShadowVariableSession {
```
ge0ffrey, line 7: This seems similar to ConstraintVerifier, so the naming should be similar.
Does the ConstraintVerifier API have anything that is a Session?

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,16 @@
...
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface ShadowVariableSessionFactory {
+    static ShadowVariableSessionFactory create(SolutionDescriptor<?> solutionDescriptor,
+            ShadowVariableProvider shadowVariableProvider) {
+        return new MockShadowVariableSessionFactory<>(solutionDescriptor, shadowVariableProvider);
+    }
+
+    ShadowVariableSession forEntities(Object... entities);
```
ge0ffrey, line 15: Feels similar to ConstraintVerifier.given(), so maybe rename it to "given"?

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,16 @@
+package ai.timefold.solver.core.preview.api.domain.variable.declarative;
+
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.declarative.MockShadowVariableSessionFactory;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface ShadowVariableSessionFactory {
```
ge0ffrey, line 9: Maybe rename to something like ShadowVariableVerifier, as it serves a a simular purpose?

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/declarative/fsr/TestdataFSRAssertionConstraintProvider.java
```diff
@@ -0,0 +1,30 @@
...
+import java.time.Duration;
+
+import ai.timefold.solver.core.api.score.buildin.hardsoft.HardSoftScore;
+import ai.timefold.solver.core.api.score.stream.Constraint;
+import ai.timefold.solver.core.api.score.stream.ConstraintFactory;
+import ai.timefold.solver.core.api.score.stream.ConstraintProvider;
+
+import org.jspecify.annotations.NonNull;
+
+public class TestdataFSRAssertionConstraintProvider implements ConstraintProvider {
```
ge0ffrey, line 12: Is FSR being copy pasted into the solver test sources?
Or is it a subset?

Christopher-Chianelli: A **heavily** simplified subset.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -111,6 +127,7 @@ private void processShadowVariableDescriptorWithListVariable(ShadowVariableDescr
 
     private void
             processShadowVariableDescriptorWithoutListVariable(ShadowVariableDescriptor<Solution_> shadowVariableDescriptor) {
+        // TODO: Set up variable listener for declarative shadow variables
```
triceo, line 130: Should this still be here? I'd expect this feature to be in a mergable state, and therefore without TODOs. Possible future improvements will be tracked separately.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/InvalidityMarker.java
triceo, line 12: TODO: Naming and Javadoc.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableSession.java
triceo: As far as I understand, this does not need to be in the public API. Please remove.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableSessionFactory.java
triceo: Likewise, please remove from public API. The user has no use for this.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariableUpdater.java
```diff
@@ -0,0 +1,12 @@
+package ai.timefold.solver.core.preview.api.domain.variable.declarative;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ShadowVariableUpdater {
```
triceo, line 10: TODO: Naming and Javadoc

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/declarative/fsr/TestdataFSRAssertionConstraintProvider.java
triceo: Please remove references to FSR. FSR has no meaning in the context of the solver.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/declarative/task_assignment/TestdataTAConstraintProvider.java
```diff
@@ -0,0 +1,27 @@
...
+import java.time.Duration;
+
+import ai.timefold.solver.core.api.score.buildin.hardsoft.HardSoftScore;
+import ai.timefold.solver.core.api.score.stream.Constraint;
+import ai.timefold.solver.core.api.score.stream.ConstraintFactory;
+import ai.timefold.solver.core.api.score.stream.ConstraintProvider;
+
+import org.jspecify.annotations.NonNull;
+
+public class TestdataTAConstraintProvider implements ConstraintProvider {
```
triceo, line 12: Just as with "FSR", "TA" is also meaningless here. The solver isn't tailored to any particular problem, and these domain classes should therefore carry generic names. We already have dozens of them, and we were able to follow this policy; I don't see the need to divert now.

triceo (commented):
First round of review, surface-level.
Next round will look deeper into the logic.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -352,7 +366,14 @@ The entityClass (%s) has a @%s annotated member (%s) that has an unsupported typ
             declaredShadowVariableDescriptorMap.put(memberName, variableDescriptor);
         } else if (variableAnnotationClass.equals(ShadowVariable.class)
                 || variableAnnotationClass.equals(ShadowVariable.List.class)) {
-            var variableDescriptor = new CustomShadowVariableDescriptor<>(nextVariableDescriptorOrdinal, this, memberAccessor);
+            ShadowVariableDescriptor<Solution_> variableDescriptor;
+            var annotation = memberAccessor.getAnnotation(ShadowVariable.class);
+            if (annotation != null && !annotation.method().isEmpty()) {
+                variableDescriptor =
```
triceo, line 372: Where are the fail-fasts for when the other fields are used? The Javadoc suggests that they exist, but I didn't notice them.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSession.java
```diff
@@ -0,0 +1,70 @@
...
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariableSession;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public class DefaultShadowVariableSession<Solution_> implements ShadowVariableSession, Supply {
+    final VariableReferenceGraph<Solution_> graph;
+
+    record EntityVariablePair<Solution_>(VariableDescriptor<Solution_> variableDescriptor, Object entity) {
```
triceo, line 18: Considering what the descriptor is used for, it can easily be replaced by VariableMetaModel.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                    });
+                    variableReferenceGraph.addAfterProcessor(sourceVariableId, (graph, toEntity) -> {
+                        alias.targetEntityFunctionStartingFromVariableEntity().accept(toEntity, fromEntity -> {
+                            graph.addEdge((EntityVariableOrFactReference<Solution_>) graph.lookup(fromVariableId, fromEntity),
+                                    (EntityVariableOrFactReference<Solution_>) graph.lookup(toVariableId, toEntity));
+                        });
+                    });
+                }
+                // Note: it is impossible to have a declarative variable affect graph edges,
+                // since accessing a declarative variable from another declarative variable is prohibited.
```
triceo, line 136: "Accessing a declarative variable from another declarative variable" clearly doesn't mean what I think it means, because accessing one variable for another is the goal of this entire "a.b" syntax.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                }
+            }
+        }
+
+        // Create variable listeners for each declarative shadow variable descriptor
+        for (var declarativeShadowVariable : declarativeShadowVariableDescriptors) {
+            var fromEntityClass = declarativeShadowVariable.getEntityDescriptor().getEntityClass();
+            var fromVariableName = declarativeShadowVariable.getVariableName();
+            final var fromVariableId = new VariableId(fromEntityClass, fromVariableName);
```
triceo, line 85: VariableMetaModel was invented earlier to serve this purpose.
(VariableDescriptor is more of a utility class, and it is a mess. VariableMetaModel is a well-defined carrier of metadata.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                                                ignored -> new LinkedHashSet<>())
+                                        .add(source);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        // Create variable listeners for each declarative shadow variable descriptor
```
triceo, line 81: Where exactly are the listeners being created? I don't see any use of `VariableListener` here. Maybe let's call it something else, to not mix concepts?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                if (entityClass.isInstance(entity)) {
+                    var toVariableId = new VariableId(entityClass, declarativeShadowVariableDescriptor.getVariableName());
+                    for (var sourceRoot : declarativeShadowVariableDescriptor.getSources()) {
+                        for (var source : sourceRoot.variableSourceReferences()) {
+                            if (source.isTopLevel() && source.isDeclarative()) {
+                                var fromEntityClass = source.entityClass();
+                                var fromVariableName = source.variableName();
+                                var fromVariableId = new VariableId(fromEntityClass, fromVariableName);
+
+                                ((BiConsumer<Object, Consumer>) (BiConsumer) sourceRoot.valueEntityFunction())
```
triceo, line 153: This is a strange cast. I don't think I've ever seen a double cast, honestly didn't know it was even possible. What's the point?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultTopologicalOrderGraph.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public int getTopologicalOrder(int node) {
+        return ord[node];
+    }
+
+    @Override
+    public void endBatchChange() {
+        AtomicInteger index = new AtomicInteger(1);
+        AtomicInteger stackIndex = new AtomicInteger(0);
```
triceo, line 83: I think I introduced a `MutableInteger` a while back. Since you don't need atomicity, you can use that one.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultTopologicalOrderGraph.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        // If node is a root node, pop the stack and generate an SCC
+        if (onStackSet[node] && lowMap[node] == indexMap[node]) {
+            BitSet out = new BitSet();
+
+            int current;
+            do {
+                current = stack[stackIndex.decrementAndGet()];
+                onStackSet[current] = false;
```
triceo, line 146: I don't understand - if `out` is a `BitSet`, why is `onStackSet` (which also stores bits) not a `BitSet` as well?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityVariableOrFactReference.java
```diff
@@ -0,0 +1,23 @@
+package ai.timefold.solver.core.impl.domain.variable.declarative;
+
+import java.util.Objects;
+
+public record EntityVariableOrFactReference<Entity_>(VariableId variableId, Entity_ entity,
```
triceo, line 5: Another possible use of VariableMetaModel.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    public static <Entity_, Value_> RootVariableSource<Entity_, Value_> from(
+            Class<? extends Entity_> rootEntityClass,
+            String targetVariableName,
+            String variablePath,
+            MemberAccessorFactory memberAccessorFactory,
+            DescriptorPolicy descriptorPolicy) {
+        var pathParts = variablePath.split("\\.");
```
triceo, line 49: If we have `COLLECTION_REFERENCE_SUFFIX`, let's make the separator a constant too.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+        boolean afterVariable = false;
+        Class<?> currentEntity = rootEntityClass;
+
+        for (var pathPart : pathParts) {
+            if (pathPart.endsWith(COLLECTION_REFERENCE_SUFFIX)) {
+                if (afterVariable) {
+                    throw new IllegalArgumentException("Cannot reference a collection on a variable.");
+                }
+                if (hasListMemberAccessor) {
+                    throw new IllegalArgumentException("Cannot reference a collection on a collection.");
```
triceo, line 64: Deserves a more descriptive message.
Where did the user make a mistake? Maybe there is a suggestion for fixing it?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+                    variableMemberAccessor.getDeclaringClass(),
+                    variableMemberAccessor.getName(),
+                    sourceVariablePath.memberAccessorsBeforeEntity,
+                    isTopLevel,
+                    isDeclarativeShadowVariable(variableMemberAccessor),
+                    new VariableId(rootEntityClass, targetVariableName),
+                    downstreamDeclarativeVariable,
+                    sourceVariablePath.getValueVisitorFromVariableEntity()));
+            isTopLevel = false;
+        }
```
triceo, line 176: Arguably can be extracted from this already very long method.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+                    for (var accessor : finalChainToVariable) {
+                        current = accessor.executeGetter(current);
+                        if (current == null) {
+                            continue outer;
+                        }
+                    }
+                    consumer.accept((Value_) current);
+                }
+            };
+        }
```
triceo, line 149: Arguably can be extracted from this already very long method.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+                Object current = entity;
+                for (var accessor : listMemberAccessors) {
+                    current = accessor.executeGetter(current);
+                    if (current == null) {
+                        return;
+                    }
+                }
+
+                Iterable<Object> iterable = (Iterable<Object>) current;
+                outer: for (var item : iterable) {
```
triceo, line 138: Can we do this without labels? My experience has been that, whenever I felt like I need to use explicit goto, there was a better code structure that avoided it.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,305 @@
...
+        changed.set(node.id());
+    }
+
+    public void updateChanged() {
+        if (updating) {
+            // An exception occurred when processing a move
+            // Variable state is probably incorrect/will trigger a different
+            // exception, so return early.
+            return;
+        }
```
triceo, line 153: If an uncaught exception occurred, why are we even here? The solver would've crashed before getting here. (The undo move issue was already fixed.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,305 @@
...
+
+        record AffectedEntity(Object entity, VariableUpdaterInfo variableUpdaterInfo) {
```
triceo, line 155: The method is long enough already, even without data types being declared inside of it. Please move this type outside - you can still make it private.

core/src/test/java/ai/timefold/solver/core/preview/api/variable/declarative/fsr/FSRShadowVariableTest.java
```diff
@@ -0,0 +1,491 @@
...
+import ai.timefold.solver.core.impl.testdata.domain.declarative.fsr.TestdataFSRConstraintProvider;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.fsr.TestdataFSRRoutePlan;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.fsr.TestdataFSRVehicle;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.fsr.TestdataFSRVisit;
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariableSessionFactory;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+public class FSRShadowVariableTest {
```
triceo, line 26: These tests are complex enough that they deserve comments.
What behavior is being tested by each group of lines, why is the expected result what it is.

core/src/test/java/ai/timefold/solver/core/preview/api/variable/declarative/task_assignment/TAShadowVariableTest.java
```diff
@@ -0,0 +1,72 @@
...
+import ai.timefold.solver.core.config.solver.SolverConfig;
+import ai.timefold.solver.core.config.solver.termination.TerminationConfig;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.task_assignment.TestdataTAConstraintProvider;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.task_assignment.TestdataTAEmployee;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.task_assignment.TestdataTASchedule;
+import ai.timefold.solver.core.impl.testdata.domain.declarative.task_assignment.TestdataTATask;
+
+import org.junit.jupiter.api.Test;
+
+public class TAShadowVariableTest {
```
triceo, line 22: Let's not overdo it with the tests that run the solver.
I'm fine with having this one, it's a good end-to-end test, but let's test future functionality on a much more granular level.

triceo (changes requested):
General comments having read the entire PR in depth:

- The code structure could be improved. There are several very long methods, which (when broken down into smaller, well-named ones) will help the algorithms to be self-documenting.
- The public API of the feature is undocumented.
- The testing infrastructure for the new mechanism should be hidden for now; we haven't discussed it at all, and we should not expose it to the users since we only have a limited amount of time to get this merged.
- The tests are complex enough to warrant comments.

We need to go through a naming excercise for the two new annotations we expose. Also, the feature name itself ("declarative shadow variables") is arguably no longer accurate. Let's leave that for the end, once everything is properly documented, at which point we'll have the most accurate information to infer correct names for concepts.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSession.java
```diff
@@ -19,14 +18,14 @@ record EntityVariablePair<Solution_>(VariableDescriptor<Solution_> variableDescr
...
-                return entity == other.entity && Objects.equals(variableDescriptor, other.variableDescriptor);
+                return entity == other.entity && variableDescriptor.getOrdinal() == other.variableDescriptor.getOrdinal();
             }
             return false;
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(System.identityHashCode(entity), variableDescriptor);
+            return (31 * System.identityHashCode(entity)) ^ variableDescriptor.getOrdinal();
```
triceo, line 28: ... but ordinal has the function. In either case, arguably we should put this logic inside `VariableDescriptor` and not externalize it here.

triceo: And maybe I would add a comment here that this class is often used as key in hashmaps, and therefore this method needs to be optimized.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -0,0 +1,246 @@
...
+                    isTopLevel,
+                    isDeclarativeShadowVariable(variableMemberAccessor),
+                    new VariableId(rootEntityClass, targetVariableName),
+                    downstreamDeclarativeVariable,
+                    sourceVariablePath.getValueVisitorFromVariableEntity()));
+            isTopLevel = false;
+        }
+
+        if (variableSourceReferences.isEmpty()) {
+            throw new IllegalArgumentException("The source path \"%s\" on entity class %s does not reference any variables."
```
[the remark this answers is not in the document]
Christopher-Chianelli: Even for arguments that are Strings?  (i.e. should Strings be represented as `"text"`, `(text)` or `("text")`?

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
Christopher-Chianelli: Fail fasts are in RootVariableSource, which does the processing of the variable paths.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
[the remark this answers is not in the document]
triceo: I don't think we add the quotes for strings. At least I don't recall a situation where we do. In fact, most things we put here are some sort of a string (class name, var name, ...) and we do not use quotation marks for that.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
Christopher-Chianelli: The older version the PR was not tested with more exotic cases, such as "previous.previous" or "factProperty.previous" or "myShadowVariable.myOtherShadowVariable".

The old impl made it harder to see the inherent issues that this impl makes obvious by removing all the indirection.

Basically, the three rules are:

- You can only access normal shadow variables from the root entity or from a collection element. Otherwise, the variable listener need to maintain an inverse set to lookup the affected entities for a shadow variable change.
- You cannot chain multiple declarative shadow variables together in a single path. For instance, "closestStation.visitCount"; this has the potential to change the dependency/topology of the graph in the middle of an update.
- Your path must have at least one variable in it, and all names must be the name of members in the class.

triceo: Two comments:

- Let's make sure these rules are listed in a Javadoc somewhere. Lately we've been touching decade-old code, and we often wished the authors of that code had the foresight to list all of their thinking, rules and assumptions somewhere. Code archeology is not fun.

- "Your path must have at least one variable in it, and all names must be the name of members in the class." What else _could_ be there in the path? I'd argue _nothing but variables_. Constants do not need to be referenced - we only need to reference what we need to build and maintain the graph.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    public void markChanged(EntityVariableOrFactReference<?> node) {
+        if (changed.isEmpty()) {
+            graph.startBatchChange();
+        }
+        changed.set(node.id());
+    }
+
+    public void updateChanged() {
```
triceo, line 147: This method is very long and also - not coincidentally - shows up as a dominator in the flamegraph. Please break this method down into smaller methods, so that we can better understand the logic, and so that the flame graph can actually meaningfully show how much each part contributes to the result.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultTopologicalOrderGraph.java
Christopher-Chianelli: onStackSet is simply a mapping from node (an int)  and does not need any of the BitSet methods. Essentially, indexMap, lowMap, and onStackSet acts as additional attributes on nodes.

out is a BitSet since we need to iterate it and access other BitSet methods.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/declarative/concurrent_values/TestdataConcurrentValue.java
```diff
@@ -0,0 +1,272 @@
...
+            for (var visit : concurrentValueGroup) {
+                if (visit.serviceReadyTime != null && startTime.isBefore(visit.serviceReadyTime)) {
+                    startTime = visit.serviceReadyTime;
+                }
+            }
+        }
+        return startTime;
+    }
+
+    @ShadowSources(sources = { "serviceStartTime" })
```
triceo, line 122: Let's try to do it without the mandatory argument name. With just one argument, this should be possible.

benchmark/src/main/resources/benchmark.xsd
```diff
@@ -2603,6 +2603,9 @@
     <xs:restriction base="xs:string">
                   
       
+      <xs:enumeration value="DECLARATIVE_SHADOW_VARIABLES"/>
```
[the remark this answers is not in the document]
Christopher-Chianelli: What about "supplier variable"? Matches the API syntax `@ShadowVariable(supplierName="...")`?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,330 @@
...
+            }
+            visited[nextNode] = true;
+            var shadowVariable = instanceList.get(nextNode);
+            var isChanged = updateShadowVariable(shadowVariable,
+                    graph.isLooped(loopedTracker, nextNode),
+                    affectedEntities);
+
+            if (isChanged) {
+                graph.nodeForwardEdges(nextNode).forEachRemaining(
+                        (int node) -> nodeHeap.add(new AffectedShadowVariable(node, graph.getTopologicalOrder(node))));
```
triceo, line 181: Performance idea as I was reading through this: could it be possible that a node is being added even though it is already `visited`? If so, let's not add it, save ourselves some iteration.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+        changed.clear();
+        return heap;
+    }
+
+    @SuppressWarnings("unchecked")
+    private void updateInvalidityStatusOfAffectedEntities(Set<AffectedEntity> affectedEntities, LoopedTracker loopedTracker) {
+        for (var affectedEntity : affectedEntities) {
+            var invalidDescriptor = affectedEntity.variableUpdaterInfo.invalidityMarkerVariableDescriptor();
```
triceo, line 262: Please look for the old name "invalidityMarker" and replace it with the new name.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,330 @@
...
+        }
+    }
+
+    private PriorityQueue<AffectedShadowVariable> createInitialChangeQueue() {
+        var heap = new PriorityQueue<AffectedShadowVariable>(instanceList.size());
+        for (var i = changed.nextSetBit(0); i >= 0; i = changed.nextSetBit(i + 1)) {
+            var topologicalOrder = graph.getTopologicalOrder(i);
+            heap.add(new AffectedShadowVariable(i, topologicalOrder));
+            if (i == Integer.MAX_VALUE) {
+                break; // or (i+1) would overflow
```
triceo, line 252: Shouldn't we rather throw an exception here?
Seems like this is a failure state, maybe even impossible state.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -0,0 +1,330 @@
...
+        fixedEdges = new HashMap<>();
+        entityToVariableReferenceMap = new IdentityHashMap<>();
+    }
+
+    public <Entity_> EntityVariablePair addVariableReferenceEntity(
+            VariableMetaModel<?, ?, ?> variableId,
+            Entity_ entity,
+            VariableUpdaterInfo variableReference) {
+        if (!variableId.entity().type().isInstance(entity)) {
+            throw new IllegalArgumentException(variableId + " " + entity);
```
triceo, line 47: Deserves a descriptive message.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
triceo: Also applies to the previous exception immediately above.

triceo (approved):
LGTM after all conversations are resolved. 
(I added a small amount of new ones.)

benchmark/src/main/resources/benchmark.xsd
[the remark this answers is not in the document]
triceo: I was thinking a bit more, and let's just keep the name for now.
While it's in preview, it's going to be called this.
Once it's no longer in preview, it's just a shadow variable, and all other options are deprecated. Even the annotation says "ShadowVariable", nothing else.
So the need to provide another name will go away together with the preview status.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -0,0 +1,98 @@
...
+     * The paths to variables the method uses to compute the value of a {@link ShadowVariable#supplierName() supplier variable}.
+     * <p>
+     * Each path is a {@link String} that is one of the following three forms:
+     *
+     * <ul>
+     * <li>
+     * "variableName", for referring any variable on the same planning entity.
+     * </li>
+     * <li>
+     * A list of names seperated by ".", such as "variableOrFact.fact.entity.supplierVariable",
```
triceo, line 27: Why would you ever reference a fact? This sounds like something that should not be possible.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
Christopher-Chianelli: Overly cautious `BitSet` iteration. How `BitSet` iteration works is by querying for the first set bit after or at an index. Then to get the next set bit, you do `nextSetBit(currentBit + 1)`. `currentBit + 1` will overflow if BitSet has the INT_MAX bit set. For our cases, this is impossible, since the JVM limit array sizes to slightly less than INT_MAX.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
Christopher-Chianelli: Leftover from debugging; removing

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
Christopher-Chianelli: FSR needs this due to the domain; the visit collection of a visit group is on its own object instead of on the entity; i.e. `visitGroup.visits[]` instead of `visitGroup[]`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
triceo: I think you've just given a reason for why this should have been treated as a failure state. :-) At the very least, there should have been a comment explaining what you just described.

Try to think of it like this: if I have a question because I don't understand something, it's probably a piece of code that deserves an explanatory comment. :-) There already is a comment, but in this case, the additional context you provided is useful.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
triceo: Then arguably let's only allow that use case, and forbid all other fact references?

You don't need to reference `factA.factB` for correctness - you only ever need to reference `factA.facts[]`. A chain of fact references is IMO unnecessary and I'd rather forbid it now, possibly re-introducing it later if there's demand. 

IMO we should not encourage code with so many levels of indirection.

Christopher-Chianelli: It is arguably more code to allow only a single level of indirection/fail otherwise and for very little benefit. IMO if we support one level of indirection, we might as well support infinite levels of indirection, since it takes no extra code to support (and it takes extra code to explicitly forbid).

Christopher-Chianelli: Java encourages indirection, so we _should_ allow it.
It is extremely reasonable to, in Java, have code structure like this:

```java
public record ConcurrentTasks(List<Task> tasks, TimeAdjuster adjuster) {}
public record DependencyTasks(List<Task> tasks, TimeAdjuster adjuster) {}
public record TaskGroup(ConcurrentTasks concurrentTasks, DependencyTasks dependencyTasks) {}

@PlanningEntity
public class Task {
    TaskGroup taskGroup;
}
```

Such a structure would require at least two level of indirections. Last I checked, we wanted users to be able to use their own classes without restrictions. Imposing a "at most one level of indirection in planning entity classes" is an unreasonable restriction that will force users to create their own, separate Solver specific classes.

triceo: We limit users every day in what sort of models they can build. We have rules about inheritance, and what methods you can and can not override. We do that to keep complexity at hand.

We also make decisions that limit users in other places. We only have quad streams, assuming no more will ever be necessary. (And for the one person who ever asked for more, we created `map(...)`.

Is one level of indirection too low a limit? Maybe. I'd rather go too low and have to increase the limit later, than go too high and enable unnecessary complexity. 

Arguably, this is a philosophical debate - are we a project that allows users unlimited power, and bets on their responsibility and wisdom? Or are we a project that has an opinion, and makes conscious choices which guide people towards certain preferred outcomes? In my opinion, we have always been the former.

Christopher-Chianelli: If we are the latter, we are doing a terrible job at it; we are effectively the former with SolverConfig + what we allow in user models.

The SolverConfig gives the user infinite freedom, is vital to provide/edit in order to get the best performance, and does not provide an opinion on anything.

We do not forbid the vast majority of user code; we only forbid user code that is incompatible with the Solver (ex: code w/o a no-arg constructor or code referring to a variable that does not exist).

PR #1422, opened 2025-02-25
Feat: Allow phase commands to read termination status

triceo:


core/src/test/java/ai/timefold/solver/core/api/solver/SolverFactoryTest.java
```diff
@@ -237,8 +237,7 @@ void validateInitializationPhases() {
         solverConfig = PlannerTestUtils
                 .buildSolverConfig(TestdataSolution.class, TestdataEntity.class)
                 .withPhases(new CustomPhaseConfig()
-                        .withCustomPhaseCommandList(List.of(scoreDirector -> {
-                        })),
+                        .withCustomPhaseCommands(new NoChangeCustomPhaseCommand()),
```
zepfred, line 240: We should consider moving the `NoChange*` related models to the test package. The main issue would be the duplication.

core/src/test/java/ai/timefold/solver/core/impl/phase/custom/DefaultCustomPhaseTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private static Duration measure(Runnable runnable) {
+        var milliTime = System.currentTimeMillis();
+        runnable.run();
+        return Duration.ofMillis(System.currentTimeMillis() - milliTime);
+    }
+
+    @NullMarked
+    private static final class NeverendingPhaseCommand implements PhaseCommand<TestdataSolution> {
```
zepfred, line 65: The name implies it never ends. Perhaps we could use a different name, such as `PhaseEndingCheckPhaseCommand` or something similar.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-to-latest-version.adoc
```diff
@@ -93,6 +93,48 @@ var solverConfig = new SolverConfig()
 ----
 ====
 
+'''
```
zepfred, line 96: Should we provide a recipe for that?

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -914,23 +908,6 @@ add the `customProperties` element and use xref:using-timefold-solver/configurat
 ----
 
 
-[#noChangeSolverPhase]
```
zepfred, line 917: I'm curious. Which rare cases would it be?

zepfred (approved):
LGTM!

zepfred:
@triceo, the tests have failed again. I hope it's not mutable global state causing this issue.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
triceo: I can not come up with any, which is why I've deprecated it for removal.
I'm guessing it's something ancient that is no longer necessary.

If people notice the deprecation, they may come back to us and tell us. At which point, we can decide if this is something we want to support, and then perhaps figure out a proper way of solving the use case.

PR #1431, opened 2025-02-28
feat: allow phase-level terminations on solver-level

triceo:
Allows terminations which were previously only possible on phases (such as diminished returns termination) to also be possible on solver level.
If a globally configured phase-level termination is not applicable to a particular phase, it is skipped.
If a phase-level termination is configured on a phase to which it is not applicable, the solver fails fast.

To achieve all of this, terminations are significantly refactored. There is now a distinction between phase terminations, solver terminations, and terminations that are both (universal).


core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
```diff
@@ -88,18 +92,32 @@ public boolean isAssertShadowVariablesAreNotStaleAfterStep() {
...
     public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        if (!solver.isTerminationSameAsSolverTermination(phaseTermination)) {
+            // Only fail if the user put the inapplicable termination on the phase, not on the solver.
+            // On the solver level, inapplicable phase terminations are skipped.
+            // Otherwise you would only be able to configure a global phase-level termination on the solver
+            // if it was applicable to all phases.
+            var unsupportedTerminationList = phaseTermination instanceof UniversalTermination<Solution_> universalTermination
+                    ? universalTermination.getPhasesTerminationsInapplicableTo(phaseScope)
+                    : Collections.emptyList();
+            if (!unsupportedTerminationList.isEmpty()) {
```
Christopher-Chianelli, line 113: Can we fail fast when the solver is being built? This will fail during solving.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -40,26 +41,30 @@ public abstract class AbstractSolver<Solution_> implements Solver<Solution_> {
     protected final BestSolutionRecaller<Solution_> bestSolutionRecaller;
     // Note that the DefaultSolver.basicPlumbingTermination is a component of this termination.
     // Called "solverTermination" to clearly distinguish from "phaseTermination" inside AbstractPhase.
-    protected final Termination<Solution_> solverTermination;
+    protected final UniversalTermination<Solution_> solverTermination;
```
Christopher-Chianelli, line 44: Maybe rename field to `universalTermination`? Since there is a class called `SolverTermination` now.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AbstractCompositeTermination.java
```diff
@@ -3,89 +3,110 @@
...
+        for (var termination : phaseTerminationList) {
             termination.phaseStarted(phaseScope);
         }
     }
 
     @Override
-    public void stepStarted(AbstractStepScope<Solution_> stepScope) {
-        for (Termination<Solution_> termination : terminationList) {
+    public final void stepStarted(AbstractStepScope<Solution_> stepScope) {
+        for (var termination : phaseTerminationList) {
```
Christopher-Chianelli, line 63: The `step` methods do not get called for the solver terminations? How does the solver terminations get updated then?

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AbstractCompositeTermination.java
```diff
@@ -3,89 +3,110 @@
...
         permits AndCompositeTermination, OrCompositeTermination {
 
     protected final List<Termination<Solution_>> terminationList;
+    protected final List<PhaseTermination<Solution_>> phaseTerminationList;
+    protected final List<SolverTermination<Solution_>> solverTerminationList;
 
     protected AbstractCompositeTermination(List<Termination<Solution_>> terminationList) {
-        this.terminationList = terminationList;
+        this.terminationList = Objects.requireNonNull(terminationList);
+        this.phaseTerminationList = terminationList.stream()
```
Christopher-Chianelli, line 32: What happens to universal terminations? Do they get put in both lists?

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTerminationTest.java
```diff
@@ -0,0 +1,132 @@
...
+                .withPhases(new ConstructionHeuristicPhaseConfig())
+                .withTerminationConfig(new TerminationConfig()
+                        .withDiminishedReturnsConfig(new DiminishedReturnsTerminationConfig()));
+        var solution = TestdataSolution.generateSolution(2, 2);
+        solution.getEntityList().forEach(entity -> entity.setValue(null)); // Uninitialize.
+        var solver = SolverFactory.<TestdataSolution> create(solverConfig)
+                .buildSolver();
+        var bestSolution = solver.solve(solution);
+        // 2 entities means 2 steps, but the step count limit is 1.
+        // Therefore the best solution is uninitialized.
```
Christopher-Chianelli, line 71: Outdated comment; no step count termination is set.

Christopher-Chianelli (changes requested):
Need to make https://github.com/TimefoldAI/timefold-solver/blob/4fa79be7f18916cf1faa9d83b85a2350456f0e8b/quarkus-integration/quarkus/runtime/src/main/java/ai/timefold/solver/quarkus/TimefoldRecorder.java#L142-L169 set Diminished returns on Solver level.

Christopher-Chianelli:
Also need to change https://github.com/TimefoldAI/timefold-solver/blob/4fa79be7f18916cf1faa9d83b85a2350456f0e8b/spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverAutoConfiguration.java#L339-L359 in spring

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/AbstractCompositeTermination.java
triceo: Yes.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
triceo: The comment explains the field name well, I think.
I agree that the seeming conflict with `SolverTermination` is not ideal, but the field name is correct - the termination will terminate the solver.

Christopher-Chianelli (approved):

PR #1437, opened 2025-03-04
docs: assorted improvements

triceo:


triceo:
@TomCools When you find a minute, would you mind taking a look?

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-optaplanner.adoc
```diff
@@ -27,7 +27,7 @@ Gradle::
 --
 [source,shell,subs=attributes+]
 ----
-curl https://timefold.ai/product/upgrade/upgrade-timefold.gradle > upgrade-timefold.gradle ; gradle -Dorg.gradle.jvmargs=-Xmx2G --init-script upgrade-timefold.gradle rewriteRun -DtimefoldSolverVersion={timefold-solver-version} ; rm upgrade-timefold.gradle
+curl https://raw.githubusercontent.com/TimefoldAI/timefold-solver/refs/tags/v{timefold-solver-version}/migration/upgrade-timefold.gradle > upgrade-timefold.gradle ; gradle -Dorg.gradle.jvmargs=-Xmx2G --init-script upgrade-timefold.gradle rewriteRun -DtimefoldSolverVersion={timefold-solver-version} ; rm upgrade-timefold.gradle
```
TomCools, line 30: This doesn't work for me. 
example gives 404, because `upgrade-timefold.gradle` doesn't exist? 

https://raw.githubusercontent.com/TimefoldAI/timefold-solver/refs/tags/v1.19.0/migration/upgrade-timefold.gradle

triceo: Correct, doesn't exist for now. It will when the next release is tagged.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-to-latest-version.adoc
```diff
@@ -33,7 +33,7 @@ Gradle::
 --
 [source,shell,subs=attributes+]
 ----
-curl https://timefold.ai/product/upgrade/upgrade-timefold.gradle > upgrade-timefold.gradle ; gradle -Dorg.gradle.jvmargs=-Xmx2G --init-script upgrade-timefold.gradle rewriteRun -DtimefoldSolverVersion={timefold-solver-version} ; rm upgrade-timefold.gradle
+curl https://raw.githubusercontent.com/TimefoldAI/timefold-solver/refs/tags/v{timefold-solver-version}/migration/upgrade-timefold.gradle > upgrade-timefold.gradle ; gradle -Dorg.gradle.jvmargs=-Xmx2G --init-script upgrade-timefold.gradle rewriteRun -DtimefoldSolverVersion={timefold-solver-version} ; rm upgrade-timefold.gradle
```
TomCools, line 36: Ditto other comment

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -140,22 +140,51 @@ Some solver configurations use the `Random` instance a lot more than others.
 For example, Simulated Annealing depends highly on random numbers, while Tabu Search only depends on it to deal with score ties.
 The environment mode influences the seed of that `Random` instance.
 
-These are the environment modes:
+[#environmentModeReproducibility]
+=== Reproducibility
 
+For the environment mode to be reproducible,
+any two runs of the same dataset with the same solver configuration will must have the same result at every step.
```
TomCools, line 147:
```suggestion
any two runs of the same dataset with the same solver configuration must have the same result at every step.
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,76 +209,78 @@ Any variable that changed between the "before move" solution and the "after move
...
 because it clones solutions before and after each move.
 
 
+[#environmentModeFullAssert]
+==== `FULL_ASSERT`
+
+The `FULL_ASSERT` mode turns on all assertions (such as assert that the incremental score calculation is uncorrupted for each move) to fail-fast on a bug in a Move implementation, a constraint, the engine itself, ...
+
+This mode is <<environmentModeReproducibility,reproducible>>.
+It is also intrusive because it calls the method `calculateScore()` more frequently than a non-assert mode.
```
TomCools, line 225:
```suggestion
It is also intrusive because it calls the method `calculateScore()` more frequently than a non-assert mode, making the `FULL_ASSERT` mode very slow.
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,76 +209,78 @@ Any variable that changed between the "before move" solution and the "after move
...
+
+The `FULL_ASSERT` mode is horribly slow,
```
TomCools, line 227: Remove these 2 with the suggestions above.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,76 +209,78 @@ Any variable that changed between the "before move" solution and the "after move
...
+because it does not rely on incremental score calculation.
+
+
 [#environmentModeNonIntrusiveFullAssert]
-=== `NON_INTRUSIVE_FULL_ASSERT`
+==== `NON_INTRUSIVE_FULL_ASSERT`
 
 The `NON_INTRUSIVE_FULL_ASSERT` turns on several assertions to fail-fast on a bug in a Move implementation,
```
TomCools, line 234: Do we have a list of the assertions somewhere? Because "ALL" (FULL_ASSERT) and "several" (NONE_INTRUSIVE_FULL_ASSERT) isn't very telling. 

Side note: having to say we only activate "several assertions" for NONE_INTRUSIVE_FULL_ASSERT is a bit weird. (But probably legacy?)

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,76 +209,78 @@ Any variable that changed between the "before move" solution and the "after move
...
+This mode is <<environmentModeReproducibility,reproducible>>
+and is negligibly slower than the `NO_ASSERT` mode.
+However, it gives you the benefit of quickly checking for score corruptions.
+If you can guarantee that your code is and will remain bug-free,
+you can switch to the `NO_ASSERT` mode for a marginal performance gain.
 
-[NOTE]
-====
-Despite the reproducible mode, your application might still not be fully reproducible because of:
+In practice, this mode uses the default, fixed <<randomNumberGenerator,random seed>> if no seed is specified,
```
TomCools, line 267: Since Seeds aren't really mentioned clearly before, I wouldn't introduce the concept here.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,76 +209,78 @@ Any variable that changed between the "before move" solution and the "after move
...
+and it also disables certain concurrency optimizations, such as work stealing.
```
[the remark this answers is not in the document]
triceo: The `NO_ASSERT` mode is negligibly faster than `PHASE_ASSERT`, because it really just removes one quick assertions. But the text makes it seems as if we're comparing `NO_ASSERT` to `NON_REPRODUCIBLE`, which is not the case - the difference between the two will likely be more significant.

I will think about rephrasing this.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
triceo: We do not, unfortunately. Which assertions are enabled also depends on what each phase supports - listing this with all the possible combinations would IMO bring even more complexity. I'll look into rewording this a bit.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -180,80 +211,96 @@ Any variable that changed between the "before move" solution and the "after move
...
+but it is not <<environmentModeReproducibility,reproducible>>.
 Avoid using it during development as it makes debugging and bug fixing painful.
 If your production environment doesn't care about reproducibility, use this mode in production.
 
-In practice, this mode uses no fixed <<randomNumberGenerator,random seed>> if no seed is specified.
+Unlike all the other modes,
+this mode doesn't use any fixed <<randomNumberGenerator,random seed>> unless one is provided.
+
+
+[#environmentModeRecommendations]
```
triceo, line 294:
```suggestion
[#environmentModeBestPractices]
```

PR #1438, opened 2025-03-04
fix: Do not call variable listeners when the solution is in an inconsistent state in undo moves.

Christopher-Chianelli:
Previously, we triggered variable listeners on
Undo(BeforeListVariableChanged) so list variable listeners events are processed immediately (since otherwise, the list change events' start and end do not match the actual list indicies).

However, this causes a problem for swap moves, since variable listeners are triggered when the solution is in an inconsistent state. Consider the following swap move for instance:

```
A = [0]
B = [1]
A[0] <-> B[0]
```

This is converted into two pairs of before/after listVariableChange:

```
beforeListVariableChanged(A, "values", 0, 1);
A.values.set(0, 1);
afterListVariableChanged(A, "values", 0, 1);

beforeListVariableChanged(B, "values", 0, 1);
B.values.set(0, 0);
afterListVariableChanged(B, "values, 0, 1);
```

When the move is undo, the listVariableChange events are replayed in reverse, and beforeListVariableChanged always triggered variable listeners. In the above example,
`beforeListVariableChanged(B, "values", 0, 1)` causes an issue, since when variable listeners are triggered, the solution has the following state:

```
A = [1]  // A[0] 0 -> 1 was not undone yet (and thus A has value 1)
B = [1]  // B[0] 1 -> 0 was undone (and thus B has value 1)
```

There are two major issues with this inconsistent state:

- 1 appears in two seperate entity lists, violating the semantics of list variable

- 0 appears in no entity lists despite not being unassigned.

This did not cause issues before since most variable listeners just look at values on the changed entity, which are in a consistent state. However, this causes issue for variable listeners that check multiple entities, such as declarative shadow variables.

To fix this, we remove the triggerVariableListeners call in Undo(BeforeListVariableChanged). To ensure list change indicies line up, we record whenever the user called triggerVariableListeners and add them as a variable change event. This works since in order to correctly implement a move, the user MUST call triggerVariableListeners everytime they modify the same entity list, and the user MUST call triggerVariableListeners when the solution is in a consistent state.


triceo (commented):
Well spotted!

Only have two comments - first of all, please fix formatting, so that we can see the entire CI run.

More importantly though, can we test this without running the solver? Here's why I don't like testing things by running the solver:
- It usually takes too long. Unit tests are supposed to be quick.
- It's not immediately clear what's being tested. The real behavior is several levels of architecture below what the test does, and the test barely gives an indication.
- As we make unrelated changes to the solver, its behavior may change and the test will silently no longer reproduce the issue. (Perhaps because the moves will no longer do exactly what they do now.)

Here's what I propose we do instead:
- Create an instance of the score director, put a tailored solution in, run a tailored move and an undo. The fail will happen immediately.
- If you foresee multiple different situations where this happens, create test for each.
- The variable listener you implemented which checks for these changes, I think that is very valuable piece of code. Maybe we finish https://github.com/TimefoldAI/timefold-solver/pull/1260 and make these fail-fasts part of it?

triceo:
Any reason why the tests do not use a real score director?

IMO the real value of this test is to check that the score director will do the right thing. Mocking away all of its logic like this is IMO counter-productive and the fact that you have to mock internal details (like the descriptor cache) proves the point I think. The test shouldn't care about this, the test needn't know about this.

triceo:
The Python fail is likely related to the other bug we have yet to merge a fix for.

triceo (approved):

PR #1463, opened 2025-03-19
chore: first change move using move streams

triceo:


core/src/main/java/ai/timefold/solver/core/impl/bavet/AbstractSession.java
```diff
@@ -3,32 +3,52 @@
...
 
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
 import ai.timefold.solver.core.impl.bavet.uni.AbstractForEachUniNode;
+import ai.timefold.solver.core.impl.bavet.uni.AbstractForEachUniNode.LifecycleOperation;
 
 public abstract class AbstractSession {
 
     private final NodeNetwork nodeNetwork;
-    private final Map<Class<?>, AbstractForEachUniNode<Object>[]> effectiveClassToNodeArrayMap;
+    private final Map<Class<?>, AbstractForEachUniNode<Object, Object>[]> initializeEffectiveClassToNodeArrayMap;
```
zepfred, line 13: Cool!

core/src/main/java/ai/timefold/solver/core/impl/bavet/AbstractSession.java
```diff
@@ -3,32 +3,52 @@
...
         this.nodeNetwork = nodeNetwork;
-        this.effectiveClassToNodeArrayMap = new IdentityHashMap<>(nodeNetwork.forEachNodeCount());
+        this.initializeEffectiveClassToNodeArrayMap = new IdentityHashMap<>(nodeNetwork.forEachNodeCount());
+        this.insertEffectiveClassToNodeArrayMap = new IdentityHashMap<>(nodeNetwork.forEachNodeCount());
+        this.updateEffectiveClassToNodeArrayMap = new IdentityHashMap<>(nodeNetwork.forEachNodeCount());
+        this.retractEffectiveClassToNodeArrayMap = new IdentityHashMap<>(nodeNetwork.forEachNodeCount());
+    }
+
+    public final void initialize(Object workingSolution) {
+        for (var node : findNodes(PlanningSolution.class, LifecycleOperation.INITIALIZE)) {
```
zepfred, line 27: We are passing an annotation class here. I'm unsure how `nodeNetwork.getForEachNodes()` will locate the solution class.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractUniDataStream.java
```diff
@@ -83,7 +80,7 @@ public UniDataStream<Solution_, A> addNull() {
         throw new UnsupportedOperationException();
     }
 
-    public AbstractDataset<Solution_, UniTuple<A>> createDataset() {
+    public UniDataset<Solution_, A> createDataset() {
```
zepfred, line 83: Should we avoid the use of specialized classes here?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DataStreamFactory.java
```diff
@@ -10,24 +9,23 @@
...
-import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.SolutionExtractor;
+import ai.timefold.solver.core.impl.move.streams.FromSolutionValueCollectingFunction;
 import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.UniDataStream;
 
-import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullMarked;
 
-public final class DefaultDataStreamFactory<Solution_> implements DataStreamFactory<Solution_> {
+@NullMarked
+public final class DataStreamFactory<Solution_> {
```
zepfred, line 18: Should this class be abstract, or will there be no default implementation?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/ForEachDataStream.java
```diff
@@ -12,48 +11,40 @@
...
+@NullMarked
 public final class ForEachDataStream<Solution_, A>
         extends AbstractUniDataStream<Solution_, A>
         implements TupleSource {
 
     private final Class<A> forEachClass;
-    private final Predicate<A> filter;
-    private final SolutionExtractor<Solution_, A> extractor;
-    private final Collection<A> source;
+    private final @Nullable FromSolutionValueCollectingFunction<Solution_, A> valueCollectingFunction;
```
zepfred, line 26: Why not to add specialized classes, one for the predicate and another for the FromSolutionValueCollectingFunction?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/BiMoveProducer.java
```diff
@@ -0,0 +1,136 @@
...
+        return new BiMoveIterable((DefaultMoveStreamSession<Solution_>) moveStreamSession);
+    }
+
+    @Override
+    public void collectActiveDataStreams(Set<AbstractDataStream<Solution_>> activeDataStreamSet) {
+        aDataset.collectActiveDataStreams(activeDataStreamSet);
+        bDataset.collectActiveDataStreams(activeDataStreamSet);
+    }
+
+    private final class BiMoveIterator implements Iterator<Move<Solution_>> {
```
zepfred, line 47: Nice!

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -81,11 +81,11 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
     private int workingInitScore = 0;
     private String undoMoveText;
 
-    // Null when tracking disabled
-    private final boolean trackingWorkingSolution;
     private final SolutionTracker<Solution_> solutionTracker;
     private final MoveDirector<Solution_> moveDirector = new MoveDirector<>(this);
 
+    // Null when tracking disabled
+    private final boolean trackingWorkingSolution;
```
zepfred, line 88: It cannot be null

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -59,6 +59,7 @@ public BavetConstraintStreamScoreDirector(
     @Override
     public void setWorkingSolution(Solution_ workingSolution) {
         session = scoreDirectorFactory.newSession(workingSolution, constraintMatchPolicy, derived);
+        session.initialize(workingSolution);
```
zepfred, line 62: Is this necessary at this stage of the implementation?

zepfred:
What is our plan for addressing the issues with Sonar?

zepfred (approved):
This new feature looks great!

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/DataStreamFactory.java
triceo: At the moment, I'm only doing things I need. If I ever need a public and private version of this, I'll change it - right now, it looks as if it will be enough if this is completely in impl, and none of it in API. Therefore no need to create abstractions.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/ForEachDataStream.java
triceo: Consistency with the rest of the streams code. (CS also does it like that.) I guess the original idea was to avoid the boilerplate of dozens of different stream classes, but I don't really remember anymore.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
triceo: It was in a previous iteration of this PR, which was doing slightly more work than this version does. Well spotted, I'll remove it.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractUniDataStream.java
triceo: Makes sense to me this way - it's a uni stream, therefore it can afford to return a uni dataset. No need to cast it later.

PR #1479, opened 2025-03-25
fix: register TimefoldJacksonModule for native

winklerm:
Register TimefoldJacksonModule for native compilation.

Switched from legacy quarkus-resteasy to quarkus-rest because of difference in service loading [1] that caused existing tests not to detect the problem with native compilation.

[1] https://quarkus.io/guides/rest-migration#service-loading

Related issue: https://github.com/TimefoldAI/timefold-solver/issues/1482


triceo:
Any chance we could have a test? We have a whole bunch of coverage for Quarkus and Spring Boot which deals with native, and could be extended to deal with whatever the issue is/was here.

winklerm:
Sure, will check the tests and add a new one.

winklerm:
The existing tests did not catch this problem because they use `quarkus-resteasy`, with `quarkus-rest` the problem appears since it does not support service loading: https://quarkus.io/guides/rest-migration#service-loading

I have switched the code to the recommended `quarkus-rest`; however, I am not sure if it is ok to do that? If it is, I will send a PR for timefold-solver-enterprise. If not, we would need to add another test module for `quarkus-rest`.

quarkus-integration/quarkus-jackson/deployment/src/main/java/ai/timefold/solver/quarkus/jackson/deployment/TimefoldJacksonProcessor.java
```diff
@@ -15,7 +17,14 @@ FeatureBuildItem feature() {
...
     @BuildStep
-    void registerTimefoldJacksonModule(BuildProducer<AdditionalBeanBuildItem> additionalBeans) {
+    ClassPathJacksonModuleBuildItem registerTimefoldJacksonModule() {
+        // Make timefold-solver-jackson discoverable by quarkus-rest
+        // https://quarkus.io/guides/rest-migration#service-loading
+        return new ClassPathJacksonModuleBuildItem(TimefoldJacksonModule.class.getName());
+    }
+
+    @BuildStep
+    void registerTimefoldObjectMapperCustomizer(BuildProducer<AdditionalBeanBuildItem> additionalBeans) {
```
winklerm, line 27: In fact, this is no longer needed, as well as `TimefoldObjectMapperCustomizer`.

triceo: I'm good with removing this.

triceo:
I honestly don't know enough about this to understand what the difference between `quarkus-rest` and `quarkus-resteasy` is. Is it a backward compatible change? Will we break anyone?

triceo:
Having discussed with the platform folks, I don't mind the migration to quarkus-rest, provided you:

- Made sure we have replaced all the -resteasy artifacts with their -rest counterparts. Incl. in the quickstarts, if there are any such dependencies.
- Seen quickstarts work, by opening the UI of one or two, and solving a problem.

winklerm:
Thanks, on it!

winklerm:
Created PRs for [quickstarts](https://github.com/TimefoldAI/timefold-quickstarts/pull/763) and [timefold-solver-enterprise](https://github.com/TimefoldAI/timefold-solver-enterprise/pull/256).

As for the `TimefoldObjectMapperCustomizer` removal, I am double checking if it is fine to have a Quarkus extension with deployment and no runtime.

winklerm:
`TimefoldObjectMapperCustomizer` removed, ready for review.

PR #1480, opened 2025-03-25
fix: call cascade listeners for unassigned values

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -57,6 +60,7 @@ public static <Solution_> VariableListenerSupport<Solution_> create(InnerScoreDi
                 .toList();
         this.listVariableDescriptor = scoreDirector.getSolutionDescriptor().getListVariableDescriptor();
         this.listVariableEventList = new ArrayList<>();
+        this.unassignedValueWithEmptyInverseEntitySet = new HashSet<>();
```
triceo, line 63: Arguably this should be an identity hash set; no entities should be considered equal here. (The class doesn't exist in the JDK, but it can be created by converting `IdentityHashMap`.)

triceo: Also however, for reproducibility, it should be a _linked_ hash set. (So that iteration order is stable.) This creates a problem, because there is no way to have a linked identity hash set, unless you implement one from scratch.

triceo: Finally, the better place for this information would be `ExternalizedListVariableStateSupply`, which already keeps all information about list variables. It doesn't have a collection of unassigned elements, but if we're adding it somewhere, it should be there. `VariableListenerSupport` can still do what it does, but it should read the information from the supply.

triceo (commented):
Will need refactoring.

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityHashSet.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Collection;
+import java.util.IdentityHashMap;
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.Set;
+
+public class IdentityHashSet<V> implements Set<V> {
+
+    private ElementAwareList<V> delegate;
+    private IdentityHashMap<V, ElementAwareListEntry<V>> identityMap;
```
triceo, line 12: Let's make them final. Clearing is cheap, final is fast.

core/src/main/java/ai/timefold/solver/core/impl/util/IdentityHashSet.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public IdentityHashSet() {
```
triceo, line 14: Should be called `LinkedIdentityHashSet`.

triceo (commented):
I'll run some benchmarks overnight.
This is likely to affect performance, I want to know by how much.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -57,6 +60,7 @@ public static <Solution_> VariableListenerSupport<Solution_> create(InnerScoreDi
                 .toList();
         this.listVariableDescriptor = scoreDirector.getSolutionDescriptor().getListVariableDescriptor();
         this.listVariableEventList = new ArrayList<>();
+        this.unassignedValueWithEmptyInverseEntitySet = new IdentityHashSet<>();
```
triceo, line 63: We should add a comment here, explaining why we chose this particular type.
(It's not obvious.)

triceo:
There is a performance regression, which is unavoidable. The code now does more than it did before. Performance regressions are acceptable, if they fix incorrect behavior.

However, I made a whole bunch of changes to _minimize_ the regression. There is a simple underlying principle to all of my changes: don't do things that don't need doing. Please review the changes and ask questions in case anything is unclear.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -240,61 +261,83 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
-        if (listVariableEventList.isEmpty() || cascadingUpdateShadowVarDescriptorList.isEmpty()) {
-            return;
-        }
         for (var cascadingUpdateShadowVariableDescriptor : cascadingUpdateShadowVarDescriptorList) {
-            for (var event : listVariableEventList) {
-                var values = listVariableDescriptor.getValue(event.entity());
-                // Evaluate all elements inside the range
-                evaluateFromIndex(values, event.fromIndex(), event.toIndex(), true, cascadingUpdateShadowVariableDescriptor);
-                // Evaluate later elements, but stops when there is no change
-                evaluateFromIndex(values, event.toIndex(), values.size(), false, cascadingUpdateShadowVariableDescriptor);
```
triceo, line 277: Both of these separate iterations have been merged together to just one iteration.
If you can iterate once, iterate once; useful for CPU caches.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -240,61 +261,83 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
+            // Outside the range, only update while the values keep changing.
+            var forceUpdate = currentIndex < toIndex;
+            if (!cascadingUpdateShadowVariableDescriptor.update(scoreDirector, value) && !forceUpdate) {
                 break;
             }
-            lastUpdated++;
+        }
+    }
+
+    private void cascadeUnassignedValues(
```
triceo, line 330: I only split this into a separate method so that I can see its impact in the profiler individually. This is not actually a perf optimization, just helps with profiling.

core/src/main/java/ai/timefold/solver/core/impl/util/LinkedIdentityHashSet.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public int size() {
+        return size;
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        if (size == 0) { // Micro-optimization; contains() on an empty map is not entirely free.
+            return false;
+        }
```
triceo, line 40: These zero checks are speculative; but if you take a look at what the map does even when it's empty, I bet this will prove itself if the methods are called often enough.

core/src/main/java/ai/timefold/solver/core/impl/util/LinkedIdentityHashSet.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public boolean add(V v) {
+        var entry = identityMap.get(v);
+        if (entry == null) {
+            identityMap.put(v, delegate.add(v));
+            size += 1;
+            return true;
+        }
```
triceo, line 51: computeIfAbsent does the same thing, but it also forces the JVM to create a lambda instance. This does not.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareList.java
```diff
@@ -155,9 +155,18 @@ public void forEach(Consumer<? super T> tupleConsumer) {
      */
     @Override
     public Iterator<T> iterator() {
+        if (size == 0) {
+            return Collections.emptyIterator();
```
triceo, line 159: Reduces allocations in some pathological cases.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -25,38 +28,51 @@
...
         this.listVariableDescriptor = scoreDirector.getSolutionDescriptor().getListVariableDescriptor();
-        this.listVariableEventList = new ArrayList<>();
+        this.cascadingUpdateShadowVarDescriptorList =
+                listVariableDescriptor != null ? scoreDirector.getSolutionDescriptor().getEntityDescriptors().stream()
+                        .flatMap(e -> e.getDeclaredCascadingUpdateShadowVariableDescriptors().stream())
+                        .toList() : Collections.emptyList();
+        var hasCascadingUpdates = !cascadingUpdateShadowVarDescriptorList.isEmpty();
+        this.listVariableChangedNotificationList = hasCascadingUpdates ? new ArrayList<>() : Collections.emptyList();
+        this.unassignedValueWithEmptyInverseEntitySet =
+                hasCascadingUpdates ? new LinkedIdentityHashSet<>() : Collections.emptySet();
```
triceo, line 75: Don't create what doesn't need creating.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -215,14 +233,17 @@ public void beforeVariableChanged(VariableDescriptor<Solution_> variableDescript
...
         if (!notifiables.isEmpty()) {
             ListVariableNotification<Solution_> notification = Notification.elementUnassigned(element);
             for (var notifiable : notifiables) {
                 notifiable.notifyAfter(notification);
             }
             notificationQueuesAreEmpty = false;
         }
+        if (!cascadingUpdateShadowVarDescriptorList.isEmpty()) { // Only necessary if there is a cascade.
+            unassignedValueWithEmptyInverseEntitySet.add(element);
+        }
```
triceo, line 246: Don't use what doesn't need using.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -240,61 +261,83 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
-            ListVariableNotification<Solution_> notification = Notification.listVariableChanged(entity, fromIndex, toIndex);
             for (var notifiable : notifiables) {
                 notifiable.notifyAfter(notification);
             }
             notificationQueuesAreEmpty = false;
         }
-        listVariableEventList.add(new ListVariableEvent(entity, fromIndex, toIndex));
+        if (!cascadingUpdateShadowVarDescriptorList.isEmpty()) { // Only necessary if there is a cascade.
+            listVariableChangedNotificationList.add(notification);
+        }
```
triceo, line 273: No need to create the event if we already have the notification.
Help the GC do less work, help yourself.

triceo:
I'll have the final benchmark results in a few hours.

triceo (approved):
I wasn't able to measure any regression anymore that I could attribute to this PR. LGTM after the Sonar issues are fixed.

PR #1493, opened 2025-03-31
chore: expose Move Streams as a experimental feature

triceo:
This feature is under heavy development.
Any use is discouraged.


core/src/main/java/ai/timefold/solver/core/config/localsearch/LocalSearchPhaseConfig.java
```diff
@@ -27,14 +27,17 @@
...
 import ai.timefold.solver.core.config.util.ConfigUtils;
+import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.MoveProviders;
 
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
 @XmlType(propOrder = {
         "localSearchType",
         "moveSelectorConfig",
+        "moveProvidersClass",
```
zepfred, line 40: Should it be named `moveProviderClass`?

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/config/localsearch/LocalSearchPhaseConfig.java
triceo: Since the class is actually called "MoveProviders", I think this is a correct field name.
But I have the name of the class, so the field name will eventually change as well.

PR #1495, opened 2025-03-31
chore: move streams dataset automatically contains null if it needs to

triceo:


core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -160,6 +160,4 @@ default UniDataStream<Solution_, A> ifNotExistsOtherIncludingUnassigned(Class<A>
         return ifNotExistsIncludingUnassigned(otherClass, allJoiners);
     }
 
-    UniDataStream<Solution_, A> addNull();
```
triceo, line 163: This is no longer necessary...

core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ChangeMoveProviderTest.java
```diff
@@ -79,8 +81,67 @@ void fromSolution() {
...
+
+        var secondMove = moveList.get(1);
+        assertSoftly(softly -> {
+            softly.assertThat(secondMove.extractPlanningEntities())
+                    .containsExactly(firstEntity);
+            softly.assertThat(secondMove.extractPlanningValues())
+                    .containsExactly(secondValue);
+        });
+
+        var thirdMove = moveList.get(2);
```
zepfred, line 126: Please provide a comment on this test. I needed to run it to realize that entity 1 holds the value 1.

zepfred (approved):
Looks good!

PR #1501, opened 2025-04-04
chore: make init score inaccessible to users

triceo:
The init score exposed by the Score interface is now always zero. To find out when the solver has finished initializing a solution, users are encouraged to catch the initialization event thrown by the solver manager.

Internally, the init score remains. A new type was created (InnerScore), which now carries the information that init score used to carry. All code which requires to know the initialization status of the solution now has to use InnerScore. Code which does not care about this information can continue using Score.


core/src/main/java/ai/timefold/solver/core/api/score/buildin/hardmediumsoft/HardMediumSoftScore.java
```diff
@@ -24,29 +24,30 @@
...
+    public static final @NonNull HardMediumSoftScore ONE_MEDIUM = new HardMediumSoftScore(0, 1, 0);
+    private static final @NonNull HardMediumSoftScore MINUS_ONE_MEDIUM = new HardMediumSoftScore(0, -1, 0);
+    public static final @NonNull HardMediumSoftScore ONE_SOFT = new HardMediumSoftScore(0, 0, 1);
+    private static final @NonNull HardMediumSoftScore MINUS_ONE_SOFT = new HardMediumSoftScore(0, 0, -1);
 
     public static @NonNull HardMediumSoftScore parseScore(@NonNull String scoreString) {
         String[] scoreTokens = ScoreUtil.parseScoreTokens(HardMediumSoftScore.class, scoreString,
                 HARD_LABEL, MEDIUM_LABEL, SOFT_LABEL);
-        int initScore = ScoreUtil.parseInitScore(HardMediumSoftScore.class, scoreString, scoreTokens[0]);
         int hardScore = ScoreUtil.parseLevelAsInt(HardMediumSoftScore.class, scoreString, scoreTokens[1]);
```
zepfred, line 38: Should we begin with index 0 instead? If so, I wonder why no test has failed.

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
```diff
@@ -189,32 +167,19 @@ default boolean isZero() {
...
-     *
-     * For solutions which do allow unassigning values,
-     * {@link #initScore()} is always zero and therefore this method always returns true.
-     *
-     * @return true if {@link #initScore()} is 0
+     * @return always true
+     * @deprecated No point in using this method anymore.
      */
+    @Deprecated(forRemoval = true, since = "1.21.0")
     default boolean isSolutionInitialized() {
```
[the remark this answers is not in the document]
triceo: This is an entirely correct observation, and you are right that this would be the natural place to check for solution initialization. But this field is actually a lie. As already explained elsewhere, this method _almost always_ returns `true` - even in cases when the solution is not initialized.

The solver doesn't know how to tell a difference between "variable uninitialized" and "variable intentionally left null". This is not a problem that can be fixed, and therefore this misleading method needs to be eliminated.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -194,29 +195,29 @@ public int getWorkingEntityCount() {
         return solverScope.getWorkingEntityCount();
     }
 
-    public <Score_ extends Score<Score_>> Score_ calculateScore() {
-        return (Score_) solverScope.calculateScore();
+    public InnerScore<?> calculateScore() {
```
zepfred, line 198:
```suggestion
    public <Score_ extends Score<Score_>> InnerScore<Score_> calculateScore() {
```

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractMoveScope.java
```diff
@@ -36,11 +37,16 @@ public Move<Solution_> getMove() {
         return move;
     }
 
-    public Score getScore() {
+    public InnerScore<?> getScore() {
```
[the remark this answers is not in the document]
zepfred: This also applies to the `calculateScore()` method. There are many places where we can eliminate the casting.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -74,8 +75,8 @@ public Long getStartingSystemTimeMillis() {
         return startingSystemTimeMillis;
     }
 
-    public <Score_ extends Score<Score_>> Score_ getStartingScore() {
-        return (Score_) startingScore;
+    public InnerScore<?> getStartingScore() {
```
zepfred, line 78: Same as before

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractStepScope.java
```diff
@@ -29,11 +30,16 @@ public int getStepIndex() {
         return stepIndex;
     }
 
-    public Score<?> getScore() {
+    public InnerScore<?> getScore() {
```
zepfred, line 33: Same as before

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ExhaustiveSearchDecider.java
```diff
@@ -138,7 +139,7 @@ private <Score_ extends Score<Score_>> void doMove(ExhaustiveSearchStepScope<Sol
         if (assertExpectedUndoMoveScore) {
             // In BRUTE_FORCE a stepScore can be null because it was not calculated
             if (stepScope.getStartingStepScore() != null) {
-                scoreDirector.assertExpectedUndoMoveScore(move, (Score_) stepScope.getStartingStepScore(),
+                scoreDirector.assertExpectedUndoMoveScore(move, (InnerScore<Score_>) stepScope.getStartingStepScore(),
```
zepfred, line 142: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ExhaustiveSearchDecider.java
```diff
@@ -153,7 +154,7 @@ private <Score_ extends Score<Score_>> void processMove(ExhaustiveSearchStepScop
         var lastLayer = moveNode.isLastLayer();
         if (!scoreBounderEnabled) {
             if (lastLayer) {
-                Score_ score = phaseScope.calculateScore();
+                InnerScore<Score_> score = (InnerScore<Score_>) phaseScope.calculateScore();
```
zepfred, line 157: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ExhaustiveSearchDecider.java
```diff
@@ -163,24 +164,24 @@ private <Score_ extends Score<Score_>> void processMove(ExhaustiveSearchStepScop
...
             if (lastLayer) {
                 // There is no point in bounding a fully initialized score
-                phaseScope.registerPessimisticBound(score);
-                bestSolutionRecaller.processWorkingSolutionDuringMove(score, stepScope);
+                phaseScope.registerPessimisticBound(innerScore);
+                bestSolutionRecaller.processWorkingSolutionDuringMove(innerScore, stepScope);
             } else {
                 var scoreDirector = phaseScope.getScoreDirector();
-                var optimisticBound = (Score_) scoreBounder.calculateOptimisticBound(scoreDirector, score);
+                var optimisticBound = (InnerScore<Score_>) scoreBounder.calculateOptimisticBound(scoreDirector, innerScore);
```
zepfred, line 178: Same as before

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ExhaustiveSearchDecider.java
```diff
@@ -163,24 +164,24 @@ private <Score_ extends Score<Score_>> void processMove(ExhaustiveSearchStepScop
...
                 moveNode.setOptimisticBound(optimisticBound);
-                var bestPessimisticBound = (Score_) phaseScope.getBestPessimisticBound();
+                var bestPessimisticBound = (InnerScore<Score_>) phaseScope.getBestPessimisticBound();
```
zepfred, line 180: Same as before

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/scope/ExhaustiveSearchStepScope.java
```diff
@@ -37,7 +37,7 @@ public void setExpandingNode(ExhaustiveSearchNode expandingNode) {
         this.expandingNode = expandingNode;
     }
 
-    public Score getStartingStepScore() {
+    public InnerScore<?> getStartingStepScore() {
```
zepfred, line 40: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/hillclimbing/HillClimbingAcceptor.java
```diff
@@ -1,19 +1,16 @@
...
-    // ************************************************************************
-    // Worker methods
-    // ************************************************************************
-
+    @SuppressWarnings({ "unchecked", "rawtypes" })
     @Override
     public boolean isAccepted(LocalSearchMoveScope<Solution_> moveScope) {
-        Score moveScore = moveScope.getScore();
-        Score lastStepScore = moveScope.getStepScope().getPhaseScope().getLastCompletedStepScope().getScore();
+        InnerScore moveScore = moveScope.getScore();
```
zepfred, line 12: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/lateacceptance/DiversifiedLateAcceptanceAcceptor.java
```diff
@@ -46,15 +46,15 @@ private void validate() {
...
+    @Override
     public boolean isAccepted(LocalSearchMoveScope<Solution_> moveScope) {
         // The acceptance and replacement strategies are based on the work:
         // Diversified Late Acceptance Search by M. Namazi, C. Sanderson, M. A. H. Newton, M. M. A. Polash, and A. Sattar
         var moveScore = moveScope.getScore();
-        var current = (Score) moveScope.getStepScope().getPhaseScope().getLastCompletedStepScope().getScore();
+        var current = (InnerScore) moveScope.getStepScope().getPhaseScope().getLastCompletedStepScope().getScore();
         var previous = current;
-        var accept = moveScore.compareTo(current) == 0 || moveScore.compareTo(lateWorse) > 0;
+        var accept = moveScore.compareTo(current) == 0 || moveScore.compareTo((InnerScore) lateWorse) > 0;
```
zepfred, line 57: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/AbstractTabuAcceptor.java
```diff
@@ -120,22 +121,21 @@ protected void adjustTabuList(int tabuStepIndex, Collection<? extends Object> ta
...
-            if (moveScope.getScore().compareTo(
-                    moveScope.getStepScope().getPhaseScope().getBestScore()) > 0) {
-                logger.trace("{}        Proposed move ({}) is tabu, but is accepted anyway due to aspiration.",
-                        logIndentation,
-                        moveScope.getMove());
-                return true;
-            }
+        // Natural comparison because shifting penalties don't apply
+        if (aspirationEnabled &&
+                moveScope.getScore().compareTo((InnerScore) moveScope.getStepScope().getPhaseScope().getBestScore()) > 0) {
```
zepfred, line 134: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/score/buildin/BendableScoreDefinition.java
```diff
@@ -67,22 +67,17 @@ public BendableScore fromLevelNumbers(int initScore, Number[] levelNumbers) {
         for (int i = 0; i < softLevelsSize; i++) {
             softScores[i] = (Integer) levelNumbers[hardLevelsSize + i];
         }
-        return BendableScore.ofUninitialized(initScore, hardScores, softScores);
+        return BendableScore.of(hardScores, softScores);
     }
 
     public BendableScore createScore(int... scores) {
-        return createScoreUninitialized(0, scores);
```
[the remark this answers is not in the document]
triceo: `ScoreDefinition` is not public API, therefore I can do anything I like to it. :-)

I kept every method in the public API, because I didn't want to break people compile-time. But this is not a method where we promise backwards compatibility, and therefore deprecation makes no sense.

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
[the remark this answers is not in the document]
zepfred: Understood, I see your point. It appears we have created a new class called `InnerScore` to hold the initialization details. So, we still need to determine if the solution is initialized at some point, and we still have `Score::isSolutionInitialized.` Have you considered creating an abstract base score class (`AbstractScore`) that implements the Score contract and includes an initialization flag?

We could set `AbstractScore#initialized` to true by default. Additionally, we could provide an extra constructor for the score classes to update the initialization flag when necessary. I argue that this could eliminate the need for `InnerScore` and all related changes while still keeping the answer for initialization in the score class.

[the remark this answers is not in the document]
triceo: The way how score is designed, this is not really an option. You don't access score through the `Score` interface; you access the specific final implementation, such as `HardSoftMediumLongScore`. Therefore whatever I put on `AbstractScore` will be available to every user.

(Btw. there still is an `AbstractScore` class, which has been deprecated for years and we still haven't deleted it.)

core/src/main/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirector.java
```diff
@@ -104,9 +102,4 @@ public void updateShadowVariables() {
         scoreDirector.triggerVariableListeners();
     }
 
-    public Score<?> doProblemChange(ProblemChange<Solution_> problemChange) {
```
zepfred, line 107: Is it safe to remove this?

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -172,19 +173,19 @@ public int getWorkingEntityCount() {
         return scoreDirector.getWorkingGenuineEntityCount();
     }
 
-    public Score calculateScore() {
+    public InnerScore<?> calculateScore() {
```
zepfred, line 176: Please see my comment about avoiding the cast to `InnerScore`.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -172,19 +173,19 @@ public int getWorkingEntityCount() {
...
         return scoreDirector.calculateScore();
     }
 
     public void assertScoreFromScratch(Solution_ solution) {
         scoreDirector.getScoreDirectorFactory().assertScoreFromScratch(solution);
     }
 
-    public Score getStartingInitializedScore() {
+    public Score<?> getStartingInitializedScore() {
```
zepfred, line 184: Same as before

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -218,11 +219,16 @@ public void setBestSolution(Solution_ bestSolution) {
         this.bestSolution.set(bestSolution);
     }
 
-    public Score getBestScore() {
+    public InnerScore<?> getBestScore() {
```
zepfred, line 222: Same as before

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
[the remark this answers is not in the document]
zepfred: It appears that there is no option other than the creation of `InnerScore `.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/result/SingleBenchmarkResult.java
```diff
@@ -49,6 +49,7 @@ public class SingleBenchmarkResult implements BenchmarkResult {
     private Integer failureCount = null;
     private Score totalScore = null;
     private Score averageScore = null;
+    private boolean hasUninitializedScores = false;
```
zepfred, line 52: In some places, we use a flag that returns true when the score is initiated. What do you think  abount maintaining this pattern and introducing a flag called `allScoresInitialized`?

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -545,9 +545,6 @@ One way to deal with this is to filter the entity selector of the placer in the
 A planning variable is considered initialized if its value is not `null` or if the variable is ``allowsUnassigned``.
 So a variable which allows unassigned values is always considered initialized.
 
-A planning entity is initialized if all of its planning variables are initialized.
```
zepfred, line 548: Does it no longer make sense?

python/python-core/src/main/python/_solver.py
```diff
@@ -34,19 +34,22 @@ class BestSolutionChangedEvent(Generic[Solution_]):
         - In real-time planning, not all ProblemChanges might be processed:
           check `is_every_problem_change_processed`.
 
-        - This `planning_solution` might be uninitialized: check `Score.init_score`.
+        - This `planning_solution` might be uninitialized: check `is_new_best_solution_initialized`.
 
         - This `planning_solution` might be infeasible: check `Score.is_feasible`.
 
+    is_new_best_solution_initialized : bool
```
zepfred, line 41: Please update the related test case to verify the new flag.

core/src/test/java/ai/timefold/solver/core/impl/score/buildin/BendableBigDecimalScoreDefinitionTest.java
```diff
@@ -96,16 +96,16 @@ void createScore() {
     @Test
     void divideBySanitizedDivisor() {
         BendableBigDecimalScoreDefinition scoreDefinition = new BendableBigDecimalScoreDefinition(1, 1);
-        BendableBigDecimalScore dividend = scoreDefinition.createScoreUninitialized(2, BigDecimal.ZERO, BigDecimal.TEN);
+        BendableBigDecimalScore dividend = scoreDefinition.createScore(BigDecimal.ZERO, BigDecimal.TEN);
```
zepfred, line 99: We are modifying several test files that can be simplified by using the `var` pattern. However, the real question is: do we want to apply this pattern to all those files?

core/src/main/java/ai/timefold/solver/core/api/solver/event/BestSolutionChangedEvent.java
```diff
@@ -66,6 +79,10 @@ public long getTimeMillisSpent() {
         return newBestScore;
     }
 
+    public boolean isNewBestSolutionInitialized() {
```
zepfred, line 82: Please add documentation for the new method. I couldn't find any tests using this new flag. Should we add a test case to cover it?

zepfred (changes requested):
The overall changes look good. +1 after addressing the comments.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/lateacceptance/DiversifiedLateAcceptanceAcceptor.java
triceo: Some of these are not fixable this way. But I'm doing my best to remove as many of these casts as possible.

core/src/main/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirector.java
triceo: Yes. Not public API, unused method.

core/src/main/java/ai/timefold/solver/core/api/score/buildin/hardmediumsoft/HardMediumSoftScore.java
triceo: Good observation! Neither I nor the tests noticed this, because the code was actually written very resiliently. Even though the init score was no longer there, it still allocated the array as if it were - therefore the code didn't notice anything, just that the init level in the array was empty.

Anyway... It's fixed now and the code now only generates the arrays it actually needs. No init score.

core/src/main/java/ai/timefold/solver/core/api/score/AbstractScore.java
```diff
@@ -21,16 +21,22 @@
...
 
-    protected static int parseInitScore(Class<? extends AbstractScore<?>> scoreClass,
-            String scoreString, String initScoreString) {
-        return ScoreUtil.parseInitScore(scoreClass, scoreString, initScoreString);
+    protected static int parseInitScore(Class<? extends AbstractScore<?>> scoreClass, String scoreString,
+            String initScoreString) {
+        try {
+            return Integer.parseInt(initScoreString);
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("The scoreString (" + scoreString
```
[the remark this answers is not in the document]
triceo: This class is deprecated, unused and I won't be touching it.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
triceo: For variables, not really. I moved the text to a section about entities, and slightly improved it.

python/python-core/src/main/python/_solver.py
triceo: There is no test coverage for this event anywhere in the Python solver.
I think I know why - these tests are notoriously unreliable (flaky), even in the Java solver.
I will not be adding one this time.

core/src/test/java/ai/timefold/solver/core/impl/score/buildin/BendableBigDecimalScoreDefinitionTest.java
triceo: I went through every single modified file and converted them to `var`.
When re-reviewing, I recommend going commit-by-commit and skipping the last few, which are nothing but hundreds of lines of `var` replacements.

triceo:
@zepfred All comments resolved, PR finalized.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScore.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public record InnerScore<Score_ extends Score<Score_>>(Score_ raw, int unassignedCount)
+        implements
+            Comparable<InnerScore<Score_>> {
+
+    public static <Score_ extends Score<Score_>> InnerScore<Score_> of(Score_ score) {
+        return ofUninitialized(score, 0);
+    }
+
+    public static <Score_ extends Score<Score_>> InnerScore<Score_> ofUninitialized(Score_ score, int uninitializedCount) {
```
zepfred, line 28: Can this method be called for a model that accepts unassigned values? If that is the case, it should not be considered as an uninitialized solution.

zepfred (approved):
Great job! Please review the Sonar issues before merging.

PR #1509, opened 2025-04-09
fix: do not reset global terminations on phase level

triceo:


core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -280,6 +290,22 @@ public SolverConfig(@NonNull SolverConfig inheritedConfig) {
         inherit(inheritedConfig);
     }
 
+    /**
```
zepfred, line 293: We are using Clock for purposes other than testing. In several locations, `System.currentTimeMillis()` has been replaced with `getSolverScope().getClock().millis()`.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -280,6 +290,22 @@ public SolverConfig(@NonNull SolverConfig inheritedConfig) {
...
+     * For testing purposes only.
```
zepfred, line 294: I would not say it is solely used for tests, as we pass it to the solver factory.

core/src/test/java/ai/timefold/solver/core/impl/solver/termination/TerminationTest.java
```diff
@@ -57,7 +66,51 @@ void stepCountTerminationAtSolverLevel() {
         // 2 entities means 2 steps, but the step count limit is 1.
         // Therefore the best solution is uninitialized, but the score is still just zero.
         // (No init score.)
-        Assertions.assertThat(bestSolution.getScore()).isEqualTo(SimpleScore.ZERO);
+        assertThat(bestSolution.getScore()).isEqualTo(SimpleScore.ZERO);
+    }
+
+    @Test
+    void globalTimeSpentTerminationAtPhaseLevel() throws InterruptedException, ExecutionException {
```
zepfred, line 73: I suggest a second test case where the global termination is less than the LS time.

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
triceo: The solver factory gets the default value, unless the config says something.
If a test needs to override it, it should use the config. Therefore I'd say that the solver config really is _only_ for testing.

PR #1513, opened 2025-04-10
chore: formalize entity inheritance rules

zepfred:


core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/inheritance/multipleinheritance/baseannotated/mixed/childannotated/TestMultipleConstraintProvider.java
```diff
@@ -0,0 +1,30 @@
+package ai.timefold.solver.core.impl.testdata.domain.inheritance.multipleinheritance.baseannotated.mixed.childannotated;
```
triceo, line 1: This package name is so insanely long that I bet this will fail on Windows due to path length limits.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/inheritance/singlevarinterface/baseannotated/childannotated/TestConstraintProvider.java
```diff
@@ -0,0 +1,22 @@
+package ai.timefold.solver.core.impl.testdata.domain.inheritance.singlevarinterface.baseannotated.childannotated;
```
triceo, line 1: Example of shortening: 
```suggestion
package ai.timefold.solver.core.impl.testdata.domain.inheritance.singlevarinterface.baseannotated.childnot;
```

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/inheritance/multipleinheritanceinterface/baseannotated/childnotannotated/TestdataBaseEntity.java
```diff
@@ -0,0 +1,20 @@
+package ai.timefold.solver.core.impl.testdata.domain.inheritance.multipleinheritanceinterface.baseannotated.childnotannotated;
```
triceo, line 1: Example of shortening:

```suggestion
package ai.timefold.solver.core.impl.testdata.domain.inheritance.multiple.interface.baseannotated.childnot;
```

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/inheritance/singlevarinterface/baseannotated/childannotatedaddvar/TestdataSolution.java
```diff
@@ -0,0 +1,86 @@
+package ai.timefold.solver.core.impl.testdata.domain.inheritance.singlevarinterface.baseannotated.childannotatedaddvar;
```
triceo, line 1: Example of shortening:

```suggestion
package ai.timefold.solver.core.impl.testdata.domain.inheritance.singlevarinterface.baseannotated.childtoo.addvar;
```

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/manytomany/TestdataManyToManyShadowedEntityUniqueEvents.java
```diff
@@ -24,25 +21,6 @@ public TestdataManyToManyShadowedEntityUniqueEvents(String code, TestdataValue p
...
-    public TestdataValue getSecondaryValue() {
-        return super.getSecondaryValue();
-    }
-
-    @Override
-    @ShadowVariable(variableListenerClass = ComposedValuesUpdatingVariableListener.class, sourceVariableName = "primaryValue")
-    @ShadowVariable(variableListenerClass = ComposedValuesUpdatingVariableListener.class, sourceVariableName = "secondaryValue")
-    public String getComposedCode() {
-        return super.getComposedCode();
-    }
```
triceo, line 44: These changes are worrisome.
Does this mean we will actually be breaking something that worked?

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -2839,3 +2839,240 @@ def generate_demo_data(demo_data: DemoData) -> Timetable:
 
 Usually, most of this data comes from your data layer,
 and your solution implementation just aggregates that data and creates the uninitialized planning entity instances for the solver to plan.
+
+[#planningModelInheritance]
+== Inheritance
```
triceo, line 2844: We'll soon be doing the same exercise for solution. :-) I'll be explicit here and say that this is inheritance of entities and variables.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -210,12 +209,13 @@ SolverConfigBuildItem recordAndRegisterBuildTimeBeans(CombinedIndexBuildItem com
...
-                    + " the Jandex index by using the jandex-maven-plugin in that dependency, or by adding"
-                    + "application.properties entries (quarkus.index-dependency.<name>.group-id"
-                    + " and quarkus.index-dependency.<name>.artifact-id).");
+            log.warn(
+                    """
+                            Skipping Timefold extension because there are no @%s or @%s annotated classes.
+                            If your domain classes are located in a dependency of this project, maybe try generating the \
+                            Jandex index by using the jandex-maven-plugin in that dependency, or by addingapplication.properties entries \
+                            (quarkus.index-dependency.<name>.group-id and quarkus.index-dependency.<name>.artifact-id)."""
+                            .formatted(PlanningSolution.class.getSimpleName(), PlanningEntity.class.getSimpleName()));
```
triceo, line 218: This is a logging message. Instead of using our own formatting, use the logger's native way of passing parameters to log messages.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -350,11 +352,6 @@ Some solver configs (%s) don't specify a %s class, yet there are multiple availa
                     "Unused classes ([%s]) found with a @%s annotation.".formatted(String.join(", ", unusedSolutionClassList),
                             PlanningSolution.class.getSimpleName()));
         }
-        // Validate the solution classes target types
-        List<AnnotationTarget> targetList = annotationInstanceCollection.stream()
-                .map(AnnotationInstance::target)
-                .toList();
-        assertTargetClasses(targetList, DotNames.PLANNING_SOLUTION);
     }
```
triceo, line 355: Who calls all this removed code now, and where from?

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorImmutablePlanningSolutionTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+class TimefoldProcessorImmutablePlanningSolutionTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .overrideConfigKey("quarkus.timefold.solver.termination.best-score-limit", "0")
```
triceo, line 23: Is this necessary if we're not actually going to be solving?
Let's not reference things we're not using; it creates problems for when those things need to change, and people start worrying why it's even used here.

triceo (commented):
I'm glad we're formalizing this!

I left several inline comments. Please apply them consistently in all places where they apply - it is not practical for me to point out all the places; you know what you touched, you'll know where it applies.

The testdata package structure needs to change, this is no longer acceptable. The neverending package names etc. Let's start inventing ways of making this shorter. I'd start by:

- Moving testdata domain classes to `ai.timefold.solver.core.testdomain` and the rest to `ai.timefold.solver.core.testutil`.
- Reducing duplication in package names, I made some suggestions.
- Introducing more descriptive class names, because those FQNs are making tests incomprehensible.

Also - albeit I'm not sure what can be done about that - I have to say I think that 3000 lines of testdata boilerplate is a bit too much to add for a relatively minor PR as this one. (I have not actually reviewed those classes.) The entire new shadow var subsystem is only twice that.

Finally, before this can be merged, we need to do our homework in validation:
- Make sure every single quickstart works, incl. Python.
- Make sure FSR works.
- Make sure ESS works.
- Talk to Radovan, explain the new rules, make sure they don't break his new extensible models.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/manytomany/TestdataManyToManyShadowedEntityUniqueEvents.java
zepfred: Yes, these changes broke the rule regarding replacing planning variables. I simply removed them, and it did not disrupt the related tests in `CustomVariableListenerTest`.

triceo: I wonder why that was even possible. There were tests for it, so it was intentional. What use case could there be?

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
zepfred: These assertions are contained within the `SolutionDescriptor` and `EntityDescriptor`. Line #1018 is responsible for instantiating the solution descriptor and triggering the related assertions.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
zepfred: We are using `org.jboss.logging.Logger`, but its API does not provide an easy way to pass parameters.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/manytomany/TestdataManyToManyShadowedEntityUniqueEvents.java
zepfred: Before the proposed changes, it was not possible to inherit planning variables when annotating both models. This is because a planning entity must have at least one planning variable, even if it inherits from another planning entity. Therefore, the only option was to redefine the planning variables and override the necessary logic. As you can see, the removed code simply calls the superclass method. However, the `setComposedCode` method has been overridden to include custom logic.

The new logic enables an entity to inherit planning variables from another entity. Additionally, the tests still pass because the expected behavior remains unchanged.

I believe this test remains relevant as it validates listeners and inheritance, and none of the new tests evaluate the listeners. Also, I understand that the new rule concerning the replacement of planning variables might disrupt the user code. Should we revert that?

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptorValidator.java
```diff
@@ -0,0 +1,151 @@
...
+import ai.timefold.solver.core.api.domain.variable.PiggybackShadowVariable;
+import ai.timefold.solver.core.api.domain.variable.PlanningListVariable;
+import ai.timefold.solver.core.api.domain.variable.PlanningVariable;
+import ai.timefold.solver.core.api.domain.variable.PreviousElementShadowVariable;
+import ai.timefold.solver.core.api.domain.variable.ShadowVariable;
+import ai.timefold.solver.core.config.util.ConfigUtils;
+
+public class EntityDescriptorValidator {
+
+    private static final Class[] PLANNING_ENTITY_ANNOTATION_CLASSES = {
```
triceo, line 26: It only makes sense to extract the validation if it leads to _better_ code. If it creates duplication of some key logic (such as this array), it's not worth it and actually makes the code worse.

core/src/test/java/ai/timefold/solver/core/impl/testdata/domain/shadow/manytomany/TestdataManyToManyShadowedEntityUniqueEvents.java
triceo: No, it's OK.

triceo:
Leaving one more comment, and Sonar makes some valid suggestions.
LGTM after those are addressed, and you confirm that this had happened:

    Finally, before this can be merged, we need to do our homework in validation:

    Make sure every single quickstart works, incl. Python.
    Make sure FSR works.
    Make sure ESS works.
    Talk to Radovan, explain the new rules, make sure they don't break his new extensible models.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptorValidator.java
zepfred: The variables are slightly different. The one from `EntityDescriptorValidator` includes a few more types, and I did that to align with what I found in the Quarkus implementation.

triceo: Any chance we could somehow merge the two arrays? Or rather make one "inherit" from the other?
I can already see us add a thing to one of them, and forget to add it to another - this is a bug waiting to happen.

triceo (approved):

PR #1525, opened 2025-04-16
chore: fix URLs all over the place

triceo:


triceo:
It appears that the Docs build is broken for some reason.
@zepfred Any idea what it means? The docs changes are trivial, I can't image they'd break anything.

zepfred (commented):
What has changed in all these images?

triceo:
The SVGs carry metadata with a URL to a license. So I fixed that too. As far as I know, this has no impact on the rendered PNG, so I didn't rebuild them.

PR #1529, opened 2025-04-16
Better explain the requirements for @PlanningId properties

mthmulders:
The current text states that an `@PlanningId` property must implement `Object#equals()` and `Object#hashCode()`. If they are instantiable types, they automatically implement those methods, although not in accordance with the respective contracts.


triceo (approved):

triceo:
Thanks for the contribution, @mthmulders!

PR #1534, opened 2025-04-17
fix: do not enumerate constraint matches which do not differ

triceo:
In some use cases, we've seen this reduce the size of diffs by as much as 90 %.


winklerm (approved):
Thank you!

zepfred (approved):
LGTM!

PR #1552, opened 2025-04-25
chore: refactor test data models

zepfred:


core/src/test/java/ai/timefold/solver/core/api/solver/SolutionManagerTest.java
```diff
@@ -17,9 +17,6 @@
 import ai.timefold.solver.core.impl.testdata.domain.TestdataEntity;
 import ai.timefold.solver.core.impl.testdata.domain.TestdataSolution;
 import ai.timefold.solver.core.impl.testdata.domain.TestdataValue;
-import ai.timefold.solver.core.impl.testdata.domain.allows_unassigned.TestdataAllowsUnassignedEntity;
```
triceo, line 20: Thinking out loud:

- Package names here are too long, we ought to shorten.
- We have two kinds of testdata - domains and utils.

So what if we:

- Move domains to `ai.timefold.solver.core.testdomain`.
- Move utils to `ai.timefold.solver.core.testutil`.

Notice how I dropped the "impl". That's 5 characters less, and no significant loss of information.
I also dropped "data", another 5 characters less.

triceo: Or maybe... thinking out of the box here... the domains can move to something like `org.acme.testdomain`?

zepfred: I liked the idea of using `org.acme.testdomain`.

zepfred: What about `ai.timefold.solver.core.domain` and `ai.timefold.solver.core.util`?

zepfred: It is already a test resource. Do we really need the prefix `test`?

triceo: Good question. Not sure.

The idea of `testdata` was to have a package which doesn't and couldn't possibly map to anything in the `src/main` of the Solver. `core.test` arguably could (and it does in the `solver-test` module), `core.testdomain` is unlikely to.

`org.acme` removes that problem altogether, but I'm not sure I like it better. It's for sure the shortest.

zepfred: I changed my mind regarding `org.acme.testdomain` and I am unsure about using a completely different pattern.

core/src/test/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptorTest.java
```diff
@@ -5,14 +5,14 @@
...
-import ai.timefold.solver.core.impl.testdata.domain.pinned.extended.TestdataExtendedPinnedEntity;
-import ai.timefold.solver.core.impl.testdata.domain.pinned.extended.TestdataExtendedPinnedSolution;
+import ai.timefold.solver.core.testdomain.TestdataEntity;
+import ai.timefold.solver.core.testdomain.immutable.TestdataSolution;
+import ai.timefold.solver.core.testdomain.inheritance.solution.baseannotated.childtoo.TestdataBothAnnotatedChildEntity;
+import ai.timefold.solver.core.testdomain.inheritance.solution.baseannotated.childtoo.TestdataBothAnnotatedExtendedSolution;
+import ai.timefold.solver.core.testdomain.invalid.noplanningvar.TestdataNoVariableSolution;
+import ai.timefold.solver.core.testdomain.pinned.TestdataPinnedEntity;
+import ai.timefold.solver.core.testdomain.pinned.extended.TestdataExtendedPinnedEntity;
+import ai.timefold.solver.core.testdomain.pinned.extended.TestdataExtendedPinnedSolution;
```
triceo, line 15: So we were not able to get rid of the `extended` domains? There are _so many_ of them that I hoped there would be significant overlap with the `inheritance` domains.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
```diff
@@ -1050,51 +1060,51 @@ private void assertDeepCloningEntityClone(TestdataFieldAnnotatedDeepCloningEntit
...
-            for (List<String> key : originalStringListToStringMap.keySet()) {
+            for (var key : originalStringListToStringMap.keySet()) {
                 assertThat(cloneStringListToStringMap.get(key)).isSameAs(originalStringListToStringMap.get(key));
             }
         }
 
-        Map<String, List<String>> originalStringToStringListMap = originalEntity.getStringToStringListMap();
-        Map<String, List<String>> cloneStringToStringListMap = cloneEntity.getStringToStringListMap();
+        var originalStringToStringListMap = originalEntity.getStringToStringListMap();
+        var cloneStringToStringListMap = cloneEntity.getStringToStringListMap();
```
triceo, line 1100: IMO we need to come up with some sort of a guideline for when we make these `var` refactors.
This PR has 800 affected files. It is near impossible to verify the main changes, if they're intertwined with these additional `var` changes as well.

I have done this in the past too, I'm not assigning any blame. But going forward, I'm thinking we'll all have to have more discipline to only stick to the related changes.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
```diff
@@ -1050,51 +1060,51 @@ private void assertDeepCloningEntityClone(TestdataFieldAnnotatedDeepCloningEntit
...
         if (originalStringToStringListMap == null) {
             assertThat(cloneStringToStringListMap).isNull();
         } else {
             assertThat(cloneStringToStringListMap).isNotSameAs(originalStringToStringListMap);
             assertThat(cloneStringToStringListMap).hasSameSizeAs(originalStringToStringListMap);
-            for (String key : originalStringToStringListMap.keySet()) {
-                assertThat(cloneStringToStringListMap.get(key)).isEqualTo(originalStringToStringListMap.get(key));
+            for (var key : originalStringToStringListMap.keySet()) {
+                assertThat(cloneStringToStringListMap).containsEntry(key, originalStringToStringListMap.get(key));
```
triceo, line 1107: Especially if the PR has changes like this, the review goes from "check imports" to "check 800 files for any possible problem in the test".

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
```diff
@@ -157,12 +157,13 @@ void noProblemFactPropertyWithEasyScoreCalculation() {
...
     @Test
     void extended() {
-        var solutionDescriptor = TestdataAnnotatedExtendedSolution.buildExtendedSolutionDescriptor();
+        var solutionDescriptor = TestdataBothAnnotatedExtendedSolution.buildSolutionDescriptor();
         assertThat(solutionDescriptor.getProblemFactMemberAccessorMap()).isEmpty();
         assertThat(solutionDescriptor.getProblemFactCollectionMemberAccessorMap()).containsOnlyKeys("valueList",
                 "subValueList");
-        assertThat(solutionDescriptor.getEntityMemberAccessorMap()).isEmpty();
-        assertThat(solutionDescriptor.getEntityCollectionMemberAccessorMap()).containsOnlyKeys("entityList", "subEntityList");
+        assertThat(solutionDescriptor.getEntityMemberAccessorMap()).containsOnlyKeys("entity", "subEntity");
```
triceo, line 164: This changes the logic of the test to the exact opposite!
This tells me that the two domains were not actually duplicates.

The test does a different thing now. That should not be allowed.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/mutation/MutationCounterTest.java
```diff
@@ -116,40 +115,48 @@ void countMutationsAll() {
...
         for (Object o : b.getRawEntityList()) {
-            ((TestdataEntity) o).setValue(val2);
+            ((TestdataBothAnnotatedChildEntity) o).setValue(val2);
+            ((TestdataBothAnnotatedChildEntity) o).setSubValue(val2);
         }
-
-        assertThat(mutationCounter.countMutations(a, b)).isEqualTo(entityListSize + subEntityListSize + rawEntityListSize + 2);
+        var baseDescriptorCount = entityListSize + subEntityListSize + rawEntityListSize + 2;
+        var childDescriptorCount = subEntityListSize * 2 + rawEntityListSize * 2 + 2;
+        assertThat(mutationCounter.countMutations(a, b)).isEqualTo(baseDescriptorCount + childDescriptorCount);
```
triceo, line 160: I do not expect tests to change behavior.
If the test needs to change, then the domains were not duplicates.

core/src/test/java/ai/timefold/solver/core/impl/solver/SolverSolutionInheritanceTest.java
```diff
@@ -38,11 +38,11 @@ void testBothClassesAnnotated() {
                 .withSolutionClass(TestdataBothAnnotatedExtendedSolution.class)
                 .withEntityClasses(TestdataBothAnnotatedChildEntity.class)
                 .withConstraintProviderClass(TestdataBothAnnotatedExtendedConstraintProvider.class)
-                .withTerminationConfig(new TerminationConfig().withMoveCountLimit(14L));
+                .withTerminationConfig(new TerminationConfig().withMoveCountLimit(100L));
 
         var problem = TestdataBothAnnotatedExtendedSolution.generateSolution(3, 2);
         var solution = PlannerTestUtils.solve(solverConfig, problem);
-        assertThat(solution.getScore()).isEqualTo(SimpleScore.of(8));
+        assertThat(solution.getScore()).isEqualTo(SimpleScore.of(16));
```
triceo, line 45: Another change in behavior. Although if this is just due to a change in the score calculator, I'm OK with it.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/inheritance/solution/TimefoldProcessorBothClassesAnnotatedTest.java
```diff
@@ -26,7 +26,7 @@ class TimefoldProcessorBothClassesAnnotatedTest {
                     .addClasses(TestdataBothAnnotatedExtendedConstraintProvider.class,
                             TestdataBothAnnotatedConstraintProvider.class, TestdataBothAnnotatedExtendedSolution.class,
                             TestdataBothAnnotatedSolution.class, TestdataBothAnnotatedChildEntity.class,
-                            TestdataBothAnnotatedBaseEntity.class))
+                            TestdataEntity.class))
```
triceo, line 29: Isn't this a change of behavior?
The entity wasn't required previously, now it is. Why?

triceo (changes requested):
I wouldn't expect any changes to assertions. Changes to assertions need to be explained first, except:

- where the assertion changes because it references a package name.
- where the assertion changes compatibly (`isEqualTo(0)` v. `isZero()`).

core/src/test/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptorTest.java
zepfred: Hmm. Missed that one. Let me double-check that.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
zepfred: Makes sense. Should I revert the `var` changes?

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
zepfred: Okay. Should I revert it or adhere to the proposed guidelines in the next related changes?

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptorTest.java
zepfred: I understand your point. However, I had to modify the existing inheritance classes to align with the deleted classes. This is why two new fields were added. I know the model change affects the tests, but the main goal of the test remains the same.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
triceo: No need anymore. I already went through all of it. But in the future, let's try to be cautious about this, on large PRs.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/mutation/MutationCounterTest.java
zepfred: I could simply move the extended classes to the inheritance package, but I attempted to merge some models with similar features. Adding new planning variables to the existing inheritance classes also changed the score, as the constraint was evaluated more times than before.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
triceo: Don't revert anymore. I already reviewed everything.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/mutation/MutationCounterTest.java
zepfred: One solution is to restore the extended classes and merge the packages. What do you think?

triceo: Merging models is fine, I like that we're removing boilerplate code. 
But why does that mean changing behavior?

triceo (approved):

PR #1560, opened 2025-04-28
chore: support pinning in Move Streams

triceo:
By default, enumerations produce results without pinned entities.
If pinned entities are required, they have to be specifically requested.


core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -37,18 +37,23 @@
     /**
      * A pinned planning entity is never changed during planning,
      * this is useful in repeated planning use cases (such as continuous planning and real-time planning).
-     * <p>
      * This applies to all the planning variables of this planning entity.
-     * To pin individual variables, see https://issues.redhat.com/browse/PLANNER-124
      * <p>
      * The method {@link PinningFilter#accept(Object, Object)} returns false if the selection entity is pinned
```
zepfred, line 42: Should it return true if the selection entity is pinned?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractForEachDataStream.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public final void collectActiveDataStreams(Set<AbstractDataStream<Solution_>> dataStreamSet) {
+        dataStreamSet.add(this);
+    }
+
+    @Override
+    public final void buildNode(DataNodeBuildHelper<Solution_> buildHelper) {
+        TupleLifecycle<UniTuple<A>> tupleLifecycle = buildHelper.getAggregatedTupleLifecycle(childStreamList);
+        int outputStoreSize = buildHelper.extractTupleStoreSize(this);
```
zepfred, line 35:
```suggestion
        var outputStoreSize = buildHelper.extractTupleStoreSize(this);
 ```

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
triceo: This is an old class, which I'm now deprecating.
We can not change its behavior due to bwd compat, so even though we may have originally designed it better, there's no point discussing it now.

PR #1564, opened 2025-04-29
fix: Allow using declarative shadow variables on extended entities

Christopher-Chianelli:
- Getters will be preferred over fields unless the field is more
  specific than the getter
    
- Modify how VariableListenerSupport notifies ShadowVariableSession
    
- Made VariableMetaclasses check for equality via variable descriptors;
  previous it checked both the entity metaclass and the variable
  descriptors, which created two different canonical VariableMetaclasses
  or a variable on a base class that is extended.
    
- Fix getting declared fields/getters

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningListVariableMetaModel.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningListVariableMetaModel.java
@@ -1,5 +1,7 @@
 package ai.timefold.solver.core.impl.domain.solution.descriptor;
 
+import java.util.Objects;
+
 import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningEntityMetaModel;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningListVariableMetaModel;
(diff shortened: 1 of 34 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningListVariableMetaModel.java
```diff
@@ -30,6 +32,22 @@ public boolean allowsUnassignedValues() {
         return false;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        // DO NOT USE entity in equality checks;
```
triceo, line 37:
```suggestion
        // Do not use entity in equality checks;
```

Let's not shout at people.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningVariableMetaModel.java
```diff
@@ -35,6 +37,22 @@ public boolean isChained() {
         return variableDescriptor.isChained();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        // DO NOT USE entity in equality checks;
```
triceo, line 42:
```suggestion
        // Do not use entity in equality checks;
```

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultShadowVariableMetaModel.java
```diff
@@ -30,10 +32,26 @@ public ShadowVariableDescriptor<Solution_> variableDescriptor() {
         return variableDescriptor;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        // DO NOT USE entity in equality checks;
```
triceo, line 37:
```suggestion
        // Do not use entity in equality checks;
```

triceo (approved):

triceo:
Please show how both of the examples in the referenced issue can be successfully upgraded to DSVs. Even with this PR, I still failed, no matter what I tried.

triceo (approved):
Since I didn't review this in time before Chris' PTO, I made some changes myself. LGTM now.

PR #1568, opened 2025-04-29
fix: Disable the C2 compiler in Python

### LINKED ISSUE #1561: chore: add support for JDK 24 in the Python module.
**Describe the bug**
The Python CI is failing when using JDK24 and needs further investigation.

**Actual behavior**
It fails without an apparent reason: https://github.com/zepfred/timefold-solver/actions/runs/14709002887

Christopher-Chianelli:
It appears there is a bug in C2 compiler for JDK 24 that cause a crash on exit.

### DIFF
--- a/python/python-core/src/main/python/_timefold_java_interop.py
+++ b/python/python-core/src/main/python/_timefold_java_interop.py
@@ -88,7 +88,9 @@ def init(*args, path: list[str] = None, include_timefold_jars: bool = True) -> N
         path = path + extract_timefold_jars()
     if len(args) == 0:
         args = [get_default_jvm_path()]
-    init(*args, path=path, include_translator_jars=False)
+
+    # JDK 24 bug: Compiler tier 4 (C2) cause a crash at exit
+    init(*args, '-XX:TieredStopAtLevel=3', path=path, include_translator_jars=False)
 
     from ai.timefold.solver.python.logging import PythonDelegateAppender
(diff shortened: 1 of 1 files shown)


python/python-core/src/main/python/_timefold_java_interop.py
```diff
@@ -88,7 +88,9 @@ def init(*args, path: list[str] = None, include_timefold_jars: bool = True) -> N
         path = path + extract_timefold_jars()
     if len(args) == 0:
         args = [get_default_jvm_path()]
-    init(*args, path=path, include_translator_jars=False)
+
+    # JDK 24 bug: Compiler tier 4 (C2) cause a crash at exit
+    init(*args, '-XX:TieredStopAtLevel=3', path=path, include_translator_jars=False)
```
triceo, line 93: Can we do this only when Java is 24?
Also, this cannot be merged, because the tests fail. The failures need to be addressed somehow.

Christopher-Chianelli: Test failures are from the Solver being too slow for integration tests without C2.
As for getting the JVM version, that can only be determined after launching the JVM, at which point, we cannot disable C2 (see https://jpype.readthedocs.io/en/latest/userguide.html#launching-the-jvm).

triceo:
This can not be merged.
We cannot merge something with failing tests, and we do not want to disable C2 and ruin the performance on account of a single version of JVM.

PR #1571, opened 2025-04-30
feat: Testing support for DSV

zepfred:


core/src/main/java/ai/timefold/solver/core/api/solver/SolutionManager.java
```diff
@@ -84,6 +86,28 @@ public interface SolutionManager<Solution_, Score_ extends Score<Score_>> {
...
+     * Unlike the latter method,
+     * it does not require the complete configuration necessary to obtain an instance of {@link SolutionManager}.
+     * <p>
+     * However, this method requires that the entity does not define any shadow variables that rely on listeners,
+     * as it requires a complete solution.
+     *
+     * @param solutionClass the solution class
+     * @param entities all entities to be updated
+     */
+    static <Solution_> void updateShadowVariables(@NonNull Class<Solution_> solutionClass,
```
triceo, line 101: Arguably we want the same, but for an entire solution.
That one can and should run the full update.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+        var session = new InternalShadowVariableSession<>(solutionDescriptor,
+                new VariableReferenceGraph<>(ChangedVariableNotifier.empty()));
+        session.init(entities);
+        // Update all built-in shadow variables
+        session.processBasicVariable(entities);
+        session.processChainedVariable(entities);
+        session.processListVariable(entities);
+        session.processCascadingVariable(entities);
+        session.triggerUpdateShadowVariables();
```
triceo, line 64: I am concerned about this. This code requires a list of all shadow variables and their types. This means that - when we introduce a new type - this code will silently fail unless we remember to update it as well. 

This is brittle, and it leads me to think... why does the solver currently do all of this work without having to have all of this information in a single place? Surely, this has to be code duplication - the solver can already do this, and we just need to hook into the code which already does it now.

triceo (commented):
Although this works and I can see it, I have shared a concern inline which has to be resolved.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
zepfred: The logic is primarily located at `VariableListenerSupport`. We could reuse it by creating a solution instance from the problem entity and facts. Let me check that.

triceo: Or we could _adapt it_ so that only a part of it requires a solution.
I imagine that the full VLS gets all the variables, and the simplified one filters out those impossible without a solution.

triceo: IMO creating new instance of a solution is a dead end. You don't know enough about the solution to be able to do that. We don't enforce public no-arg constructors, we don't enforce mutability... This is just trading one complexity for another.

triceo (commented):
Some comments, mostly about code layout.
Functionality looks good, once we solve the whole solution, or decide we don't do it.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
```diff
@@ -0,0 +1,360 @@
...
+            super(solutionDescriptor);
+        }
+
+        @Override
+        public AbstractScoreDirector.AbstractScoreDirectorBuilder<Solution_, Score_, ?, ?> createScoreDirectorBuilder() {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    private static class InternalScoreDirector<Solution_, Score_ extends Score<Score_>, Factory_ extends AbstractScoreDirectorFactory<Solution_, Score_, Factory_>>
```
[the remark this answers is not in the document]
zepfred: I want to make sure they are not being used unless it is necessary.

triceo (approved):

PR #1582, opened 2025-05-07
fix: avoid more situations where a K-OPT cannot be found

triceo:
CC @rsynek @winklerm


rsynek (approved):

winklerm (approved):
Thank you!

triceo:
@winklerm promised to provide a reproducer later.

PR #1583, opened 2025-05-07
fix: clone the solution before sending it to the consumer

zepfred:
This PR addresses an issue where we return the same solution instance as the SolverScope to external consumers. Therefore, if changes are made to the solution, it may be possible that these changes will affect the solver phases, because the current best solution from the solver scope is restored at the end of each phase.

```
AbstractSolver::runPhases(SolverScope<Solution_> solverScope) {
            ...
            // If there is a next phase, it starts from the best solution, which might differ from the working solution.
            // If there isn't, no need to planning clone the best solution to the working solution.
            if (it.hasNext()) {
                solverScope.setWorkingSolutionFromBestSolution();
            }
```

It is worth noting that I have analyzed the use of `SolverScope::getBestSolution` and external actions. The solution is now cloned before being shared with these external consumers.


core/src/test/java/ai/timefold/solver/core/impl/solver/BestSolutionHolderTest.java
```diff
@@ -186,6 +189,34 @@ void problemChangeBarrageIntermediateBestSolutionConsumer() throws InterruptedEx
...
+                            intermediate -> {
+                                intermediate.getEntityList().add(new TestdataEntity("Bad Entity 3"));
+                                countDownLatch.countDown();
+                            })
+                    .withConfigOverride(new SolverConfigOverride<TestdataSolution>()
+                            .withTerminationConfig(new TerminationConfig().withMoveCountLimit(100L)))
+                    .run()
+                    .getFinalBestSolution();
+            countDownLatch.await();
+            assertThat(solution.getEntityList()).hasSize(2);
```
triceo, line 216: This deserves a comment explaining why the solution only has 2 elements.
(Essentially a comment to explain what the test tries to do.)

triceo (changes requested):
We can still improve this a bit.

triceo (approved):

PR #1592, opened 2025-05-14
perf: improve many aspects of DSV performance

triceo:
Less memory allocations.
Less map access.
Cleaner code structure.

Commits are cleanly isolated, so that it is possible to review each idea independently.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/util/DynamicIntArray.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/util/DynamicIntArray.java
@@ -9,6 +9,11 @@ import java.util.Arrays;
  */
 public final class DynamicIntArray {
 
+    // Growth factor for array expansion; not too much, the point of this class is to avoid excessive memory use.
+    private static final double GROWTH_FACTOR = 1.2;
+    // Minimum capacity increment to avoid small incremental growth
+    private static final int MIN_CAPACITY_INCREMENT = 10;
+
     private final int maxLength;
     private int[] array;
     private int firstIndex;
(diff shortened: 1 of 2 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -0,0 +1,199 @@
...
+        for (var node : entityVariablePairFunction.apply(affectedEntity)) {
+            if (graph.isLooped(loopedTracker, node.graphNodeId())) {
+                isEntityLooped = true;
+                break;
+            }
+        }
+        var oldValue = shadowVariableLoopedDescriptor.getValue(affectedEntity);
+        if (!Objects.equals(oldValue, isEntityLooped)) {
+            // TODO what if we don't let users treat this as a shadow var?
+            //  Nobody can hook up to it => we need not trigger events?
```
Christopher-Chianelli, line 117: We need to trigger events for constraint streams; i.e.
```java
cf.forEach(ShadowEntity.class)
  .filter(ShadowEntity::isEntityLooped)
  .penalize(HardSoftScore.ONE_HARD);
```

Christopher-Chianelli (changes requested):
Changes look good; suggestions within.

triceo:
@Christopher-Chianelli I addressed all your comments.
I also added extra commits with more improvements.

Please check the final commit - it seems to have broken a test, although I wasn't able to figure out how. As far as I can tell, the change is functionally equivalent to what was there previously. Please check if you can find out what went wrong - worst case, I revert the final commit.

(But I'd rather not to, because that commit takes MJS from 4.5G heap permanently occupied down to 3G heap.)

Christopher-Chianelli (approved):

PR #1593, opened 2025-05-15
feat: enable Basic var and List var to coexist for the construction heuristic

zepfred:
This PR allows for the construction of an initial solution using both basic and list variables within the same model. The proposal does not introduce any new configurations for the construction heuristic; rather, it modifies the entity placer definition to accept a list instead of just a single element. 

There are some limitations concerning the definition of the entity placer. Therefore, the current behavior for a single placer configuration will remain unchanged. However, a new option has been added to define a placer for values (list variable) and a placer for entities (basic variable).

```
  <constructionHeuristic>
    <!-- List Variable -->
    <queuedValuePlacer>
      <valueSelector id="placerValueSelector">
        <cacheType>PHASE</cacheType>
        <selectionOrder>SORTED</selectionOrder>
        <sorterManner>INCREASING_STRENGTH</sorterManner>
      </valueSelector>
      <changeMoveSelector>
        <entitySelector>
          <cacheType>PHASE</cacheType>
          <selectionOrder>SORTED</selectionOrder>
          <sorterManner>DECREASING_DIFFICULTY</sorterManner>
        </entitySelector>
        <valueSelector mimicSelectorRef="placerValueSelector"/>
      </changeMoveSelector>
    </queuedValuePlacer>
    <!-- Basic Variable -->
    <queuedEntityPlacer>
      <entitySelector id="placerEntitySelector">
        <cacheType>PHASE</cacheType>
        <selectionOrder>SORTED</selectionOrder>
        <sorterManner>DECREASING_DIFFICULTY</sorterManner>
      </entitySelector>
      <changeMoveSelector>
        <entitySelector mimicSelectorRef="placerEntitySelector"/>
        <valueSelector>
          <cacheType>PHASE</cacheType>
          <selectionOrder>SORTED</selectionOrder>
          <sorterManner>INCREASING_STRENGTH</sorterManner>
        </valueSelector>
      </changeMoveSelector>
    </queuedEntityPlacer>
  </constructionHeuristic>
```

By default, the solving process combines the default strategies from both placers. For each list variable placement, we then test all possible placements for the basic variables and select the best combination. It is important to note that the list variables impose some restrictions on the CH settings, limiting it to only defining `QueuedValuePlacer`. The proposed approach retains these limitations, and the mixed model can only be configured as shown in the previous snippet.


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/placer/QueuedMultiplePlacer.java
```diff
@@ -0,0 +1,213 @@
...
+            this.queuedPlacerList = new ArrayList<>(queuedPlacerList);
+            reset();
+        }
+
+        /**
+         * The method uses a strategy similar to {@link CartesianProductMoveSelector},
+         * but it uses placer iterators instead.
+         */
+        private Move<Solution_> nextMove() {
+            if (cachedMove == null) {
```
triceo, line 65: Invert the logic, avoid nesting.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/placer/QueuedMultiplePlacer.java
```diff
@@ -0,0 +1,213 @@
...
+            return index;
+        }
+
+        /**
+         * Recreate all move iterators starting from #lastValidIteratorIndex.
+         *
+         * @param lastValidIteratorIndex the index of the last iterator that generated a valid move
+         * @param move the move array to be loaded
+         */
+        private boolean recreateNextIterators(int lastValidIteratorIndex, Move<?>[] move) {
```
triceo, line 123: Let's avoid side effects.
The method can return an array on success, and null on failure.
That way, it can create the array internally, keeping logic more focused and isolated.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/placer/QueuedMultiplePlacer.java
```diff
@@ -0,0 +1,213 @@
...
+        private void clearCache() {
+            this.cachedMove = null;
+        }
+
+        private void reset() {
+            placementIteratorList = new ArrayList<>(queuedPlacerList.size());
+            moveIteratorList = new ArrayList<>(queuedPlacerList.size());
+            for (EntityPlacer<Solution_> placements : queuedPlacerList) {
+                moveIteratorList.add(null);
+                placementIteratorList.add(null);
```
triceo, line 149: This is strange code. If the default value has to be null, why not use an array instead? Or at the very least explain that this is done so that you can do `add(i, ...)` later without getting an exception.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -727,11 +727,17 @@ private void validateListVariableDescriptors() {
 
         var listVariableDescriptor = listVariableDescriptorList.get(0);
         var listVariableEntityDescriptor = listVariableDescriptor.getEntityDescriptor();
-        if (listVariableEntityDescriptor.getGenuineVariableDescriptorList().size() > 1) {
+        var hasChained = listVariableEntityDescriptor.getGenuineVariableDescriptorList().stream()
```
triceo, line 730: I think this check needs to be smarter. The chained variable can be defined on _any_ entity in the solution.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -877,6 +883,10 @@ public PlanningSolutionMetaModel<Solution_> getMetaModel() {
         return planningSolutionMetaModel;
     }
 
+    public boolean hasBothBasicAndListVariables() {
+        return getMetaModel().hasBothBasicAndListVariables();
```
triceo, line 887: Careful with the metamodel. It's supposed to be a simple interface for users - not all methods should go there. Only those that the user would actually need. Arguably, this is not one of them.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -119,11 +120,33 @@ public void linkVariableListeners() {
...
+    private boolean isOnlyShadowBasicVariable(ShadowVariableDescriptor<Solution_> shadowVariableDescriptor) {
+        return !(shadowVariableDescriptor instanceof InverseRelationShadowVariableDescriptor<Solution_>)
+                && !isOnlyShadowListVariable(shadowVariableDescriptor);
+    }
+
+    private boolean isOnlyShadowListVariable(ShadowVariableDescriptor<Solution_> shadowVariableDescriptor) {
+        return shadowVariableDescriptor instanceof IndexShadowVariableDescriptor<Solution_> ||
+                shadowVariableDescriptor instanceof PreviousElementShadowVariableDescriptor<Solution_> ||
+                shadowVariableDescriptor instanceof NextElementShadowVariableDescriptor<Solution_>;
+    }
```
triceo, line 148: Arguably these methods belong on the variable descriptor directly.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -156,6 +179,23 @@ private void processShadowVariableDescriptorWithListVariable(ShadowVariableDescr
         }
     }
 
+    private void processShadowVariableDescriptorWithBothBasicListVariable(
+            ShadowVariableDescriptor<Solution_> shadowVariableDescriptor,
+            ListVariableStateSupply<Solution_> listVariableStateSupply) {
+        if (shadowVariableDescriptor instanceof InverseRelationShadowVariableDescriptor<Solution_> inverseRelationShadowVariableDescriptor) {
+            // If the source type is a collection, then it is a list variable
+            if (inverseRelationShadowVariableDescriptor.isSourceVariableCollectionType()) {
```
triceo, line 187: Isn't there a more direct way of checking for this? Doesn't the shadow var descriptor point to its list var descriptor, or something like that? This feels like you're checking for a side effect, and not for the root cause.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/metamodel/PlanningEntityMetaModel.java
```diff
@@ -131,6 +131,26 @@ default boolean hasVariable(String variableName) {
         return false;
     }
 
+    /**
+     * Checks whether a basic variable is defined by the entity.
+     * 
+     * @return True if present, false otherwise.
+     */
+    default boolean hasBasicVariable() {
```
triceo, line 139: `PlanningVariable`

core/src/main/java/ai/timefold/solver/core/preview/api/domain/metamodel/PlanningEntityMetaModel.java
```diff
@@ -131,6 +131,26 @@ default boolean hasVariable(String variableName) {
...
+        return variables().stream()
+                .anyMatch(variable -> variable.isGenuine() && !variable.isList());
+    }
+
+    /**
+     * Checks whether a list variable is defined by the entity.
+     * 
+     * @return True if present, false otherwise.
+     */
+    default boolean hasListVariable() {
```
triceo, line 149: `PlanningListVariable`

core/src/test/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseTest.java
```diff
@@ -341,4 +358,164 @@ void constructionHeuristicAllocateToValueFromQueue() {
...
+
+        var problem = TestdataListMultiVarSolution.generateUninitializedSolution(2, 2, 2);
+        var solution = PlannerTestUtils.solve(solverConfig, problem);
+        assertThat(solution.getEntityList().stream()
+                .filter(e -> e.getBasicValue() == null || e.getSecondBasicValue() == null || e.getValueList().isEmpty()))
+                .isEmpty();
+    }
+
+    @Test
+    void solveCustomConfigurationWithListVariables() {
```
triceo, line 496: I'm missing coverage for unassigned values, both basic and list.
I'm missing coverage for pinning.
Both have broken CH in the past, I'd be surprised if they didn't do it again this time.

docs/src/modules/ROOT/pages/optimization-algorithms/construction-heuristics.adoc
```diff
@@ -702,6 +702,76 @@ is supported too.
...
+== Mixed modeling and construction heuristics
+
+If the model includes both xref:using-timefold-solver/modeling-planning-problems.adoc#planningVariable[basic] and xref:using-timefold-solver/modeling-planning-problems#planningListVariable[list] variables,
+the default behavior of the construction heuristic is
+to use strategies based on the allocation of xref:optimization-algorithms/construction-heuristics.adoc#allocateEntityFromQueue[entities] and xref:optimization-algorithms/construction-heuristics.adoc#allocateToValueFromQueue[values]
+for solving all related variables.
+
+=== Algorithm description
+
+The allocations works like this:
```
triceo, line 715:
```suggestion
The allocations work like this:
```

docs/src/modules/ROOT/pages/optimization-algorithms/construction-heuristics.adoc
```diff
@@ -702,6 +702,76 @@ is supported too.
...
+
+. Put all values in a round-robin queue.
+. Put all entities in a second queue.
+. Combine the moves from the allocation of entity and values.
+.. Assign the first entity from the queue to the best value.
+.. Assign the best entity from the second queue to the first value.
+. Repeat until all entities are assigned.
```
triceo, line 722: This deserves a picture. There are plenty of pictures around CH docs, you should be able to reuse them.

triceo (commented):
First round of review finished. Will do a deeper reading when I have more time.

One general comment: until we can deliver this for LS as well, this needs to be disabled - people shouldn't be able to configure this, if it's going to fail later. It also should not be in the docs.

Let's keep the PR as is for now, but before we merge it, we add a kill switch. We will remove the kill switch when we merge the LS part.

Also - did you make sure enterprise works? Including nearby?

zepfred:
> did you make sure enterprise works? Including nearby?

Not yet. I've planned it as a later step for when the LS is working.

benchmark/src/main/resources/benchmark.xsd
```diff
@@ -674,7 +674,7 @@
           <xs:element minOccurs="0" name="valueSorterManner" type="tns:valueSorterManner"/>
                               
           
-          <xs:choice minOccurs="0">
+          <xs:choice maxOccurs="unbounded" minOccurs="0">
```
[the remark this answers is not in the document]
zepfred: Yes, but I didn't find a way to do that when using a list.

core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/ConstructionHeuristicPhaseConfig.java
```diff
@@ -110,12 +111,35 @@ public void setValueSorterManner(@Nullable ValueSorterManner valueSorterManner)
...
 
-    public void setEntityPlacerConfig(@Nullable EntityPlacerConfig entityPlacerConfig) {
-        this.entityPlacerConfig = entityPlacerConfig;
+    /**
+     * @deprecated Use {@link #setEntityPlacerConfigList(List)}} instead.
+     */
+    @Deprecated(forRemoval = true, since = "1.22.0")
+    public void setEntityPlacerConfig(EntityPlacerConfig entityPlacerConfig) {
+        setEntityPlacerConfigList(List.of(entityPlacerConfig));
+    }
```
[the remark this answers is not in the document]
zepfred: Yes, it is. However, we have a new property that is a list and accepts varargs. There are two ways to achieve the same configuration.

core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/ConstructionHeuristicPhaseConfig.java
```diff
@@ -36,7 +37,7 @@
         "constructionHeuristicType",
         "entitySorterManner",
         "valueSorterManner",
-        "entityPlacerConfig",
+        "entityPlacerConfigList",
```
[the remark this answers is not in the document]
zepfred: This configuration remains valid:

```
<?xml version="1.0" encoding="UTF-8"?>
<solver xmlns="https://timefold.ai/xsd/solver" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="https://timefold.ai/xsd/solver https://timefold.ai/xsd/solver/solver.xsd">
    <environmentMode>FULL_ASSERT</environmentMode>
    <constructionHeuristic>
        <queuedEntityPlacer>
            <entitySelector id="placerEntitySelector">
                <cacheType>PHASE</cacheType>
            </entitySelector>
            <cartesianProductMoveSelector>
                <changeMoveSelector>
                    <entitySelector mimicSelectorRef="placerEntitySelector"/>
                    <valueSelector variableName="timeslot">
                        <cacheType>PHASE</cacheType>
                    </valueSelector>
                </changeMoveSelector>
                <changeMoveSelector>
                    <entitySelector mimicSelectorRef="placerEntitySelector"/>
                    <valueSelector variableName="room">
                        <cacheType>PHASE</cacheType>
                    </valueSelector>
                </changeMoveSelector>
            </cartesianProductMoveSelector>
        </queuedEntityPlacer>
    </constructionHeuristic>
    <localSearch>
    </localSearch>
</solver>
```

Why do you think it is not backward compatible?

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -58,9 +58,9 @@ public void solve(SolverScope<Solution_> solverScope) {
 
         var solutionDescriptor = solverScope.getSolutionDescriptor();
         var listVariableDescriptor = solutionDescriptor.getListVariableDescriptor();
-        var hasListVariable = listVariableDescriptor != null;
+        var hasOnlyListVariable = listVariableDescriptor != null && !solutionDescriptor.hasBothBasicAndListVariables();
```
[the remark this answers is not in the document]
zepfred: Yes, it is still necessary because we should enter the if block when there is a list variable.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -58,9 +58,9 @@ public void solve(SolverScope<Solution_> solverScope) {
...
         var maxStepCount = -1;
-        if (hasListVariable) {
+        if (hasOnlyListVariable) {
```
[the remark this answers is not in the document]
zepfred: Good catch! The previous change would have caused it to run endlessly and needs to be reverted.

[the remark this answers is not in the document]
zepfred: We still assign a planning value to the best entity per step.

core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/ConstructionHeuristicPhaseConfig.java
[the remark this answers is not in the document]
triceo: I was assuming that the name `list` would be part of the tag. If it's not, then OK with me.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/ShadowVariableDescriptor.java
```diff
@@ -64,6 +64,11 @@ public boolean hasVariableListener() {
         return true;
     }
 
+    /**
+     * return true if the source variable is a basic variable; otherwise, return false.
+     */
+    public abstract boolean isListVariableSource();
```
triceo, line 70: The name of the method suggests the exact opposite of what the Javadoc says.

core/src/test/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseTest.java
```diff
@@ -342,4 +362,240 @@ void constructionHeuristicAllocateToValueFromQueue() {
...
+                TestdataListMultiVarSolution.class, TestdataListMultiVarEntity.class, TestdataListMultiVarValue.class,
+                TestdataListMultiVarOtherValue.class)
+                .withPhases(new ConstructionHeuristicPhaseConfig()
+                        .withTerminationConfig(new TerminationConfig().withStepCountLimit(16)))
+                .withEasyScoreCalculatorClass(TestdataListMultiVarEasyScoreCalculator.class);
+
+        var problem = TestdataListMultiVarSolution.generateUninitializedSolution(2, 2, 2);
+        // Pin the first entity
+        problem.getEntityList().get(0).setPinned(true);
+        problem.getEntityList().get(0).setPinnedIndex(0);
```
triceo, line 451: Let's also add a test which includes a partially pinned list variable.
And a test with an entity with a basic variable, which is pinned.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/ShadowVariableDescriptor.java
zepfred: Yep. It is wrong.

docs/src/modules/ROOT/pages/optimization-algorithms/construction-heuristics.adoc
```diff
@@ -702,6 +702,76 @@ is supported too.
 For scaling out, see <<scalingConstructionHeuristics,scaling construction heuristics>>.
 
 
+[#mixedModelConstructionHeuristics]
+== Mixed modeling and construction heuristics
```
triceo, line 706: I wonder if this should be here now.
Yes, it should be here when the feature is fully delivered, but now when it only works for CH? Maybe let's not tell people to start using it, if they can really only use it for CH and not for LS.

triceo (approved):
One comment, otherwise LGTM.

PR #1606, opened 2025-05-23
feat: enable Basic var and List var to coexist for the local search

zepfred:
This PR completes the feature and enables the mixed model for local search.


zepfred:
@triceo I plan to add documentation about the mixed model once I finish the quick-start change, as I'll have concrete models to use in the explanation.

triceo (commented):
This continues to be suspiciously simple. I am really surprised.

Did you test what happens when strength/difficulty comparisons are used? Did you check entity/value sorter manner? We do not support some of it with list variables, and we need to make sure that this does not change here; and the use cases which work still need to work in the mixed mode.

zepfred:
> This continues to be suspiciously simple. I am really surprised.
> 
> Did you test what happens when strength/difficulty comparisons are used? Did you check entity/value sorter manner? We do not support some of it with list variables, and we need to make sure that this does not change here; and the use cases which work still need to work in the mixed mode.

If there is coverage for the mentioned features, then we know the default behavior remains unchanged. I'll ensure to add tests for the relevant features.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/placer/QueuedValuePlacer.java
```diff
@@ -31,6 +32,10 @@ public Iterator<Placement<Solution_>> iterator() {
         return new QueuedValuePlacingIterator();
     }
 
+    public boolean isListChangeMoveSelector() {
```
triceo, line 35: The correct name in this case would be `hasListChangeMoveSelector`.

core/src/test/java/ai/timefold/solver/core/testdomain/mixed/singleentity/MixedCustomBasicVariableFactory.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Iterator;
+import java.util.Random;
+import java.util.stream.Stream;
+
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+import ai.timefold.solver.core.impl.heuristic.selector.move.factory.MoveIteratorFactory;
+
+public class MixedCustomBasicVariableFactory
+        implements MoveIteratorFactory<TestdataMixedSolution, MixedCustomBasicVariableSwapMove> {
```
triceo, line 11: This is not a variable factory. This name is confusing.

core/src/test/java/ai/timefold/solver/core/testdomain/mixed/singleentity/MixedCustomPhase.java
```diff
@@ -0,0 +1,18 @@
+package ai.timefold.solver.core.testdomain.mixed.singleentity;
+
+import java.util.function.BooleanSupplier;
+
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+import ai.timefold.solver.core.api.solver.phase.PhaseCommand;
+
+public class MixedCustomPhase implements PhaseCommand<TestdataMixedSolution> {
```
triceo, line 8: This is not a phase, it's a phase command.

core/src/test/java/ai/timefold/solver/core/testdomain/mixed/singleentity/MixedCustomBasicVariableSwapMove.java
```diff
@@ -0,0 +1,31 @@
+package ai.timefold.solver.core.testdomain.mixed.singleentity;
+
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+import ai.timefold.solver.core.impl.heuristic.move.AbstractMove;
+
+public class MixedCustomBasicVariableSwapMove extends AbstractMove<TestdataMixedSolution> {
```
triceo, line 6: Why do we even need this? Looks like a simple swap move could have been reused. There is no special logic here.

core/src/test/java/ai/timefold/solver/core/testdomain/mixed/singleentity/TestdataMixedValueListener.java
```diff
@@ -0,0 +1,44 @@
+package ai.timefold.solver.core.testdomain.mixed.singleentity;
+
+import ai.timefold.solver.core.api.domain.variable.VariableListener;
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+
+import org.jspecify.annotations.NonNull;
+
+public class TestdataMixedValueListener implements VariableListener<TestdataMixedSolution, TestdataMixedValue> {
```
triceo, line 8: Not a value listener, it is a variable listener.
Also, I'd prefer we not use any new variable listeners - we have DSVs now, let's use them.

triceo (approved):
LGTM after comments resolved.

PR #1608, opened 2025-05-23
chore: filtering move selector honoring the phase termination

zepfred:
This PR ensures that the `FilteringMoveSelector` will honor the phase termination setting. There may be use cases where a move filtering filters out all moves, resulting in the iterator analyzing a large number of moves before returning no valid upcoming selection.

It is important to note that the proposed strategy may not work for every termination configuration, and the phase will only end after the move selector bails out. Termination not based on time can cause the process to stuck during the generation of the first valid move in the phase, which may result in no move being returned after bailing out.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
@@ -9,6 +9,7 @@ import ai.timefold.solver.core.impl.heuristic.selector.common.iterator.UpcomingS
 import ai.timefold.solver.core.impl.heuristic.selector.move.AbstractMoveSelector;
 import ai.timefold.solver.core.impl.heuristic.selector.move.MoveSelector;
 import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
+import ai.timefold.solver.core.impl.solver.termination.PhaseTermination;
 
 public final class FilteringMoveSelector<Solution_> extends AbstractMoveSelector<Solution_> {
 
@@ -24,6 +25,7 @@ public final class FilteringMoveSelector<Solution_> extends AbstractMoveSelector
     private final MoveSelector<Solution_> childMoveSelector;
     private final SelectionFilter<Solution_, Move<Solution_>> filter;
     private final boolean bailOutEnabled;
+    private AbstractPhaseScope<Solution_> phaseScope;
 
     private ScoreDirector<Solution_> scoreDirector = null;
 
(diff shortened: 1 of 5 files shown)


triceo (commented):
I am not sure that we want to do this.
To solve a pathological corner case, we slow down almost every move selector, filtering is used _very_ often.

IMO the solution is worse than the problem.

zepfred:
> I am not sure that we want to do this. To solve a pathological corner case, we slow down almost every move selector, filtering is used _very_ often.
> 
> IMO the solution is worse than the problem.

I was also concerned about the performance. In the worst-case scenario, if all moves are filtered out, there will be no drop in performance. However, if no moves are removed, there will be two calls to check the termination.

We could consider adding a triggering condition. For example, we would check that only if the filtered move count is equal to some value, such as 10% of the selector size. Also, if we can ensure the filtered selector is used in the move repository, we could remove the termination check from the decider implementation.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhase.java
```diff
@@ -125,6 +125,7 @@ public void phaseStarted(LocalSearchPhaseScope<Solution_> phaseScope) {
         super.phaseStarted(phaseScope);
         decider.phaseStarted(phaseScope);
         assertWorkingSolutionInitialized(phaseScope);
+        phaseScope.setTermination(phaseTermination);
```
triceo, line 128: This is on `AbstractPhaseScope`, yet only `LocalSearchPhase` actually sets a value.
This is a recipe for an NPE - either make it only available on `LocalSearchPhaseScope`, or make sure the termination is set for every phase type.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractPhaseScope.java
```diff
@@ -36,6 +37,11 @@ public abstract class AbstractPhaseScope<Solution_> {
 
     protected int bestSolutionStepIndex;
 
+    /**
+     * The solver termination configuration
```
triceo, line 41: Not correct. Solver termination is a different thing.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/FilteringMoveSelector.java
```diff
@@ -68,23 +74,53 @@ public long getSize() {
...
         }
 
         @Override
         protected Move<Solution_> createUpcomingSelection() {
             Move<Solution_> next;
             long attemptsBeforeBailOut = bailOutSize;
+            // To reduce the impact of checking for termination on each move,
+            // we only check for termination
+            // after filtering out a total number of moves equal to size(childMoveSelector).
+            long attemptsBeforeCheckTermination = bailOutSize / BAIL_OUT_MULTIPLIER;
```
triceo, line 123: I'm wondering if this bail-out size-based approach isn't too complicated. Why not check for termination every Nth invocation? Say 1000th? 

Otherwise if you have to do millions of iterations before checking the iteration, it will look as if the termination is ignored anyway.

PR #1616, opened 2025-05-27
fix: Update looped status of entity even if no variables changed

Christopher-Chianelli:
We previously only added a non-looped entity to affected entities if one of its declarative variables changed. However, if null is a valid value (such as for unassigned entities), then it might be the case the looped status of an entity changed from looped to non-looped. We now add an entity to the affected entities list if its loop status changed.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
@@ -118,13 +118,18 @@ final class AffectedEntitiesUpdater<Solution_>
         var entity = entityVariable.entity();
         var shadowVariableReference = entityVariable.variableReference();
         var oldValue = shadowVariableReference.memberAccessor().executeGetter(entity);
+        var loopDescriptor = shadowVariableReference.shadowVariableLoopedDescriptor();
+        if (loopDescriptor != null) {
+            var oldLooped = (boolean) loopDescriptor.getValue(entity);
+            if (oldLooped != isLooped) {
+                // Loop status change; add to affected entities
+                affectedEntities.add(entityVariable);
+            }
+        }
 
         if (isLooped) {
-            // null might be a valid value, and thus it could be the case
-            // that is was not looped and null, then turned to looped and null,
-            // which is still considered a change.
-            affectedEntities.add(entityVariable);
             if (oldValue != null) {
+                affectedEntities.add(entityVariable);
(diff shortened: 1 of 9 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -354,6 +354,12 @@ public void triggerVariableListenersInNotificationQueues() {
         }
         if (shadowVariableSession != null) {
             shadowVariableSession.updateVariables();
+            // Some internal variable listeners (such as those used
+            // to check for solution corruption) might have a declarative
+            // shadow variable as a source and need to be triggered here.
+            for (var notifiable : notifiableRegistry.getAll()) {
+                notifiable.triggerAllNotifications();
+            }
```
triceo, line 362: This code is becoming very confusing - we trigger, then the cascade, then we update shadows again, and then we trigger again. 

We need to start simplifying it - cascade can easily be converted to DSV. And eventually all the other listeners too. Something to look forward to.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
triceo, line 1: This won't help with performance. :-( Well, correctness first.

triceo (approved):
Thanks, this must have been painful to track down!
I've left some comments, but mostly stating the obvious, nothing to be done now.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
Christopher-Chianelli: This is strictly for our own variable listeners; user variable listeners cannot have declarative variable listeners as a source, but our own internal ones might. In particular, the variable listeners that track the before/after variable changed calls to check for solution corruption.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
Christopher-Chianelli: It can probably be disabled if we can check the assert mode, since these variable listeners are only added in assert mode.

triceo: Let's look into the performance of this class holistically. It's on the agenda for later.

PR #1619, opened 2025-05-29
chore: simplify DSV's affected entity update

triceo:


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -122,21 +89,19 @@ private boolean updateShadowVariable(EntityVariablePair<Solution_> entityVariabl
         if (loopDescriptor != null) {
             var oldLooped = (boolean) loopDescriptor.getValue(entity);
             if (oldLooped != isLooped) {
-                // Loop status change; add to affected entities
-                affectedEntities.add(entityVariable);
```
triceo, line 126: This was indirection. Put into a set, affect later. Seemingly unnecessary, since the invariant is that the loop status changed from previous, and therefore we need to execute the update even if we had already executed it on this entity before. (The set wouldn't save any work.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -122,21 +89,19 @@ private boolean updateShadowVariable(EntityVariablePair<Solution_> entityVariabl
...
+                // Loop status change; mark it
+                changeShadowVariableAndNotify(loopDescriptor, entity, isLooped);
             }
         }
 
         if (isLooped) {
             if (oldValue != null) {
-                affectedEntities.add(entityVariable);
```
triceo, line 132: This is effectively dead code. The `add` method would check if the loop descriptor is not null and only scheduled the update in that case. Since the update in this case already happened above, this code is no longer necessary.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -122,21 +89,19 @@ private boolean updateShadowVariable(EntityVariablePair<Solution_> entityVariabl
...
                 changeShadowVariableAndNotify(shadowVariableReference, entity, null);
             }
             return true;
         } else {
             var newValue = shadowVariableReference.calculator().apply(entity);
             if (!Objects.equals(oldValue, newValue)) {
-                affectedEntities.add(entityVariable);
```
triceo, line 139: Same code, same logic.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -90,30 +81,6 @@ private void initializeChangeQueue(BitSet changed) {
         changed.clear();
     }
 
-    private void updateLoopedStatusOfAffectedEntity(Object affectedEntity) {
```
triceo, line 93: Because we no longer defer the updates (see below), this is no longer necessary.

Christopher-Chianelli (changes requested):
This will not work; the `isLooped` parameter is true if a particular shadow variable is looped on the entity, whereas the ShadowVariableLooped shadow variable is true if any shadow variable is looped on the entity. For instance, if you have distance (which cannot be looped) and serviceStartTime (which can be looped), doing the update without checking the loop status of all variables on the entity can result into an entity being incorrectly marked as looped/not looped.

PR #1621, opened 2025-05-29
feat: Support accessing non-declarative shadow variables from a fact

### LINKED ISSUE #1548: Declarative Shadow Variables: Struggles with overridden entities
Consider a situation which occurs in the old implementation of CVRPTW:

- `Customer` does not have `arrivalTime`. It doesn't need it, because it doesn't understand time windows.
- `TimeWindowedCustomer` extends `Customer` and brings `arrivalTime` variable.
- `Customer` has `previousCustomer` shadow, which `TimeWindowedCustomer` needs to compute its `arrivalTime`. Therefore this needs to be a source.

This currently results in an exception, because `TimeWindowedCustomer` doesn't see `previousCustomer.arrivalTime`, because `previousCustomer` is `Customer` and not `TimeWindowedCustomer`.

Attempts to solve this by overriding getter/setter for `previousCustomer` on `TimeWindowedCustomer` to provide `TimeWindowedCustomer` only results in a different runtime exception.

The issue will be resolved when the following code can be migrated to DSV, without changing the data model:

- https://github.com/TimefoldAI/timefold-solver-benchmarks/tree/main/src/main/java/ai/timefold/solver/benchmarks/examples/vehiclerouting
- https://github.com/TimefoldAI/timefold-solver-benchmarks/tree/main/src/main/java/ai/timefold/solver/benchmarks/examples/examination

Christopher-Chianelli:
Fixes #1548

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/VariableDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/VariableDescriptor.java
@@ -2,6 +2,7 @@ package ai.timefold.solver.core.impl.domain.variable.descriptor;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
 import ai.timefold.solver.core.api.domain.variable.PlanningVariable;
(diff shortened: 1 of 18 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/VariableDescriptor.java
```diff
@@ -139,4 +140,22 @@ public final boolean isGenuineAndUninitialized(Object entity) {
...
+    public boolean equals(Object o) {
+        if (o instanceof VariableDescriptor<?> other) {
+            // VariableDescriptors should be compared by identity, but
+            // it appears two different instances of the same VariableDescriptor
+            // get created somewhere
+            // VariableDescriptors can be compared by their entity classes and names
+            return Objects.equals(entityDescriptor.getEntityClass(), other.entityDescriptor.getEntityClass())
+                    && Objects.equals(variableName, other.variableName);
+        }
+        return false;
```
triceo, line 154: Let's not put a band-aid on a problem. Let's find where it happens, and fix it at the source please.

triceo (approved):

PR #1622, opened 2025-05-30
feat: new SolutionUpdatePolicy to force update of shadow vars

triceo:


core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -358,8 +360,52 @@ protected void clearVariableListenerEvents() {
...
+                if (shadowVariableDescriptors.isEmpty()) {
+                    return;
+                }
+                for (var shadowVariableDescriptor : shadowVariableDescriptors) {
+                    var value = shadowVariableDescriptor.getValue(entity);
+                    if (value == null) {
+                        continue;
+                    }
+                    if (shadowVariableDescriptor instanceof InverseRelationShadowVariableDescriptor
+                            && value instanceof Collection<?> collection) {
```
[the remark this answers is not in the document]
triceo: But in this case, it can not be, because there is no way how an inverse shadow is a map.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorGeneratedGizmoSupplierTest.java
```diff
@@ -99,13 +99,17 @@ public interface DummyInterfaceEntity {
...
+        void setLength(Integer length);
     }
 
     @PlanningEntity
     public abstract static class DummyAbstractEntity {
         @ShadowVariable(variableListenerClass = DummyVariableListener.class,
                 sourceEntityClass = TestdataEntity.class, sourceVariableName = "value")
         abstract Integer getLength();
+
+        abstract void setLength(Integer length);
```
triceo: @Christopher-Chianelli I tried to fix this myself, but I wasn't able to find the correct place to do this. Can I please ask you to:

- Make this fail fast if getter is declared, but setter is not. (And vice versa.)
- When the setter and getter is declared, to generate both of them. (Currently a test is complaining that a setter is missing.)

Technically, this is a bwd incompatible behavior. But in practice, I consider this a bug - if a getter is declared, a setter must be declared too. It's either both or neither.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -323,44 +321,27 @@ private void processValueRangeProviderAnnotation(DescriptorPolicy descriptorPoli
         }
     }
 
+    @SuppressWarnings("unchecked")
     private void processPlanningVariableAnnotation(MutableInt variableDescriptorCounter, DescriptorPolicy descriptorPolicy,
             Member member) {
-        var variableAnnotationClass = ConfigUtils.extractAnnotationClass(
-                member, VARIABLE_ANNOTATION_CLASSES);
+        var variableAnnotationClass = ConfigUtils.extractAnnotationClass(member, VARIABLE_ANNOTATION_CLASSES);
```
triceo, line 327: This method was unnecessarily complex. A variable is read+write. Therefore it's either a field, or it has both getter and a setter. Nothing in between.

quarkus-integration/quarkus/deployment/src/test/java/ai/timefold/solver/quarkus/TimefoldProcessorGeneratedGizmoSupplierTest.java
Christopher-Chianelli: For a normal shadow variable, a setter isn't strictly needed; it is the responsibility of the VariableListener to change the value. Am I understanding correctly you want to change the behaviour of normal shadow variables to require a getter/setter (or be a field)?

triceo: Yes. For a mutable field, the presence of a getter should require the presence of a setter. Or the field is exposed, requiring neither - those two are acceptable options. 
You see the same behavior on Jackson etc.

triceo:
Closing the PR as it still doesn't work in FSR. It is too late to get it into the next release, and it will likely need a different approach anyway.

I will take the accessor changes and create a new PR out of them.

PR #1636, opened 2025-06-09
fix: Benchmarker best score graph no longer empty

triceo:
During the init score removal (cf87aa7), we did not notice that this impacts metrics as well. Benchmarker reads its data from metrics, and therefore lost the information on what score is or is not initialized. As a result, the default value was always false, resulting in empty score graph.

This commit fixes that issue by registering a metric for the unassigned count, which will be reused by benchmarker to provide the correct initialization state.

### DIFF
--- a/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/StatisticRegistry.java
+++ b/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/StatisticRegistry.java
@@ -12,13 +12,14 @@ import java.util.function.Function;
 import java.util.function.ObjLongConsumer;
 import java.util.stream.Collectors;
 
-import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.score.constraint.ConstraintRef;
 import ai.timefold.solver.core.config.solver.monitoring.SolverMetric;
 import ai.timefold.solver.core.impl.phase.event.PhaseLifecycleListener;
 import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
 import ai.timefold.solver.core.impl.phase.scope.AbstractStepScope;
 import ai.timefold.solver.core.impl.score.definition.ScoreDefinition;
+import ai.timefold.solver.core.impl.score.director.InnerScore;
+import ai.timefold.solver.core.impl.solver.monitoring.SolverMetricUtil;
 import ai.timefold.solver.core.impl.solver.scope.SolverScope;
 
 import io.micrometer.core.instrument.Meter;
@@ -88,21 +89,17 @@ public class StatisticRegistry<Solution_> extends SimpleMeterRegistry
                 .collect(Collectors.toSet());
     }
 
-    public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<Score<?>> scoreConsumer) {
-        var labelNames = scoreDefinition.getLevelLabels();
-        for (var i = 0; i < labelNames.length; i++) {
-            labelNames[i] = labelNames[i].replace(' ', '.');
-        }
-        var levelNumbers = new Number[labelNames.length];
-        for (var i = 0; i < labelNames.length; i++) {
-            var scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
+    public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<InnerScore<?>> scoreConsumer) {
+        var score = SolverMetricUtil.extractScore(metric, scoreDefinition, id -> {
+            var gaugeId = metric.getMeterId() + "." + id;
+            var scoreLevelGauge = this.find(gaugeId).tags(runId).gauge();
             if (scoreLevelGauge != null && Double.isFinite(scoreLevelGauge.value())) {
-                levelNumbers[i] = scoreLevelNumberConverter.apply(scoreLevelGauge.value());
+                return scoreLevelNumberConverter.apply(scoreLevelGauge.value());
             } else {
-                return;
+                return scoreLevelNumberConverter.apply(0.0);
             }
-        }
-        scoreConsumer.accept(scoreDefinition.fromLevelNumbers(levelNumbers));
+        });
+        scoreConsumer.accept(score);
     }
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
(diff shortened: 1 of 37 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/event/DefaultBestSolutionChangedEvent.java
```diff
@@ -0,0 +1,23 @@
+package ai.timefold.solver.core.impl.solver.event;
+
+import ai.timefold.solver.core.api.solver.Solver;
+import ai.timefold.solver.core.api.solver.event.BestSolutionChangedEvent;
+import ai.timefold.solver.core.impl.score.director.InnerScore;
+
+import org.jspecify.annotations.NonNull;
+
+public final class DefaultBestSolutionChangedEvent<Solution_> extends BestSolutionChangedEvent<Solution_> {
```
zepfred, line 9: Nice refactoring! Should we make the class `BestSolutionChangedEvent` abstract?

core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/statistic/PickedMoveBestScoreDiffStatistic.java
```diff
@@ -80,12 +81,10 @@ private void localSearchStepEnded(LocalSearchStepScope<Solution_> stepScope) {
                 var newBestScore = stepScope.<Score_> getScore().raw();
                 var bestScoreDiff = newBestScore.subtract(oldBestScore);
                 oldBestScore = newBestScore;
-                SolverMetric.registerScoreMetrics(SolverMetric.PICKED_MOVE_TYPE_BEST_SCORE_DIFF,
-                        stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
-                                .and("move.type", moveType),
-                        scoreDefinition,
-                        tagsToMoveScoreMap,
-                        bestScoreDiff);
+                Tags tags = stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
```
zepfred:
```suggestion
                var tags = stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
```

core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/statistic/PickedMoveStepScoreDiffStatistic.java
```diff
@@ -82,12 +83,10 @@ private void localSearchStepEnded(LocalSearchStepScope<Solution_> stepScope) {
             var stepScoreDiff = newStepScore.subtract(oldStepScore);
             oldStepScore = newStepScore;
 
-            SolverMetric.registerScoreMetrics(SolverMetric.PICKED_MOVE_TYPE_STEP_SCORE_DIFF,
-                    stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
-                            .and("move.type", moveType),
-                    scoreDefinition,
-                    tagsToMoveScoreMap,
-                    stepScoreDiff);
+            Tags tags = stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
```
zepfred:
```suggestion
            var tags = stepScope.getPhaseScope().getSolverScope().getMonitoringTags()
```

core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/ScoreLevels.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+public final class ScoreLevels {
+
+    final AtomicReference<Number>[] levelValues;
+    final AtomicInteger unnassignedCount;
+
+    @SuppressWarnings("unchecked")
+    ScoreLevels(int unnassignedCount, Number[] levelValues) {
```
zepfred:
```suggestion
    ScoreLevels(int unassignedCount, Number[] levelValues) {
```

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/bestscore/BestScoreSubSingleStatistic.java
```diff
@@ -32,7 +32,7 @@ public void open(StatisticRegistry<Solution_> registry, Tags runTag) {
         registry.addListener(SolverMetric.BEST_SCORE,
                 timestamp -> registry.extractScoreFromMeters(SolverMetric.BEST_SCORE, runTag,
                         score -> pointList
-                                .add(new BestScoreStatisticPoint(timestamp, score, subSingleBenchmarkResult.isInitialized()))));
+                                .add(new BestScoreStatisticPoint(timestamp, score.raw(), score.fullyAssigned()))));
```
zepfred: I wonder if we should rename `score.fullyAssigned()` to `score.isFullyAssigned()`. What do you think?

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/StatisticRegistry.java
```diff
@@ -88,21 +89,17 @@ public Set<Meter.Id> getMeterIds(SolverMetric metric, Tags runId) {
...
-        var labelNames = scoreDefinition.getLevelLabels();
-        for (var i = 0; i < labelNames.length; i++) {
-            labelNames[i] = labelNames[i].replace(' ', '.');
-        }
-        var levelNumbers = new Number[labelNames.length];
-        for (var i = 0; i < labelNames.length; i++) {
-            var scoreLevelGauge = this.find(metric.getMeterId() + "." + labelNames[i]).tags(runId).gauge();
+    public void extractScoreFromMeters(SolverMetric metric, Tags runId, Consumer<InnerScore<?>> scoreConsumer) {
+        var score = SolverMetricUtil.extractScore(metric, scoreDefinition, id -> {
+            var gaugeId = metric.getMeterId() + "." + id;
```
zepfred: Let's reuse the logic from `SolverMetricUtil::getGaugeName`.

core/src/main/java/ai/timefold/solver/core/impl/solver/event/DefaultBestSolutionChangedEvent.java
triceo: Eventually I want to make it an interface; but for now, I think there's no point in making any changes. We don't need to expose users to change here, so let's not.

PR #1637, opened 2025-06-09
chore: temporally disabling the mixed model

zepfred:


benchmark/src/main/resources/benchmark.xsd
```diff
@@ -674,7 +674,7 @@
           <xs:element minOccurs="0" name="valueSorterManner" type="tns:valueSorterManner"/>
                               
           
-          <xs:choice minOccurs="0">
+          <xs:choice maxOccurs="unbounded" minOccurs="0">
```
triceo: Is this correct? There was only ever one placer allowed here, no?

How did this even happen? I don't see any changes to any of the configs, so I don't understand why the XSD changes.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -166,6 +166,11 @@ public static <Solution_> SolutionDescriptor<Solution_> buildSolutionDescriptor(
             solutionDescriptor.constraintWeightSupplier.initialize(solutionDescriptor,
                     descriptorPolicy.getMemberAccessorFactory(), descriptorPolicy.getDomainAccessType());
         }
+        // Temporally disabling the mixed model
+        if (solutionDescriptor.hasBothBasicAndListVariables()) {
+            throw new IllegalStateException(
+                    "The mixed model is currently unavailable for general use. Please modify the model to utilize either a list variable or only basic variables.");
```
triceo:
```suggestion
                    "Combining list variable and basic variables is currently not supported.");
```

triceo (commented):
The XSD change is fishy. Otherwise LGTM.

benchmark/src/main/resources/benchmark.xsd
zepfred: That's certainly odd. It happened because I ran the tests locally, which updated the XSD file. I don't understand why that occurred, and it is inconsistent with the `solver.xsd` file. Rerunning the Maven `verify` action one more time reverted the change.

PR #1640, opened 2025-06-10
chore: enforce consistency between variable getters and setters

triceo:
If a getter exists, a setter has to.
If a setter exists, a getter has to.
Both getter and setter need to have the same visibility.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
@@ -4,7 +4,9 @@ import java.lang.annotation.Annotation;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
+import java.util.function.IntPredicate;
 
 import ai.timefold.solver.core.impl.domain.common.ReflectionHelper;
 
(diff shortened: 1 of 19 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
```diff
@@ -34,33 +36,38 @@ public ReflectionBeanPropertyMemberAccessor(Method getterMethod, boolean getterO
...
                 } catch (IllegalAccessException e) {
                     throw new IllegalStateException("""
                             Impossible state: method (%s) not accessible.
-                            %s
-                            """
-                            .strip()
+                            %s"""
                             .formatted(setterMethod, MemberAccessorFactory.CLASSLOADER_NUDGE_MESSAGE), e);
                 }
             } else {
```
zepfred: At this point, there is no setter method, but we do not fail fast. Is that correct?

core/src/test/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessorTest.java
```diff
@@ -26,4 +28,22 @@ void methodAnnotatedEntity() throws NoSuchMethodException {
         assertThat(e1.getValue()).isSameAs(v2);
     }
 
+    @Test
```
zepfred, line 32: Is there a test for scenarios where either only the getter or only the setter method exists?

zepfred:
@triceo it seems to me the changes do not completely match the PR description. I found no validation for scenarios where the getter or setter methods are missing.

core/src/test/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessorTest.java
triceo: Considering how much testdomain boilerplate would be required to test for this trivial aspect, I decided against it.

zepfred: Okay. I suggest adding a new variable to `TestdataDifferentGetterSetterVisibilityEntity` and renaming the class to something like `TestdataInvalidGetterSetterEntity`.

zepfred (approved):

PR #1646, opened 2025-06-12
fix: fail-fast on @PlanningEntity being used in a @ProblemFactCollectionProperty

Christopher-Chianelli:
The old test classes had different issues that caused a different fail-fast; those issues are now removed.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -591,7 +591,8 @@ public class SolutionDescriptor<Solution_> {
             if (type.isArray()) {
                 problemFactType = type.getComponentType();
             } else {
-                problemFactType = ConfigUtils.extractGenericTypeParameterOrFail("", memberAccessor.getDeclaringClass(),
+                problemFactType = ConfigUtils.extractGenericTypeParameterOrFail(PlanningSolution.class.getSimpleName(),
+                        memberAccessor.getDeclaringClass(),
                         type, memberAccessor.getGenericType(), annotationClass, memberAccessor.getName());
             }
         } else {
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -587,7 +587,13 @@ public class SolutionDescriptor<Solution_> {
                                         Collection.class.getSimpleName()));
             }
             problemFactCollectionMemberAccessorMap.put(memberAccessor.getName(), memberAccessor);
-            problemFactType = type;
+
+            if (type.isArray()) {
+                problemFactType = type.getComponentType();
+            } else {
+                problemFactType = ConfigUtils.extractGenericTypeParameterOrFail("", memberAccessor.getDeclaringClass(),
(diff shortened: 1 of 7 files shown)


core/src/test/java/ai/timefold/solver/core/testdomain/invalid/entityannotatedasproblemfact/TestdataEntityAnnotatedAsProblemFactSolution.java
```diff
@@ -24,11 +26,16 @@ public static SolutionDescriptor<TestdataEntityAnnotatedAsProblemFactSolution> b
     public TestdataEntityAnnotatedAsProblemFactSolution() {
     }
 
-    @ProblemFactProperty
+    @PlanningEntityProperty
```
zepfred, line 29: Should we also test the single problem fact?

core/src/test/java/ai/timefold/solver/core/testdomain/invalid/entityannotatedasproblemfact/TestdataEntityAnnotatedAsProblemFactCollectionSolution.java
```diff
@@ -25,6 +27,11 @@ public TestdataEntityAnnotatedAsProblemFactCollectionSolution() {
     }
 
     @ProblemFactCollectionProperty
+    public List<TestdataEntity> getEntitiesAsFacts() {
+        return entities;
+    }
+
+    @PlanningEntityCollectionProperty
```
zepfred, line 34: Since this is an invalid class, do we need to define `@PlanningEntityCollectionProperty`?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -587,7 +587,13 @@ private void processProblemFactPropertyAnnotation(DescriptorPolicy descriptorPol
                                         Collection.class.getSimpleName()));
             }
             problemFactCollectionMemberAccessorMap.put(memberAccessor.getName(), memberAccessor);
-            problemFactType = type;
+
+            if (type.isArray()) {
```
zepfred, line 591: Let's include a unit test for array types as well

core/src/test/java/ai/timefold/solver/core/testdomain/invalid/entityannotatedasproblemfact/TestdataEntityAnnotatedAsProblemFactSolution.java
Christopher-Chianelli: That test already exists (in fact it was accidentally duplicated for the collection test)

core/src/test/java/ai/timefold/solver/core/testdomain/invalid/entityannotatedasproblemfact/TestdataEntityAnnotatedAsProblemFactCollectionSolution.java
Christopher-Chianelli: Yep; otherwise it would fail (for the wrong reason); a planning solution class must have at least one @PlanningEntityProperty or @PlanningEntityCollectionProperty. This is one of the reason why the test pass despite the behaviour being incorrect (the other reasons being: missing @PlanningScore annotation and a test being accidentally duplicated).

zepfred (approved):

zepfred:
Thank you for looking into and resolving the issue!

PR #1648, opened 2025-06-12
chore: Revert "fix: clone the solution before sending it to the consumer"

zepfred:


core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -59,6 +59,10 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
 
     /**
      * Sets the best solution consumer, which may be called multiple times during the solving process.
+     * <p>
+     * We strongly recommend not applying any changes to the solution instance while the solver runs.
```
triceo:
```suggestion
     * We strongly recommend not applying any changes to the intermediate best solution while the solver runs.
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -764,6 +764,93 @@ The current `SolverManager` implementation runs on a single computer node,
...
+----
+class SolverManager(Generic[Solution_, ProblemId_]):
+    ...
+    def solve_builder(self) -> SolverJobBuilder[Solution_, ProblemId_]:
+        ...
+----
+====
+
+
+A `SolverJobBuilder` includes various settings, as shown in the following code snippet:
```
triceo: Add a note explaining that they are optional and the user does not need to use them.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -764,6 +764,93 @@ The current `SolverManager` implementation runs on a single computer node,
...
+        .with_best_solution_consumer(on_best_solution_changed)
+        .with_final_best_solution_consumer(on_final_solution_changed)
+        .with_exception_handler(on_exception_handler)
+        .with_config_override(config_override)
+...
+----
+====
+
+The job's unique ID is specified using `withProblemId(problemId)`,
+while the problem is defined with `withProblem(problem)`.
```
triceo: Doesn't make it clear what the distinction between these two is.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -764,6 +764,93 @@ The current `SolverManager` implementation runs on a single computer node,
...
+A consumer for the first initialized solution can be configured with `withFirstInitializedSolutionConsumer(...)`.
+The solution is returned by the last phase that immediately precedes the first local search phase.
+Whenever a new best solution is generated by the solver,
+it can be consumed by configuring it with `withBestSolutionConsumer(...)`.
```
triceo: Don't we already have a section about this somewhere? If not, then I agree it belongs here. But even then - this paragraph deals with many options and arguably, each should get its own paragraph. For clarity.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -764,6 +764,93 @@ The current `SolverManager` implementation runs on a single computer node,
...
+The final best solution consumer,
+which is called at the end of the solving process,
+can be set using `withFinalBestSolutionConsumer(...)`.
+To handle errors that may arise during the solving process,
+set up the  handling logic by defining `withExceptionHandler(...)`.
+Finally, some solver settings, such as the termination configuration, can be overridden using `withConfigOverride(...)`.
```
triceo: Please explain that they come from the solver config that the manager was created with, and that these overrides apply for a particular job.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
triceo: Also, since you're already talking about this, you might as well reference the throttling best solution consumer from the enterprise guide.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -764,6 +764,93 @@ The current `SolverManager` implementation runs on a single computer node,
...
+
+[WARNING]
+====
+The solutions returned by the events in `withFirstInitializedSolutionConsumer(...)`
+and `withBestSolutionConsumer(...)` can be reused during the solving process.
+Therefore, do not apply any changes to the solution instance while the solver is running,
+as any modifications may lead to solver corruption.
```
triceo: IMO this text implies the opposite of what you want to say. ("can be reused during the solving process")

You mean to say _the solver_ can reuse them, and you want to let people know that the user _must not_. Flip the sentences. First say what the user must not do, then explain why.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
zepfred: I couldn't find any reference to `withFirstInitializedSolutionConsumer`, except for an upgrade note.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -830,25 +830,45 @@ solver_manager.solve_builder()
 ----
 ====
 
-The job's unique ID is specified using `withProblemId(problemId)`,
-while the problem is defined with `withProblem(problem)`.
+The job's unique ID is specified using `withProblemId(problemId)`.
+The provided ID allows for the identification of a specific problem,
+enabling actions such as checking the solving status or terminating its execution.
+In addition to the unique ID, we must specify the problem to solve using `withProblem(problem)`.
+Typically, the problem involves uninitialized planning variables that the solver will optimize.
```
triceo: I'd remove this last line. No need to state the obvious.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -830,25 +830,45 @@ solver_manager.solve_builder()
...
 A consumer for the first initialized solution can be configured with `withFirstInitializedSolutionConsumer(...)`.
 The solution is returned by the last phase that immediately precedes the first local search phase.
+
 Whenever a new best solution is generated by the solver,
 it can be consumed by configuring it with `withBestSolutionConsumer(...)`.
 The final best solution consumer,
 which is called at the end of the solving process,
 can be set using `withFinalBestSolutionConsumer(...)`.
+Additionally,
+an improved solution consumer capable of throttling events is available in the xref:enterprise-edition/enterprise-edition.adoc#throttlingBestSolutionEvents[enterprise product].
```
triceo:
```suggestion
an improved solution consumer capable of throttling events is available in the xref:enterprise-edition/enterprise-edition.adoc#throttlingBestSolutionEvents[Enterprise Edition].
```

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -830,25 +830,45 @@ solver_manager.solve_builder()
...
+
+Finally, to build an instance of the solver,
+xref:using-timefold-solver/configuration.adoc[a configuration step] is necessary.
+These settings are static and applied to any related solving execution.
+If you want to override certain settings for a particular job,
+such as the termination configuration, you can use the `withConfigOverride(...)` method.
+
+[NOTE]
+====
+The `SolverJobBuilder` contract includes many settings, but only two are required: `withProblemId(...)` and `withProblem(...)`.
```
triceo:
```suggestion
The `SolverJobBuilder` contract includes many optional methods, but only two are required: `withProblemId(...)` and `withProblem(...)`.
```

triceo: Also, maybe we do this differently? Have two headlines - mandatory settings, and optional ones. That way you avoid the note, and the information is obvious from document structure.

docs/src/modules/ROOT/pages/using-timefold-solver/running-the-solver.adoc
```diff
@@ -830,25 +830,45 @@ solver_manager.solve_builder()
...
+The solver also permits the configuration of multiple solver managers with distinct settings in xref:integration/integration.adoc#integrationWithQuarkusMultipleResources[Quarkus] or xref:integration/integration.adoc#integrationWithSpringBootMultipleResources[Spring Boot].
+====
 
 [WARNING]
 ====
-The solutions returned by the events in `withFirstInitializedSolutionConsumer(...)`
-and `withBestSolutionConsumer(...)` can be reused during the solving process.
-Therefore, do not apply any changes to the solution instance while the solver is running,
-as any modifications may lead to solver corruption.
+Do not modify the solutions returned by the events in `withFirstInitializedSolutionConsumer(...)` and `withBestSolutionConsumer(...)`. These instances are still utilized during the solving process, and any modifications may lead to solver corruption.
```
triceo: I'd put this closer to where you actually talk about the two. Vertical density.

PR #1650, opened 2025-06-16
perf: marginal improvements to cloning

triceo:


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
```diff
@@ -297,87 +341,152 @@ private <C> C cloneCollectionsElementIfNeeded(C original, Map<Object, Object> or
...
-                                        .formatted(f.getName(), declaringClass.getCanonicalName(),
-                                                f.getType().getCanonicalName(), DeepPlanningClone.class.getSimpleName()));
-                            } else {
-                                f.setAccessible(true);
-                            }
-                        })
-                        .map(ShallowCloningFieldCloner::of)
-                        .toArray(ShallowCloningFieldCloner[]::new);
+            if (copiedFieldArray == null) {
+                synchronized (this) {
```
Christopher-Chianelli, line 435: Do move threads/solver manager share the same solution cloner?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
```diff
@@ -51,14 +57,14 @@ public FieldAccessingSolutionCloner(SolutionDescriptor<Solution_> solutionDescri
 
     @Override
     public @NonNull Solution_ cloneSolution(@NonNull Solution_ originalSolution) {
-        Map<Object, Object> originalToCloneMap = new IdentityHashMap<>();
-        Queue<Unprocessed> unprocessedQueue = new ArrayDeque<>();
-        Solution_ cloneSolution = clone(originalSolution, originalToCloneMap, unprocessedQueue,
+        var originalToCloneMap = CollectionUtils.newIdentityHashMap(EXPECTED_OBJECT_COUNT);
```
Christopher-Chianelli: I wonder, can you make `EXPECTED_OBJECT_COUNT` an instance field, and set it to the actual size of the `originalToCloneMap` after validateCloneSolution?

Christopher-Chianelli (approved):
Are SolutionCloners shared between multiple threads (i.e. move threads or SolverManager)? If they are not, the synchronized and double-checked locking checks seems redundant.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
triceo: Anything that shares the solution descriptor will share the cloner. (There is a 1:1 relationship.) This class was specifically marked as thread-safe before this PR, so I made sure that continues being true.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/FieldAccessingSolutionCloner.java
triceo: Hmmm, interesting idea. I wonder what the downsides are.

If we first cloned a big solution, and then cloned a small solution, the map would stay big. But If I divided the count by 2, then this problem would be less, and paying the price of doubling the array once isn't really much of a cost.

Thoughts?

Christopher-Chianelli: Are SolutionDescriptors cached between solvers? If each Solver had it own  SolutionDescriptor, each Solver would have it own cloner with it own `EXPECTED_OBJECT_COUNT`, and the size of the solution does not change during solving. If it is cached, then there are two strategies we can do:

- `EXPECTED_OBJECT_COUNT = max(ACTUAL_OBJECT_COUNT)`, which cause additional memory pressure when solving large then  only small, but the collections created should be short-lived.
- `EXPECTED_OBJECT_COUNT = previous(ACTUAL_OBJECT_COUNT)`, which creates additional memory pressure when concurrently solving bigs and small solutions (since the collections need to be doubled many times when a big solution is cloned after a small solution is cloned).

triceo: Solution descriptors are long-lived. All the things built from the same solver factory will share one solution descriptor. So it's reasonable to expect that solver managers for the same solution class will share the same cloner. (Which is intentional - cloners deduce and hold a lot of metadata about the solution, we want to reuse it.)

I'll see what I can do.

PR #1651, opened 2025-06-16
feat: enable support for mixed models and R&R moves with basic variables

zepfred:
The proposed approach adds two new properties: `entitySelector` and `variableName`. The `entitySelector` property allows the configuration of the entity when the model has multiple entities, and the second property defines the variable to be used when there are multiple basic variables.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveSelectorFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveSelectorFactory.java
@@ -51,7 +51,7 @@ public final class RuinRecreateMoveSelectorFactory<Solution_>
                     .filter(v -> v.getVariableName().equals(config.getVariableName())).findFirst().orElse(null);
         }
         if (variableDescriptor == null) {
-            throw new UnsupportedOperationException("The entity class %s has no variable named as %s."
+            throw new UnsupportedOperationException("The entity class %s has no variable named %s."
                     .formatted(ruinRecreateEntitySelector.getEntityDescriptor().getEntityClass(), config.getVariableName()));
         }
         var nestedEntitySelectorConfig =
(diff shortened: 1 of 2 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMoveSelectorFactory.java
```diff
@@ -25,18 +26,39 @@ protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<So
...
         var variableDescriptor = genuineVariableDescriptorList.get(0);
-
-        var constructionHeuristicPhaseBuilder = RuinRecreateConstructionHeuristicPhaseBuilder.create(configPolicy);
-        return new RuinRecreateMoveSelector<>(entitySelector, variableDescriptor, constructionHeuristicPhaseBuilder,
+        if (genuineVariableDescriptorList.size() > 1) {
+            variableDescriptor = genuineVariableDescriptorList.stream()
+                    .filter(v -> v.getVariableName().equals(config.getVariableName())).findFirst().orElse(null);
+        }
+        if (variableDescriptor == null) {
+            throw new UnsupportedOperationException("The entity class %s has no variable named as %s."
```
triceo:
```suggestion
            throw new UnsupportedOperationException("The entity class %s has no variable named %s."
```

zepfred:
> I am not sure about the approach taken here. Shouldn't R&R ruin the entire solution, as opposed to just a single variable?
> 
> Let's think about the pros and cons of each of the approaches, make the best choice.

Yes, that's a good point. I honestly don't recall the use of R&R in models with multiple variables. The ones I can remember now are related to VRP problems and focus on ruining and recreating routes.

The current implementation for basic variables can only handle one variable, but I believe it can be updated to manage multiple variables. 

1. We could permit the user to define just the entity selector. 
2. By doing this, we will ruin the values of all variables associated with that entity. 
3. We need to modify the R&R logic for generating the move list and likely create a Cartesian Product of moves, following the default CH approach.
4. Should we make this more complicated by allowing multiple entities?

In contrast, the one-variable-per-move approach may be more advantageous if one variable is prioritized, allowing users to focus exclusively on that variable. While it is still possible to ruin multiple variables, this cannot happen within the same move.

triceo (approved):

PR #1658, opened 2025-06-23
chore: Move streams allow value ranges on entities

triceo:


core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ChangeMoveProvider.java
```diff
@@ -0,0 +1,47 @@
...
+    private final SolutionViewTriPredicate<Solution_, Entity_, @Nullable Value_> entityValueFilter;
+
+    public ChangeMoveProvider(PlanningVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel) {
+        this.variableMetaModel = Objects.requireNonNull(variableMetaModel);
+        this.entityValueFilter = (solutionView, entity, value) -> {
+            Value_ oldValue = solutionView.getValue(variableMetaModel, entity);
+            if (Objects.equals(oldValue, value)) {
+                return false;
+            }
+            return solutionView.isValueInRange(variableMetaModel, entity, value);
```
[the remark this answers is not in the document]
triceo: This is unnecessary.
`Objects.equals` will filter out every case where the two values are equal.
Therefore we only process when values are different, regardless if either of the values is null - null is a valid value, therefore it needs to be checked for range.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -116,7 +117,10 @@ public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> ge
                     metricsRequiringConstraintMatchSet);
         }
 
+        var valueRangeState = new ValueRangeState<Solution_>();
```
triceo, line 120: @zepfred You should be able to reuse this instance in selectors - it should eliminate the need to worry about what's in which range. You simply ask this class, and it tells you yes/no.

If necessary, we can also add a method to expose the range, so that you don't need to create your own range instances. (It can be expensive.) That could possibly simplify the selector code - instead of EntityDependent+EntityIndependent selector, you'd just have one and that one would defer to this new state class.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -313,6 +321,30 @@ protected void setWorkingEntityListDirty() {
         workingEntityListRevision++;
     }
 
+    @Override
+    public <Entity_, Value_> boolean isValueInRange(GenuineVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
+            @Nullable Entity_ entity, @Nullable Value_ value) {
+        var valueRangeDescriptor = extractValueRangeDescriptor(variableMetaModel);
+        if (entity == null && valueRangeDescriptor.isEntityIndependent()) {
```
zepfred: Should it be `!valueRangeDescriptor.isEntityIndependent()` instead?

zepfred: If that’s the case, I wonder why it isn’t failing.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractForEachDataStream.java
```diff
@@ -15,13 +15,16 @@
...
         extends AbstractUniDataStream<Solution_, A>
         implements TupleSource
-        permits ForEachIncludingPinnedDataStream, ForEachExcludingPinnedDataStream, ForEachFromSolutionDataStream {
+        permits ForEachIncludingPinnedDataStream, ForEachExcludingPinnedDataStream {
 
     protected final Class<A> forEachClass;
+    private final boolean shouldIncludeNull;
 
-    protected AbstractForEachDataStream(DataStreamFactory<Solution_> dataStreamFactory, Class<A> forEachClass) {
+    protected AbstractForEachDataStream(DataStreamFactory<Solution_> dataStreamFactory, Class<A> forEachClass,
```
zepfred, line 23: We are implementing the builder pattern in some parts of the code. I wonder if we should adopt it as the default approach when there are three or more properties.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ChangeMoveProvider.java
zepfred, line 30: I understand the approach, but I'm thinking if it might include some unnecessary overhead by performing a lookup every time.

zepfred: In this filter, we could assume that the planning value comes from a valid stream. Therefore, I don't understand why we should execute the lookup if the value range is placed within the solution class.

zepfred: Since we can determine whether the range descriptor is entity-independent, I suggest avoiding the execution of the filter in these cases. This approach would maintain the logic of ValueRangeState in relation to the lookup logic, while also eliminating unnecessary calls for value ranges placed in solutions.

zepfred: One issue I am trying to address in the `QueuedValuePlace` is related to executing moves with values that does not belongs the entity range. This filter approach may assist me with that.

core/src/test/java/ai/timefold/solver/core/impl/move/streams/dataset/UniDatasetStreamTest.java
```diff
@@ -327,4 +536,85 @@ void forEachListVariableExcludingPinnedValues() {
...
+        var value4 = solution.getValueList().get(4); // Initially unassigned.
+        // 1 value, entity pinned.
+        var fullyPinnedEntity = solution.getEntityList().get(0);
+        fullyPinnedEntity.setPinned(true);
+        fullyPinnedEntity.setValueList(List.of(value0));
+        // 2 values, 1 pinned.
+        var partiallyPinnedEntity = solution.getEntityList().get(1);
+        partiallyPinnedEntity.setPlanningPinToIndex(1);
+        partiallyPinnedEntity.setValueList(List.of(value1, value2));
+        // 1 value, not pinned.
```
zepfred, line 562:
```suggestion
```

core/src/test/java/ai/timefold/solver/core/impl/move/streams/dataset/UniDatasetStreamTest.java
```diff
@@ -327,4 +536,85 @@ void forEachListVariableExcludingPinnedValues() {
...
+        var value3 = solution.getValueList().get(3);
+        var value4 = solution.getValueList().get(4); // Initially unassigned.
+        // 1 value, entity pinned.
+        var fullyPinnedEntity = solution.getEntityList().get(0);
+        fullyPinnedEntity.setPinned(true);
+        fullyPinnedEntity.setValueList(List.of(value0));
+        // 2 values, 1 pinned.
+        var partiallyPinnedEntity = solution.getEntityList().get(1);
+        partiallyPinnedEntity.setPlanningPinToIndex(1);
+        partiallyPinnedEntity.setValueList(List.of(value1, value2));
```
zepfred:
```suggestion
        // 1 value, not pinned.
        partiallyPinnedEntity.setValueList(List.of(value1, value2));
```

core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ChangeMoveProviderTest.java
```diff
@@ -84,6 +95,133 @@ void fromSolutionBasicVariable() {
         });
     }
 
+    @Test
+    void fromSolutionBasicVariableIncompleteValueRange() {
```
zepfred, line 99: This test addresses the question I had regarding `ChangeMoveProvider`. However, I am not entirely convinced that we should apply the filter in any circumstances. So, here are two questions:
1. Can we identify unused problem fact generators and exclude them from the related streams?
2. Can we fail fast if any unused fact generators like `getValueListNotInValueRange` are identified?

My main concern is to avoid filtering values in large datasets, even with the optimizations from `ListValueRange` being merged.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ChangeMoveProvider.java
triceo: Resolving in favor of a similar comment further down.

core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ChangeMoveProviderTest.java
triceo: You're assuming that this is a problem. It's not. Having values which are not part of any value range is perfectly valid - and preventing it would have been difficult on its own.

The constant filtering of large value ranges is a problem, one which this API will have to eventually solve. But at this stage of the work, I'd rather move on to start solving other problems and implementing more move types - I need to get a better idea of how much work it's going to take, and I will never know that, if I never get to swap moves, pillar moves etc. I need to start uncovering the difficult problems, so that I know how much time it will take to fix them later.

This certainly is one difficult problem.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
triceo: I checked. No code actually passes `null` there.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractForEachDataStream.java
triceo: I think 3 is too low. The amount of overhead for the builder pattern is significant, and arguably 3 arguments is not bad enough.

triceo:
Will be replaced shortly with a new version, updated to the latest code and significantly refactored for improved performance.

PR #1662, opened 2025-06-26
fix: process each entity only once and in topological order

Christopher-Chianelli:
Previously, the code relied on eventually consistency to avoid collections (that is, sometimes suppliers will be called with stale data, but will always eventually be called with correct data).
Now, an initial sort is done to ensure no data is stale, and a map and comparator are utilized to prevent unnecessary computations.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
@@ -6,7 +6,6 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.UnaryOperator;
@@ -14,6 +13,7 @@ import java.util.function.UnaryOperator;
 import ai.timefold.solver.core.preview.api.domain.metamodel.VariableMetaModel;
 
 public final class SingleDirectionalParentVariableReferenceGraph<Solution_> implements VariableReferenceGraph {
+
     private final Set<VariableMetaModel<?, ?, ?>> monitoredSourceVariableSet;
     private final VariableUpdaterInfo<Solution_>[] sortedVariableUpdaterInfos;
     private final UnaryOperator<Object> successorFunction;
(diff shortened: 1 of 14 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
```diff
@@ -55,9 +63,9 @@ public SingleDirectionalParentVariableReferenceGraph(
             }
         }
 
-        for (var shadowEntity : shadowEntities) {
-            updateChanged(shadowEntity);
-        }
+        changedEntities.addAll(List.of(shadowEntities));
```
triceo, line 66: `Arrays.asList(...)` will be free here, nothing will get copied.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
```diff
@@ -70,15 +78,29 @@ public SingleDirectionalParentVariableReferenceGraph(
     @Override
     public void updateChanged() {
         isUpdating = true;
+        changedEntities.sort(topologicalOrderComparator);
+        Map<Object, Object> processed = new IdentityHashMap<>();
         for (var changedEntity : changedEntities) {
```
triceo, line 83: I wonder... if we loop this `while (changedEntity.isEmpty())` and we remove items from the end, we get to avoid the `clear()` at the end. (Of course we'd have to sort the list in reverse order.)

Could it perform better? Maybe, maybe not.

triceo: (Removing items from the start would perform horribly, as the items in the array would have to shift.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
triceo: But you won't be able to clear it later, so nevermind. :-)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
triceo: Nevermind, array list will shrink the array at removal, so there is no way how we can save any time.

triceo (approved):

PR #1673, opened 2025-07-03
perf: Use a single node to represent multiple shadow variables in graphs

### LINKED ISSUE #1521: Investigate and improve runtime performance of Declarative Shadow Variables
We have seen some cases where the performance of DSV was widely inadequate. We need to analyze those cases and see what we can do to improve the performance there.

Christopher-Chianelli:
When there is only a single declarative entity and all non-declarative parents go in the same direction (ex: previous), then some variables can share the same node in the graph. In particular, there is a static topological order when only the variables on a single entity is considered, and as long as the next declarative shadow variable does not have any new non-declarative parents (i.e. group), it can be grouped into the same node as the previous one. All nodes share the same node if no groups are used.

Also fixed a bug in the topological sort done to determine variable trigger order when groups are used (previously it added group variables, which caused it to create a cycle and make all topological orders valid).

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
@@ -51,7 +51,7 @@ final class AffectedEntitiesUpdater<Solution_>
             }
             visited.set(nextNode);
             var shadowVariable = instanceList.get(nextNode);
-            var isChanged = updateShadowVariable(shadowVariable, graph.isLooped(loopedTracker, nextNode));
+            var isChanged = updateEntityShadowVariables(shadowVariable, graph.isLooped(loopedTracker, nextNode));
 
             if (isChanged) {
                 var iterator = graph.nodeForwardEdges(nextNode);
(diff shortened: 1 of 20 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -127,21 +127,26 @@ private boolean updateShadowVariable(EntityVariablePair<Solution_> entityVariabl
...
+                }
+                anyChanged = true;
+            } else {
+                var newValue = shadowVariableReference.calculator().apply(entity);
+                if (!Objects.equals(oldValue, newValue)) {
+                    affectedEntities.add(entityVariable);
+                    changeShadowVariableAndNotify(shadowVariableReference, entity, newValue);
+                    anyChanged = true;
+                }
             }
```
triceo: Please extract this to a new function. Will make profiling easier in the future.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -143,7 +147,7 @@ private static <Solution_> VariableReferenceGraph buildArbitraryGraph(
             VariableReferenceGraphBuilder<Solution_> variableReferenceGraphBuilder, Object[] entities,
             IntFunction<TopologicalOrderGraph> graphCreator) {
         var declarativeShadowVariableDescriptors = solutionDescriptor.getDeclarativeShadowVariableDescriptors();
-        var variableIdToUpdater = new HashMap<VariableMetaModel<?, ?, ?>, VariableUpdaterInfo<Solution_>>();
+        var variableIdToUpdater = new HashMap<VariableMetaModel<?, ?, ?>, VariableUpdaterInfo<Solution_>[]>();
```
triceo: IMO this is complex enough now to warrant its own type.
Even if it just wraps the map, the names of the methods and fields of the new type will make it more obvious what's happening here.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
```diff
@@ -99,12 +128,33 @@ public static <Solution_> GraphStructureAndDirection determineGraphStructure(
...
             return new GraphStructureAndDirection(NO_DYNAMIC_EDGES, null, null);
         } else {
             // Cannot use a single successor function if there are multiple entity classes
-            return multipleDeclarativeEntityClasses ? new GraphStructureAndDirection(ARBITRARY, null, null)
-                    : new GraphStructureAndDirection(SINGLE_DIRECTIONAL_PARENT, parentMetaModel, directionalType);
+            return new GraphStructureAndDirection(SINGLE_DIRECTIONAL_PARENT, parentMetaModel, directionalType);
+        }
+    }
+
+    private static <Solution_> boolean doesEntitiesUseDeclarativeShadowVariables(Object[] entities,
```
triceo: `doEntities...`?
Also, maybe switch the order of the arguments, and make entities a vararg?

triceo (approved):
LGTM, small comments.

triceo:
Oh and also, a green Sonar rating, please.

PR #1684, opened 2025-07-07
chore: prevent NPE in solution annotation processing

triceo:
Was bringing our Sonar rating down to C.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -396,7 +396,7 @@ public class SolutionDescriptor<Solution_> {
     }
 
     private void processSolutionAnnotations(DescriptorPolicy descriptorPolicy) {
-        var annotation = extractTopMostPlanningSolutionAnnotation();
+        var annotation = extractMostRelevantPlanningSolutionAnnotation();
         autoDiscoverMemberType = annotation.autoDiscoverMemberType();
         var solutionClonerClass = annotation.solutionCloner();
         if (solutionClonerClass != PlanningSolution.NullSolutionCloner.class) {
@@ -407,7 +407,7 @@ public class SolutionDescriptor<Solution_> {
                 new LookUpStrategyResolver(descriptorPolicy, lookUpStrategyType);
     }
 
-    private @NonNull PlanningSolution extractTopMostPlanningSolutionAnnotation() {
+    private @NonNull PlanningSolution extractMostRelevantPlanningSolutionAnnotation() {
         var solutionAnnotation = solutionClass.getAnnotation(PlanningSolution.class);
         if (solutionAnnotation != null) {
             return solutionAnnotation;


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -395,20 +396,9 @@ Maybe add a getScore() method with a @%s annotation."""
     }
 
     private void processSolutionAnnotations(DescriptorPolicy descriptorPolicy) {
```
triceo, line 398: This is the fix.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -579,7 +592,7 @@ private void processProblemFactPropertyAnnotation(DescriptorPolicy descriptorPol
             var type = memberAccessor.getType();
             if (!(Collection.class.isAssignableFrom(type) || type.isArray())) {
                 throw new IllegalStateException(
-                        "The solutionClass (%s) has a @%s annotated member (%s) that does not return a %s or an array."
+                        "The solutionClass (%s) has a @%s-annotated member (%s) that does not return a %s or an array."
```
triceo, line 595: This is an unrelated improvement to a fail-fast message which I noticed in a different report.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -417,6 +407,29 @@ private void processSolutionAnnotations(DescriptorPolicy descriptorPolicy) {
                 new LookUpStrategyResolver(descriptorPolicy, lookUpStrategyType);
     }
 
+    private @NonNull PlanningSolution extractTopMostPlanningSolutionAnnotation() {
```
zepfred: We're returning the child class first, not the parent class. The method's name suggests that we always return the parent first.

triceo: See, in my mind, the "top-most" class is the child. But yeah, you're right, it's ambiguous.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -395,20 +396,9 @@ Maybe add a getScore() method with a @%s annotation."""
...
-        if (solutionAnnotation == null && parentSolutionAnnotation == null) {
-            throw new IllegalStateException(
-                    "The solutionClass (%s) has been specified as a solution in the configuration, but does not have a @%s annotation."
-                            .formatted(solutionClass, PlanningSolution.class.getSimpleName()));
-        }
-        var annotation = solutionAnnotation != null ? solutionAnnotation : parentSolutionAnnotation;
+        var annotation = extractTopMostPlanningSolutionAnnotation();
         autoDiscoverMemberType = annotation.autoDiscoverMemberType();
-        // We accept only the child class cloner
-        var solutionClonerClass =
```
zepfred, line 410: This is not the behavior anymore. Since we can return the parent class from `extractTopMostPlanningSolutionAnnotation`, the cloner may not be the one from the child class.

zepfred: We might also need to extract that logic into a separate method to maintain consistency.

triceo: We will only return the parent if the child is null.
Therefore we always use the most relevant cloner.
The comment no longer applies, which is why it was removed - but the behavior remains.

zepfred: ~Yep. You're right!~

zepfred: The current logic does not accept the cloner from the parent class in any case. Therefore, the new logic may end up using it instead of `PlanningSolution.NullSolutionCloner.class`. Does it make sense?

zepfred: We should avoid using the parent cloner because it might be inconsistent with the child class. I believe that's the reason we have that comment.

triceo: The way annotations are done, you really have no way of telling if the user requested a value from parent, or from child. You have to assume that child overrides parent.

Therefore if parent specifies a cloner and the child does not, IMO it makes sense to consider the parent cloner as overridden, therefore no cloner as a result.

zepfred: That's the current strategy, or am I missing something?
```
var solutionClonerClass = solutionAnnotation != null ? solutionAnnotation.solutionCloner() : PlanningSolution.NullSolutionCloner.class;
```

zepfred: The new logic is changing that by possibly using the parent cloner instead of `PlanningSolution.NullSolutionCloner.class`.

zepfred:
```
private @NonNull PlanningSolution extractClonerPlanningSolutionAnnotation() {
        var solutionAnnotation = solutionClass.getAnnotation(PlanningSolution.class);
        if (solutionAnnotation != null) {
            return solutionAnnotation.solutionCloner();
        }
        return PlanningSolution.NullSolutionCloner.class;
}
```

zepfred (approved):

PR #1691, opened 2025-07-08
chore: minor update

zepfred:


zepfred:
The solution must not cripple the code

PR #1707, opened 2025-07-23
chore: move streams value range on entity for basic vars

triceo:
Also addresses issues with composite ranges which could possibly duplicate elements.


core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ChangeMoveProviderTest.java
```diff
@@ -113,47 +299,83 @@ void fromSolutionBasicVariableAllowsUnassigned() {
                 .toList();
         assertThat(moveList).hasSize(4);
 
-        // First entity is assigned to null, therefore the applicable moves assign to firstValue and secondValue.
+        // TODO There is a strange issue here that needs to be investigated,
+        //   as it potentially breaks difficulty comparators.
```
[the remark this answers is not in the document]
triceo: Will be handled during code review at the end of the first stage of development.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractForEachDataStream.java
```diff
@@ -34,6 +38,10 @@ public final void buildNode(DataNodeBuildHelper<Solution_> buildHelper) {
         TupleLifecycle<UniTuple<A>> tupleLifecycle = buildHelper.getAggregatedTupleLifecycle(childStreamList);
         var outputStoreSize = buildHelper.extractTupleStoreSize(this);
         var node = getNode(tupleLifecycle, outputStoreSize);
+        if (shouldIncludeNull && !(node instanceof ForEachFromSolutionUniNode<?, ?>)) {
```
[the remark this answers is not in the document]
triceo: Considered, rejected.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseFactory.java
```diff
@@ -79,7 +80,7 @@ protected DefaultConstructionHeuristicPhaseBuilder<Solution_> createBuilder(
     @Override
     public ConstructionHeuristicPhase<Solution_> buildPhase(int phaseIndex, boolean lastInitializingPhase,
             HeuristicConfigPolicy<Solution_> solverConfigPolicy, BestSolutionRecaller<Solution_> bestSolutionRecaller,
-            SolverTermination<Solution_> solverTermination) {
+            SolverTermination<Solution_> solverTermination, ValueRangeManager<Solution_> valueRangeManager) {
```
zepfred, line 83: We have a significant number of parameters now. Should we consider adding a Builder?

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/AbstractIdentityValueRange.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Same as {@link ListValueRange}, but it employs a different caching strategy and remains immutable.
+ *
+ * @param <T> the value type
+ */
+@NullMarked
+public abstract class AbstractIdentityValueRange<T, Range_ extends CountableValueRange<T>>
```
zepfred: The name implies that it is derived from the identity cache. A more suitable name might be `AbstractValueRange`.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/AbstractIdentityValueRange.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Random;
+
+import ai.timefold.solver.core.api.domain.valuerange.CountableValueRange;
+import ai.timefold.solver.core.impl.domain.valuerange.AbstractCountableValueRange;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Same as {@link ListValueRange}, but it employs a different caching strategy and remains immutable.
```
zepfred: The comment appears to be incorrect.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractDatasetInstance.java
```diff
@@ -0,0 +1,37 @@
...
+        this.parent = Objects.requireNonNull(parent);
+        this.inputStoreIndex = inputStoreIndex;
+    }
+
+    public AbstractDataset<Solution_, Tuple_> getParent() {
+        return parent;
+    }
+
+    @Override
+    public void update(Tuple_ tuple) {
```
zepfred, line 29: Should we delegate the implementation to the child classes?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractForEachDataStream.java
zepfred: I recommend updating the `AbstractForEachUniNode` contract to include a method similar to `acceptNullValue()` instead of relying on `instanceof`.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ChangeMoveProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public MoveProducer<Solution_> apply(MoveStreamFactory<Solution_> moveStreamFactory) {
+        var defaultMoveStreamFactory = (DefaultMoveStreamFactory<Solution_>) moveStreamFactory;
+        var variableDescriptor = ((DefaultPlanningVariableMetaModel<Solution_, Entity_, Value_>) variableMetaModel)
+                .variableDescriptor();
+        var dataStream = defaultMoveStreamFactory.enumerateEntityValuePairs(variableMetaModel)
+                .filter((entity, value) -> {
+                    var currentValue = variableDescriptor.getValue(entity);
+                    return !Objects.equals(currentValue, value);
```
zepfred, line 33: Why not just compare the references?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultBiFromUnisMoveStream.java
```diff
@@ -0,0 +1,32 @@
...
+
+import ai.timefold.solver.core.impl.move.streams.dataset.UniDataset;
+import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.BiMoveConstructor;
+import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.BiMoveStream;
+import ai.timefold.solver.core.impl.move.streams.maybeapi.stream.MoveProducer;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public final class DefaultBiFromUnisMoveStream<Solution_, A, B> implements BiMoveStream<Solution_, A, B> {
```
zepfred, line 14: Should the name be `DefaultBiFromUniMoveStream`?

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhaseFactory.java
triceo: Maybe at some point. I don't want to make this PR bigger than it already is, with unrelated changes.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/AbstractIdentityValueRange.java
triceo: I disagree. This class only has two impls, `IdentitySetValueRange` and `IdentityListValueRange` - therefore either the name `AbstractIdentityValueRange` is correct, or all three classes are named incorrectly.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractDatasetInstance.java
triceo: Why? I expect this method will never be needed, and therefore there is no point in complicating the child classes with it.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ChangeMoveProvider.java
triceo: Values could be anything. `OffsetDateTime` etc. Equality is necessary here, identity is a problem. The existing moves and move selectors do the same.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultBiFromUnisMoveStream.java
triceo: Honestly, I'm not sure. This code will evolve, it will flex with all the features that are still missing. I don't think either of these names are final, and I don't think I have a good way right now of knowing what the right naming structure is here.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/AbstractIdentityValueRange.java
triceo: The new code changes this anyway.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -36,6 +44,22 @@ public final class ValueRangeManager<Solution_> {
...
+     * This requires {@link #reset(Object)} to be called before the first call to this method,
+     * and therefore this method will throw an exception if called before the score director is instantiated.
+     *
+     * @throws IllegalStateException if called before {@link #reset(Object)} is called
+     */
+    public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_> valueRangeDescriptor) {
+        if (cachedWorkingSolution == null) {
+            throw new IllegalStateException(
+                    "Impossible state: value range (%s) requested before the working solution is known."
+                            .formatted(valueRangeDescriptor));
```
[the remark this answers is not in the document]
triceo: It says "impossible state", which means a bug in the solver. Nothing the user could do.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -348,7 +348,8 @@ public void setWorkingSolutionFromBestSolution() {
         scoreDirector.setWorkingSolution(scoreDirector.cloneSolution(getBestSolution()));
     }
 
-    public SolverScope<Solution_> createChildThreadSolverScope(ChildThreadType childThreadType) {
+    public SolverScope<Solution_> createChildThreadSolverScope(ChildThreadType childThreadType,
```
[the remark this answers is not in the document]
triceo: Not in public API, therefore we are free to do what we please.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultMoveStreamFactory.java
```diff
@@ -36,50 +44,53 @@ public DefaultMoveStreamSession<Solution_> createSession(Solution_ workingSoluti
...
+                                && planningListVariableMetaModel.allowsUnassignedValues();
+        if (valueRangeDescriptor.canExtractValueRangeFromSolution()) {
+            // No need for filtering the value range; all values from solution are valid.
+            var stream = dataStreamFactory.forEachFromSolution(variableMetaModel, includeNull);
+            return entityDataStream.join(stream);
         } else {
-            throw new UnsupportedOperationException("Value range on entity is not yet supported.");
+            var stream = dataStreamFactory.forEachExcludingPinned(variableMetaModel.type(), includeNull);
+            return entityDataStream.join(stream, Joiners.filtering(
+                    (entity, value) -> valueRangeManager.getFromEntity(valueRangeDescriptor, entity).contains(value)));
```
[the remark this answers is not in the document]
triceo: That's the whole point of the value range manager.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
```diff
@@ -5,54 +5,63 @@
...
-import ai.timefold.solver.core.impl.domain.valuerange.cache.HashSetValueRangeCache;
-import ai.timefold.solver.core.impl.domain.valuerange.cache.ValueRangeCacheStrategy;
+import ai.timefold.solver.core.impl.domain.valuerange.ValueRangeCache;
 import ai.timefold.solver.core.impl.heuristic.selector.common.iterator.CachedListRandomIterator;
 
-import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
 
+@NullMarked
```
zepfred, line 14: The implementation appears to be incorrect. The method `get(int)` uses `list` to retrieve the element, but it is copied once the cache is generated, resulting in outdated data afterward.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Iterates in random order, does not terminate.
+     */
+    public Iterator<Value_> iterator(Random workingRandom) {
+        return new CachedListRandomIterator<>(valuesWithFastRandomAccess, workingRandom);
+    }
+
+    public enum Builder {
```
zepfred, line 75: This builder simplifies caching strategy generation. However, upon reviewing the code, I don't see a reason for having both `ListValueRange` and `SetValueRange`. Both implementations could be replaced with a single `CollectionValueRange`.

The field `valuesWithFastRandomAccess` allows indexed access, and we are already copying the collections when creating the cache. There is no performance gain when using `ListValueRange` as the list is copied.

zepfred: Additionally, please refer to my comment about the `ListValueRange` as it might indicate a bug.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
```diff
@@ -0,0 +1,119 @@
...
+
+        /**
+         * Use when {@link #FOR_TRUSTED_VALUES} is not suitable.
+         */
+        FOR_USER_VALUES {
+            @Override
+            public <Value_> ValueRangeCache<Value_> buildCache(int size) {
```
zepfred, line 82: I appreciated how you solved that issue, as I was also worried that using a smaller size could lead to resizing operations.

core/src/main/java/ai/timefold/solver/core/preview/api/move/SolutionView.java
```diff
@@ -61,4 +64,38 @@ <Entity_, Value_> Value_ getValueAtIndex(PlanningListVariableMetaModel<Solution_
...
+     *
+     * @param variableMetaModel variable in question
+     * @param value value to check
+     * @return true if the value is acceptable for the variable
+     * @param <Entity_> generic type of the entity that the variable is defined on
+     * @param <Value_> generic type of the value that the variable can take
+     * @throws IllegalArgumentException if the value range is on an entity as opposed to a solution;
+     *         use {@link #isValueInRange(GenuineVariableMetaModel, Object, Object)} instead.
+     */
+    default <Entity_, Value_> boolean isValueInRange(GenuineVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
```
zepfred, line 79: The methods `isValueInRange` are not being used.

core/src/main/java/ai/timefold/solver/core/impl/move/director/MoveDirector.java
```diff
@@ -122,6 +124,23 @@ public final <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Sol
         return value;
     }
 
+    @Override
+    public <Entity_, Value_> boolean isValueInRange(GenuineVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
```
zepfred, line 128: It looks like we are not using this method as well.

zepfred: We are using the cache strategy to check if a value is in the range. E.g., `DefaultMoveStreamFactory.enumerateEntityValuePairs`

triceo: These methods are not for us. They are for users. They will be used, once people start writing their own move streams.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractDatasetInstance.java
zepfred: The contract from `TupleLifecycle` includes the update method. Ignoring it does not seem correct to me because if it is called, then the caller expects the tuple to be updated.

zepfred: Is this method even called?

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
triceo: I'd argue that's not correct. The list and set ranges are both based on different collections - the set range will only create the cache if the list is needed, and the list range vice versa.

In both cases, there is significant possibility that the cache will never get created.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
triceo: IMO There is no outdated data. The list will be copied, but it will be an exact copy, and the cache will no longer change after copied. Therefore list.get(X) and cache.get(X) will always be the same.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultBiFromUnisMoveStream.java
zepfred: It seems that there is an unnecessary "s" in the name DefaultBiFromUni"s"MoveStream.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -25,8 +24,17 @@
  * Outside a {@link ProblemChange}, value ranges are not allowed to change.
  * Call {@link #reset(Object)} every time the working solution changes through a problem fact,
  * so that all caches can be invalidated.
+ * 
+ * <p>
+ * Two score directors can never share the same instance of this class;
```
zepfred, line 29: Please note that we will reuse this during certain assertions in `AbstractScoreDirector.assertScoreFromScratch`.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -146,7 +146,7 @@ public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> ge
                 .build();
         var basicPlumbingTermination = new BasicPlumbingTermination<Solution_>(isDaemon);
         var termination = buildTermination(basicPlumbingTermination, configPolicy, configOverride);
-        var phaseList = buildPhaseList(configPolicy, bestSolutionRecaller, termination);
+        var phaseList = buildPhaseList(configPolicy, bestSolutionRecaller, termination, valueRangeManager);
```
zepfred, line 149: Why do we need to share the value range manager during the construction of the phases? Can we retrieve it during the lifecycle events?

This appears inconsistent with the approach used by other components of the solver. I feel we are passing information that could be obtained in another way.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/AbstractDatasetInstance.java
triceo: Most likely it is. But the update doesn't matter. In this case, we do not care what's _in_ the tuple - we only care that the tuple is there.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultBiFromUnisMoveStream.java
triceo: It's not unnecessary. :-) It's a Bi from two Unis.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
triceo: We must not. Seriously, we mustn't. If we do, it's a bug.
I originally had a comment in your PR that this must be made _impossible_ - clearly we forgot about it. Once this PR is merged, I will refactor this to make sure ranges can never be reused.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
zepfred: Makes sense. Why are we copying the collection for `ListValueRange`?

 I suggest updating `ValueRangeCache` to accept `List` and `Set` collections and perform actions based on their type. We should avoid copying the list if it is immutable.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
triceo: I was looking for another way. Didn't find any.
The entire value manager handling will need reworking; the current way how it's exposed is too brittle.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
zepfred: The cache is likely to be created when entity ranges are used.

triceo: Which is not always. We optimize where we can.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
triceo: It's not guaranteed to be mutable. Yet in the cache, it needs to be mutable - because the cache may add things later. Therefore the copy is absolutely necessary.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
zepfred: Also, using a `Set` implementation for the value range will likely trigger cache creation because of `createRandomIterator`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: I'd argue we should share the manager in this specific case; otherwise, we'll need to populate the range manager every time we make an assertion from scratch. This will make some environments slower when using entity value ranges.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
zepfred: `MoveStreamsBasedMoveRepository` is invoked during lifecycle actions. Can we retrieve the range manager during these lifecycle actions?

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
triceo: I am aware of that. And I prefer it over introducing bugs by sharing things that should not be shared. Correctness first, performance second.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
triceo: The manager is created completely outside of any phase, and then somehow transferred everywhere where it needs to go. This is a horrible thing, and it needs refactoring.

First, we must make sure it is never reused. When we make reuse fail fast, we will immediately see all the places where we need to inject new instances of the manager. And when we know that, we will figure out a proper way to provide those instances.

I can do that in this PR, but it feels like it's already big enough.

zepfred: Yep. It is indeed complex enough.

zepfred: Let's address that another time.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
triceo: For now, probably.
Move Streams will not use these iterators at all.

PR #1708, opened 2025-07-24
ci: report test failures in Github Actions

triceo:


triceo:
Won't work on PRs from forks, therefore useless. :-(

PR #1716, opened 2025-07-28
fix: centralize the ownership of ValueRangeManager to avoid subtle bugs

triceo:
`ValueRangeManager` is now owned and exclusively instantiated by `ScoreDirector`.
Therefore all access to it is only possible within the context of a working solution.
In order to achieve this, Move Streams get a special kind of filtering, with access to `SolutionView` - this will be useful anyway, since I've been meaning to do this at a later stage of Move Streams anyway.


core/src/test/java/ai/timefold/solver/core/impl/move/streams/dataset/UniDatasetStreamTest.java
```diff
@@ -389,21 +366,12 @@ void forEachListVariableIncludingPinnedValues() {
...
-            assertThat(uniDatasetInstance.iterator())
-                    .toIterable()
-                    .map(t -> t.factA)
-                    .containsExactly(value1, value2, value3, value4, unassignedValue);
-
-            // Make incremental changes.
-            var entity4 = new TestdataPinnedWithIndexListEntity("entity4", unassignedValue);
-            datasetSession.insert(entity4); // This will add the value to the dataset.
-            datasetSession.retract(partiallyPinnedEntity); // This will remove the pin on value3.
-            datasetSession.settle();
```
triceo, line 405: This was testing situations which are forbidden, and the new code exposed it.
No entity can be inserted or retracted within the session, therefore it need not be tested.

core/src/test/java/ai/timefold/solver/core/testdomain/list/pinned/index/TestdataPinnedWithIndexListEntity.java
```diff
@@ -31,7 +31,7 @@ public static TestdataPinnedWithIndexListEntity createWithValues(String code, Te
         return new TestdataPinnedWithIndexListEntity(code, values).setUpShadowVariables();
     }
 
-    TestdataPinnedWithIndexListEntity setUpShadowVariables() {
+    public TestdataPinnedWithIndexListEntity setUpShadowVariables() {
```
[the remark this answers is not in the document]
triceo: We're good as is.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -39,11 +53,271 @@
...
+        return valueRangeManager;
+    }
+
+    /**
+     * It is not recommended for code other than {@link ScoreDirector} to create instances of this class.
+     * See class-level documentation for more details.
+     * For safety, prefer using {@link #of(SolutionDescriptor, Object)} to create an instance of this class
+     * with a solution already set.
+     */
+    public ValueRangeManager(SolutionDescriptor<Solution_> solutionDescriptor) {
```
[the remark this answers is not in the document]
triceo: The constructor is a suggestion, but there is nothing wrong with instantiating the class.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -39,11 +53,271 @@
...
+            }
+        }
+        return out.longValue();
+    }
+
+    /**
+     * Calculates an indication on how big this problem instance is.
+     * This is approximately the base 10 log of the search space size.
+     * 
+     * @return {@code >= 0}
```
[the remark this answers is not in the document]
triceo: Useful, with tweaks.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -39,11 +53,271 @@
...
+            result += MathUtils.getPossibleArrangementsScaledApproximateLog(MathUtils.LOG_PRECISION, logBase,
+                    totalListMovableValueCount, possibleTargetsForListValue);
+        }
+        var scale = (result / (double) MathUtils.LOG_PRECISION) / MathUtils.getLogInBase(logBase, 10d);
+        if (Double.isNaN(scale) || Double.isInfinite(scale)) {
+            return 0;
+        }
+        return scale;
+    }
+
```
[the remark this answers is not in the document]
triceo: Useful.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
```diff
@@ -87,7 +87,6 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
 
             var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
             nestedSolverScope.setSolver(solverScope.getSolver());
```
zepfred, line 89: Why are we not passing the value range manager? This may cause the R&R to recalculate the unique list of values for entity ranges with every move.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMove.java
```diff
@@ -54,7 +54,6 @@ protected void doMoveOnGenuineVariables(ScoreDirector<Solution_> scoreDirector)
 
         var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
         nestedSolverScope.setSolver(solverScope.getSolver());
-        nestedSolverScope.setValueRangeManager(solverScope.getValueRangeManager());
         nestedSolverScope.setScoreDirector(recordingScoreDirector.getBacking());
```
zepfred, line 57: Same as list R&R.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
zepfred: During my previous validation, I found a bug where the unique list of elements was always recalculated, causing a performance regression.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
```diff
@@ -39,7 +41,39 @@ public GenuineVariableDescriptor<Solution_> getVariableDescriptor() {
...
+
+    @Override
+    public void stepEnded(AbstractStepScope<Solution_> stepScope) {
+        this.valueRangeManager = null;
+        super.stepEnded(stepScope);
+    }
+
+    @Override
+    public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
+        this.valueRangeManager = null;
```
zepfred: Let's call the super method first.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
```diff
@@ -39,7 +41,39 @@ public GenuineVariableDescriptor<Solution_> getVariableDescriptor() {
...
+        super.phaseEnded(phaseScope);
+    }
+
+    @Override
+    public void solvingEnded(SolverScope<Solution_> solverScope) {
+        this.valueRangeManager = null;
```
zepfred: Let's call the super method first.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
triceo: Every score director has its own instance of the VRM.
Therefore when we set the score director (on the very next line), its VRM will be used.

I see no issues here. Prove me wrong.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/RuinRecreateMove.java
triceo: Same answer.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultMoveStreamFactory.java
```diff
@@ -28,19 +27,16 @@ public final class DefaultMoveStreamFactory<Solution_>
...
-        this.valueRangeManager = valueRangeManager;
     }
 
-    public DefaultMoveStreamSession<Solution_> createSession(Solution_ workingSolution, SupplyManager supplyManager) {
-        var session = datasetSessionFactory.buildSession();
-        session.initialize(workingSolution, supplyManager);
-        return new DefaultMoveStreamSession<>(session, workingSolution);
+    public DefaultMoveStreamSession<Solution_> createSession(SessionContext<Solution_> context) {
+        var session = datasetSessionFactory.buildSession(context);
+        session.initialize(context);
```
zepfred: Is it possible to store the working solution during the initialization phase?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultMoveStreamFactory.java
```diff
@@ -28,19 +27,16 @@ public final class DefaultMoveStreamFactory<Solution_>
...
+        return new DefaultMoveStreamSession<>(session, context.workingSolution());
```
zepfred, line 38: Can we retrieve the working solution from the session instance?

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -55,7 +56,7 @@ public void clearShadowVariablesListenerQueue() {
     @Override
     public void setWorkingSolution(Solution_ workingSolution) {
         session = scoreDirectorFactory.newSession(workingSolution, constraintMatchPolicy, derived);
-        session.initialize(workingSolution, this.getSupplyManager());
+        session.initialize(new SessionContext<>(this));
```
zepfred, line 59: Do we reset the session after a real-time event occurs?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMove.java
zepfred: At this point in the review, I understand that.

zepfred (approved):
The new `SessionContext` and related changes look great to me. I still think there will be a performance regression when asserting the solution from scratch, but let's measure that in the future.

Please check the failing CI tests. The issue with the quickstart should have been identified before this change.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
triceo: The real-time event results in resetting the working solution, so yes - you are looking at the place where the session reset happens.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultMoveStreamFactory.java
triceo: That one doesn't store the solution; it doesn't need it for anything.
Therefore I only store the solution in `MoveStreamSession`, where it is actually being used.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/DefaultMoveStreamFactory.java
triceo: I refactored this code a bit more. This particular line no longer exists.

triceo:
The CI failure in benchmarks was caused by a stale branch - I rebased and pushed, the problem will now disappear.

PR #1718, opened 2025-07-28
feat: allowing specify a group key to remove redundant calculations for entities that will be assigned the same value

### LINKED ISSUE #1521: Investigate and improve runtime performance of Declarative Shadow Variables

Christopher-Chianelli:
If a calculation only involves facts/variables on indirect variables, then that calculation will result in the same value for entities that share the same indirect variables/facts. To prevent redundant recalculations, the user can specify a group key, and entities that share the same group key for a shadow will only be calculated once.


triceo:
This is going to need some very thorough documentation, with plenty of examples. 
As a user, I need to be able to answer several key questions:

- What happens to variables if I don't group them, and what is the difference in behavior after I do?
- Are there any red flags that indicate two entities should not be put in the same group?
- How can I design groups for maximum performance?

The current documentation doesn't deal at all with direct/indirect variables/facts. It also doesn't deal with directionality, or loops. Those are concepts that need to be clearly explained together with grouping, ideally with pictures.


I don't mind if there's a separate pull request for the docs, your call. I'll likely submit my code review tomorrow.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -95,4 +95,20 @@
...
+     * for one entity with that group key, and all other entities with that group key will
+     * have their shadows set to that value.
+     * <p>
+     * When the group key is unspecified or null, the entity is not considered to
+     * be part of a group and will not share variable calculations with other
+     * entities.
+     *
+     * @return A property on the entity to group by, or the empty string to not performing grouping.
+     */
+    String groupKey() default "";
```
triceo: Does this mean that every DSV can possibly have its own groupKey?

Considering how many DSVs there can possibly be—as we've already seen—I wonder if this is the best approach. Once an entity has 3, 4 groupKeys, it will easily become unmaintainable.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -1106,26 +1106,37 @@ private static void generateDomainAccessorsForShadowSources(IndexView indexView,
...
+        }
+    }
+
+    private static void generateDomainAccessorsForSourcePath(IndexView indexView, Class<?> rootType, String source,
+            List<AnnotationInstance> membersToGeneratedAccessorsForCollection) {
+        for (var iterator = RootVariableSource.pathIterator(rootType, source); iterator.hasNext();) {
+            var member = iterator.next().member();
+            AnnotationTarget target;
+
+            if (member instanceof Field field) {
```
triceo, line 1127: How would this code decide whether to do the accessor on methods or on the field? The field name is declared in an annotation on another field, so we cannot make the decision based on whether or not the field itself is annotated, as we do elsewhere.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupportTest.java
```diff
@@ -292,20 +292,18 @@ void shadowVariableListGraphEvents() {
         for (var visit : solution.getValues()) {
 
             // If a visit does not have a concurrent group, all variables of that visit share the same node.
-            if (visit.getConcurrentValueGroup() != null) {
-                // serviceStartTime and serviceFinishTime are in the same group, so no edge between them!
-                verifyAddEdge.accept(serviceReadyTime, visit, serviceStartTime, visit);
+            if (visit.getConcurrentValueGroup() != null && visit.equals(visit.getConcurrentValueGroup().get(0))) {
```
triceo, line 295: I see a variant of this code already existed before. How did the groups work before this PR?

Also, arguably this now deserves a better name. (But the `groupKey` naming is undecided yet.)

core/src/test/java/ai/timefold/solver/core/preview/api/variable/declarative/concurrent/ConcurrentValuesShadowVariableTest.java
```diff
@@ -112,12 +113,12 @@ void groupChain() {
...
-        var valueC = new TestdataConcurrentValue("c");
+        var valueA1 = Mockito.spy(new TestdataConcurrentValue("a1"));
+        var valueA2 = Mockito.spy(new TestdataConcurrentValue("a2"));
+        var valueB1 = Mockito.spy(new TestdataConcurrentValue("b1"));
+        var valueB2 = Mockito.spy(new TestdataConcurrentValue("b2"));
+        var valueB3 = Mockito.spy(new TestdataConcurrentValue("b3"));
+        var valueC = Mockito.spy(new TestdataConcurrentValue("c"));
 
         var concurrentGroupA = List.of(valueA1, valueA2);
         var concurrentGroupB = List.of(valueB1, valueB2, valueB3);
```
triceo, line 124: How were the values for the groups selected? Based on what criteria? How can I verify that the grouping is correct?

I see values of the same type, with the same sources. There is nothing which gives me any indication whether these should all be grouped together, or whether they should be grouped at all.

Can we fail fast if we see some cases of grouping which we know should not be allowed?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DeclarativeShadowVariableDescriptor.java
```diff
@@ -80,6 +83,19 @@ A shadow variable must have at least one source (since otherwise it a constant).
...
+
+        if (shadowVariableUpdater.groupKey() != null && !shadowVariableUpdater.groupKey().isEmpty()) {
+            var groupKey = shadowVariableUpdater.groupKey();
+            Member member = RootVariableSource.getMember(entityDescriptor.getEntityClass(),
+                    groupKey, entityDescriptor.getEntityClass(),
+                    groupKey);
+            groupKeyMap =
+                    entityDescriptor.getSolutionDescriptor().getMemberAccessorFactory().buildAndCacheMemberAccessor(member,
+                            MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD, ShadowSources.class,
+                            descriptorPolicy.getDomainAccessType())::executeGetter;
```
triceo: This member accessor loads a map? How?
The entity should have a field with the group key, I don't understand how it suddenly reads a map, and where from.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -207,16 +217,22 @@ public List<VariableUpdaterInfo<Solution_>> getUpdatersForEntity(Object entity)
...
         var sortedDeclarativeVariableDescriptors =
                 topologicallySortedDeclarativeShadowVariables(declarativeShadowVariableDescriptors);
         var groupIndexToVariables = new HashMap<Integer, List<DeclarativeShadowVariableDescriptor<Solution_>>>();
         var groupVariables = new ArrayList<DeclarativeShadowVariableDescriptor<Solution_>>();
         groupIndexToVariables.put(0, groupVariables);
         for (var declarativeShadowVariableDescriptor : sortedDeclarativeVariableDescriptors) {
-            if (!groupVariables.isEmpty() && Arrays.stream(declarativeShadowVariableDescriptor.getSources())
-                    .anyMatch(rootVariableSource -> rootVariableSource.parentVariableType() == ParentVariableType.GROUP)) {
-                // Create a new variable group, since the group might reference prior variables
+            if (!groupVariables.isEmpty() && (Arrays.stream(declarativeShadowVariableDescriptor.getSources())
```
triceo: This condition is becoming kinda crazy. Let's split it into its component parts, name the variables, that way we can better read what it does.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityVariableUpdaterLookup.java
```diff
@@ -4,63 +4,124 @@
...
+    private final BiFunction<Object, VariableMetaModel<Solution_, ?, ?>, Object> entityVariableToKey;
+    private final Map<Object, Lookup<Solution_>> keyToEntityLookup;
     private final Supplier<Lookup<Solution_>> entityLookupSupplier;
     private int nextId = 0;
 
-    private record Lookup<Solution_>(Function<Object, List<VariableUpdaterInfo<Solution_>>> getter,
-            BiConsumer<Object, Supplier<List<VariableUpdaterInfo<Solution_>>>> setter) {
+    private record Lookup<Solution_>(
+            BiFunction<Object, VariableMetaModel<Solution_, ?, ?>, List<VariableUpdaterInfo<Solution_>>> getter,
+            TriConsumer<Object, VariableMetaModel<Solution_, ?, ?>, Supplier<List<VariableUpdaterInfo<Solution_>>>> setter) {
```
triceo: These types are now officially crazy enough to warrant their own functional interfaces—that way, the full signature doesn't proliferate throughout the code; instead, something much shorter and appropriately named will.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityVariableUpdaterLookup.java
```diff
@@ -4,63 +4,124 @@
...
+                        return valueMap.get(entity);
+                    };
+
+            TriConsumer<Object, VariableMetaModel<Solution_, ?, ?>, Supplier<List<VariableUpdaterInfo<Solution_>>>> valueSetter =
+                    (entity, variableMetaModel, valueSupplier) -> {
+                        var groupMapper = variableToGroupKeyMapper.apply(variableMetaModel);
+                        if (groupMapper != null) {
+                            var groupKey = groupMapper.apply(entity);
+                            if (groupKey != null) {
+                                groupValueMap.computeIfAbsent(groupKey, ignored -> valueSupplier.get());
```
triceo: If this is called often enough on the hot path, computeIfAbsent will create a capturing lambda every time it is called. The good old get/put-if-null works better in these cases.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -95,4 +95,20 @@
...
+     * If non-empty, this is a property on the entity that will be used
+     * to identify the entity's group.
+     * <p>
+     * When the group key is non-null, the shadow variable will only be calculated
+     * for one entity with that group key, and all other entities with that group key will
+     * have their shadows set to that value.
+     * <p>
+     * When the group key is unspecified or null, the entity is not considered to
+     * be part of a group and will not share variable calculations with other
+     * entities.
```
triceo, line 111: Can the group key ever change? The Javadoc should mention that.
And if it can't (which is what I'm assuming), are we detecting it and failing fast?

This feature opens up a treasure trove of subtle corruption bugs. We need to do as much as we can to prevent the user from shooting themselves in the foot with this.

triceo (changes requested):
There isn't much I can say about this code. But I have concerns about this feature in general.

I spent not insignificant time with this, and I still don't understand how I should design my groups. DSVs are already very cryptic, and the total lack of guidance around group keys doesn't help. Documentation with plenty of examples will be absolutely necessary - if I can't easily understand this, nobody else will.

In fact, I wonder if this feature is worth the pain it will bring. What are the numbers here? How much does it help? And if it's not applicable to FSR, are we certain there are enough meaningful use cases where it would be applicable? Can we apply it on any of the quickstarts (assuming they are converted)?

I still haven't given up on the idea of auto-detection, somehow. If the user can build the groups ahead of time, so should we be able to - we understand the entire graph. Once I understand how to build these groups, I will have a better idea of what we can do to help the user build them as well.

For the moment, I can not merge this. Not because of the code, but because I do not understand the core principles, and I am unsure about exposing this complexity.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DeclarativeShadowVariableDescriptor.java
Christopher-Chianelli: A `map` as in `mapper`/`function`; think of `stream().map(...)`.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
Christopher-Chianelli: Yes; if two seperate DSV use different "groups", they need different group keys. Note you don't need "groups" to use `groupKey`; i.e. in ExamScheduling/leader-follwer, you can do this:

```java
@ShadowSources(value="leader.value", groupKey="leader")
public TestdataValue getValue() {
    return leader.value;
}
```

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
Christopher-Chianelli: The more specific of the two are preferred, using the getter if the types are the same: https://github.com/TimefoldAI/timefold-solver/blob/72d48c258f0cb4130e13c7dfd4cb978844fc7793/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java#L311-L340

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupportTest.java
Christopher-Chianelli: Before, each visit in a group get its own node and calculate its own value.
Now the first visit in iteration order for each group gets its own node, and the other visits in that group shares it calculations.

Which means, before the graph looked like this:

```
a1.arrival->a1.startTime->a1.finishTime
a2.arrival->a2.startTime->a2.finishTime
a3.arrival->a3.startTime->a3finishTime
```

Now it looks more like this:

```
a1.arrival--->a1.startTime-|->a1.finishTime
a2.arrival-|               |->a2.finishTime
a3.arrival-|               |->a3finishTime
```

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityVariableUpdaterLookup.java
Christopher-Chianelli: Not on hot path; only called when generating the graph structure; not called during solving (during solving, we use an array of entities with the same group key store in the updater).

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
Christopher-Chianelli: The group key can never change. We do not detect if it a variable currently (although even detecting that does not prevent them from changing the field inside the setter of another variable).

Christopher-Chianelli:
FWIW, auto-detection will result in more bugs, since it require including used facts/properties inside `@ShadowSources`.

i.e. if you changed startTime from

```java
// this can be group; each member will calculate the same value
@ShadowSources({"arrivalTime", "group[].arrivalTime"})
public int startTime() {
    return group.stream().map(Visit::arrivalTime).min().get();
}
```

to

```java
int minStartTime; // this is a fact/property

// this cannot be grouped; each member may calculate a different value
// minStartTime is required to be in the sources for auto-detection to work.
@ShadowSources({"arrivalTime", "group[].arrivalTime", "minStartTime"})
public int startTime() {
    var groupStartTime = group.stream().map(Visit::arrivalTime).min().get();
    if (groupStartTime < minStartTime) {
        return minStartTime;
    }
    return groupStartTime
}
```

`ShadowSources` will change to include the used fact so auto-detection would work.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
triceo: Let's fail fast if we detect it's on a variable. It's as much as we can do without any performance impact.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
triceo: I see the point, although it is a big confusing. Shadow sources cannot be facts, but group keys _must_ be facts.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupportTest.java
triceo: So, naming... "Group Key" arguably misses some context (group for what, what does the text mean), while including information that's obvious (it's a key). I'm currently thinking 

    "updateGroupSupplierName" (to mimic "supplierName" on @ShadowVariable)

The group is for updates, it references a supplier, by name.
I'm not happy with this name though - "updateGroup" doesn't say anything about the intended use either. Things other than "update" on my mind: deduplication, merging.

Any other suggestions?

triceo:
I see your point about auto-detection. But it seems to me that both auto-detection and status quo have the same problem - if you depend on something that's not declared, things will break. It doesn't much matter what exactly breaks - something will.

triceo:
So, 3 things before I can merge this:

- Name of the "groupKey" field. Tom and I discussed this and we gravitated towards "resultCacheKeySource". It says that there is a result cache, that this is a key in that cache, and that it points to the source of that key. Thoughts?
- Docs. This new field needs a section in the DSV docs; it should explain the behavior with and without the key, the performance implications, and it should list the limitations (such as the expected immutability of the key). A specific example, such as with visit groups, will help - and so will a picture.
- Sonar. The expected rating is no worse than B, preferrably A.

triceo:
Thinking out loud - since the `resultCacheKeySource` is already on something called `ShadowSources`, maybe we can drop the `Source` off the end.

Christopher-Chianelli:
I personally dislike `resultCacheKey`/`resultCacheKeySource`, since it can make users think "this value won't be recomputed unless this particular property changes". This is fundamentally a grouping operation, similar to `groupBy(keys, accumulators)`, where the entities are the tuples, and the `groupKey` is the keys.

Christopher-Chianelli:
@triceo What do you think of `alignmentKey` or `alignWith`?

triceo:
> I personally dislike `resultCacheKey`/`resultCacheKeySource`, since it can make users think "this value won't be recomputed unless this particular property changes". This is fundamentally a grouping operation, similar to `groupBy(keys, accumulators)`, where the entities are the tuples, and the `groupKey` is the keys.

That's a fair point. `alignmentKey` works, I think. Let's update the API, the implementation, and the docs to all use the new name.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
 @ShadowSources marked methods do not need to check @ShadowVariableLooped properties, since they are only called if all their dependencies are not looped.
 ====
 
+==== Grouping Shadow Variables
+
+`@ShadowSources` has an optional `groupKey` parameter that can be used to reuse calculation results for related entities.
```
triceo: IMO this sentence is not enough to understand what this feature will do, at least it wasn't for me, and it doesn't even make it particularly clear that this exists to _improve performance_.

I recommend:

- Specifically mentioning that this feature is optional and can be enabled when an increased performance is necessary.
- Showing a sequence of calculations before and after; much like you did for me in our earlier DMs; change in X triggers computations in Y1+Y2+Y3, but if this is enabled, only Y1 is triggered.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
...
+To be used, a few requirements must be met:
+
+* `groupKey` must be the name of a problem fact on the entity; its value cannot change during solving.
```
triceo: `groupKey` must _contain_ the name... ? Also, let's not use "its" here, it may not be clear what we refer to. "As with any problem fact, the value of this fact must not change during solving."

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
...
+
+* For all entities with the same, non-null `groupKey` value, calling the supplier method will result in the same value regardless of what entity is called.
```
triceo: Here, I would stress that it is the user's responsibility to only use this feature if they can guarantee that every aligned variable really does share this one value. This is a big opportunity for shooting themselves in the foot, they should be aware.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
...
+
+    @ShadowSources(value = {"leader.value"}, groupKey = "leader")
+    public Integer valueSupplier() {
+      return leader.value;
+    }
+}
+----
+====
+
+Meets the requirements since the `leader` field is not a variable, and entities with the same `leader` will get the same value.
```
triceo: Let's put this _before_ the code sample. "In the following example, ..."

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
...
+      if (leader.value == null || leader.value < minValue) {
+          return minValue;
+      }
+      return leader.value;
+    }
+}
+----
+====
+
+Does not meet the requirement, since `Follower` entities with the same `leader` but a different `minValue` may calculate different values.
```
triceo: Same applies here.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1018,6 +1018,132 @@ Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSo
 @ShadowSources marked methods do not need to check @ShadowVariableLooped properties, since they are only called if all their dependencies are not looped.
 ====
 
+==== Grouping Shadow Variables
```
triceo: Value alignment?

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -95,4 +95,22 @@
      * @return A non-empty list of variables the supplier method accesses.
      */
     String[] value();
+
+    /**
+     * If non-empty, this is the property name of a problem fact on the entity that will be used
+     * to identify the entity's group.
```
triceo: ... that will be used to define a group of entities to align values for.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -95,4 +95,22 @@
...
+     * <p>
+     * When the group key is non-null, the shadow variable will only be calculated
+     * for one entity with that group key, and all other entities with that group key will
+     * have their shadows set to that value.
```
triceo: This is awkwardly phrased, IMO.
I'd start by saying that, if the value is null, the entities are considered independent and their values will not be aligned. Then continue with the rest.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowSources.java
```diff
@@ -95,4 +95,22 @@
...
+     * <p>
+     * When the group key is unspecified or null, the entity is not considered to
+     * be part of a group and will not share variable calculations with other
+     * entities.
+     * <p>
+     * Important: the group key must not be a variable and must not change during solving.
+     *
+     * @return The property name of a problem fact on the entity to group by, or the empty string to not performing grouping.
```
triceo:
```suggestion
     * @return The name of a problem fact on the entity to use for value alignment, or empty string to not align values with any other entity.
```

triceo (approved):
LGTM after the suggested changes to docs and Javadocs are made.

When changing the implementation code to no longer mention group key, please pay attention to mention "alignment group" instead of just "group". "Group" is too vague and could mean anything; alignment groups are now a well-defined concept and we should refer to it as such.

PR #1722, opened 2025-07-31
chore: improve entity range for list variables

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -138,6 +138,10 @@ private void processWorkingSolutionDuringStep(ConstructionHeuristicStepScope<Sol
         solver.getBestSolutionRecaller().processWorkingSolutionDuringConstructionHeuristicsStep(stepScope);
     }
 
+    public void enableAssertions(EnvironmentMode environmentMode) {
```
triceo: I don't think we should do this.

We have assertions to guard _user_ code to prevent _users_ from making mistakes. Our own code doesn't require assertions - our own code is guaranteed correct by the unit tests.

Fail fasts are typically not used on the hot path, which is where this would be.

I can see how this was useful during development. Now, it needs to be replaced with test coverage. Please remove.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractMove.java
```diff
@@ -20,6 +21,42 @@
  */
 public abstract class AbstractMove<Solution_> implements Move<Solution_> {
 
+    private Boolean cachedDoableEvaluation;
```
triceo: Why are we caching doability?

isMoveDoable() is called once per move. Therefore there should be absolutely no reason to cache this.

Also, even though moves are not part of public API, we treat them as such. People write custom moves. Therefore you cannot just write additional functionality and hope that people magically find it and start using it.

IMO this should be removed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrixDemand.java
```diff
@@ -0,0 +1,84 @@
...
+        this.workingSolution = workingSolution;
+        this.valueRangeManager = valueRangeManager;
+        this.entityDescriptor = entityDescriptor;
+        this.valueRangeDescriptor = valueRangeDescriptor;
+    }
+
+    @Override
+    public MemoizingSupply<ReachableValueMatrix> createExternalizedSupply(SupplyManager supplyManager) {
+        Supplier<ReachableValueMatrix> supplier = () -> {
+            var entityList = entityDescriptor.extractEntities(workingSolution);
```
triceo: That's a long method. I recommend splitting the iteration into subroutines, and maybe even move this entire code to the matrix itself.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrixDemand.java
```diff
@@ -0,0 +1,84 @@
...
+
+@NullMarked
+public final class ReachableValueMatrixDemand<Solution_> implements Demand<MemoizingSupply<ReachableValueMatrix>> {
+
+    private final Solution_ workingSolution;
+    private final ValueRangeManager<Solution_> valueRangeManager;
+    private final EntityDescriptor<Solution_> entityDescriptor;
+    private final ValueRangeDescriptor<Solution_> valueRangeDescriptor;
+
+    public ReachableValueMatrixDemand(Solution_ workingSolution, ValueRangeManager<Solution_> valueRangeManager,
```
triceo: I'm not exactly sure why we need this supply-demand mechanics, if we already have `ValueRangeManager` instance.
Have the matrix be part of `ValueRangeInstance`, and instantiate it on-demand there?

Key benefit is that this code will all be conceptually in one place, making it easy to find.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrix.java
```diff
@@ -0,0 +1,78 @@
...
+        this.valueToEntityMap = valueToEntityMap;
+        this.randomAccessValueToEntityMap = new IdentityHashMap<>(this.valueToEntityMap.size());
+        for (var entry : this.valueToEntityMap.entrySet()) {
+            randomAccessValueToEntityMap.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+        }
+        this.valueToValueMap = valueToValueMap;
+        this.randomAccessValueToValueMap = new IdentityHashMap<>(this.valueToValueMap.size());
+        for (var entry : this.valueToValueMap.entrySet()) {
+            randomAccessValueToValueMap.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+        }
```
triceo: All four of these maps could be filled on-demand.
This happens during move selection—if we end up picking the first move we select, enumerating all values for all entities is pointless, it will never be used.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -73,4 +96,16 @@ private DestinationSelector<Solution_> applyNearbySelection(HeuristicConfigPolic
         return TimefoldSolverEnterpriseService.loadOrFail(TimefoldSolverEnterpriseService.Feature.NEARBY_SELECTION)
                 .applyNearbySelection(config, configPolicy, minimumCacheType, resolvedSelectionOrder, destinationSelector);
     }
+
+    private EntitySelector<Solution_> applyEntityValueRangeFiltering(HeuristicConfigPolicy<Solution_> configPolicy,
```
triceo, line 100: Doesn't this impact the solver config?

Or let me ask another way—what does the solver config look like to configure filtering on values and entities, if list var used with value range on entities? And how does it different from the same config when value ranges are only on solution?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -0,0 +1,375 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(nonReplayingValueSelector, replayingValueSelector);
+    }
+
+    private abstract class AbstractFilteringValueRangeIterator extends UpcomingSelectionIterator<Object> {
+
+        private final ListVariableStateSupply<Solution_> listVariableStateSupply;
+        private final boolean assertBothSides;
+        private final boolean assertValue;
```
triceo: These need a better name. "Assert" is reserved for assertions in environment mode; this is unrelated to that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -0,0 +1,375 @@
...
+            Object next;
+            var bailoutSize = maxBailoutSize;
+            do {
+                if (bailoutSize <= 0) {
+                    return noUpcomingSelection();
+                }
+                bailoutSize--;
+                var index = workingRandom.nextInt(valueList.size());
+                next = valueList.get(index);
+            } while (!isEntityReachable(next));
```
triceo, line 362: This is still filtering!
The idea is to iterate the right collection; having to skip items at runtime is inefficient, regardless of where it happens.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
```diff
@@ -41,39 +40,23 @@ public GenuineVariableDescriptor<Solution_> getVariableDescriptor() {
...
     public void solvingEnded(SolverScope<Solution_> solverScope) {
         super.solvingEnded(solverScope);
-        this.valueRangeManager = null;
+        this.scoreDirector = null;
+        this.cachedSolution = null;
+        this.cachedValueRange = null;
+    }
+
+    private void checkCachedValueRange() {
+        if (cachedSolution == null || cachedSolution != scoreDirector.getWorkingSolution()) {
```
triceo: Why would the solution ever differ from the working solution? Have you ever seen this happen?

triceo:
Sorry for undrafting the PR, that was a misclick in the IDE.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
zepfred: Okay. We have coverage for these validations on the classes `ListChangeMoveSelectorTest` and `ListSwapMoveSelectorTest`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/AbstractMove.java
zepfred: We usually call the method `isMoveDoable` twice, once during the selector filtering and again when the director applies the move. The goal was to avoid calling the method twice, as it may cause additional CPU cycles when validating the entity range.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrixDemand.java
zepfred: I followed the same concept used in distance matrix initialization. However, adding this to the `ValueRangeManager` wouldn't be difficult.

zepfred: One thing we need to consider is that the `ValueRangeManager` is becoming bloated and more complex.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
```diff
@@ -0,0 +1,253 @@
...
+
+    private long cachedEntityListRevision;
+    private ReachableValueMatrix reachableValueMatrix;
+
+    public FilteringEntityValueRangeSelector(EntitySelector<Solution_> childEntitySelector,
+            IterableValueSelector<Solution_> replayingValueSelector, boolean randomSelection) {
+        this.replayingValueSelector = replayingValueSelector;
+        this.childEntitySelector = childEntitySelector;
+        this.randomSelection = randomSelection;
+        phaseLifecycleSupport.addEventListener(new SelectionCacheLifecycleBridge<>(SelectionCacheType.STEP, this));
```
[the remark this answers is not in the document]
zepfred: Not applying.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrixDemand.java
```diff
@@ -0,0 +1,84 @@
...
+        Supplier<ReachableValueMatrix> supplier = () -> {
+            var entityList = entityDescriptor.extractEntities(workingSolution);
+            var allValues = valueRangeManager.getFromSolution(valueRangeDescriptor, workingSolution);
+            var valuesSize = allValues.getSize();
+            if (valuesSize > Integer.MAX_VALUE) {
+                throw new IllegalStateException(
+                        "The matrix %s cannot be built for the entity %s (%s) because value range has a size (%d) which is higher than Integer.MAX_VALUE."
+                                .formatted(ReachableValueMatrixDemand.class.getSimpleName(),
+                                        entityDescriptor.getEntityClass().getSimpleName(),
+                                        valueRangeDescriptor.getVariableDescriptor().getVariableName(), valuesSize));
```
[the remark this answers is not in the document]
zepfred: Not applying.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/demand/ReachableValueMatrixDemand.java
triceo: In my opinion, this is pretty much the same functionality and therefore it belongs in the same place. I think that centralizing this will be beneficial overall, and outweighs the downside of bloating the top-level class.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -0,0 +1,375 @@
...
+        @Override
+        protected Object createUpcomingSelection() {
+            initialize();
+            if (!hasData) {
+                return noUpcomingSelection();
+            }
+            Object next;
+            var bailoutSize = maxBailoutSize;
+            do {
+                if (bailoutSize <= 0) {
```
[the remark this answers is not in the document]
zepfred: No need for logging

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
zepfred: Yes, I understand that, but the filtering call is necessary. Let’s assume the following reachable values:
```
E1: [V1, V2]
E2: [V1, V2, V3]
E3: [V2, V3]
```

Reachable matrix:
```
V1: V2, V3
V2: V1, V3
V3: V1, V2
```

Assuming the following solution:
```
E1(V1)
E2(V3)
E3(V2)
```

A swap move `(V1, V2)` is not valid because `E3` does not accept `V1`. The same move would be valid for a different solution:
```
E1()
E2(V1, V3)
E3(V2)
```

zepfred: I want to address this filtering call in the performance pull request.

zepfred:
> Sorry for undrafting the PR, that was a misclick in the IDE.

No worries!

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
zepfred: This can occur due to a real-time event that alters the solution. In such cases, the cached value range becomes outdated. I have created a class to abstract this simple caching function.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
zepfred: We have a flag `enableEntityValueRangeFilter` that enables the automatic configuration of the decorators `FilteringEntityValueRangeSelector` and `FilteringValueRangeSelector`.

Currently, only `SubListChangeMoveSelectorFactory` and `ListChangeMoveSelectorFactory` call `DestinationSelectorFactory`. Of these two, only `ListChangeMoveSelectorFactory` applies the settings to build a destination selector for entity ranges. When enabling the entity range setting, we ensure the recorder is configured (`ListChangeMoveSelectorFactory`), which is a requirement for configuring the decorators in the destination selector.

The current method will automatically add a filter layer on top of the user configuration, if one is provided.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValueMatrix.java
```diff
@@ -0,0 +1,99 @@
...
+        this.valueToEntityMap = valueToEntityMap;
+        this.randomAccessValueToEntityMap = new IdentityHashMap<>(this.valueToEntityMap.size());
+        //        for (var entry : this.valueToEntityMap.entrySet()) {
+        //            randomAccessValueToEntityMap.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+        //        }
+        this.valueToValueMap = valueToValueMap;
+        this.randomAccessValueToValueMap = new IdentityHashMap<>(this.valueToValueMap.size());
+        //        for (var entry : this.valueToValueMap.entrySet()) {
+        //            randomAccessValueToValueMap.put(entry.getKey(), new ArrayList<>(entry.getValue()));
+        //        }
```
triceo: Please remove commented-out code.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -423,9 +414,74 @@ public long countOnEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor,
                 .getSize();
     }
 
+    public ReachableValueMatrix getReachableValeMatrix(ListVariableDescriptor<Solution_> listVariableDescriptor) {
```
triceo: Let's find a different name than `Matrix`. We already have a nearby distance matrix, and we don't want the two to be confused.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
```diff
@@ -0,0 +1,232 @@
...
+    // ************************************************************************
+    // Worker methods
+    // ************************************************************************
+
+    public EntitySelector<Solution_> getChildEntitySelector() {
+        return childEntitySelector;
+    }
+
+    @Override
+    public @NonNull ReachableValueMatrix buildCacheItem(@NonNull InnerScoreDirector<Solution_, ?> scoreDirector) {
```
triceo: This caching mechanism seems pointless.
We're already caching the matrix inside of the manager.
I'd rather remove this complexity, as this code is IMO unlikely to lead to significant performance improvements.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/SubListChangeMove.java
```diff
@@ -79,6 +79,7 @@ public int getDestinationIndex() {
 
     @Override
     public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
+        // The move selector can still produce invalid moves, so we need to enable the assertion by default
```
triceo: What assertion?
(There are other places in the PR where this code seems equally out of place.)

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -0,0 +1,379 @@
...
+            var destinationValid = false;
+            var sourceValid = true;
+            // Test if the destination entity accepts the selected value
+            var assignedDestinationPosition = listVariableStateSupply.getElementPosition(destinationValue);
+            if (assignedDestinationPosition instanceof PositionInList elementPosition) {
+                destinationValid = entitiesSet.contains(elementPosition.entity());
+            } else {
+                // Unassigned element is valid
+                destinationValid = true;
+            }
```
triceo, line 238: Arguably you can save yourself some work if you return when detecting invalid destination.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/AbstractCachingEnabledSelector.java
```diff
@@ -0,0 +1,52 @@
...
+    public void constructCache(SolverScope<Solution_> solverScope) {
+        var scoreDirector = solverScope.getScoreDirector();
+        if (cachedItem == null || scoreDirector.isWorkingEntityListDirty(revision)) {
+            this.cachedItem = buildCacheItem(scoreDirector);
+            this.revision = scoreDirector.getWorkingEntityListRevision();
+        }
+    }
+
+    @Override
+    public void disposeCache(SolverScope<Solution_> solverScope) {
```
triceo: As discussed above, I believe the caching mechanism may be pointless.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/SubListChangeMove.java
zepfred: Comment removed.

zepfred:
> * A review of the documentation. It mentions in several places what features the list variable is missing, and that list is outdated by now.

I believe we addressed that in the first PR. I could find only one place mentioning the limitation of using entity ranges with list variables. Is there anything else you believe is missing?

> * If we're adding the seed to Quarkus/Spring, that field needs to be documented with the other fields we already document.
> 

That change won't be included. Used for testing purposes.

triceo:
> > * A review of the documentation. It mentions in several places what features the list variable is missing, and that list is outdated by now.
> 
> I believe we addressed that in the first PR. I could find only one place mentioning the limitation of using entity ranges with list variables. Is there anything else you believe is missing?

You're probably right, I'm misremembering. I did a quick check of the docs and couldn't find anything else obvious.

PR #1725, opened 2025-08-05
perf: group declarative variables after aligned variables when possible

Christopher-Chianelli:
- When there are declarative variables after aligned variables, they are grouped into the same graph node.
- `ShadowVariableLooped` is now guarantee to be updated before the calculator method is called.
- `LoopedTracker` now has access to an `int[][] entityIdToNodeIds` that it can use to determine if an entity is looped (an entity is looped if any of its variables are looped, and variables may correspond to different graph nodes).
-  Each node in the graph now have an `int entityId` and an `int[] groupEntityIds` which is computed when a node is created.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
@@ -189,14 +189,14 @@ public class DefaultShadowVariableSessionFactory<Solution_> {
             List<DeclarativeShadowVariableDescriptor<Solution_>> sortedDeclarativeVariableDescriptors,
             List<VariableUpdaterInfo<Solution_>> allUpdaters,
             List<VariableUpdaterInfo<Solution_>> groupedUpdaters,
-            Map<DeclarativeShadowVariableDescriptor<Solution_>, Map<Object, VariableUpdaterInfo<Solution_>>> variableToEntityToGroupUpdater) {
+            Map<DeclarativeShadowVariableDescriptor<Solution_>, Map<Object, List<VariableUpdaterInfo<Solution_>>>> variableToEntityToGroupUpdater) {
 
         public List<VariableUpdaterInfo<Solution_>> getUpdatersForEntityVariable(Object entity,
                 DeclarativeShadowVariableDescriptor<Solution_> declarativeShadowVariableDescriptor) {
             if (variableToEntityToGroupUpdater.containsKey(declarativeShadowVariableDescriptor)) {
-                var updater = variableToEntityToGroupUpdater.get(declarativeShadowVariableDescriptor).get(entity);
-                if (updater != null) {
-                    return List.of(updater);
+                var updaters = variableToEntityToGroupUpdater.get(declarativeShadowVariableDescriptor).get(entity);
+                if (updaters != null) {
+                    return updaters;
                 }
             }
             for (var shadowVariableDescriptor : sortedDeclarativeVariableDescriptors) {
@@ -225,24 +225,7 @@ public class DefaultShadowVariableSessionFactory<Solution_> {
         var groupVariables = new ArrayList<DeclarativeShadowVariableDescriptor<Solution_>>();
         groupIndexToVariables.put(0, groupVariables);
         for (var declarativeShadowVariableDescriptor : sortedDeclarativeVariableDescriptors) {
-            // If a @ShadowSources has a group source (i.e. "visitGroup[].arrivalTimes"),
-            // create a new group since it must wait until all members of that group are processed
-            var hasGroupSources = Arrays.stream(declarativeShadowVariableDescriptor.getSources())
-                    .anyMatch(rootVariableSource -> rootVariableSource.parentVariableType() == ParentVariableType.GROUP);
-
-            // If a @ShadowSources has an alignment key,
-            // create a new group since multiple entities must be updated for this node
-            var hasAlignmentKey = declarativeShadowVariableDescriptor.getAlignmentKeyMap() != null;
-
-            // If the previous @ShadowSources has an alignment key,
-            // create a new group since we are updating a single entity again
-            // NOTE: Can potentially be optimized/share a node if VariableUpdaterInfo
-            //       update each group member independently after the alignmentKey
-            var previousHasAlignmentKey = !groupVariables.isEmpty() && groupVariables.get(0).getAlignmentKeyMap() != null;
-
-            if (!groupVariables.isEmpty() && (hasGroupSources
-                    || hasAlignmentKey
-                    || previousHasAlignmentKey)) {
+            if (shouldCreateNewGroupForVariable(declarativeShadowVariableDescriptor, groupVariables)) {
                 groupVariables = new ArrayList<>();
                 groupIndexToVariables.put(groupIndexToVariables.size(), groupVariables);
             }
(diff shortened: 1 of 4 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -301,12 +265,77 @@ public List<VariableUpdaterInfo<Solution_>> getUpdatersForEntityVariable(Object
...
+        var hasGroupSources = Arrays.stream(declarativeShadowVariableDescriptor.getSources())
+                .anyMatch(rootVariableSource -> rootVariableSource.parentVariableType() == ParentVariableType.GROUP);
+
+        // If a @ShadowSources has an alignment key,
+        // create a new group since multiple entities must be updated for this node
+        var alignmentKey = declarativeShadowVariableDescriptor.getAlignmentKeyName();
+
+        // If the previous @ShadowSources has an alignment key,
+        // create a new group since we are updating a single entity again
+        // NOTE: Can potentially be optimized/share a node if VariableUpdaterInfo
```
triceo: Is this talking of the optimization which we agreed not to do, because it would expose too much of the internals?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -301,12 +265,77 @@ public List<VariableUpdaterInfo<Solution_>> getUpdatersForEntityVariable(Object
...
+            if (alignmentGroup.getKey() == null) {
+                var updaterCopy = updater.withGroupId(updaterKey);
+                updaters.add(updaterCopy);
+                allUpdaters.add(updaterCopy);
+                updaterKey++;
+            } else {
+                final var newAlignmentUpdaterKey = updaterKey;
+                final var alignmentUpdaterKey = alignmentKeyToGroupIndex.computeIfAbsent(alignmentGroup.getKey(),
+                        ignored -> newAlignmentUpdaterKey);
+                if (alignmentUpdaterKey == newAlignmentUpdaterKey) {
```
triceo, line 317: Shouldn't this be Objects.equals()?
`newAlignmentUpdaterKey` is `int`, but `alignmentUpdaterKey` is `Integer` - based on Java's `Integer` cache, I think this will only work correctly for integers under 128.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -301,12 +265,77 @@ public List<VariableUpdaterInfo<Solution_>> getUpdatersForEntityVariable(Object
...
+        for (var alignmentGroup : alignmentKeyToAlignedEntitiesMap.entrySet()) {
+            if (alignmentGroup.getKey() == null) {
+                var updaterCopy = updater.withGroupId(updaterKey);
+                updaters.add(updaterCopy);
+                allUpdaters.add(updaterCopy);
+                updaterKey++;
+            } else {
+                final var newAlignmentUpdaterKey = updaterKey;
+                final var alignmentUpdaterKey = alignmentKeyToGroupIndex.computeIfAbsent(alignmentGroup.getKey(),
+                        ignored -> newAlignmentUpdaterKey);
```
triceo, line 316: I question the use of computeIfAbsent here.
It forces us to deal with this `final` boilerplate here, and it also introduces the `ignored` boilerplate. Maybe the old ways of get/put would be better here?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/LoopedTracker.java
```diff
@@ -24,6 +26,15 @@ public void mark(int node, LoopedStatus status) {
         looped.set(node, status.ordinal());
     }
 
+    public boolean isEntityLooped(BaseTopologicalOrderGraph graph, int entityId) {
```
triceo: What is the interaction with `status(int)` here? It seems that the two are asking the same information, and one could rely on the other.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -91,47 +106,39 @@ private void initializeChangeQueue(BitSet changed) {
...
-                anyChanged = true;
+            var groupEntities = shadowVariableReferences.get(0).groupEntities();
+            var groupEntityIds = entityVariable.groupEntityIds();
+
+            if (groupEntities != null) {
+                for (var i = 0; i < groupEntityIds.length; i++) {
+                    var groupEntity = groupEntities[i];
+                    var groupEntityId = groupEntityIds[i];
+                    var oldLooped = loopDescriptor.getValue(groupEntity);
+                    var isEntityLooped = loopedTracker.isEntityLooped(graph, groupEntityId);
```
triceo: The code for looped var update is identical to the one in the else branch. Please extract to a method; code reuse is a good thing, and it helps with profiling, where this will no longer account as self time, but will show up properly attributed to looped var update.

triceo (changes requested):
I like this! Do we have some performance numbers around this?

With the looped update, can we now use "looped" as a shadow source? If so, we should remove the fail-fast. (Thinking out loud here: if depending on looped is now technically possible, then we should teach people to depend on it; otherwise we create a strange exception to the rule, where they have to declare all their sources, except for "looped". I'd rather not have such exceptions, they muddy the waters.)

Speaking of "looped" - what made this change (guaranteeing the update _before_ the shadow variable) possible? Or asked differently - was there a particular reason why it was not done like this already? Some performance implication perhaps?

The entire loop situation deserves documentation, but we already have an issue for that, so I won't make you add it to this PR. Same goes for the other documentation/logging changes we discussed on Slack yesterday.

Sonar actually caught a bug this time, I think. You know the rule - B-rated at worst before merging.

Christopher-Chianelli:
Depends on #1725

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/LoopedTracker.java
Christopher-Chianelli: `status` is called indirectly via `graph.isLooped`; `isEntityLooped = any(graph.isLooped(node) for node in entityIdToNodes[entityId])`

Christopher-Chianelli:
In regards to `looped`; people really should not read looped in their suppliers, since it is now guaranteed to always be false when a supplier is called.

triceo:
I will review when https://github.com/TimefoldAI/timefold-solver/pull/1726 is merged, and this PR is rebased. By then, Sonar results will have changed, and the CI will likely no longer fail.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
Christopher-Chianelli: I prefer the `computeIfAbsent`; less logic to go wrong. When minimal Java is 22, we can finally use unused parameter via [JEP 456](https://openjdk.org/jeps/456).

triceo (approved):
LGTM when Sonar is green.

PR #1726, opened 2025-08-06
perf: simplify handling of @ShadowVariableLooped 

Christopher-Chianelli:
- @ShadowVariableLooped is now updated before the
  calculator is called

- Nodes in the graph now have entityId and groupEntityIds

- A mapping is created from entityId to graphNodeId, which
  LoopedTracker will used to lookup the nodes corresponding to
  a given entityId.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
@@ -23,6 +23,7 @@ import ai.timefold.solver.core.impl.domain.policy.DescriptorPolicy;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningSolutionMetaModel;
 import ai.timefold.solver.core.preview.api.domain.metamodel.VariableMetaModel;
 import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowSources;
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariableLooped;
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -388,18 +387,6 @@ private static ParentVariableType determineParentVariableType(Class<?> rootClass
...
-                    The source path (%s) starting from root class (%s) accesses a @%s property (%s).
-                    @%s properties cannot be used as a source, since they are not guaranteed to
-                    be updated when the supplier is called. Supplier methods are only called when
-                    none of their dependencies are looped, so reading @%s properties are not needed.
-                    Maybe remove the source path (%s) from the @%s?
-                    """.formatted(
-                    variablePath, rootClass.getCanonicalName(), ShadowVariableLooped.class.getSimpleName(),
-                    memberName, ShadowVariableLooped.class.getSimpleName(), ShadowVariableLooped.class.getSimpleName(),
-                    variablePath, ShadowSources.class.getSimpleName()));
-        }
```
triceo: In the other PR, you say you think nobody should depend on `looped`, because it will be `false` when the supplier is called. This makes sense - that tells me we should actually _keep_ this fail-fast.

triceo (commented):
The code looks good. See one comment on the `looped` fail-fast.

Also:
- Sonar still complains. If you need me to disable some false positives, I can.
- CI is failing. Maybe the PR needs to be rebased on latest `main`?

triceo (approved):

PR #1727, opened 2025-08-07
chore: move streams change moves for list variable

triceo:
Introduces `ListChangeMoveProvider` that is capable of generating change moves, assign moves and unassign moves.

To do that, several new capabilities are introduced to data streams, which mimic capabilities of constraint streams and use the same underlying Bavet nodes.

With the new value range work, nodes and data sessions no longer need initialization, because everything we need, we can now find in the value range manager.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/bi/UniMapBiDataStream.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/bi/UniMapBiDataStream.java
@@ -13,15 +13,15 @@ import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
 @NullMarked
-final class UniMapBiDataStream<Solution_, A, B, NewA, NewB>
+final class UniMapBiDataStream<Solution_, A, B, NewA>
         extends AbstractUniDataStream<Solution_, NewA> {
 
     private final BiDataMapper<Solution_, A, B, NewA> mappingFunction;
     private @Nullable AftBridgeUniDataStream<Solution_, NewA> aftStream;
 
-    public UniMapBiDataStream(DataStreamFactory<Solution_> constraintFactory, AbstractBiDataStream<Solution_, A, B> parent,
+    public UniMapBiDataStream(DataStreamFactory<Solution_> dataStreamFactory, AbstractBiDataStream<Solution_, A, B> parent,
             BiDataMapper<Solution_, A, B, NewA> mappingFunction) {
-        super(constraintFactory, parent);
+        super(dataStreamFactory, parent);
         this.mappingFunction = mappingFunction;
     }
 
@@ -51,7 +51,7 @@ final class UniMapBiDataStream<Solution_, A, B, NewA, NewB>
             return true;
         if (object == null || getClass() != object.getClass())
             return false;
-        UniMapBiDataStream<?, ?, ?, ?, ?> that = (UniMapBiDataStream<?, ?, ?, ?, ?>) object;
+        UniMapBiDataStream<?, ?, ?, ?> that = (UniMapBiDataStream<?, ?, ?, ?>) object;
         return Objects.equals(parent, that.parent) &&
                 Objects.equals(mappingFunction, that.mappingFunction);
     }
@@ -63,7 +63,7 @@ final class UniMapBiDataStream<Solution_, A, B, NewA, NewB>
 
     @Override
     public String toString() {
-        return "BiMap()";
+        return "UniMap()";
     }
 
 }
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/BiMapUniDataStream.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/BiMapUniDataStream.java
@@ -20,9 +20,9 @@ final class BiMapUniDataStream<Solution_, A, NewA, NewB>
     private final UniDataMapper<Solution_, A, NewB> mappingFunctionB;
     private @Nullable AftBridgeBiDataStream<Solution_, NewA, NewB> aftStream;
 
-    public BiMapUniDataStream(DataStreamFactory<Solution_> constraintFactory, AbstractUniDataStream<Solution_, A> parent,
+    public BiMapUniDataStream(DataStreamFactory<Solution_> dataStreamFactory, AbstractUniDataStream<Solution_, A> parent,
             UniDataMapper<Solution_, A, NewA> mappingFunctionA, UniDataMapper<Solution_, A, NewB> mappingFunctionB) {
-        super(constraintFactory, parent);
+        super(dataStreamFactory, parent);
         this.mappingFunctionA = mappingFunctionA;
         this.mappingFunctionB = mappingFunctionB;
     }
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
@@ -14,16 +14,35 @@ import ai.timefold.solver.core.preview.api.domain.metamodel.ElementPosition;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningListVariableMetaModel;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PositionInList;
 import ai.timefold.solver.core.preview.api.domain.metamodel.UnassignedElement;
+import ai.timefold.solver.core.preview.api.move.SolutionView;
 
 import org.jspecify.annotations.NullMarked;
 
+/**
+ * For each unassigned value, creates a move to assign it to some position of some list variable.
+ * For each assigned value that is not pinned, creates:
+ * 
+ * <ul>
+ * <li>A move to unassign it.</li>
+ * <li>A move to reassign it to another position if assigned.</li>
+ * </ul>
+ * 
+ * To assign or reassign a value, creates:
+ * 
+ * <ul>
+ * <li>A move for every unpinned value in every entity's list variable to assign the value before that position.</li>
+ * <li>A move for every entity to assign it to the last position in the list variable.</li>
+ * </ul>
+ *
+ * This is a generic move provider that works with any list variable;
+ * user-defined change move providers needn't be this complex, as they understand the specifics of the domain.
+ */
 @NullMarked
 public class ListChangeMoveProvider<Solution_, Entity_, Value_>
         implements MoveProvider<Solution_> {
 
     private final PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel;
     private final BiDataFilter<Solution_, Entity_, Value_> isValueInListFilter;
-    private final BiDataFilter<Solution_, Value_, ElementPosition> validChangeFilter;
 
     public ListChangeMoveProvider(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel) {
         this.variableMetaModel = Objects.requireNonNull(variableMetaModel);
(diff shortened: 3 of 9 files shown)


core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/bi/UniMapBiDataStream.java
```diff
@@ -0,0 +1,69 @@
...
+import ai.timefold.solver.core.impl.move.streams.dataset.common.DataNodeBuildHelper;
+import ai.timefold.solver.core.impl.move.streams.dataset.common.bridge.AftBridgeUniDataStream;
+import ai.timefold.solver.core.impl.move.streams.dataset.uni.AbstractUniDataStream;
+import ai.timefold.solver.core.impl.move.streams.maybeapi.BiDataMapper;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+final class UniMapBiDataStream<Solution_, A, B, NewA, NewB>
```
zepfred: Do we need `NewB` here?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/bi/UniMapBiDataStream.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(parent, mappingFunction);
+    }
+
+    @Override
+    public String toString() {
+        return "BiMap()";
```
zepfred: UniMap?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/AbstractForEachDataStream.java
```diff
@@ -38,15 +40,13 @@ public final void collectActiveDataStreams(Set<AbstractDataStream<Solution_>> da
     public final void buildNode(DataNodeBuildHelper<Solution_> buildHelper) {
         TupleLifecycle<UniTuple<A>> tupleLifecycle = buildHelper.getAggregatedTupleLifecycle(childStreamList);
         var outputStoreSize = buildHelper.extractTupleStoreSize(this);
-        var node = getNode(tupleLifecycle, outputStoreSize);
```
zepfred, line 41: Accessing the child class in this context appears to be an anti-pattern. It seems more effective to move it to `ForEachIncludingUnassignedUniNode`.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/BiMapUniDataStream.java
```diff
@@ -0,0 +1,73 @@
...
+
+@NullMarked
+final class BiMapUniDataStream<Solution_, A, NewA, NewB>
+        extends AbstractBiDataStream<Solution_, NewA, NewB> {
+
+    private final UniDataMapper<Solution_, A, NewA> mappingFunctionA;
+    private final UniDataMapper<Solution_, A, NewB> mappingFunctionB;
+    private @Nullable AftBridgeBiDataStream<Solution_, NewA, NewB> aftStream;
+
+    public BiMapUniDataStream(DataStreamFactory<Solution_> constraintFactory, AbstractUniDataStream<Solution_, A> parent,
```
zepfred: Is the variable name `constraintFactory` correct?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        implements MoveProvider<Solution_> {
+
+    private final PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel;
+    private final BiDataFilter<Solution_, Entity_, Value_> isValueInListFilter;
+    private final BiDataFilter<Solution_, Value_, ElementPosition> validChangeFilter;
+
+    public ListChangeMoveProvider(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel) {
+        this.variableMetaModel = Objects.requireNonNull(variableMetaModel);
+        this.isValueInListFilter = (solution, entity, value) -> {
+            if (entity == null || value == null) {
```
zepfred, line 50: This condition bothers me. Can the entity or value be null? Is there an alternative method to verify if all dependencies are complete?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        // For each assigned value that is not pinned, we need to create:
+        // - A move to unassign it.
+        // - A move to reassign it to another position if assigned.
+        // To assign or reassign a value, we need to create:
+        // - A move for every unpinned value in every entity's list variable to assign the value before that position.
+        // - A move for every entity to assign it to the last position in the list variable.
+        var unpinnedEntities =
+                moveStreamFactory.enumerate(variableMetaModel.entity().type(), variableMetaModel.allowsUnassignedValues());
+        var unpinnedValues = moveStreamFactory.enumerate(variableMetaModel.type(), true)
+                .filter((solutionView, value) -> value == null
```
zepfred, line 70: Does `value == null` mean unassigned?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+                        || solutionView.getPositionOf(variableMetaModel, value) instanceof PositionInList);
+        var entityValuePairs = unpinnedEntities.join(unpinnedValues, DataJoiners.filtering(isValueInListFilter))
```
zepfred, line 75: The `value` can be null here, and `isValueInListFilter` will accept it. However, it seems that `value` cannot be null if `allowUnassignedValues` is set to `false`. Is that correct?

zepfred: Nevermind, `unpinnedValues` will always include the null value.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
zepfred: After reviewing the move generation logic, the comment is now clearer and makes more sense.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+                        return ElementPosition.unassigned();
+                    }
+                    var valueCount = solutionView.countValues(variableMetaModel, entity);
+                    if (value == null || valueCount == 0) { // This will trigger assignment of the value at the end of the list.
+                        return ElementPosition.of(entity, valueCount);
+                    } else { // This will trigger assignment of the value immediately before this value.
+                        return solutionView.getPositionOf(variableMetaModel, value);
+                    }
+                })
+                .distinct();
```
zepfred, line 87: I wonder if we should always execute a distinct operation, as it may incur additional costs. Producing moves for local search wouldn’t be a problem having duplicate destinations.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+        var dataStream = moveStreamFactory.enumerate(variableMetaModel.type(), false)
```
zepfred, line 90: ~~I am confused about the join operation. It seems to create a Cartesian product, even though we have already filtered valid positions for all unpinned values.~~

zepfred: ~~Additionally, where are we generating moves for unassigned values?~~ Please include comments to clearly identify each of these generated streams.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+                .join(entityValuePairs, DataJoiners.filtering(validChangeFilter));
+        return moveStreamFactory.pick(dataStream)
```
zepfred, line 94: Very cool!

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -123,4 +124,76 @@ default UniDataStream<Solution_, A> ifNotExistsOther(Class<A> otherClass, BiData
...
+     * There are several recommendations for implementing the mapping function:
+     *
+     * <ul>
+     * <li>Purity.
+     * The mapping function should only depend on its input.
+     * That is, given the same input, it always returns the same output.</li>
+     * <li>Bijectivity.
+     * No two input tuples should map to the same output tuple,
+     * or to tuples that are {@link Object#equals(Object) equal}.
+     * Not following this recommendation creates a constraint stream with duplicate tuples,
```
zepfred:
```suggestion
     * Not following this recommendation creates a data stream with duplicate tuples,
```

core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ListChangeMoveProviderTest.java
```diff
@@ -0,0 +1,425 @@
...
+        var solution = TestdataListEntityProvidingSolution.generateSolution();
+        var e1 = solution.getEntityList().get(0);
+        var v1 = solution.getValueList().get(0);
+        var v2 = solution.getValueList().get(1);
+        var v3 = solution.getValueList().get(2);
+        e1.getValueList().clear();
+        var e2 = solution.getEntityList().get(1);
+        var initiallyAssignedValue = e2.getValueRange().get(0);
+        e2.getValueList().add(initiallyAssignedValue);
+        solution.getEntityList().forEach(TestdataListEntityProvidingEntity::setUpShadowVariables);
```
zepfred, line 132: Let's use `SolutionManager.updateShadowVariables(...)`.

zepfred (commented):
The work on the list move provider looks great! 

However, I have one major concern about always applying the distinct operation. At some point, it will require maintaining the state of records and avoiding duplicates in the stream. This could become an issue with larger datasets, even if we can afford the duplicates, such as when running the LS phase. I would like to hear your opinion before making approving the PR.

triceo:
The concern regarding `distinct()` is not well founded, for three reasons:

- This is an operation in the underlying Bavet network, and is fully incremental. (Under the hood, it is implemented using a `GroupBy` node.) It does have some overhead, but _any_ Bavet node has overhead. In this case, the overhead is comparable to `map`, which does a very similar thing.

- CS uses the same underlying Bavet network, and its network is updated on every move. MS, on the other hand, only updates its Bavet network once at the end of every step. Therefore, if we assume that the performance of CS is sufficient, the performance of MS will be pretty much irrelevant, because it will be executed much less than CS.

- This is an implementation detail. If it ever proves insufficient, we will change the implementation of the provider. At the moment, I am worried about functionality and the programming model that we expose, not minor performance issues.

core/src/test/java/ai/timefold/solver/core/impl/move/streams/maybeapi/provider/ListChangeMoveProviderTest.java
triceo: Although that's a good idea, this is a pattern used throughout the code and I'd rather not introduce even more unrelated changes to this PR.

We should do this in another PR though.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
triceo: Yes, it would be a problem. Producing duplicate destinations skews selection probability - some destinations would appear more often than others, and therefore the solver would unfairly prefer some destinations.

Distinct is really not a problem here anyway, as explained in another comment.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
triceo: At this point in time, it just means that null value needs to survive until later.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/AbstractForEachDataStream.java
triceo: I don't understand. I don't see any child access here.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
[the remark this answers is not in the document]
triceo: Wrong. This is not a Java stream.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListChangeMoveProvider.java
```diff
@@ -0,0 +1,118 @@
...
+                .asMove((solutionView, value, targetPosition) -> {
+                    var currentPosition = solutionView.getPositionOf(variableMetaModel, Objects.requireNonNull(value));
+                    if (targetPosition instanceof UnassignedElement) {
+                        var currentElementPosition = currentPosition.ensureAssigned();
+                        return new ListUnassignMove<>(variableMetaModel, value, currentElementPosition.entity(),
+                                currentElementPosition.index());
+                    }
+                    var targetElementPosition = Objects.requireNonNull(targetPosition).ensureAssigned();
+                    if (currentPosition instanceof UnassignedElement) {
+                        return new ListAssignMove<>(variableMetaModel, value, targetElementPosition.entity(),
```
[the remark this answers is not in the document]
triceo: Although technically correct, this is actually a performance improvement.
Reducing the size of the joins early is beneficial, even though it theoretically results in the method to be called more often.

zepfred:
> The concern regarding `distinct()` is not well founded, for three reasons:
> 
> * This is an operation in the underlying Bavet network, and is fully incremental. (Under the hood, it is implemented using a `GroupBy` node.) It does have some overhead, but _any_ Bavet node has overhead. In this case, the overhead is comparable to `map`, which does a very similar thing.
> * CS uses the same underlying Bavet network, and its network is updated on every move. MS, on the other hand, only updates its Bavet network once at the end of every step. Therefore, if we assume that the performance of CS is sufficient, the performance of MS will be pretty much irrelevant, because it will be executed much less than CS.
> * This is an implementation detail. If it ever proves insufficient, we will change the implementation of the provider. At the moment, I am worried about functionality and the programming model that we expose, not minor performance issues.

Thank you for the explanation. Having the `GroupBy` at the end of the step is indeed better than having it at the end of each applied move. However, if it is not necessary, avoiding it altogether is even better. 

Nevertheless, I liked the functionality and programming model provided at this stage.

zepfred (approved):

core/src/main/java/ai/timefold/solver/core/impl/move/streams/dataset/uni/AbstractForEachDataStream.java
zepfred: I mixed up the names `ForEachIncludingPinnedDataStream` and `ForEachIncludingUnassignedUniNode`. 

Currently, the only child class is `ForEachIncludingPinnedDataStream`. Is it correct to return `ForEachIncludingUnassignedUniNode` by default?

triceo: Yes. Whether something is or is not assigned does not make a difference in the context of Move Streams - we need to operate on both. Therefore we ask for the node which will return both.

PR #1728, opened 2025-08-07
docs: add note on optimizing performance of declarative shadow variables

Christopher-Chianelli:


docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1011,10 +1011,8 @@ Constraint penalizeLoopedJobs(ConstraintFactory factory) {
 
 ====
 
-[IMPORTANT]
+[NOTE]
 ====
-Do not use a @ShadowVariableLooped property in a method annotated with @ShadowSources.
-@ShadowVariableLooped properties can be updated after the @ShadowSources marked method is called, causing score corruption.
 @ShadowSources marked methods do not need to check @ShadowVariableLooped properties, since they are only called if all their dependencies are not looped.
 ====
```
triceo:
```suggestion
NOTE: `@ShadowSources`-annotated methods do not need to check `@ShadowVariableLooped`-annotated properties. These methods are only called if neither of their dependencies are looped, and therefore the value of such properties at that time is guaranteed to be `false`.
```

triceo: The simplified `NOTE:` syntax works with AsciiDoc and does the same thing as the more verbose `[NOTE]`.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1150,6 +1148,25 @@ public class Job {
 ----
 ====
 
+==== Optimizing Shadow Variables
+
+The Solver analyzes the graph formed `@ShadowVariable` to decide what is the most optimal strategy for updating them.
```
triceo:
```suggestion
Each custom Shadow Variable on each planning entity constitutes one node of a reference graph. 
The Solver analyzes this graph to decide what is the most optimal strategy for updating these variables. 
In general, the less variables and the less dependencies between them, the faster the processing.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1150,6 +1148,25 @@ public class Job {
...
+For the best performance, follow these guidelines:
+
+- Only declare custom Shadow Variables on a single entity class.
```
triceo:
```suggestion
- Keep all custom Shadow Variables on the same entity class.
If you have a custom Shadow Variable on another entity class that is used to group entities of another class, consider moving the shadow variable to the grouped entity class and use `alignmentKey` to avoid recomputation.
```

The example is somewhat vague and, if I did not already know what this was talking about, I wouldn't have understood it from this description. I wonder if the use case is so common that it deserves a mention; if you keep it, please reference the alignment section, instead of specifically mentioning the `alignmentKey`.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1150,6 +1148,25 @@ public class Job {
...
+If you have a custom Shadow Variable on another entity class that is used to group entities of another class, consider moving the shadow variable to the grouped entity class and use `alignmentKey` to avoid recomputation.
+
+- Do not use both `@PreviousElementShadowVariable` and `@NextElementShadowVariable` in your sources.
+This includes using `@PreviousElementShadowVariable` as as source in one custom Shadow Variable, and use `@NextElementShadowVariable` as a source in another custom Shadow Variable.
```
triceo:
```suggestion
This includes using `@PreviousElementShadowVariable` as a source in one custom Shadow Variable while using `@NextElementShadowVariable` as a source in another custom Shadow Variable.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1150,6 +1148,25 @@ public class Job {
...
+
+- Compute as much as possible with a custom Shadow Variable.
+Do not needlessly create additional custom Shadow Variables when it is possible to compute their values directly from another custom Shadow Variable.
```
triceo:
```suggestion
- Compute as much as possible within a single custom Shadow Variable supplier method.
Do not create intermediate custom Shadow Variables when it is possible to compute the value directly from another custom Shadow Variable.
```

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1150,6 +1148,25 @@ public class Job {
...
+
+[NOTE]
+====
+These are guidelines and don't necessary need to be followed; for instance, the Solver will still work correctly if you have custom Shadow Variables on multiple entity classes.
+====
```
triceo:
```suggestion
NOTE: These are tips and tricks for optimal performance and needn't be followed to the letter. The solver will work correctly even if you decide not to follow any of the advice, albeit with slightly lesser performance.
```

triceo (approved):
Made some suggestions, hopefully improving clarity.
LGTM after resolved.

PR #1729, opened 2025-08-08
feat: improve the Nearby logic for entity ranges

zepfred:
Companion for the PR in the enterprise repo.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
@@ -62,7 +62,8 @@ public final class FilteringEntityValueRangeSelector<Solution_> extends Abstract
         super.phaseStarted(phaseScope);
         this.entitiesSize = childEntitySelector.getEntityDescriptor().extractEntities(phaseScope.getWorkingSolution()).size();
         this.reachableValues = phaseScope.getScoreDirector().getValueRangeManager()
-                .getReachableValeMatrix(childEntitySelector.getEntityDescriptor().getGenuineListVariableDescriptor());
+                .getReachableValueMatrix(
+                        childEntitySelector.getEntityDescriptor().getGenuineListVariableDescriptor().getValueRangeDescriptor());
         this.childEntitySelector.phaseStarted(phaseScope);
     }
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
@@ -76,7 +76,7 @@ public final class FilteringValueRangeSelector<Solution_> extends AbstractDemand
         this.nonReplayingValueSelector.phaseStarted(phaseScope);
         this.replayingValueSelector.phaseStarted(phaseScope);
         this.reachableValues = phaseScope.getScoreDirector().getValueRangeManager()
-                .getReachableValeMatrix(listVariableStateSupply.getSourceVariableDescriptor());
+                .getReachableValueMatrix(listVariableStateSupply.getSourceVariableDescriptor().getValueRangeDescriptor());
         valuesSize = reachableValues.getSize();
     }
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
@@ -129,20 +129,31 @@ public final class FilteringValueRangeSelector<Solution_> extends AbstractDemand
     @Override
     public Iterator<Object> iterator() {
         if (randomSelection) {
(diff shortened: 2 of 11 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -307,15 +318,71 @@ protected Object createUpcomingSelection() {
...
+            return next;
+        }
+    }
+
+    /**
+     * The optimized iterator only traverses reachable values from the current selection.
+     * Unlike {@link RandomFilteringValueRangeIterator},
+     * it does not use an outer iterator to filter out non-reachable values.
+     */
+    private class OptimizedRandomFilteringValueRangeIterator extends AbstractFilteringValueRangeIterator {
```
triceo, line 361: I don't understand this class. It still does filtering. Therefore it is still wasteful, as opposed to taking values directly from the reachable matrix and therefore not requiring filtering.

I am confused why this class needs to exist, if both iterators still filter.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
```diff
@@ -62,7 +62,8 @@ public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
         super.phaseStarted(phaseScope);
         this.entitiesSize = childEntitySelector.getEntityDescriptor().extractEntities(phaseScope.getWorkingSolution()).size();
         this.reachableValues = phaseScope.getScoreDirector().getValueRangeManager()
-                .getReachableValeMatrix(childEntitySelector.getEntityDescriptor().getGenuineListVariableDescriptor());
+                .getReachableValueMatrix(
```
triceo: There is only one list variable descriptor, and therefore you can avoid the boilerplate of finding the value range. You _know_ the value range, the lookup is unnecessary overhead.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -414,15 +415,15 @@ public long countOnEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor,
                 .getSize();
     }
 
-    public ReachableValues getReachableValeMatrix(ListVariableDescriptor<Solution_> listVariableDescriptor) {
+    public ReachableValues getReachableValueMatrix(ValueRangeDescriptor<Solution_> valueRangeDescriptor) {
```
triceo: I'd rename this to `getReachableValues`. That's what the type is.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -414,15 +415,15 @@ public long countOnEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor,
...
+        var reachableValues = fromReachableValuesMap.get(valueRangeDescriptor);
         if (reachableValues == null) {
             if (cachedWorkingSolution == null) {
                 throw new IllegalStateException(
                         "Impossible state: the matrix %s requested before the working solution is known."
                                 .formatted(ReachableValues.class.getSimpleName()));
```
triceo:
```suggestion
"Impossible state: value reachability requested before the working solution is known.");
```

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
zepfred: Consider a scenario in which the user customizes the filtering for the secondary value selector for a swap operation. In this case, we still want to apply the filtering operation and subsequently filter out non-reachable values. To achieve this, we should use the `RandomFilteringValueRangeIterator` so that the user's filters are applied effectively. 

On the other hand, if the secondary value selector does not have any user customization, we can simply query only the reachable values, as there are no user-configured filters to consider.

triceo: If we have user filtering, why not apply the filter _after_ the value ranges?
I see that there is a loop which iterates _while reachable_ - which tells me that the value ranges happen last, not first.

triceo: See:


            do {
                if (bailoutSize <= 0 || !valueIterator.hasNext()) {
                    return noUpcomingSelection();
                }
                bailoutSize--;
                next = valueIterator.next();
            } while (!isValueOrEntityReachable(next));

I wouldn't expect this to be there. The reachable values ought to be the bottom, the source of the data. Not a thing slapped onto the top.

triceo: Anyway... I really want this in the next release, so let's put a pin in this. If what we have works as intended, we'll improve it further after the release.

zepfred: Yes, in the current implementation, the value range logic is applied last. That's why we need the filtering function.

triceo (approved):
Approving with the understanding that we will still improve this in the next release.

PR #1734, opened 2025-08-11
feat: allow using builtin variable after builtin variable source in declarative shadow variables

Christopher-Chianelli:
Required to convert [LineOperatorUpdatingVariableListener](https://github.com/TimefoldAI/timefold-quickstarts/blob/stable/java/food-packaging/src/main/java/org/acme/foodpackaging/domain/LineOperatorUpdatingVariableListener.java) to declarative shadow variables.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
@@ -1,9 +1,13 @@
 package ai.timefold.solver.core.impl.domain.variable.declarative;
 
+import java.util.Collections;
 import java.util.function.BiConsumer;
 
 import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.CollectionInverseVariableDemand;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.CollectionInverseVariableSupply;
 import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.preview.api.domain.metamodel.VariableMetaModel;
 
 import org.jspecify.annotations.Nullable;
 
@@ -16,6 +20,17 @@ public record ChangedVariableNotifier<Solution_>(BiConsumer<VariableDescriptor<S
             (a, b) -> {
             },
             null);
+
+    public CollectionInverseVariableSupply getCollectionInverseVariableSupply(VariableMetaModel<?, ?, ?> variableMetaModel) {
+        if (innerScoreDirector == null) {
+            return entity -> Collections.emptyList();
+        } else {
+            var solutionDescriptor = innerScoreDirector.getSolutionDescriptor();
+            var variableDescriptor = solutionDescriptor.getEntityDescriptorStrict(variableMetaModel.entity().type())
+                    .getVariableDescriptor(variableMetaModel.name());
+            return innerScoreDirector.getSupplyManager().demand(new CollectionInverseVariableDemand<>(variableDescriptor));
+        }
+    }
 
     @SuppressWarnings("unchecked")
     public static <Solution_> ChangedVariableNotifier<Solution_> empty() {
--- a/core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSourceTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSourceTest.java
@@ -436,6 +436,55 @@ class RootVariableSourceTest {
     }
 
     @Test
+    void pathUsingBuiltinShadowAfterBuiltinShadowAfterFact() {
+        var rootVariableSource = RootVariableSource.from(
+                planningSolutionMetaModel,
+                TestdataInvalidDeclarativeValue.class,
+                "shadow",
+                "fact.previous.previous",
+                DEFAULT_MEMBER_ACCESSOR_FACTORY,
+                DEFAULT_DESCRIPTOR_POLICY);
+
+        assertThat(rootVariableSource.rootEntity()).isEqualTo(TestdataInvalidDeclarativeValue.class);
+        assertThat(rootVariableSource.variableSourceReferences()).hasSize(2);
+        assertThat(rootVariableSource.parentVariableType()).isEqualTo(ParentVariableType.INDIRECT);
+        var previousSource = rootVariableSource.variableSourceReferences().get(0);
+
+        assertChainToVariableEntity(previousSource, "fact");
+        assertThat(previousSource.variableMetaModel()).isEqualTo(previousElementMetaModel);
+        assertThat(previousSource.isTopLevel()).isTrue();
+        assertThat(previousSource.isBottomLevel()).isFalse();
+        assertThat(previousSource.onRootEntity()).isFalse();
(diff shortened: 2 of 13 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
```diff
@@ -17,6 +21,17 @@ public record ChangedVariableNotifier<Solution_>(BiConsumer<VariableDescriptor<S
             },
             null);
 
+    public CollectionInverseVariableSupply getCollectionInverseVariableSupply(VariableMetaModel<?, ?, ?> variableMetaModel) {
+        if (innerScoreDirector == null) {
```
triceo, line 25: Why would the score director be null, and isn't this situation more of an error state?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
```diff
@@ -17,6 +21,17 @@ public record ChangedVariableNotifier<Solution_>(BiConsumer<VariableDescriptor<S
...
+            return entity -> Collections.emptyList();
+        } else {
+            var solutionDescriptor = innerScoreDirector.getSolutionDescriptor();
+            var variableDescriptor = solutionDescriptor.getEntityDescriptorStrict(variableMetaModel.entity().type())
+                    .getVariableDescriptor(variableMetaModel.name());
+            return innerScoreDirector.getSupplyManager().demand(new CollectionInverseVariableDemand<>(variableDescriptor));
```
triceo, line 31: Is it possible for the method to be called multiple times with the same variable descriptor? This method should, under no circumstances, cause the demands to accumulate.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSourceTest.java
```diff
@@ -389,36 +389,67 @@ void pathUsingDeclarativeShadowAfterBuiltinShadowAfterGroup() {
...
     @Test
-    void invalidPathMissingProperty() {
-        assertThatCode(() -> RootVariableSource.from(
+    void pathUsingBuiltinShadowAfterBuiltinShadow() {
+        var rootVariableSource = RootVariableSource.from(
                 planningSolutionMetaModel,
                 TestdataInvalidDeclarativeValue.class,
                 "shadow",
-                "missing",
+                "previous.previous",
```
triceo, line 397: This is strange. 

- If I want to reference 5 values back, do I have to do previous.previous.previous.previous.previous? Also, this is done at build-time - how can I reference an arbitrary previous reference?
- What happens if any of the values in this chain are null?

Honestly, it seems to me that this should not be allowed.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -871,7 +871,7 @@ These paths must follow 1 of 3 syntactic forms:
 
 |Chained Property Path
 |"a.b.c"
-|Refers to a supplier variable reachable via chained properties. Intermediate elements must be facts or variables; the final one must be a supplier variable.
+|Refers to a supplier variable reachable via chained properties. Intermediate elements must be facts or variables; the final one must be a variable.
```
triceo: What is the difference between a variable and a supplier variable?
The path "refers to a supplier variable", yet "final one must be a variable". It seems like the terms are used interchangeably.

triceo (commented):
I have concerns, let's discuss.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
Christopher-Chianelli: Can/should change "supplier variable" to "variable" here; supplier variables are declarative shadow variables; variable is any variable.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSourceTest.java
Christopher-Chianelli: - I think we limit it to 2 somewhere, although that might only be for indirect facts
- Null -> it will be called, but there will be no edge in the graph

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
Christopher-Chianelli: Yes; but wouldn't the demands share the same supply since the variable descriptor the same instance? The demands are also Solver lifecycle long: https://github.com/TimefoldAI/timefold-solver/blob/16e2464fda033694b962d06e1c6895e1e838a74b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/supply/AbstractVariableDescriptorBasedDemand.java#L25-L40

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
Christopher-Chianelli: Nope; you can create a ShadowVariableSession without a planning solution or a score director; `SolutionManager.updateShadowVariables(Object entities)` uses it.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
triceo: Let's. The Javadoc on `ShadowSources` has a similar issue I think.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
triceo: And how do we know then that the correct value to return is an empty list? I'd like to see a test for this situation.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
triceo: Yes, the demands would result in the same supply. At the same time though, a counter would be increased; so this effectively creates "a leak", a supply that will never be un-demanded.

Is it a problem? Probably not.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSourceTest.java
triceo: Let's check if there's a limit. If not, let's introduce one. 2 sounds reasonable.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
Christopher-Chianelli: Any value that does not crash the updater is correct; score director is null only in the case we are updating the shadow variables of a `Solution` via `SolutionManager`. The supply is used to register change listeners. Since there is no score director, there are no change listeners that need to be registered. The test would be for one of the `SolutionManager` update methods, and there is one existing tests that check the behaviour.

triceo (approved):

PR #1736, opened 2025-08-13
fix: start the termination counter at the first step

zepfred:
When the phase start takes a long time (such as when nearby distance matrix is computed), that time will eat into the grace period of diminished returns, and possibly entirely eradicate it. Therefore we only start the grace period at first step start.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
@@ -22,8 +22,8 @@ final class DiminishedReturnsTermination<Solution_, Score_ extends Score<Score_>
     private final double minimumImprovementRatio;
 
     private boolean isGracePeriodActive;
-    boolean locked = true;
-    long gracePeriodStartTimeNanos;
+    private boolean isGracePeriodStarted = false;
+    private long gracePeriodStartTimeNanos;
     private double gracePeriodSoftestImprovementDouble;
 
     private final DiminishedReturnsScoreRingBuffer<Score_> scoresByTime;
@@ -98,7 +98,7 @@ final class DiminishedReturnsTermination<Solution_, Score_ extends Score<Score_>
     private void resetGracePeriod(long currentTime, InnerScore<Score_> startingScore) {
         gracePeriodStartTimeNanos = currentTime;
         isGracePeriodActive = true;
-        locked = false;
+        isGracePeriodStarted = true;
 
         // Remove all entries in the map since grace is reset
         scoresByTime.clear();
@@ -108,7 +108,7 @@ final class DiminishedReturnsTermination<Solution_, Score_ extends Score<Score_>
     }
 
     public boolean isTerminated(long currentTime, InnerScore<Score_> endScore) {
-        if (locked) {
+        if (!isGracePeriodStarted) {
             return false;
         }
         if (isGracePeriodActive) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
```diff
@@ -97,6 +98,7 @@ public void step(long time, InnerScore<Score_> bestScore) {
     private void resetGracePeriod(long currentTime, InnerScore<Score_> startingScore) {
         gracePeriodStartTimeNanos = currentTime;
         isGracePeriodActive = true;
+        locked = false;
```
triceo: Why can't we not `isGracePeriodActive` for this? Seems to be the same thing.

And if we can't, this IMO needs a better name; Like "isGracePeriodActive". :-)

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
```diff
@@ -22,7 +22,8 @@ final class DiminishedReturnsTermination<Solution_, Score_ extends Score<Score_>
     private final double minimumImprovementRatio;
 
     private boolean isGracePeriodActive;
-    private long gracePeriodStartTimeNanos;
+    boolean locked = true;
```
triceo: I don't like that we're exposing internal synchronization details of the algorithm.

If you need to expose something for tests, please expose the current state - such as whether the period is active, or its start time.

core/src/test/java/ai/timefold/solver/core/impl/solver/termination/UnimprovedTimeMillisSpentTerminationTest.java
```diff
@@ -131,4 +135,35 @@ void phaseTerminationWithConstructionHeuristic() { // CH ignores unimproved time
...
+    @Test
+    void phaseTerminationAndLongInitializationPeriod() {
+        // The goal is to verify termination after a long initialization,
+        // ensuring the termination does not occur until all necessary resources,
+        // such as distance matrices, are fully loaded.
+        var solverScope = new SolverScope<TestdataSolution>();
+        var phaseScope = spy(new LocalSearchPhaseScope<>(solverScope, 0));
+        var stepScope = mock(AbstractStepScope.class);
+
+        Clock clock = mock(Clock.class);
```
triceo: See `MockClock` in test utils.

triceo (commented):
I think the code could be simplified. But other than that, LGTM.

core/src/main/java/ai/timefold/solver/core/impl/solver/termination/DiminishedReturnsTermination.java
zepfred: No, `isGracePeriodActive` means it hasn't started yet or it has already ended. We need a new variable.

triceo: But that's the point - if you reset it back to `false` when it ends, it serves exactly the same purpose as `locked`.

zepfred: The `locked` variable will be updated only once when the grace period is activated for the first time.

There are two evaluation phases: one during the grace period and another after it ends. If we begin using the activation variable, we will not trigger the second evaluation, which occurs outside the grace period evaluation block.

Returning `false` for `!isGracePeriodActivated` would cause us to skip the second termination evaluation phase, creating an inconsistency.

triceo (approved):

PR #1738, opened 2025-08-13
chore: create filtering value range selectors as early as possible

zepfred:
This PR updates the value and entity factories to create the entity filtering-related selectors as early as possible. This way, the reachable values are returned first and then processed by outer node selectors. Additionally, it also includes a small change to minimize the hash lookup.

(The next PR will focus on these hash lookups)

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
@@ -99,7 +99,7 @@ public class EntityDescriptor<Solution_> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(EntityDescriptor.class);
 
-    private final int ordinal;
+    private final int ordinalId;
     private final SolutionDescriptor<Solution_> solutionDescriptor;
     private final Class<?> entityClass;
     private final List<Class<?>> declaredInheritedEntityClassList = new ArrayList<>();
@@ -147,8 +147,8 @@ public class EntityDescriptor<Solution_> {
     // Constructors and simple getters/setters
     // ************************************************************************
 
-    public EntityDescriptor(int ordinal, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
-        this.ordinal = ordinal;
+    public EntityDescriptor(int ordinalId, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
+        this.ordinalId = ordinalId;
         SolutionDescriptor.assertMutable(entityClass, "entityClass");
         this.solutionDescriptor = solutionDescriptor;
         this.entityClass = entityClass;
@@ -170,8 +170,8 @@ public class EntityDescriptor<Solution_> {
      *
      * @return zero or higher
      */
-    public int getOrdinal() {
-        return ordinal;
+    public int getOrdinalId() {
+        return ordinalId;
     }
 
     /**
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
@@ -45,6 +45,7 @@ import ai.timefold.solver.core.impl.domain.common.ReflectionHelper;
 import ai.timefold.solver.core.impl.domain.common.accessor.MemberAccessor;
 import ai.timefold.solver.core.impl.domain.policy.DescriptorPolicy;
 import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.valuerange.descriptor.CompositeValueRangeDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.anchor.AnchorShadowVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.cascade.CascadingUpdateShadowVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.custom.CustomShadowVariableDescriptor;
@@ -99,7 +100,7 @@ public class EntityDescriptor<Solution_> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(EntityDescriptor.class);
 
-    private final int ordinalId;
+    private final int ordinal;
     private final SolutionDescriptor<Solution_> solutionDescriptor;
     private final Class<?> entityClass;
     private final List<Class<?>> declaredInheritedEntityClassList = new ArrayList<>();
@@ -147,8 +148,8 @@ public class EntityDescriptor<Solution_> {
     // Constructors and simple getters/setters
     // ************************************************************************
 
-    public EntityDescriptor(int ordinalId, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
-        this.ordinalId = ordinalId;
+    public EntityDescriptor(int ordinal, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
+        this.ordinal = ordinal;
         SolutionDescriptor.assertMutable(entityClass, "entityClass");
         this.solutionDescriptor = solutionDescriptor;
         this.entityClass = entityClass;
@@ -170,8 +171,8 @@ public class EntityDescriptor<Solution_> {
      *
      * @return zero or higher
      */
-    public int getOrdinalId() {
-        return ordinalId;
+    public int getOrdinal() {
+        return ordinal;
     }
 
     /**
@@ -321,8 +322,7 @@ public class EntityDescriptor<Solution_> {
         if (((AnnotatedElement) member).isAnnotationPresent(ValueRangeProvider.class)) {
             var memberAccessor = descriptorPolicy.getMemberAccessorFactory().buildAndCacheMemberAccessor(member,
                     FIELD_OR_READ_METHOD, ValueRangeProvider.class, descriptorPolicy.getDomainAccessType());
-            descriptorPolicy.addFromEntityValueRangeProvider(
-                    memberAccessor);
+            descriptorPolicy.addFromEntityValueRangeProvider(memberAccessor);
         }
     }
 
@@ -738,6 +738,20 @@ public class EntityDescriptor<Solution_> {
 
     public long getGenuineVariableCount() {
         return effectiveGenuineVariableDescriptorList.size();
+    }
+
+    public int getValueRangeCount() {
+        var count = 0;
+        for (var genuineVariableDescriptor : effectiveGenuineVariableDescriptorList) {
+            if (genuineVariableDescriptor
+                    .getValueRangeDescriptor() instanceof CompositeValueRangeDescriptor<Solution_> compositeValueRangeDescriptor) {
+                // sum the child descriptors size
+                // and add one more unit to the composite descriptor itself at the end of the iter
+                count += compositeValueRangeDescriptor.getValueRangeCount();
+            }
+            count++;
+        }
+        return count;
     }
 
     public Collection<ShadowVariableDescriptor<Solution_>> getShadowVariableDescriptors() {
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
@@ -30,6 +30,8 @@ public class DescriptorPolicy {
     private Set<PreviewFeature> enabledPreviewFeatureSet = EnumSet.noneOf(PreviewFeature.class);
     @Nullable
     private MemberAccessorFactory memberAccessorFactory;
+    private int entityOrdinalCount = 0;
+    private int valueRangeOrdinalCount = 0;
 
     public void addFromSolutionValueRangeProvider(MemberAccessor memberAccessor) {
         String id = extractValueRangeProviderId(memberAccessor);
@@ -155,4 +157,19 @@ public class DescriptorPolicy {
         return valueRangeProviderIds;
     }
 
+    public int reserveEntityId() {
+        return entityOrdinalCount++;
+    }
+
+    public int getEntityIdCount() {
+        return entityOrdinalCount;
+    }
+
+    public int reserveValueRangeId() {
+        return valueRangeOrdinalCount++;
+    }
+
+    public int getValueRangeIdCount() {
+        return valueRangeOrdinalCount;
+    }
 }
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
@@ -72,7 +72,6 @@ public class DescriptorPolicy {
     @Nullable
     private MemberAccessorFactory memberAccessorFactory;
     private int entityDescriptorCount = 0;
-    private int scoreDescriptorCount = 0;
     private int valueRangeDescriptorCount = 0;
 
     public <Solution_> EntityDescriptor<Solution_> buildEntityDescriptor(SolutionDescriptor<Solution_> solutionDescriptor,
@@ -88,7 +87,7 @@ public class DescriptorPolicy {
         PlanningScore annotation = extractPlanningScoreAnnotation(scoreMemberAccessor);
         ScoreDefinition<Score_> scoreDefinition =
                 buildScoreDefinition(solutionClass, scoreMemberAccessor, scoreType, annotation);
-        return new ScoreDescriptor<>(scoreDescriptorCount++, scoreMemberAccessor, scoreDefinition);
+        return new ScoreDescriptor<>(scoreMemberAccessor, scoreDefinition);
     }
 
     public <Solution_> CompositeValueRangeDescriptor<Solution_> buildCompositeValueRangeDescriptor(
@@ -115,17 +114,18 @@ public class DescriptorPolicy {
             Class<?> solutionClass) {
         Class<?> memberType = scoreMemberAccessor.getType();
         if (!Score.class.isAssignableFrom(memberType)) {
-            throw new IllegalStateException("The solutionClass (" + solutionClass
-                    + ") has a @" + PlanningScore.class.getSimpleName()
-                    + " annotated member (" + scoreMemberAccessor + ") that does not return a subtype of Score.");
+            throw new IllegalStateException(
+                    "The solutionClass (%s) has a @%s annotated member (%s) that does not return a subtype of Score."
+                            .formatted(solutionClass, PlanningScore.class.getSimpleName(), scoreMemberAccessor));
         }
         if (memberType == Score.class) {
-            throw new IllegalStateException("The solutionClass (" + solutionClass
-                    + ") has a @" + PlanningScore.class.getSimpleName()
-                    + " annotated member (" + scoreMemberAccessor
-                    + ") that doesn't return a non-abstract " + Score.class.getSimpleName() + " class.\n"
-                    + "Maybe make it return " + HardSoftScore.class.getSimpleName()
-                    + " or another specific " + Score.class.getSimpleName() + " implementation.");
+            throw new IllegalStateException(
+                    """
+                            The solutionClass (%s) has a @%s annotated member (%s) that doesn't return a non-abstract %s class.
+                            Maybe make it return %s or another specific %s implementation."""
+                            .formatted(solutionClass, PlanningScore.class.getSimpleName(), scoreMemberAccessor,
+                                    Score.class.getSimpleName(), HardSoftScore.class.getSimpleName(),
+                                    Score.class.getSimpleName()));
         }
         return (Class<Score_>) memberType;
     }
@@ -153,12 +153,10 @@ public class DescriptorPolicy {
         if (!Objects.equals(scoreDefinitionClass, PlanningScore.NullScoreDefinition.class)) {
             if (bendableHardLevelsSize != PlanningScore.NO_LEVEL_SIZE
                     || bendableSoftLevelsSize != PlanningScore.NO_LEVEL_SIZE) {
-                throw new IllegalArgumentException("The solutionClass (" + solutionClass
-                        + ") has a @" + PlanningScore.class.getSimpleName()
-                        + " annotated member (" + scoreMemberAccessor
-                        + ") that has a scoreDefinition (" + scoreDefinitionClass
-                        + ") that must not have a bendableHardLevelsSize (" + bendableHardLevelsSize
-                        + ") or a bendableSoftLevelsSize (" + bendableSoftLevelsSize + ").");
+                throw new IllegalArgumentException(
+                        "The solutionClass (%s) has a @%s annotated member (%s) that has a scoreDefinition (%s) that must not have a bendableHardLevelsSize (%d) or a bendableSoftLevelsSize (%d)."
+                                .formatted(solutionClass, PlanningScore.class.getSimpleName(), scoreMemberAccessor,
+                                        scoreDefinitionClass, bendableHardLevelsSize, bendableSoftLevelsSize));
             }
             return ConfigUtils.newInstance(() -> scoreMemberAccessor + " with @" + PlanningScore.class.getSimpleName(),
                     "scoreDefinitionClass", scoreDefinitionClass);
@@ -166,12 +164,10 @@ public class DescriptorPolicy {
         if (!IBendableScore.class.isAssignableFrom(scoreType)) {
             if (bendableHardLevelsSize != PlanningScore.NO_LEVEL_SIZE
                     || bendableSoftLevelsSize != PlanningScore.NO_LEVEL_SIZE) {
-                throw new IllegalArgumentException("The solutionClass (" + solutionClass
-                        + ") has a @" + PlanningScore.class.getSimpleName()
-                        + " annotated member (" + scoreMemberAccessor
-                        + ") that returns a scoreType (" + scoreType
-                        + ") that must not have a bendableHardLevelsSize (" + bendableHardLevelsSize
-                        + ") or a bendableSoftLevelsSize (" + bendableSoftLevelsSize + ").");
+                throw new IllegalArgumentException(
+                        "The solutionClass (%s) has a @%s annotated member (%s) that returns a scoreType (%s) that must not have a bendableHardLevelsSize (%d) or a bendableSoftLevelsSize (%d)."
+                                .formatted(solutionClass, PlanningScore.class.getSimpleName(), scoreMemberAccessor, scoreType,
+                                        bendableHardLevelsSize, bendableSoftLevelsSize));
             }
             if (scoreType.equals(SimpleScore.class)) {
                 return (ScoreDefinition_) new SimpleScoreDefinition();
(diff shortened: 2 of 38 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -147,8 +147,8 @@ public class EntityDescriptor<Solution_> {
     // Constructors and simple getters/setters
     // ************************************************************************
 
-    public EntityDescriptor(int ordinal, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
-        this.ordinal = ordinal;
+    public EntityDescriptor(int ordinalId, SolutionDescriptor<Solution_> solutionDescriptor, Class<?> entityClass) {
```
triceo: I don't see a need for this change. "Ordinal" already implies a specific unique ordering.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
```diff
@@ -30,6 +30,8 @@ public class DescriptorPolicy {
     private Set<PreviewFeature> enabledPreviewFeatureSet = EnumSet.noneOf(PreviewFeature.class);
     @Nullable
     private MemberAccessorFactory memberAccessorFactory;
+    private int entityOrdinalCount = 0;
+    private int valueRangeOrdinalCount = 0;
```
triceo: Arguably this should be `entityCount` and `valueRangeCount`.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
```diff
@@ -155,4 +157,19 @@ public Collection<String> getValueRangeProviderIds() {
         return valueRangeProviderIds;
     }
 
+    public int reserveEntityId() {
```
triceo: This is not an entity ID, this is an entity descriptor ID.
Also arguably, it's not an ID, it's an ordinal.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
```diff
@@ -155,4 +157,19 @@ public Collection<String> getValueRangeProviderIds() {
...
+        return entityOrdinalCount++;
+    }
+
+    public int getEntityIdCount() {
+        return entityOrdinalCount;
+    }
+
+    public int reserveValueRangeId() {
```
triceo: Same comment about the ID.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -154,7 +153,8 @@ public static <Solution_> SolutionDescriptor<Solution_> buildSolutionDescriptor(
             updatedEntityClassList.addAll(filteredInheritedEntityClasses);
         }
         for (var entityClass : sortEntityClassList(updatedEntityClassList)) {
-            var entityDescriptor = new EntityDescriptor<>(ordinal++, solutionDescriptor, entityClass);
+            var entityDescriptor =
+                    new EntityDescriptor<>(descriptorPolicy.reserveEntityId(), solutionDescriptor, entityClass);
```
triceo: If we create the entity descriptors from within the policy, we will be able to completely hide the reservation mechanism from the outside code.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -267,6 +270,9 @@ private static boolean hasAnyAnnotatedMembers(Class<?> solutionClass) {
     private AutoDiscoverMemberType autoDiscoverMemberType;
     private LookUpStrategyResolver lookUpStrategyResolver;
 
+    private int maxEntityOrdinal = 0;
+    private int maxValueRangeOrdinal = 0;
```
triceo: `entityDescriptorCount`, `valueRangeCount`.

Also, I'm a bit surprised these don't exist yet. The collections of descriptors should have the same information, no?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -636,6 +642,11 @@ private void processPlanningEntityPropertyAnnotation(DescriptorPolicy descriptor
         }
     }
 
+    private void processOrdinalIds(DescriptorPolicy descriptorPolicy) {
+        this.maxEntityOrdinal = descriptorPolicy.getEntityIdCount();
+        this.maxValueRangeOrdinal = descriptorPolicy.getValueRangeIdCount();
+    }
```
triceo: Given the comment above, this method is maybe unnecessary?
It introduces another duplicate source of the same information.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/AbstractValueRangeDescriptor.java
```diff
@@ -14,12 +14,19 @@ public abstract sealed class AbstractValueRangeDescriptor<Solution_>
         implements ValueRangeDescriptor<Solution_>
         permits AbstractFromPropertyValueRangeDescriptor, CompositeValueRangeDescriptor {
 
+    private final int ordinalId;
     protected final GenuineVariableDescriptor<Solution_> variableDescriptor;
 
-    protected AbstractValueRangeDescriptor(GenuineVariableDescriptor<Solution_> variableDescriptor) {
+    protected AbstractValueRangeDescriptor(int ordinalId, GenuineVariableDescriptor<Solution_> variableDescriptor) {
+        this.ordinalId = ordinalId;
```
triceo: `ordinal` ought to be enough.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/NotifiableRegistry.java
```diff
@@ -68,7 +68,7 @@ Collection<VariableListenerNotifiable<Solution_>> get(VariableDescriptor<?> vari
 
     Collection<ListVariableListenerNotifiable<Solution_>> get(ListVariableDescriptor<?> variableDescriptor) {
         var notifiables =
-                sourceVariableToNotifiableListArray[variableDescriptor.getEntityDescriptor().getOrdinal()][variableDescriptor
+                sourceVariableToNotifiableListArray[variableDescriptor.getEntityDescriptor().getOrdinalId()][variableDescriptor
```
triceo: See inconsistency here. One descriptor has `ordinalId`, the other has `ordinal`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
```diff
@@ -100,18 +103,30 @@ public boolean isNeverEnding() {
...
+            replayedValue = iterator.next();
+        }
+        return replayedValue;
+    }
+
     @Override
     public Iterator<Object> endingIterator() {
-        return new OriginalFilteringValueRangeIterator(replayingValueSelector.iterator(), reachableValues);
+        // The ending iterator returns all values from the child iterator and ignores the reachable values
+        return getChildEntitySelector().endingIterator();
```
triceo: If the iterator ignores reachable values, shouldn't there be a filter here which checks for those? How else are they filtered out?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -33,37 +28,26 @@ private DestinationSelectorFactory(DestinationSelectorConfig destinationSelector
...
 
     private IterableValueSelector<Solution_> buildIterableValueSelector(
             HeuristicConfigPolicy<Solution_> configPolicy, EntityDescriptor<Solution_> entityDescriptor,
-            SelectionCacheType minimumCacheType, SelectionOrder inheritedSelectionOrder) {
+            SelectionCacheType minimumCacheType, SelectionOrder inheritedSelectionOrder, String entityValueRangeRecorderId) {
+        // Destination selector does not require asserting both sides,
+        // which means checking only if the destination entity accept the selected value
         ValueSelector<Solution_> valueSelector = ValueSelectorFactory
                 .<Solution_> create(Objects.requireNonNull(config.getValueSelectorConfig()))
                 .buildValueSelector(configPolicy, entityDescriptor, minimumCacheType, inheritedSelectionOrder,
```
triceo, line 54: This method was a bit crazy even before it got 2 more arguments. It's got 8 now. 

Considering how many builders there would be for these value selectors, maybe the builder pattern isn't the best choice here? Some context class instead? Up to you.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementDestinationSelector.java
```diff
@@ -171,27 +157,14 @@ public EntityDescriptor<Solution_> getEntityDescriptor() {
...
-        } else {
-            entityIterator = entitySelector.endingIterator();
-            valueIterator = valueSelector.endingIterator(null);
-        }
         return Stream.concat(
-                StreamSupport.stream(Spliterators.spliterator(entityIterator, entitySelector.getSize(), 0), false),
-                StreamSupport.stream(Spliterators.spliterator(valueIterator, valueSelector.getSize(), 0), false))
+                StreamSupport.stream(Spliterators.spliterator(entitySelector.endingIterator(), entitySelector.getSize(), 0),
+                        false),
+                StreamSupport.stream(Spliterators.spliterator(valueSelector.endingIterator(null), valueSelector.getSize(), 0),
```
triceo, line 163: How is it correct to not specify the entity?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorFactory.java
```diff
@@ -47,31 +47,39 @@ protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<So
...
-                        : mimicSelectorConfig.getId();
+                entityValueRangeRecorderId =
+                        mimicSelectorConfig.getMimicSelectorRef() != null ? mimicSelectorConfig.getMimicSelectorRef()
+                                : mimicSelectorConfig.getId();
             }
         }
+        if (enableEntityValueRangeFilter && entityValueRangeRecorderId == null) {
+            throw new IllegalStateException(
+                    "When enabling entity value range, the origin value selector ID is required for the destination selector %s."
+                            .formatted(destinationSelectorConfig));
```
triceo, line 73: Arguably this solver config requires documentation.
Clearly there is something special you need to do with your solver configs to prevent this from happening.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -56,8 +58,8 @@
 public final class ValueRangeManager<Solution_> {
 
     private final SolutionDescriptor<Solution_> solutionDescriptor;
-    private final Map<ValueRangeDescriptor<Solution_>, CountableValueRange<?>> fromSolutionMap = new IdentityHashMap<>();
-    private final Map<Object, Map<ValueRangeDescriptor<Solution_>, CountableValueRange<?>>> fromEntityMap =
+    private final List<CountableValueRange<?>> fromSolutionList;
```
triceo: Nice.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -383,8 +386,10 @@ public <T> CountableValueRange<T> getFromEntity(ValueRangeDescriptor<Solution_>
                     "Impossible state: value range (%s) on planning entity (%s) requested before the working solution is known."
                             .formatted(valueRangeDescriptor, entity));
         }
-        var valueRangeMap = fromEntityMap.computeIfAbsent(entity, e -> new IdentityHashMap<>());
-        var valueRange = valueRangeMap.get(valueRangeDescriptor);
+        var valueRangeList =
+                fromEntityMap.computeIfAbsent(entity,
+                        e -> CollectionUtils.newNullList(solutionDescriptor.countValueRangeDescriptor()));
+        CountableValueRange<?> valueRange = valueRangeList.get(valueRangeDescriptor.getOrdinalId());
```
triceo: Why not `var`? `valueRangeList` has the same generic.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -478,7 +483,8 @@ private static void updateValueMap(Map<Object, Set<Object>> valueMatrix, Countab
     }
 
     public void reset(@Nullable Solution_ workingSolution) {
-        fromSolutionMap.clear();
+        fromSolutionList.clear();
+        fromSolutionList.addAll(CollectionUtils.newNullList(solutionDescriptor.countValueRangeDescriptor()));
```
triceo: Strange hack.
If this is to avoid AIOOBE, why not just check the index on access instead?

Also, if we absolutely need this, surely this is not such a common use case to create a util method for this.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
```diff
@@ -445,10 +445,10 @@ void randomWithEntityValueRangeAndFiltering() {
         // and that causes the B destination to become an invalid destination
         assertCodesOfNeverEndingMoveSelector(moveSelector,
                 "1 {A[1]->A[1]}",
-                "3 {B[0]->A[1]}",
-                "1 {A[1]->A[1]}",
+                "3 {B[0]->B[0]}",
                 "1 {A[1]->A[1]}",
-                "1 {A[1]->A[0]}");
+                "3 {B[0]->B[0]}",
+                "3 {B[0]->B[0]}");
```
triceo, line 451: Why these changes?
Also, specifically, why changes which clearly result in moves that do nothing? (I realize it was already like this before, but I kinda hoped that the value ranges would prevent it.)

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
zepfred: They don't share a common base class. Let me add one.

triceo: I probably misspoke. 

There are collections in the solution descriptor which carry every entity descriptor; therefore `maxEntityOrdinal` is not really necessary, because the collection size is exactly that number.

zepfred: Yes, we can use it instead of creating a new variable. I've updated the code to use it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityValueRangeSelector.java
zepfred: That makes sense. I had to change it because Nearby needs all values to construct the distance matrix.

triceo: You just made me realize something - you cannot return a non-filtered iterator from a filtering selector. You're breaking user expectations, it makes no sense. This needs a different approach.

zepfred: This seems difficult to resolve. When you look at `EntityNearbyDistanceMatrixDemand#L64`, you see it calls the replaying selector, which invokes this method. At that point, we need to read all values, but `OriginalFilteringValueRangeIterator` throws an error because there is no value to replay at all.

I see no other solution but a new method for the entity selector contract: `childEndingIterator()`.

zepfred: This issue also applies to `FilteringValueRangeSelector`

triceo: This is very specific to filtering selectors.
Instead of making every selector implement this complexity, you can do some instanceof checks in the places where you actually need to get this iterator.

But there is a different question - if there are no values to replay, why is the selector trying to replay anything? Sounds like this could be the actual solution -- finding out a way to cut this problem off at the root.

zepfred: I realize we will need to adjust how we build the nearby selectors, as they should receive a non-filtering-entity-value-range selector to iterate over the values.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
zepfred: ~~A context class seems simpler~~

zepfred: Honestly, the context pattern didn't improve the code. I'll leave it as it is now.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementDestinationSelector.java
zepfred: Could you please explain it more clearly?

zepfred: Also, this is the original code before any of these entity range adjustments.

https://github.com/TimefoldAI/timefold-solver/blob/8c5f9e495e3be00bbd49233aea75647dd88db0f8/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementDestinationSelector.java#L160

triceo: The `null` in there - `valueSelector.endingIterator(null)`.
The code expects an entity, and we don't give any.

It's good that it was already like this before, but the question remains. How is this safe? And consequently, if we don't need the entity, maybe there should be a better interface?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorFactory.java
zepfred: This shouldn't be possible. The previous block should guarantee the condition is satisfied.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
zepfred: Comparing the execution of both branches, I noted that this [call](https://github.com/TimefoldAI/timefold-solver/blob/8c0ae7fc8d09d98139ff117848420396306281ae/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java#L58) returns different numbers, which explains that.

I attribute this difference to the evaluation order of `FilteringEntityValueRangeSelector`, which now occurs first.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
```diff
@@ -30,6 +71,173 @@ public class DescriptorPolicy {
     private Set<PreviewFeature> enabledPreviewFeatureSet = EnumSet.noneOf(PreviewFeature.class);
     @Nullable
     private MemberAccessorFactory memberAccessorFactory;
+    private int entityDescriptorCount = 0;
+    private int scoreDescriptorCount = 0;
```
triceo: If there's ever more than 1 score descriptor, I'd say that's a fail-fast situation.
Also, do we allow solutions without this? If not, then this can only ever be 1, and therefore it's kinda pointless.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
```diff
@@ -30,6 +71,173 @@ public class DescriptorPolicy {
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <Score_ extends Score<Score_>> Class<Score_> extractScoreType(MemberAccessor scoreMemberAccessor,
+            Class<?> solutionClass) {
+        Class<?> memberType = scoreMemberAccessor.getType();
+        if (!Score.class.isAssignableFrom(memberType)) {
+            throw new IllegalStateException("The solutionClass (" + solutionClass
+                    + ") has a @" + PlanningScore.class.getSimpleName()
+                    + " annotated member (" + scoreMemberAccessor + ") that does not return a subtype of Score.");
```
triceo: Since this is already new code in a new class, let's update exceptions in this class to the `formatted(...)` pattern.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -87,7 +87,7 @@
  * @param <Solution_> the solution type, the class with the {@link ai.timefold.solver.core.api.domain.solution.PlanningSolution}
```
triceo: Please remove the FQN.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
```diff
@@ -95,12 +106,18 @@ public EntitySelector<Solution_> buildEntitySelector(HeuristicConfigPolicy<Solut
...
         if (nearbySelectionConfig != null) {
             // TODO Static filtering (such as movableEntitySelectionFilter) should affect nearbySelection
             entitySelector = applyNearbySelection(configPolicy, nearbySelectionConfig, minimumCacheType,
                     resolvedSelectionOrder, entitySelector);
+        } else {
+            // The nearby selector will implement its own logic to filter out unreachable elements.
+            // Therefore, we only apply entity value range filtering if the nearby feature is not enabled;
+            // otherwise, we would end up applying the filtering logic twice.
+            entitySelector = applyEntityValueRangeFiltering(configPolicy, entitySelector, entityValueRangeRecorderId,
+                    minimumCacheType, inheritedSelectionOrder, baseRandomSelection);
```
triceo, line 119: The user should still be able to apply their own filtering _on top_ of the value range filtering, no?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -82,30 +67,40 @@ private IterableValueSelector<Solution_> buildIterableValueSelector(
...
-            SelectionOrder selectionOrder, boolean randomSelection) {
-        var valueSelectorConfig = new ValueSelectorConfig()
-                .withMimicSelectorRef(valueSelectorId);
-        var replayingValueSelector = (IterableValueSelector<Solution_>) ValueSelectorFactory
-                .<Solution_> create(valueSelectorConfig)
-                .buildValueSelector(configPolicy, entitySelector.getEntityDescriptor(), minimumCacheType, selectionOrder);
-        return new FilteringEntityValueRangeSelector<>(entitySelector, replayingValueSelector, randomSelection);
+        // The nearby selector will implement its own logic to filter out unreachable elements.
+        // It requires the child selectors to not be FilteringEntityValueRangeSelector or FilteringValueRangeSelector,
+        // as it needs to iterate over all available values to construct the distance matrix.
```
triceo, line 84: Same question. Value ranges are great, but that doesn't mean the user loses the capability to filter them even further.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -112,13 +125,18 @@ public ValueSelector<Solution_> buildValueSelector(HeuristicConfigPolicy<Solutio
...
         if (nearbySelectionConfig != null) {
             // TODO Static filtering (such as movableEntitySelectionFilter) should affect nearbySelection too
             valueSelector = applyNearbySelection(configPolicy, entityDescriptor, minimumCacheType,
                     resolvedSelectionOrder, valueSelector);
+        } else {
+            // The nearby selector will implement its own logic to filter out unreachable elements.
+            // Therefore, we only apply entity value range filtering if the nearby feature is not enabled;
+            // otherwise, we would end up applying the filtering logic twice.
+            valueSelector = applyValueRangeFiltering(configPolicy, valueSelector, entityDescriptor, minimumCacheType,
+                    inheritedSelectionOrder, randomSelection, entityValueRangeRecorderId, assertBothSides);
```
triceo, line 138: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -56,8 +57,8 @@
 public final class ValueRangeManager<Solution_> {
 
     private final SolutionDescriptor<Solution_> solutionDescriptor;
-    private final Map<ValueRangeDescriptor<Solution_>, CountableValueRange<?>> fromSolutionMap = new IdentityHashMap<>();
-    private final Map<Object, Map<ValueRangeDescriptor<Solution_>, CountableValueRange<?>>> fromEntityMap =
+    private final CountableValueRange<?>[] fromSolutionList;
```
triceo: It's not a list, it's an array.

core/src/main/java/ai/timefold/solver/core/impl/domain/policy/DescriptorPolicy.java
zepfred: We already adopt a fail-fast approach with various planning score definitions, and  I removed the ordinal field.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
zepfred: Yes, the user still can. The next line will handle the custom filtering. The main idea is that we filter out unreachable values before applying filtering, sorting, and other decorators.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
zepfred: Same answer as before. Note that the value and entity creation rely on `config.getEntitySelectorConfig()` and `config.getValueSelectorConfig()`, respectively. Therefore, if there are any custom user settings, they will be applied by the respective factories.

On the other hand, we don't apply the entity value range decorator, since the Nearby selector will handle that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
zepfred: Same idea as entity factory. You can see the next line evaluates the user filtering setting:
```
        var randomSelection = determineBaseRandomSelection(variableDescriptor, resolvedCacheType, resolvedSelectionOrder);
        var valueSelector =
                buildBaseValueSelector(variableDescriptor, SelectionCacheType.max(minimumCacheType, resolvedCacheType),
                        randomSelection);
        var instanceCache = configPolicy.getClassInstanceCache();
        if (nearbySelectionConfig != null) {
            // TODO Static filtering (such as movableEntitySelectionFilter) should affect nearbySelection too
            valueSelector = applyNearbySelection(configPolicy, entityDescriptor, minimumCacheType,
                    resolvedSelectionOrder, valueSelector);
        } else {
            // The nearby selector will implement its own logic to filter out unreachable elements.
            // Therefore, we only apply entity value range filtering if the nearby feature is not enabled;
            // otherwise, we would end up applying the filtering logic twice.
            valueSelector = applyValueRangeFiltering(configPolicy, valueSelector, entityDescriptor, minimumCacheType,
                    inheritedSelectionOrder, randomSelection, entityValueRangeRecorderId, assertBothSides);
        }
---->        valueSelector = applyFiltering(valueSelector, instanceCache);     <-----
```

zepfred:
> Getting closer.
> 
> Are we sure this doesn't change the moves that are generated? I mean - of course it does, that's the point, but are we sure no moves are lost that should not have been lost?
> 
> It is difficult to evaluate the logic of the move selector structure.

Our FSR validation shows that invalid moves are no longer generated and the overall solution quality has improved, indicating that no valid moves are missed. Also, the coverage using the original iterators confirms that all expected moves are generated, and since the random iterators rely on the same approach, I don't think that's a problem.

triceo (approved):

PR #1739, opened 2025-08-13
fix: Make ShadowVariableUpdateHelper only add PlanningEntity classes as planning entities

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
```diff
@@ -84,6 +84,7 @@ public void updateShadowVariables(Solution_ solution) {
                 .stream()
                 .map(Object::getClass)
                 .distinct()
+                .filter(clazz -> clazz.isAnnotationPresent(PlanningEntity.class))
                 .toArray(Class[]::new);
```
triceo: Using `SolutionDescriptor.visitAllEntities()` will minimize iteration and reuse logic which already exists. The result should be the exact same, because `getAllEntitiesAndProblemFacts()` calls `visitAllEntities()` under the hood anyway.

Ideally we'd iterate just the entity descriptors, but I realize that - at this point - they might not yet exist.

core/src/test/java/ai/timefold/solver/core/testdomain/declarative/basic/TestdataBasicVarSolution.java
```diff
@@ -26,15 +27,20 @@ public static SolutionDescriptor<TestdataBasicVarSolution> buildSolutionDescript
     @ValueRangeProvider
     List<TestdataBasicVarValue> values;
 
+    @ProblemFactCollectionProperty
+    List<Object> problemFacts;
```
triceo, line 31: This change shouldn't be necessary once the above change is applied.

triceo (changes requested):
Please simplify, see comments. Good to go after that.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
triceo: Also, considering how expensive these operations are (especially the construction of `SolutionDescriptor`, _twice_) I wonder if we should cache the final descriptor.

Since it's unlikely people would have more than 1 solution class within the same JVM, such a static cache shouldn't present a problem.

core/src/test/java/ai/timefold/solver/core/testdomain/declarative/basic/TestdataBasicVarSolution.java
Christopher-Chianelli: It so we have a test that verify that `SolutionManager.updateShadowVariables` work on a solution with problem facts. That wasn't tested before.

triceo (approved):

PR #1748, opened 2025-08-18
ci: enable SBOMs

triceo:
Once we have this agreed, I'll do the same for Enterprise.

I have verified that the SBOMs are attached to the Maven project.
The attestations are likely impossible to verify without actually running the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,7 @@ jobs:
       contents: write  # IMPORTANT: required for action to create release branch
       pull-requests: write # IMPORTANT: so release PR can be created
       id-token: write  # IMPORTANT: mandatory for trusted publishing
+      attestations: write  # IMPORTANT: mandatory for attestations
     steps:
       - name: Checkout timefold-solver
         uses: actions/checkout@v5
@@ -56,12 +57,6 @@ jobs:
           java-version: '17'
           distribution: 'temurin'
           cache: 'maven'
-
-      # Need Maven 3.9.0+ to recognize MAVEN_ARGS.
-      - name: Set up Maven
-        uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
-        with:
-          maven-version: 3.9.3
 
       # We skip tests in dry run, to make the process faster.
       # Technically, this goes against the main reason for doing a dry run; to eliminate potential problems.


pom.xml
```diff
@@ -141,14 +147,29 @@
...
     </profile>
     <profile>
       <!--
-        Migration to Timefold 9 involves upgrading to Quarkus 3, Spring Boot 3 and migrating to jakarta.* packages.
-        Timefold 8 remains compatible with Quarkus 2, Spring Boot 2 and javax.* packages.
-        Both Timefold 8 and 9 are functionally equal and will be released simultaneously.
+        Migration to Timefold 0.9 involves upgrading to Quarkus 3, Spring Boot 3 and migrating to jakarta.* packages.
+        Timefold 0.8 remains compatible with Quarkus 2, Spring Boot 2 and javax.* packages.
+        Both Timefold 0.8 and 0.9 are functionally equal.
       -->
```
triceo, line 173: Unrelated, but I noticed this at the same time, so I did it anyway.

jreleaser.yml
```diff
@@ -45,10 +51,6 @@ deploy:
         stagingRepositories:
           - "target/staging-deploy"
         artifactOverrides:
-          - groupId: ai.timefold.solver
-            artifactId: timefold-solver-core
-            sourceJar: false
-            javadocJar: false
```
triceo, line 51: No longer applicable. This comes from a long time ago, when `core` was an empty module.

jreleaser.yml
```diff
@@ -6,24 +6,30 @@ signing:
...
+    excludes:
+      - '**/*.asc'
+
 release:
   github:
     commitAuthor:
       name: "Timefold Release Bot"
       email: "release@timefold.ai"
-    releaseName: "Timefold Solver Community Edition {{projectVersion}}"
+    releaseName: "Timefold Solver {{projectVersion}}"
```
triceo, line 21: I update this manually every release anyway.

jreleaser.yml
```diff
@@ -6,24 +6,30 @@ signing:
...
     milestone:
       close: true
-      name: "v{{projectVersion}}"
     changelog:
       formatted: ALWAYS
       preset: "conventional-commits"
       contentTemplate: ".github/workflows/release-changelog-template.md"
       contributors:
-        format: "- {{contributorName}}{{#contributorUsernameAsLink}} ({{.}}){{/contributorUsernameAsLink}}"
+        format: '- {{contributorName}} ({{contributorUsernameAsLink}})'
```
triceo, line 32: Dtto.

PR #1749, opened 2025-08-18
chore: rename @ShadowVariableLooped to @ShadowVariableInconsistent

### LINKED ISSUE #1527: Finalize @ShadowVariableLooped
New name was proposed: @ShadowVariablesInconsistent.
Two things come to mind:

- Do we enable Boolean? I seem to remember we intentionally only allowed boolean. That'd need to change.
- The inconsistency now speaks of not only looped, but also whether or not the shadow variables were set by the solver, incl. the built-in shadows. This means that not only would we need to somehow hook built-ins into this, but the annotation would now also make sense even without DSV.

The latter suggests that really, somehow, the built-ins should become DSVs; plugged into the same graph. That would solve this problem, because there would never be anything else but DSVs.

Needs to be discussed and finalized.

Christopher-Chianelli:
- Solving will automatically update all shadow variables, so you can now pass in solutions with incorrect/stale shadow variables and they'll be correctly updated.

- SolutionManager.update(SolutionUpdatePolicy.UPDATE_SCORE_ONLY, ...) will now fail fast if any @ShadowVariableInconsistent member is null or if any @InverseRelationVariable is null or incorrect.


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -229,6 +231,45 @@ public void resetWorkingSolution() {
             notifiable.resetWorkingSolution();
         }
 
+        if (scoreDirector.expectShadowVariablesInCorrectState()) {
+            var solutionDescriptor = scoreDirector.getSolutionDescriptor();
+            solutionDescriptor.visitAllEntities(scoreDirector.getWorkingSolution(),
+                    entity -> {
+                        var entityDescriptors = solutionDescriptor.getEntityDescriptors();
```
triceo: Can be avoided; solution descriptor has methods to retrieve the correct descriptor for any entity class.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -239,6 +280,85 @@ public void resetWorkingSolution() {
...
 
+    private void assertSingletonInverseDescriptor(InverseRelationShadowVariableDescriptor<Solution_> inverseDescriptor,
+            Solution_ workingSolution, Object shadowEntity) {
+        var inverse = inverseDescriptor.getValue(shadowEntity);
+        var sourceDescriptor = inverseDescriptor.getSourceVariableDescriptorList().get(0);
+        if (sourceDescriptor.isListVariable()) {
+            return; // ExternalizedListInverseVariableProcessor asserts this in setInverseAsserted
+        }
+        if (inverse == null) {
+            scoreDirector.getSolutionDescriptor().visitAllEntities(workingSolution, maybeSourceEntity -> {
```
triceo: I agree that, if done like this, these assertions become a massive performance issue. Iterating `entities^2` cannot happen, and thankfully it is avoidable.

You can do this in a single pass, if you build up the information in helper maps. Those maps can then be iterated once again, to mix and match the information. This will turn the problem from `entities^2` to `entities*2`, maybe `entities*3`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -239,6 +280,85 @@ public void resetWorkingSolution() {
...
+    private void assertCollectionInverseDescriptor(InverseRelationShadowVariableDescriptor<Solution_> inverseDescriptor,
+            Solution_ workingSolution, Object shadowEntity) {
+        var inverseCollection = (Collection<?>) inverseDescriptor.getValue(shadowEntity);
+        var sourceDescriptor = inverseDescriptor.getSourceVariableDescriptorList().get(0);
+        if (inverseCollection == null) {
+            throw new IllegalStateException("""
+                    The entity (%s) has a collection @%s that is null.
+                    Verify the consistency of your input solution.
+                    """.formatted(shadowEntity, InverseRelationShadowVariable.class.getSimpleName()));
+        } else {
```
triceo: Same comment on the complexity of this. But also...

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -239,6 +280,85 @@ public void resetWorkingSolution() {
...
+                                    The entity (%s) has a collection @%s (%s) that points to a source entity (%s) but that source entity points to (%s) instead.
+                                    Verify the consistency of your input solution.
+                                    """
+                                    .formatted(shadowEntity, InverseRelationShadowVariable.class.getSimpleName(),
+                                            inverseCollection,
+                                            source, sourceValue));
+                }
+            }
+
+            scoreDirector.getSolutionDescriptor().visitAllEntities(workingSolution, maybeSourceEntity -> {
```
triceo: ... this iteration should be part of the same loop as the previous one. The separation of concerns (different assertions) stands in the way of performance, and in this case, performance has to win.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -229,6 +231,45 @@ public void resetWorkingSolution() {
             notifiable.resetWorkingSolution();
         }
 
+        if (scoreDirector.expectShadowVariablesInCorrectState()) {
+            var solutionDescriptor = scoreDirector.getSolutionDescriptor();
+            solutionDescriptor.visitAllEntities(scoreDirector.getWorkingSolution(),
```
triceo: I'd move this entire code to score director. I'd hook this iteration over all entities to the other code there that already iterates over all entities - it may require some refactoring, as the code over there is not the cleanest and best structured. But these loops are expensive; we need to do everything we can to reduce their amount.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -229,6 +231,45 @@ public void resetWorkingSolution() {
             notifiable.resetWorkingSolution();
         }
 
+        if (scoreDirector.expectShadowVariablesInCorrectState()) {
```
triceo: This checking of assertions is so much code that I'd extract it to a special class. Especially if it's refactored to all happen in the same pass, the increased complexity of that will warrant separation.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -209,6 +209,16 @@ static <Score_ extends Score<Score_>> ConstraintAnalysis<Score_> getConstraintAn
...
+    /**
+     * Disables the checks controlled by {@link #expectShadowVariablesInCorrectState()}.
+     */
+    void disableShadowVariablesInCorrectStateChecks();
+
+    /**
+     * Enables the checks controlled by {@link #expectShadowVariablesInCorrectState()}.
+     */
+    void enableShadowVariablesInCorrectStateChecks();
+
```
triceo: Arguably when the assertion logic is moved to the score director, exposing this will no longer be necessary.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -332,7 +332,12 @@ public long getMoveEvaluationSpeed() {
 
     public void setWorkingSolutionFromBestSolution() {
         // The workingSolution must never be the same instance as the bestSolution.
+        scoreDirector.disableShadowVariablesInCorrectStateChecks();
         scoreDirector.setWorkingSolution(scoreDirector.cloneSolution(getBestSolution()));
+        // The best solution might have some stale shadow variables, so
+        // update all shadow variables
+        scoreDirector.forceTriggerVariableListeners();
+        scoreDirector.enableShadowVariablesInCorrectStateChecks();
     }
```
triceo: Arguably, this method is the wrong place to do this. This method is being called often - between phases, during problem changes, in partitioned solver, ... 

You want to put this code where the variables can potentially be in incorrect states, therefore when the working solution is read from the outside; not when it is copied from the best solution. Best solution is a product of the solver, not of the user.

triceo (changes requested):
I have not done a full review yet. The changes I'm requesting will alter this PR enough that I will re-review from scratch when done.

triceo:
Also, I wonder... the performance impact of this could be mitigated if we know that these checks will be done elsewhere. If we _know_ that var listeners will be triggered, and that those have all of those assertions, can we not skip this work?

Christopher-Chianelli:
> Var listeners will not be triggered, since this code is only run for `SolutionManager.update(SolutionUpdatePolicy.UPDATE_SCORE_ONLY, ...)`, which skip variable listener calls.

The code doesn't make that at all clear. There is no comment, nor does the structure of the code suggests that. If we moved this code to a place where the `SolutionManager.update()` happens, there could be no discussion. But since the correct place for this is probably in the score director for performance reasons, at least there ought to be a series of comments explaining the purpose of this.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
Christopher-Chianelli: An Entity class might have multiply entity descriptors; especially when the entity classes are interfaces, which in turn make `getEntityDescriptorStrict` return `null`/not contain all the variables. (Recall: a class implementing a PlanningEntity interface does not need to be a PlanningEntity unless it declares new geninue or shadow variables).

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+        if (scoreDirector.expectShadowVariablesInCorrectState()) {
+            var inconsistentDescriptor = entityDescriptor.getShadowVariablesInconsistentDescriptor();
+            if (inconsistentDescriptor != null) {
+                if (scoreDirector.expectShadowVariablesInCorrectState()
+                        && inconsistentDescriptor.getValue(entity) == null) {
+                    throw new IllegalStateException(
+                            """
+                                    Shadow variables update is disabled, but the entity (%s) has a null @%s annotated field (%s).
+                                    Maybe enable shadow variable updates?
+                                    """
```
triceo: I am confused. If shadow var update is disabled, shouldn't this code be completely skipped?
Also, why checking the `expectShadowVariablesInCorrectState()` twice? This smells of a bug.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+import ai.timefold.solver.core.api.score.Score;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.BasicVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.InverseRelationShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariablesInconsistent;
+
+final class EntityValidator<Solution_, Score_ extends Score<Score_>> implements Consumer<Object> {
```
triceo: There's something I probably don't understand. This PR is meant to do the following:

- If shadow var update is enabled, shadow vars are always recomputed to their correct values.
- If shadow var update is disabled, shadow var value consistency is ignored, and the user is left to suffer the consequences of their choices.

This begs a question: why does the `EntityValidator` exist then? From the two points above, it follows that one-time validation is no longer necessary.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+                            ignored -> entityDescriptor.getShadowVariableDescriptors()
+                                    .stream()
+                                    .filter(descriptor -> descriptor instanceof InverseRelationShadowVariableDescriptor)
+                                    .toArray(InverseRelationShadowVariableDescriptor[]::new));
+            for (var inverseRelationDescriptor : inverseRelationDescriptors) {
+                assertInverseRelationConsistent(scoreDirector, inverseRelationDescriptor, entity);
+            }
+        }
+
+        if (listVariableDescriptor == null) { // Only basic variables.
```
triceo: I guess this answers my earlier question - the validator exists so that it can do this. Two follow-ups then:

- Wouldn't it be cleaner, instead of checking for `scoreDirector.expectShadowVariablesInCorrectState` here, to have two different implementations of the validator, with the correct one chosen at instantiation time? One extending the other, perhaps. Or just two different validators, chained together using `andThen(...)`.
- It's still unclear why the shadow var validation exists. Those two points from the previous comment still stand.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.InverseRelationShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariablesInconsistent;
+
+final class EntityValidator<Solution_, Score_ extends Score<Score_>> implements Consumer<Object> {
+    final InnerScoreDirector<Solution_, Score_> scoreDirector;
+    final SolutionDescriptor<Solution_> solutionDescriptor;
+    final ListVariableDescriptor<Solution_> listVariableDescriptor;
+    final Map<Class<?>, InverseRelationShadowVariableDescriptor<Solution_>[]> entityClassToInverseDescriptors;
+    final Map<Object, Set<Object>> valueToInverse;
```
triceo: `valueToInverseMap`

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.InverseRelationShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.preview.api.domain.variable.declarative.ShadowVariablesInconsistent;
+
+final class EntityValidator<Solution_, Score_ extends Score<Score_>> implements Consumer<Object> {
+    final InnerScoreDirector<Solution_, Score_> scoreDirector;
+    final SolutionDescriptor<Solution_> solutionDescriptor;
+    final ListVariableDescriptor<Solution_> listVariableDescriptor;
+    final Map<Class<?>, InverseRelationShadowVariableDescriptor<Solution_>[]> entityClassToInverseDescriptors;
```
triceo: `entityClassToInverseDescriptorMap`

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
```diff
@@ -0,0 +1,183 @@
...
+                        Verify the consistency of your input solution.
+                        """.formatted(entity, InverseRelationShadowVariable.class.getSimpleName()));
+            }
+            var actualCollection = valueToInverse.getOrDefault(entity, Collections.emptySet());
+            if (!inverseCollection.containsAll(actualCollection) || !actualCollection.containsAll(inverseCollection)) {
+                throw new IllegalStateException(
+                        """
+                                The entity (%s) has a collection @%s (%s) that does not match the actual inverse set (%s).
+                                Verify the consistency of your input solution.
+                                """
```
triceo: This check is very expensive and I wonder if it's worth its weight. Maybe a simpler checking of the sizes would suffice?

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -83,6 +83,14 @@ static <Score_ extends Score<Score_>> ConstraintAnalysis<Score_> getConstraintAn
                 .toList();
     }
 
+    /**
+     * The {@link PlanningSolution working solution} must never be the same instance as the
+     * {@link PlanningSolution initial solution}, it should be a clone with all shadow variables updated.
+     *
+     * @param initialSolution never null
+     */
+    void setInitialSolution(Solution_ initialSolution);
```
triceo: So, there are now two ways to set the working solution?

- This.
- `setWorkingSolution()`

IMO, this creates a situation that will cause subtle bugs, we don't want to have to choose between these. We should find a different way of figuring out the shadow var update - one that doesn't affect the solution's entry point.

Idea:
- `setWorkingSolution()` maintains its current contract.
- External code can decide to build the score director in such a way that `setWorkingSolution(...)` either will or will not perform the shadow var update.
- _Internal code_ will get a different method `setWorkingSolutionWithoutShadowVarUpdate(...)` that it will call instead of `setWorkingSolution(...)` when we know the shadow vars are correct (= we control them).
- This new method would be `protected` inside `AbstractScoreDirector`, therefore never exposed to external users. That way, this complexity can stay hidden inside of the score director, where it belongs. The external users still only ever see `setWorkingSolution()`.

Another option, perhaps simpler: `setWorkingSolution(...)` gets an optional overload with a boolean argument, allowing the user to choose if they want the shadow var update or not. All solver uses of `setWorkingSolution(...)` then need to be evaluated to make sure they are making the correct choice.

core/src/main/java/ai/timefold/solver/core/preview/api/domain/variable/declarative/ShadowVariablesInconsistent.java
```diff
@@ -30,18 +30,18 @@
...
  * Important:
- * Do not use a {@link ShadowVariableLooped} property in a method annotated
- * with {@link ShadowSources}. {@link ShadowVariableLooped} properties can
+ * Do not use a {@link ShadowVariablesInconsistent} property in a method annotated
+ * with {@link ShadowSources}. {@link ShadowVariablesInconsistent} properties can
  * be updated after the {@link ShadowSources} marked method is called, causing
  * score corruption. {@link ShadowSources} marked methods do not need to check
- * {@link ShadowVariableLooped} properties, since they are only called if all
+ * {@link ShadowVariablesInconsistent} properties, since they are only called if all
  * their dependencies are not looped.
```
triceo: I suggest doing a check on the word "loop", as clearly some of them slipped through the cracks.

Also, instead of saying "if all of their dependencies are not looped", the human way of speaking would be "if all of their dependencies are consistent."
"Inconsistent" has a good inverse, whereas "looped" did not - we can use that inverse.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -396,16 +396,16 @@ private static ParentVariableType determineParentVariableType(Class<?> rootClass
             return ParentVariableType.VARIABLE;
         }
 
-        if (getAnnotation(declaringClass, memberName, ShadowVariableLooped.class) != null) {
+        if (getAnnotation(declaringClass, memberName, ShadowVariablesInconsistent.class) != null) {
             throw new IllegalArgumentException("""
                     The source path (%s) starting from root class (%s) accesses a @%s property (%s).
-                    Supplier methods are only called when none of their dependencies are looped,
+                    Supplier methods are only called when none of their dependencies are inconsistent,
```
triceo: See the other comment on this - "none are inconsistent" = "all are consistent."

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/LoopedTracker.java
```diff
@@ -34,23 +35,24 @@ public void mark(int node, LoopedStatus status) {
         looped.set(node, status.ordinal());
     }
 
-    public boolean isEntityLooped(BaseTopologicalOrderGraph graph, int entityId, boolean wasEntityLooped) {
+    public boolean isEntityInconsistent(BaseTopologicalOrderGraph graph, int entityId,
+            @Nullable Boolean wasEntityInconsistent) {
         for (var entityNode : entityIdToNodes[entityId]) {
             if (graph.isLooped(this, entityNode)) {
```
triceo, line 41: Question: do we want to keep `loop` in the graph, but use consistency everywhere else? This proliferates elsewhere in the code - `LoopedTracker`, `looped.set(...)`, ... 

I understand that, within the context of the graph, "loop" is more descriptive than "inconsistent". Still, somehow, I don't like mixing the two. Dunno.

triceo (commented):
Leaving my review.

I also pushed a small commit adding Javadocs to the methods that have to do with update policies. I figured it'd be more effective to do it than to spend even more time explaining it so that you can do it.

triceo:
Also, tests are failing.

triceo:
Having reviewed the Sonar suggestions, I'd say they're useful.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
Christopher-Chianelli: `expectShadowVariablesInCorrectState` is set to true if shadow variable updates are disabled; i.e. when `SolutionManager.update(SoultionUpdatePolicy.UPDATE_SCORE_ONLY, ...)`. It is also set to true after the initial solution is set.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
Christopher-Chianelli: This is old validation code from the ScoreDirectorFactory that was merged with the new validation code.

core/src/main/java/ai/timefold/solver/core/impl/score/director/EntityValidator.java
Christopher-Chianelli: `entityClassToInverseDescriptorsMap`; When multiple basic planning variables are involved, there can be multiple inverse descriptors.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
Christopher-Chianelli: I prefer having a method with a different name than a method with overloads/that uses a parameter to change behaviour. Would `setWorkingSolutionAndUpdateShadows` suffice? Also, isn't `InnerScoreDirector`  an implementation interface? Users should see only `ScoreDirector`.

triceo: By "users" in this case I mean call sites in the solver itself - the solver is growing and we ought to keep complexity in check even for the sake of the three of us. You are correct that "users" as in "human solver users" should never see either of these methods.

If you want to do it as another method, so be it - but I am really wary of creating multiple entry points to the functionality. Multiple entry points complicate things.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/LoopedTracker.java
Christopher-Chianelli: I prefer looped with graphs, since it an implementation detail, and since "inconsistent" is vague when talking about a graph.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
triceo: Wrt. `setWorkingSolutionAndUpdateShadows` - I'll argue for the inverse. The default (`setWorkingSolution`) should update the shadows, because that is the expected behavior in vast majority of use cases. `setWorkingSolutionWithoutUpdatingShadows` is an edge case, and therefore it should use an alternative method, not the main method.

Please make the name change, and review every call to setWorkingSolution...() if it does the right thing; in other words - if we expect shadows not to be called, the right method should be called, preferrably without relying on some global variable.

triceo:
I have just added my final comment. Once that's resolved, I'll approve, provided that those changes don't result in anything questionable.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
```diff
@@ -87,7 +87,7 @@ public void updateShadowVariables(Solution_ solution) {
                 entityClassSet.toArray(new Class<?>[0]));
         try (var scoreDirector = new InternalScoreDirector.Builder<>(solutionDescriptor).build()) {
             // When we have a solution, we can reuse the logic from VariableListenerSupport to update all variable types
-            scoreDirector.setWorkingSolution(solution);
+            scoreDirector.setWorkingSolutionWithoutUpdatingShadows(solution);
             scoreDirector.forceTriggerVariableListeners();
```
triceo: Why do we call "without updating", when we update immediately after?

core/src/test/java/ai/timefold/solver/core/impl/score/director/incremental/IncrementalScoreDirectorTest.java
```diff
@@ -73,6 +73,11 @@ void variableListener() {
             inOrder.verify(incrementalScoreCalculator, times(1)).afterVariableChanged(a3, "chainedObject");
             inOrder.verify(incrementalScoreCalculator, times(1)).beforeVariableChanged(b1, "nextEntity");
             inOrder.verify(incrementalScoreCalculator, times(1)).afterVariableChanged(b1, "nextEntity");
+            // Note: Anchor was not updated with scoreDirector.setWorkingSolutionWithoutUpdatingShadows, despite the fact a change is expected
+            // In scoreDirector.setWorkingSolutionWithoutUpdatingShadows, a3.anchor is null and should be changed to b0 (but no event is triggered; a3.anchor remained null!).
+            // in scoreDirector.setWorkingSolution, a3.anchor is a0 and is changed to b0.
```
triceo: This comment is unnecessary; it mentions what used to happen in the test, but no longer happens. I appreciate the clarification, but it should not survive the merge.

If you want to keep the part explaining the expected changes, that is fair game.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -332,7 +332,16 @@ public long getMoveEvaluationSpeed() {
...
-        scoreDirector.setWorkingSolution(scoreDirector.cloneSolution(getBestSolution()));
+        // The bestSolution is set by the Solver, and thus have updated shadows
+        scoreDirector.setWorkingSolutionWithoutUpdatingShadows(scoreDirector.cloneSolution(getBestSolution()));
+    }
+
+    public void setInitialSolution(Solution_ initialSolution) {
+        // The workingSolution must never be the same instance as the bestSolution.
+        scoreDirector.setWorkingSolution(scoreDirector.cloneSolution(initialSolution));
+
+        // Set the best solution to the solution with shadow variable updated!
```
triceo:
```suggestion
        // Set the best solution to the solution with shadow variable updated.
```

Let's not shout at ourselves.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -85,12 +85,20 @@ static <Score_ extends Score<Score_>> ConstraintAnalysis<Score_> getConstraintAn
 
     /**
      * The {@link PlanningSolution working solution} must never be the same instance as the
-     * {@link PlanningSolution best solution}, it should be a (un)changed clone.
+     * {@link PlanningSolution initial solution}, it should be a clone with all shadow variables updated.
```
triceo: Should it have all shadow vars updated?
This method _will udpate shadow vars_.

core/src/test/java/ai/timefold/solver/core/impl/score/director/incremental/IncrementalScoreDirectorTest.java
Christopher-Chianelli: I think the old behaviour is worth investigating, since the `before/after` variable change should of been triggered for anchor with the old behaviour, but it never was and the anchor remained null.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
Christopher-Chianelli: It was worded oddly to try to match the same style as the original, which state an expected condition between working solution and the best solution.

core/src/test/java/ai/timefold/solver/core/impl/score/director/incremental/IncrementalScoreDirectorTest.java
triceo: I agree. Feel free to look into it.

triceo (approved):

PR #1750, opened 2025-08-20
chore: simplify reachable values logic

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,72 +20,108 @@
...
 public final class ReachableValues {
 
-    private final Map<Object, Set<Object>> valueToEntityMap;
-    private final Map<Object, Set<Object>> valueToValueMap;
-    private final Map<Object, List<Object>> randomAccessValueToEntityMap;
-    private final Map<Object, List<Object>> randomAccessValueToValueMap;
+    private final Map<Object, ReachableItemValue> values;
     private final @Nullable Class<?> valueClass;
+    private @Nullable Object chachedObject;
+    private @Nullable ReachableItemValue cachedItemValue;
```
triceo: If `ReachableItemValue` includes the `value` itself, you can avoid having to specifically store the `cachedObject`.
Arguably, it would make sense to add the value, because otherwise the type feels "incomplete" - it contains a bunch of values, but not the key that ties them together.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,72 +20,108 @@
...
-            } else {
-                result = Collections.emptyList();
-            }
-            randomAccessValueToEntityMap.put(value, result);
+        var itemValue = fetchItemValue(value, true);
+        if (itemValue == null) {
+            return Collections.emptyList();
+        }
+        if (itemValue.randomAccessEntityList == null) {
+            itemValue.randomAccessEntityList = new ArrayList<>(itemValue.entitySet);
```
triceo: Do we need to modify the list?
Is `null` ever part of the list?

If the answer to both is "no", then `List.copyOf(...)` will do a better job than creating a new `ArrayList`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,72 +20,108 @@
...
-                result = Collections.emptyList();
-            }
-            randomAccessValueToValueMap.put(value, result);
+        var itemValue = fetchItemValue(value, true);
+        if (itemValue == null) {
+            return Collections.emptyList();
         }
-        return result;
+        if (itemValue.randomAccessValueList == null) {
+            itemValue.randomAccessValueList = new ArrayList<>(itemValue.valueSet);
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,72 +20,108 @@
...
+            if (destinationItemValue != null) {
+                // Check if the destination accepts the origin entity
+                return originEntity == null || destinationItemValue.entitySet.contains(originEntity);
+            }
+        }
+        return false;
+    }
+
+    @NullMarked
+    public static class ReachableItemValue {
```
triceo: final

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,72 +20,108 @@
...
+    public boolean isEntityReachable(Object origin, @Nullable Object originEntity, Object destination,
+            @Nullable Object destinationEntity) {
+        var originItemValue = fetchItemValue(origin, true);
+        if (originItemValue == null) {
+            return false;
+        }
+        // Check if the origin accepts the destination entity
+        var isOriginValid = destinationEntity == null || originItemValue.entitySet.contains(destinationEntity);
+        if (isOriginValid) {
+            var destinationItemValue = fetchItemValue(destination, false);
```
triceo: Why not delegate to `isEntityReachable(destination, originEntity)` instead? Seems like this is code duplication.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -144,18 +143,18 @@ private Object selectReplayedValue() {
     @Override
     public Iterator<Object> iterator() {
         if (randomSelection) {
-            return new RandomFilteringValueRangeIterator(this::selectReplayedValue, reachableValues, listVariableStateSupply,
-                    workingRandom, (int) getSize(), checkSourceAndDestination);
+            return new RandomFilteringValueRangeIterator(this::selectReplayedValue, reachableValues,
+                    listVariableStateSupply, workingRandom, (int) getSize(), checkSourceAndDestination);
```
triceo: I don't understand... We're dealing with basic variables, yet for some reason, we are adding _list variable_ supply?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -270,21 +235,26 @@ private class OriginalFilteringValueRangeIterator extends AbstractFilteringValue
 
         private OriginalFilteringValueRangeIterator(Supplier<Object> upcomingValueSupplier, ReachableValues reachableValues,
                 ListVariableStateSupply<Solution_> listVariableStateSupply, boolean checkSourceAndDestination) {
-            super(reachableValues, listVariableStateSupply, checkSourceAndDestination, false);
+            super(reachableValues, listVariableStateSupply, checkSourceAndDestination);
             this.upcomingValueSupplier = upcomingValueSupplier;
         }
 
+        @Override
+        void loadValues() {
```
triceo: I honestly did not notice the introduction of this method in the previous commits. I'm not too excited about this contract - it's got too many side effects.

I'd prefer a method that actually returns something, and then the rest of the code can do something with that return value; as opposed to storing some opaque state somewhere. State leads to spaghetti code.

Also, considering how important it seems to be, it is entirely undocumented.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -435,48 +435,40 @@ public ReachableValues getReachableValues(ListVariableDescriptor<Solution_> list
...
                 var valuesIterator = allValues.createOriginalIterator();
                 var range = getFromEntity(listVariableDescriptor.getValueRangeDescriptor(), entity);
                 while (valuesIterator.hasNext()) {
                     var value = valuesIterator.next();
                     if (range.contains(value)) {
-                        updateEntityMap(entityMatrix, entity, value, entityList.size());
-                        updateValueMap(valueMatrix, range, value, (int) valuesSize);
+                        var item = initReachableMap(reachableValuesMap, value, entityList.size(), (int) valuesSize);
+                        updateMap(item, entity, value, range);
                     }
```
triceo: Not sure why these two steps are separate.
They are one logical operation and separation seems to just confuse the logic.

triceo (commented):
I have to say I don't understand how this does anything for basic variables.

On Slack, you said:

> The concern separation property allowed me to create the reachable values for basic variables without any significant updates.

Is this that PR? If so, where exactly does it deal with basic variables?
And if so, why does the enterprise counterpart also only deal with list variables?

I am confused.

zepfred:
> I have to say I don't understand how this does anything for basic variables.
> 
> On Slack, you said:
> 
> > The concern separation property allowed me to create the reachable values for basic variables without any significant updates.
> 
> Is this that PR? If so, where exactly does it deal with basic variables? And if so, why does the enterprise counterpart also only deal with list variables?
> 
> I am confused.

This PR is still incomplete and only simplifies the ReachableValues class. Regarding the concern separation, I only needed to change the parameter type from `ValueRangeManager::getReachableValues` to `GenuineVariableDescriptor` to make it work for basic variables.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
zepfred: The list will now be populated during the creation of reachable values without any additional overhead.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
zepfred: The basic variable isn't using this decorator, and I'm unsure if it will. The method for generating basic variables differs and might require a different approach to generate only valid moves. This is still in progress.

triceo:
> This is still in progress.

The PR is now a draft, and that's fair. If it had been a draft from the beginning, I wouldn't have asked some of these questions.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListSwapMoveSelectorTest.java
```diff
@@ -333,9 +335,9 @@ void randomWithEntityValueRange() {
 
         assertCodesOfNeverEndingMoveSelector(moveSelector,
                 "2 {A[0]} <-> 3 {B[0]}",
-                "1 {A[1]} <-> 2 {A[0]}",
-                "3 {B[0]} <-> 2 {A[0]}",
-                "1 {A[1]} <-> 2 {A[0]}",
+                "No change",
+                "No change",
+                "No change",
```
triceo: This is not ideal. 

We spent so much time making sure filtering is not necessary, only to end up in a place where we generate useless moves anyway? Sounds like we wasted time.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+ * <p>
+ * e1 = [e2]
+ * e2 = [e1, e3]
+ * e3 = [e2]
+ * <p>
+ * 1. e1(null) and e2(null) - they are reachable
+ * <p>
+ * 2. e1(v2) and e2(v1) - they are not reachable
+ * e1: e1 accepts v1 and e2 is reachable
+ * e2: e2 does not accept v2 and e1 is not reachable
```
triceo: Whitespace will be ignored in the rendered Javadoc. Do not rely on it for formatting.
Also, if you need an ordered list (1, 2, ...), that's what `<ol>` is for.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+ * The decorator returns a list of reachable entities for a specific entity.
+ * It enables the creation of a filtering tier when using entity selectors,
+ * ensuring only valid and reachable entities are returned.
+ * <p>
+ * e1 = entity_range[v1, v2, v3]
+ * e2 = entity_range[v1, v4]
+ * e3 = entity_range[v4, v5]
+ * <p>
+ * By default, one entity can be reached by another if assigned values are null.
+ * When values are assigned, both sides must accept these values.
```
triceo: It is not clear what this "reachable" means. Who tries to reach it? What for? Does it rely on some other solver concept?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+        this.replayedEntity = null;
+        this.valueRangeManager = null;
+        this.basicVariableDescriptors = null;
+    }
+
+    // ************************************************************************
+    // Worker methods
+    // ************************************************************************
+
+    public EntitySelector<Solution_> getChildEntitySelector() {
```
triceo: I don't particularly like that we're exposing this.
Accessing child selectors from outside of the parent selector breaks encapsulation and has the potential to cause bugs.
We should not enable people to do that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+
+    /**
+     * The expected replayed entity corresponds to the selected entity when the replaying selector has the next value.
+     * Once it is selected, it will be reused until a new entity is replayed by the recorder selector.
+     */
+    private Object selectReplayedEntity() {
+        var iterator = replayingEntitySelector.iterator();
+        if (iterator.hasNext()) {
+            var entity = iterator.next();
+            if (replayedEntity == null || entity != replayedEntity) {
```
triceo: I don't see the point of this condition.
Isn't it equivalent to just always doing `this.replacedEntity = entity`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+        return new OriginalFilteringValueRangeIterator<>(this::selectReplayedEntity, childEntitySelector.listIterator(),
+                basicVariableDescriptors, valueRangeManager);
+    }
+
+    @Override
+    public Iterator<Object> iterator() {
+        if (randomSelection) {
+            if (!childEntitySelector.isNeverEnding()) {
+                throw new IllegalArgumentException(
+                        "Impossible state: childEntitySelector must provide a never ending approach.");
```
triceo: never-ending

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+        boolean isReachable(Object otherEntity) {
+            if (replayedEntity == otherEntity) {
+                // Same entity cannot be swapped
+                return false;
+            }
+
+            if (basicVariableDescriptors.length == 1) {
+                return isReachable(replayedEntity, otherEntity, basicVariableDescriptors[0], valueRangeManager);
+            } else {
+                for (BasicVariableDescriptor<Solution_> basicVariableDescriptor : basicVariableDescriptors) {
```
triceo: `var`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
```diff
@@ -22,10 +22,19 @@ public class SwapMove<Solution_> extends AbstractMove<Solution_> {
...
     protected final Object rightEntity;
 
+    private final boolean checkValueRange;
+
     public SwapMove(List<GenuineVariableDescriptor<Solution_>> variableDescriptorList, Object leftEntity, Object rightEntity) {
+        // We don't enable the value range validation by default
+        this(variableDescriptorList, leftEntity, rightEntity, false);
+    }
+
+    public SwapMove(List<GenuineVariableDescriptor<Solution_>> variableDescriptorList, Object leftEntity, Object rightEntity,
```
triceo: Let's make this one `protected`. We're not exposing this detail.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
```diff
@@ -22,10 +22,19 @@ public class SwapMove<Solution_> extends AbstractMove<Solution_> {
     protected final Object leftEntity;
     protected final Object rightEntity;
 
+    private final boolean checkValueRange;
+
     public SwapMove(List<GenuineVariableDescriptor<Solution_>> variableDescriptorList, Object leftEntity, Object rightEntity) {
+        // We don't enable the value range validation by default
```
triceo: We can see that. The comment would be more useful if it said why not.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
```diff
@@ -606,11 +606,11 @@ void randomAllowsUnassignedValuesWithEntityValueRange() {
         phaseStarted(solverScope, moveSelector);
 
         assertCodesOfNeverEndingMoveSelector(moveSelector,
-                "1 {A[1]->A[0]}",
+                "1 {A[1]->A[1]}",
```
triceo: Why are we generating moves that don't do anything?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+            return pickSelected();
+        }
+
+        @Override
+        public int nextIndex() {
+            return entityIterator.nextIndex();
+        }
+
+        @Override
+        public int previousIndex() {
```
triceo: Do we have any actual use for previous(), previousIndex() etc.? Where is it used?
IMO this should not be implemented, we will never use these iterators to go back in time.

Arguably, all methods of these iterators which we do not need should just throw `UnsupportedOperationException`. The selectors are complex enough, even without implementing methods we do not need.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/AbstractOriginalSwapIterator.java
```diff
@@ -40,17 +40,14 @@ public AbstractOriginalSwapIterator(ListIterable<SubSelection_> leftSubSelector,
 
     @Override
     protected Move_ createUpcomingSelection() {
-        if (!rightSubSelectionIterator.hasNext()) {
+        while (!rightSubSelectionIterator.hasNext()) {
```
triceo, line 43: Why is this change only being made to the swap iterator?
Do no other moves suffer from this problem?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+            if (initialized) {
+                return;
+            }
+            checkReplayedEntity();
+            initialized = true;
+        }
+
+        void checkReplayedEntity() {
+            var updatedReplayedEntity = upcomingEntitySupplier.get();
+            if (replayedEntity == null || replayedEntity != updatedReplayedEntity) {
```
triceo: Same point on the usefulness of the condition.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+        /**
+         * The other entity is reachable if it accepts all assigned values from the replayed entity, and vice versa.
+         */
+        boolean isReachable(Object otherEntity) {
+            if (replayedEntity == otherEntity) {
+                // Same entity cannot be swapped
+                return false;
+            }
+
+            if (basicVariableDescriptors.length == 1) {
```
triceo: Is this a measurable performance improvement?
If not, I'd rather use the loop always, to simplify code.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+        @Override
+        public boolean hasNext() {
+            initialize();
+            return entityIterator.hasNext();
+        }
+
+        @Override
+        public Object next() {
+            initialize();
+            if (!entityIterator.hasNext()) {
```
triceo: This can be `if (!hasNext())` instead.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,411 @@
...
+                // If no reachable entity is found, we return the currently selected entity,
+                // which will result in a non-doable move
+                return currentReplayedEntity();
```
triceo: Not too excited here. Typically in this situation, we return `noUpcomingSelection()`, not an invalid value.

triceo (commented):
I am a bit worried about the continued development of the selector framework. What we had before was inefficient and didn't do many of the things that this now does, but at least I understood how it worked - I can no longer confidently say that.

We should invest time into documenting how the selectors work now. The interactions between the individual components, the parent-child relationships, the specific implementations for value ranges etc.

Also, before we merge this, let's confirm that FSR and MJS behave correctly, as that is where we've seen issues in the past.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: I did not assess the improvement of this and remove it for now.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: I understand your point, but that's the price we must pay to avoid executing the lookup operations during the `hasNext()` call. This particular change has enhanced performance.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: This method is no longer needed, so I have removed it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
zepfred: I realized that both basic and chained selectors will only generate valid moves, so this flag is no longer needed.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
zepfred: This concerns the problem of random iterators failing to find a reachable value or entity within the bailout size. When this occurs, we return the value on the left side, as it results in a non-doable move.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: The only places using `ListIterator::previous` are `VariableChangeRecordingScoreDirector::undoChanges` and `MimicRecordingEntitySelector`. The first is not directly related to selectors, and the second seems unnecessary.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/AbstractOriginalSwapIterator.java
zepfred: This issue only affects the original iterators. The change move selector already uses a while loop. However, it seems to also impact `OriginalPillarChangeMoveIterator`. I'll add a test to verify this.

zepfred: `OriginalPillarChangeMoveIterator` has no issues, but I have included a test case.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,414 @@
...
+ * <p>
+ * The decorator can only be applied to basic variables.
+ * <code>
+ * <p>
+ * e1 = entity_range[v1, v2, v3]
+ * <p>
+ * e2 = entity_range[v1, v4]
+ * <p>
+ * e3 = entity_range[v1, v4, v5]
+ * <p>
```
triceo: Pre-formatted blocks must not contain HTML formatting.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,414 @@
...
+            return result;
+        }
+
+        @Override
+        public Object next() {
+            return pickSelected();
+        }
+
+        @Override
+        public boolean hasPrevious() {
```
triceo: So we're not removing these implementations?
My earlier commit is marked as resolved, but I see no change.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
```diff
@@ -4,35 +4,64 @@
...
- * It enables the creation of a filtering tier when using entity value selectors,
+ * It enables the creation of a filtering tier when using entity-provided value ranges,
  * ensuring only valid and reachable entities are returned.
- *
+ * An entity is considered reachable to a value if its value range includes that value.
+ * <p>
+ * The decorator can only be applied to list variables.
+ * <p>
+ * <code>
+ * <p>
```
triceo: Same comment on pre-formatted code.
IntelliJ has a feature to show you right in the text editor how the Javadoc would render - I recommend checking it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
triceo: In fact... why is this the only place where I'm seeing this?
All other places seem to be using a plain old `Iterator`, whereas this is `ListIterator` for some reason.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/chained/ChainedSwapMove.java
```diff
@@ -22,6 +22,7 @@ public class ChainedSwapMove<Solution_> extends SwapMove<Solution_> {
 
     public ChainedSwapMove(List<GenuineVariableDescriptor<Solution_>> variableDescriptorList,
             List<SingletonInverseVariableSupply> inverseVariableSupplyList, Object leftEntity, Object rightEntity) {
+        // We enable the value-range validation for chained moves
```
triceo: These comments don't seem necessary anymore.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
```diff
@@ -42,27 +43,17 @@ public Object getRightEntity() {
...
-        var movable = false;
+        if (leftEntity == rightEntity) {
+            return false;
+        }
         for (var variableDescriptor : variableDescriptorList) {
             var leftValue = variableDescriptor.getValue(leftEntity);
             var rightValue = variableDescriptor.getValue(rightEntity);
             if (!Objects.equals(leftValue, rightValue)) {
-                movable = true;
-                if (!variableDescriptor.canExtractValueRangeFromSolution()) {
```
triceo, line 51: I'm not sure how I feel about this getting removed.
It seems useful - people can produce these moves from their own selectors.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -26,15 +31,42 @@
...
  * It enables the creation of a filtering tier when using entity-provided value ranges,
  * ensuring only valid and reachable values are returned.
+ * A value is considered reachable to another value if both exist within their respective entity value ranges.
+ * <p>
+ * The decorator can only be applied to list variables.
+ * <p>
+ * <code>
  * <p>
  * e1 = entity_range[v1, v2, v3]
+ * <p>
```
triceo: See above.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -103,27 +105,28 @@ public boolean matchesValueClass(Object value) {
...
-        private final Set<Object> valueSet;
+        private final Map<Object, Object> entitySet;
+        private final Map<Object, Object> valueSet;
         private final List<Object> randomAccessEntityList;
         private final List<Object> randomAccessValueList;
 
         public ReachableItemValue(Object value, int entityListSize, int valueListSize) {
             this.value = value;
-            this.entitySet = new LinkedHashSet<>(entityListSize);
+            this.entitySet = new IdentityHashMap<>(entityListSize);
```
triceo: Collections.newSetFromMap() solves the hacks in this code.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -436,26 +438,41 @@ public ReachableValues getReachableValues(GenuineVariableDescriptor<Solution_> v
                                         entityDescriptor.getEntityClass().getSimpleName(),
                                         variableDescriptor.getVariableName(), valuesSize));
             }
-            var reachableValuesMap = new IdentityHashMap<Object, ReachableItemValue>((int) valuesSize);
+            Class<?> valueClass = null;
```
triceo: This whole nested logic is incredibly long. Invert it - return `values` if not null, then deal with the other case without nesting.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -436,26 +438,41 @@ public ReachableValues getReachableValues(GenuineVariableDescriptor<Solution_> v
...
+                if (value == null) {
+                    continue;
+                }
+                valueClass = value.getClass();
+                break;
+            }
+            Map<Object, ReachableItemValue> reachableValuesMap = ConfigUtils.isGenericTypeImmutable(valueClass)
+                    ? new HashMap<>((int) valuesSize)
+                    : new IdentityHashMap<>((int) valuesSize);
             for (var entity : entityList) {
```
triceo: This is a triple-nested loop. The scalability of this for large datasets will cause problems. Long-term, this needs a fix.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: I have removed the use of `ListIterator`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
zepfred: I'm using it, and the use of `<p>` inside code blocks is rendering as expected.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMove.java
zepfred: I understand your point. However, I also removed the value range check from the other moves because this work is focused on generating valid moves. You're right about custom move generators, but re-adding the checks will introduce some overhead that I want to measure.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
zepfred: That would require maintaining a separate Map while loading the data. I don't see any benefit in creating a Set from it, and also keeping a Map for data loading. The current logic only uses one Map.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -420,45 +426,63 @@ public long countOnEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor,
...
-                                        entityDescriptor.getEntityClass().getSimpleName(),
-                                        variableDescriptor.getVariableName(), valuesSize));
+        if (values != null) {
+            return values;
+        }
+        if (cachedWorkingSolution == null) {
+            throw new IllegalStateException(
+                    "Impossible state: value reachability requested before the working solution is known.");
+        }
+        logger.warn("Creating reachable values structure for {}", variableDescriptor);
```
triceo: Please remove.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -420,45 +426,63 @@ public long countOnEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor,
...
                 }
             }
-            values = new ReachableValues(reachableValuesMap,
-                    variableDescriptor.getValueRangeDescriptor().acceptsNullInValueRange());
-            reachableValues[variableDescriptor.getValueRangeDescriptor().getOrdinal()] = values;
         }
+        values = new ReachableValues(reachableValuesMap, valueClass,
+                variableDescriptor.getValueRangeDescriptor().acceptsNullInValueRange());
+        reachableValues[variableDescriptor.getValueRangeDescriptor().getOrdinal()] = values;
+        logger.warn("Finished creating reachable values structure for {}", variableDescriptor);
```
triceo: Same here.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -0,0 +1,376 @@
...
+                    basicVariableDescriptors, valueRangeManager, 10);
+        } else {
+            return new OriginalFilteringValueRangeIterator<>(this::selectReplayedEntity, childEntitySelector.iterator(),
+                    basicVariableDescriptors, valueRangeManager);
+        }
+    }
+
+    @Override
+    public ListIterator<Object> listIterator() {
+        throw new UnsupportedOperationException();
```
triceo, line 196: What is this being used for?
If we can afford not to implement it, why is it even there?

triceo (approved):
LGTM after comments resolved and Javadoc fixed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
zepfred: I initially included it in the node logic because I also added tests for swap moves using original iterators. The class `AbstractOriginalSwapIterator` requires a list iterator. 

Its purpose is to return an iterator with elements in the proper order and to allow moving forward and backward. However, I could find only a few nodes using it: `AbstractOriginalSwapIterator`, `FilteringEntitySelector`, and `MimicRecordingEntitySelector`.

My understanding of that is if we need to analyze the elements in the original order, we don't create a node with the random selection flag enabled. Also, I couldn't find any use of the `previous()` call. So, it is also not clear to me why we include the `ListIterator` in our contracts.

triceo: Yeah, we won't spend time on this now, but in the future, I think the list iterator can just be replaced by a plain iterator.

PR #1764, opened 2025-08-28
feat: forEach and forEachUnassigned exclude inconsistent entities by default; forEachUnfiltered to select all entities including inconsistent entities

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -65,6 +67,14 @@ public interface ConstraintFactory {
      */
     <A> @NonNull UniConstraintStream<A> forEach(@NonNull Class<A> sourceClass);
 
+    /**
+     * As defined by {@link #forEach(Class)},
+     * but including filtered entities that match the supplied {@link ForEachInclude includes}.
+     *
+     * @param <A> the type of the matched problem fact or {@link PlanningEntity planning entity}
+     */
+    <A> @NonNull UniConstraintStream<A> forEachIncluding(@NonNull Class<A> sourceClass, ForEachInclude... includes);
```
triceo: So, the code would look like this:

    forEachIncluding(Visit.class, UNASSIGNED, INCONSISTENT);

I have to say I still prefer the following:

    forEach(Visit.class, INCLUDE_UNASSIGNED, INCLUDE_INCONSISTENT);

This way, it's crystal clear that the includes are "AND" (because they both have the "INCLUDE_" prefix). The "Including" suffix in the name of the method is IMO not sufficient for that, because there is no obvious connection between the name of the method and the enum values.

Yes, the user could read the Javadoc, and arguably they should.
But arguably, code which can be understood even without reading any documentation is better than code which does not.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -65,6 +67,14 @@ public interface ConstraintFactory {
      */
     <A> @NonNull UniConstraintStream<A> forEach(@NonNull Class<A> sourceClass);
 
+    /**
+     * As defined by {@link #forEach(Class)},
+     * but including filtered entities that match the supplied {@link ForEachInclude includes}.
```
triceo: Should mention that the enum values imply an "AND" relationship.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ForEachInclude.java
```diff
@@ -0,0 +1,13 @@
+package ai.timefold.solver.core.api.score.stream;
+
+/**
+ * In constraint streams, {@link ConstraintFactory#forEach(Class)}
+ * will filter inconsistent and unassigned entities by default.
+ * In order to select the filtered entities,
```
triceo:
```suggestion
 * In order to include these entities,
```

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -913,6 +901,74 @@ public boolean hasNoNullVariables(Object entity) {
         };
     }
 
+    private Predicate<Object> forEachIncludingPredicateBasicVarSupplier(EnumSet<ForEachInclude> includeSet) {
+        // TODO: Use a supply for this?
+        final var inconsistentDescriptor = this.getShadowVariablesInconsistentDescriptor();
+        final Predicate<Object> isUnassignedPredicate = this::hasNoNullVariables;
+        final Predicate<Object> isInconsistentPredicate = (inconsistentDescriptor == null) ? ignored -> true
+                : entity -> Boolean.FALSE.equals(inconsistentDescriptor.getValue(entity));
```
triceo: Whether or not the entity is consistent is a property of the entity, and it exists regardless of any external annotation.

With these changes to `forEach`, using this annotation became pretty much unnecessary. (Unless the user actually needs to export inconsistent solutions, and therefore needs to have the information in their model.) Therefore this operation needs to work correctly even without the annotation being declared.

Arguably the answer to the question on line 905 is "yes". The docs also need to be updated to reflect the new role of the annotation. (This probably also touches the CS docs, which would now need backlinks to the chapter about DSV.)

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -913,6 +901,74 @@ public boolean hasNoNullVariables(Object entity) {
...
+
+        if (includeSet.isEmpty()) {
+            return entity -> isUnassignedPredicate.test(entity) && isInconsistentPredicate.test(entity);
+        }
+
+        var excludeToPredicate = Map.of(
```
triceo: `EnumMap`? Doesn't matter much in this case though.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -913,6 +901,74 @@ public boolean hasNoNullVariables(Object entity) {
...
+        }
+
+        // This predicate is only used when includeSet does not contain ForEachInclude.UNASSIGNED
+        // so applicableShadowDescriptor cannot be null.
+        final Predicate<Object> isUnassignedPredicate = getForEachFilterPredicateIncludingBasicVar(EnumSet.complementOf(
+                EnumSet.of(ForEachInclude.UNASSIGNED)).toArray(ForEachInclude[]::new))
+                .and(entity -> applicableShadowDescriptor.getValue(entity) != null);
+        final Predicate<Object> isInconsistentPredicate = getForEachFilterPredicateIncludingBasicVar(
+                EnumSet.complementOf(
+                        EnumSet.of(ForEachInclude.INCONSISTENT)).toArray(ForEachInclude[]::new));
```
triceo: We don't use final local variables, unless lambdas force us to, which I don't think is the case here.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -913,6 +901,74 @@ public boolean hasNoNullVariables(Object entity) {
...
+        if (excludeSet.isEmpty()) {
+            return entity -> true;
+        }
+
+        return excludeSet.stream()
+                .map(exclude -> {
+                    var predicate = excludeToPredicateMap.get(exclude);
+                    if (predicate == null) {
+                        throw new IllegalStateException(
+                                "Missing predicate for %s %s.".formatted(ForEachInclude.class.getSimpleName(), exclude));
```
triceo: "Impossible state"?

core/src/test/java/ai/timefold/solver/core/testconstraint/TestConstraintFactory.java
```diff
@@ -33,6 +34,11 @@ public SolutionDescriptor<Solution_> getSolutionDescriptor() {
...
     }
 
+    @Override
+    public <A> @NonNull UniConstraintStream<A> forEachIncluding(@NonNull Class<A> sourceClass, ForEachInclude... includes) {
+        throw new UnsupportedOperationException();
+    }
+
     @Override
     public <A> @NonNull UniConstraintStream<A> forEachIncludingUnassigned(@NonNull Class<A> sourceClass) {
         throw new UnsupportedOperationException();
```
triceo: Arguably the override is unnecessary, if we give the original method a `default` implementation. IMO it's cleaner, because it doesn't force anyone to override deprecated methods.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ForEachInclude.java
```diff
@@ -0,0 +1,13 @@
...
+ * {@link ConstraintFactory#forEachIncluding(Class, ForEachInclude...)} must
+ * be used instead.
+ */
+public enum ForEachInclude {
```
triceo: IMO this should describe the use case - `ForEachFilteringCriteria`?

triceo (commented):
Good start! Left a bunch of comments.

Besides what's inline, we also need migration support for this; we should be able to write an OpenRewrite recipe to migrate away from `forEachIncludingUnassigned`. Check `migration` module - if memory serves, there's already recipes around `forEach` from which you can take inspiration.

Christopher-Chianelli:
@triceo I been thinking, maybe we should add another `forEach` method; `forEachOnly`. Unlike `forEachIncluding`, it would not include assigned/consistent elements in the stream, and the enum values imply an OR relationship. This allow user to change their code from:

```java
constraintFactory.forEachIncluding(Visit.class, UNASSIGNED)
    .filter(v -> v.isUnassigned())
    .penalize(HardSoftScore.ONE_HARD)
    .asConstraint("Unassigned");
```

to

```java
constraintFactory.forEachOnly(Visit.class, UNASSIGNED)
    .penalize(HardSoftScore.ONE_HARD)
    .asConstraint("Unassigned");
```

(and likewise for `isInconsistent` constraints)

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
Christopher-Chianelli: It an OR relationship, but the filter is negated, and by De Morgan's laws, the negation of an OR is an AND of their negations.
As for why the filter is negated, it because we take the complement set, and we need to include all assigned/consistent entities that would be rejected by an "only unassigned"/"only inconsistent" filter.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
Christopher-Chianelli: `EnumMap` does not have an `of` API, which makes code using it harder to read/modify.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -140,6 +140,10 @@ public class EntityDescriptor<Solution_> {
                 var moveDirector = (MoveDirector<Solution_, ?>) solutionView;
                 return !moveDirector.isPinned(this, entity);
             };
+    private final Map<EnumSet<ForEachFilteringCriteria>, Predicate<Object>> forEachIncludesToPredicateBasicVarMap =
+            new LinkedHashMap<>();
```
triceo: I don't like seeing these complex types, especially when they're being used more than once. I recommend creating a new class this complexity; that class can then also be tested separately and isolated.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -92,8 +95,8 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
     // from
     // ************************************************************************
 
-    @Override
-    public <A> @NonNull UniConstraintStream<A> forEach(@NonNull Class<A> sourceClass) {
+    public <A> @NonNull UniConstraintStream<A> forEachUnfilteringSpecified(@NonNull Class<A> sourceClass,
```
triceo: IMO a better name is in order. 

`forEachWithFilters`?
`forEachUnfilteredWithIncludes`?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -104,31 +107,45 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
...
     @Override
     public <A> @NonNull UniConstraintStream<A> forEachIncludingUnassigned(@NonNull Class<A> sourceClass) {
-        assertValidFromType(sourceClass);
-        return share(new BavetForEachUniConstraintStream<>(this, sourceClass, null, RetrievalSemantics.STANDARD));
+        return forEachUnfilteringSpecified(sourceClass, ForEachFilteringCriteria.INCLUDE_UNASSIGNED);
+    }
+
+    @Override
+    public <A> @NonNull UniConstraintStream<A> forEachUnfiltered(@NonNull Class<A> sourceClass) {
+        return forEachUnfilteringSpecified(sourceClass, ForEachFilteringCriteria.INCLUDE_UNASSIGNED,
```
triceo: I wonder if the logic should be inverted.
If no criteria is specified = naked forEach without any exclusions.
Therefore the default forEach would specify criteria.

IMO this makes the logic clearer. ForEach internally returns everything, unless we choose to exclude things. The default state is to not exclude, minimalism.

docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
```diff
@@ -263,13 +263,30 @@ This constraint stream penalizes each known and initialized instance of `Shift`.
...
+----
+    private Constraint penalizeAllConsistentShifts(ConstraintFactory factory) {
+        return factory.forEachIncludingUnassigned(Shift.class)
+                .penalize(HardSoftScore.ONE_SOFT)
+                .asConstraint("A shift");
+    }
+----
+====
+
+To include inconsistent and unassigned instances, replace the `forEach()` building block by `forEachUnfiltered()`:
```
triceo, line 288: The example below still uses `forEachIncludingUnassigned`.

triceo (commented):
I have left my final comments. (Assuming that nothing else important shows up as the changes are being made.)

In the interest of ensuring this makes it to the release next week, I propose:
- We merge this without the inconsistent supply.
- We work on the supply immediately after this is merged.
- If we don't finish the supply in time, making the annotation optional is a backwards compatible change we can introduce later.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityForEachFilter.java
```diff
@@ -0,0 +1,86 @@
...
+                listVariableDescriptor != null && listVariableDescriptor.acceptsValueType(entityDescriptor.getEntityClass());
+        if (isListVariableValue) {
+            var listInverseVariable = listVariableDescriptor.getInverseRelationShadowVariableDescriptor();
+            if (listInverseVariable != null) {
+                return entity -> entityDescriptor.hasNoNullVariables(entity) && listInverseVariable.getValue(entity) != null;
+            } else {
+                return ignored -> {
+                    throw new IllegalStateException("""
+                            Impossible state: assigned predicate for list variable value should not be used
+                            when there no inverse relation shadow variable descriptor
```
triceo:
```suggestion
                            when there is no inverse relation shadow variable descriptor.
```

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityForEachFilter.java
```diff
@@ -0,0 +1,86 @@
...
+    EntityForEachFilter(EntityDescriptor<?> entityDescriptor) {
+        var solutionDescriptor = entityDescriptor.getSolutionDescriptor();
+        var listVariableDescriptor = solutionDescriptor.getListVariableDescriptor();
+        var declarativeShadowVariables = getDeclarativeShadowVariables(entityDescriptor);
+        var shadowVariableInconsistentDescriptor = entityDescriptor.getShadowVariablesInconsistentDescriptor();
+
+        if (!declarativeShadowVariables.isEmpty() && shadowVariableInconsistentDescriptor == null) {
+            // TODO: Remove this when supplier is present
+            throw new IllegalStateException(
+                    "Entity class %s has declarative shadow variables (%s) but does not have a @%s member."
```
triceo, line 33: Do we want to keep calling it "declarative shadow variables" if there are no other ones?

Even in the docs, we'll change it. The word "declarative" will go away - it's just shadow variables.

triceo (commented):
LGTM. Final minor comments.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityForEachFilter.java
Christopher-Chianelli: I say for Solver 2.0, since Solver 1.x will have both declarative and (deprecated) non-declarative shadow variables.

triceo: Tom and I will be doing a docs review tomorrow. I expect some changes to naming to come from that. Let's see.

triceo (approved):

PR #1765, opened 2025-08-28
chore: move streams support for swap moves

triceo:


core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/move/ListChangeMove.java
```diff
@@ -122,12 +125,8 @@ private Value_ getMovedValue() {
...
     @Override
     public void execute(MutableSolutionView<Solution_> solutionView) {
-        if (sourceEntity == destinationEntity) {
-            planningValue = solutionView.moveValueInList(variableMetaModel, sourceEntity, sourceIndex, destinationIndex);
-        } else {
-            planningValue = solutionView.moveValueBetweenLists(variableMetaModel, sourceEntity, sourceIndex, destinationEntity,
-                    destinationIndex);
-        }
+        planningValue = solutionView.moveValueBetweenLists(variableMetaModel, sourceEntity, sourceIndex, destinationEntity,
+                destinationIndex);
```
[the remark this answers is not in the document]
triceo: The check now happens in the underlying method, ignoring.

core/src/main/java/ai/timefold/solver/core/preview/api/move/Move.java
```diff
@@ -114,21 +115,28 @@ default Collection<?> extractPlanningEntities() {
...
-     * @return Each value only once.
+     * @return Each value only once. May contain null.
+     *         TODO convert to SequencedCollection when on Java 21.
      */
     default Collection<?> extractPlanningValues() {
         throw new UnsupportedOperationException("The move (" + this + ") does not support tabu search.");
     }
 
     /**
+     * TODO this probably needs a better API,
```
zepfred, line 126: Adding a record with predefined fields can help standardize the process.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
```diff
@@ -119,14 +119,41 @@ <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Solution_, Entit
...
-            PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel, Entity_ entity, int sourceIndex,
+            PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel, Entity_ sourceEntity, int sourceIndex,
             int destinationIndex);
 
+    /**
+     * Swaps two values between two entities' {@link PlanningListVariable planning list variable}.
+     *
+     * @param variableMetaModel Describes the variable to be changed.
+     * @param leftEntity The first entity whose variable value is to be swapped.
+     * @param leftIndex >= 0
```
zepfred: The description of the index fields are not clear. Let's improve their explanations here and in other methods where we have index fields.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
```diff
@@ -119,14 +119,41 @@ <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Solution_, Entit
...
+    /**
+     * Swaps two values within one entity's {@link PlanningListVariable planning list variable}.
+     *
+     * @param variableMetaModel Describes the variable to be changed.
+     * @param entity The entity whose variable values are to be swapped.
+     * @param leftIndex >= 0
+     * @param rightIndex >= 0
+     * @throws IndexOutOfBoundsException if the index is out of bounds
+     */
+    <Entity_, Value_> void swapValuesInList(
```
zepfred, line 145: I have not checked the implementation logic yet, but I wonder why we don't use a default implementation that calls the previous method. This way, we can delegate the responsibility of identifying whether it is the same entity or not to the previous method.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
```diff
@@ -166,6 +166,10 @@ protected <Score_ extends Score<Score_>> void restoreWorkingSolution(ExhaustiveS
         restoreMoveList.addAll(oldMoveList);
         Collections.reverse(newMoveList);
         restoreMoveList.addAll(newMoveList);
+        if (restoreMoveList.isEmpty()) {
```
zepfred, line 169: Do we have a test case for that?

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/decider/ConstructionHeuristicDecider.java
```diff
@@ -89,9 +89,9 @@ public void decideNextStep(ConstructionHeuristicStepScope<Solution_> stepScope,
         var terminatedPrematurely = false;
         while (moveIterator.hasNext()) {
             var move = moveIterator.next();
+            var moveDirector = stepScope.getMoveDirector();
```
zepfred: Why are we changing its scope?

core/src/main/java/ai/timefold/solver/core/impl/move/director/MoveDirector.java
```diff
@@ -108,21 +108,56 @@ public final <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Sol
...
+            externalScoreDirector.afterListVariableChanged(variableDescriptor, rightEntity, rightIndex, rightIndex + 1);
+        }
+    }
+
+    @Override
+    public <Entity_, Value_> void swapValuesInList(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
+            Entity_ entity, int leftIndex, int rightIndex) {
+        if (leftIndex == rightIndex) {
+            // Do nothing.
+        } else if (leftIndex > rightIndex) { // Always start from the lower index.
```
zepfred: |We have updated the logic in `moveValueInList` to handle the indexes order. Let's reuse its logic approach for consistency.

core/src/main/java/ai/timefold/solver/core/impl/move/director/MoveDirector.java
```diff
@@ -108,21 +108,56 @@ public final <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Sol
...
+            swapValuesInList(variableMetaModel, entity, rightIndex, leftIndex);
+        } else {
+            var variableDescriptor = extractVariableDescriptor(variableMetaModel);
```
zepfred: I suggest moving the logic to the previous method and implementing the suggested changes to the contract definition. This will lead to a single method that remains non complex.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/move/AbstractMove.java
```diff
@@ -54,9 +49,7 @@ public final String describe() {
 
     public abstract String toString();
 
-    protected List<VariableMetaModel<Solution_, ?, ?>> getVariableMetaModels() {
```
zepfred, line 57: I'm uncertain about removing the prefix "get" for consistency.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/move/ListSwapMove.java
```diff
@@ -0,0 +1,175 @@
...
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Swaps two elements of a {@link PlanningListVariable list variable}.
+ * Each element is identified by an entity instance and an index in that entity's list variable.
+ * The swap move has two sides called left and right. The element at {@code leftIndex} in {@code leftEntity}'s list variable
+ * is replaced by the element at {@code rightIndex} in {@code rightEntity}'s list variable and vice versa.
+ * Left and right entity can be the same instance.
+ * <p>
+ * An undo move is created by flipping the left and right-hand entity and index.
```
zepfred: Does it make sense to mention the undo move?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/move/ListSwapMove.java
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
```
zepfred, line 28: Let's include the descriptor of the other types: `Entity_` and `Value_`

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public ListSwapMoveProvider(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel) {
+        this.variableMetaModel = Objects.requireNonNull(variableMetaModel);
+    }
+
+    @Override
+    public MoveProducer<Solution_> apply(MoveStreamFactory<Solution_> moveStreamFactory) {
+        var assignedValueStream = moveStreamFactory.enumerate(variableMetaModel.type(), false)
+                .filter((solutionView,
+                        value) -> solutionView.getPositionOf(variableMetaModel, value) instanceof PositionInList);
```
zepfred, line 30: We need to verify if the position is assigned in other places. What do you think about updating the `ElementPosition` contract to include a new method called `isAssigned`? This approach seems preferable to casting.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        var validAssignedValuePairStream = assignedValueStream.join(assignedValueStream,
```
zepfred, line 31: CS includes a method for generating distinct pairs of values, and it seems like a good fit here. What do you think about adding something similar for MS?

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
```diff
@@ -0,0 +1,65 @@
...
+                        .distinct()
+                        .map((solutionView, pair) -> FullElementPosition.of(variableMetaModel, solutionView, pair.first()),
+                                (solutionView, pair) -> FullElementPosition.of(variableMetaModel, solutionView, pair.second()));
+        // Eliminate pairs that cannot be swapped due to value range restrictions.
+        var result = validAssignedValueUniquePairStream
+                .filter((solutionView, leftPosition, rightPosition) -> solutionView.isValueInRange(variableMetaModel,
+                        rightPosition.entity(), leftPosition.value())
+                        && solutionView.isValueInRange(variableMetaModel, leftPosition.entity(), rightPosition.value()));
+        // Finally pick the moves.
+        return moveStreamFactory.pick(result)
```
zepfred, line 48: It's nice to see the simplification compared to the move selectors!

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        // Ensure unique pairs; without demanding PlanningId, this becomes tricky.
+        // Convert values to their locations in list.
+        var validAssignedValueUniquePairStream =
+                validAssignedValuePairStream
+                        .map((solutionView, leftValue, rightValue) -> new UniquePair<>(leftValue, rightValue))
+                        .distinct()
+                        .map((solutionView, pair) -> FullElementPosition.of(variableMetaModel, solutionView, pair.first()),
+                                (solutionView, pair) -> FullElementPosition.of(variableMetaModel, solutionView, pair.second()));
+        // Eliminate pairs that cannot be swapped due to value range restrictions.
+        var result = validAssignedValueUniquePairStream
```
zepfred, line 43: One improvement we could consider is using `ReachableValues` to limit the values that can be swapped and decrease the number of generated moves. If we can find a way to enumerate the reachable values based on the left node, we may generate fewer moves while also likely producing more moves that are accepted by the value range filter node.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/SwapMoveProvider.java
```diff
@@ -0,0 +1,95 @@
...
+    public MoveProducer<Solution_> apply(MoveStreamFactory<Solution_> moveStreamFactory) {
+        var entityType = entityMetaModel.type();
+        var dataStream = moveStreamFactory.enumerate(entityType, false)
+                .join(entityType,
+                        filtering((SolutionView<Solution_> solutionView, Entity_ leftEntity, Entity_ rightEntity) -> {
+                            if (leftEntity == rightEntity) {
+                                return false;
+                            }
+                            var change = false;
+                            for (var variableMetaModel : variableMetaModelList) {
```
zepfred, line 71: We need to check all variables before returning. If any of them fail to accept the swap, then the change must not be accepted. For example, given e1(v1, v2) and e2(v3, v4). Let's assume that swapping v1 with v3 is accepted, but swapping v2 with v4 is not.

zepfred: We probably need to add coverage to the described use-case.

zepfred (changes requested):
This PR was indeed simpler than the previous one. However, I would like you to confirm a possible bug in the basic swap provider.

core/src/main/java/ai/timefold/solver/core/preview/api/move/Move.java
triceo: Could be. Not something to be solved now, that's why I'm adding the TODO.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
triceo: There are two ways of looking at this - one from the user's point of view, and the other from ours. In this case, I choose the user's.

In their move code, they want to have the least possible logic. And therefore if they really want to swap values in the same list, they want to do:

     swap(list, x, y)

And not:

     swap(list, x, list, y)

If they choose the latter anyway, we can convert it to the first under the hood, which we do.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
triceo: Yes, in fact this is how this was caught. With the composite move no longer accepting empty list of moves, the previous code was throwing exceptions now.

Anyway... exhaustive search will have to be refactored anyway, when we add support for list vars there. Coming soon.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/move/AbstractMove.java
triceo: This is a problem that the JDK's decisions have created.

The convention for 20 years has been that `get...()` retrieves and `set...()` updates.
However, with the introduction of records, getters are no more and we now have _two_ conventions for retrieval, one of which doesn't use the `get` prefix.

There is no consistency anymore. Code that mixes records and classes will inevitably be inconsistent. In my mind, it is more consistent to apply the `get` pattern where there are setters, and use the record style when there are no setters. As is the case here.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
triceo: This is not actually a cast - it's pattern matching. Java is very clearly going in this direction:

https://openjdk.org/jeps/507
https://openjdk.org/jeps/441
https://openjdk.org/jeps/440

For sealed types, this is how they're supposed to be used.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
triceo: I'm actually thinking of _removing_ it from CS. :-) It does more harm than good, but that's for a longer conversation.

I have a suspicion that this code won't stay here for long anyway; it's brittle and it will need a better solution long-term. Good enough for the first preview.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/ListSwapMoveProvider.java
triceo: The benefit of using Bavet for this is that this is all really only evaluated when necessary, as the underlying data changes. Therefore even though the filtering is technically costly, it doesn't actually matter - the filter will only be recomputed when the value changes, which is once in a blue moon.

With a typical swap move in the solver, out of all the values, only 2 will be recomputed every step.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/provider/SwapMoveProvider.java
triceo: Well spotted. 

This is very much a corner case though - the vast majority of our users only have one basic variable, in which case the code worked. And of those who have multiple basic variables, an even smaller number of them are using entity ranges. But you're right of course, it has to work regardless.

triceo:
The integration test failure is expected; we merged something on main and it temporarily broke the benchmarks.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
zepfred: I agree that we should have both methods, but it seems we could also have just one implementation:

```
    default <Entity_, Value_> @Nullable Value_ moveValueInList(
            PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel, Entity_ entity, int sourceIndex,
            int destinationIndex) {
        return moveValueBetweenLists(variableMetaModel, entity, sourceIndex, entity, destinationIndex);
    }
```

`moveValueBetweenLists` would be responsible for determining whether it is the same entity or not.

PR #1774, opened 2025-09-01
docs: add programmatic benchmark aggregation example to documentation

PatrickDiallo23:

PR #1777, opened 2025-09-03
feat: add high-level aggregation methods to BenchmarkAggregator

PatrickDiallo23:
## Summary
This PR adds five new high-level methods to `BenchmarkAggregator` that simplify common benchmark aggregation flows by automatically handling configuration loading and setup.

### New Methods Added:

- **`aggregateBenchmarks(PlannerBenchmarkConfig)`** - Aggregates all benchmark results found in the configured directory
- **`aggregateSelectedBenchmarks(PlannerBenchmarkConfig, List<String>)`** - Aggregates benchmark results from specific subdirectories by name that were selected by user
- **`aggregateFrameSelectedBenchmarks(List<SingleBenchmarkResult>, Map<SolverBenchmarkResult, String>)`** - Aggregates user-selected benchmark results (for GUI/BenchmarkAggregatorFrame integration)
- **`getAvailableBenchmarkDirectories(PlannerBenchmarkConfig)`** - Returns list of available benchmark subdirectories with valid results
- **`aggregateAndShowReportInBrowser(PlannerBenchmarkConfig)`** - Aggregates all benchmarks and opens the HTML report in the default browser

### Additional Changes:
- Added `benchmarkReport` instance variable to store generated reports
- Made the original `aggregate()` method private as it's now an internal implementation detail
- Updated `aggregate()` method to store the benchmark report for browser viewing
- Added required imports for Desktop integration and automatic configuration loading
- Updated `doInBackground()` method with the new implemented `aggregateFrameSelectedBenchmarks` method to ensure backward compatibility

## Motivation
The existing `aggregate()` method requires manual setup of the aggregator (setting benchmark directory, report config, etc.) before use. These new methods provide a more convenient API by:
- Automatically loading configuration from `benchmarkConfig.xml`
- Handling directory validation and setup
- Providing browser integration for immediate report viewing
- Supporting both "aggregate all" and "aggregate selected" use cases

The existing `BenchmarkAggregator` required manual setup (setting benchmark directory, report config, etc.) and only supported aggregating pre-selected results. These new methods provide more flexible aggregation options:

- Select benchmarks by subdirectory names rather than individual results
- Find available benchmark directories automatically
- View reports immediately after generation in browser
- Support both "aggregate all" and "selective aggregation" use cases
- Automatically loading configuration from benchmark config XML file
- Handling directory validation and setup

## Usage Examples
```java
PlannerBenchmarkConfig config = PlannerBenchmarkConfig.createFromXmlResource("benchmarkConfig.xml");
BenchmarkAggregator aggregator = new BenchmarkAggregator();

// Aggregate all benchmark results
File reportFile = aggregator.aggregateBenchmarks(config);

// Discover available benchmark directories
List<String> available = aggregator.getAvailableBenchmarkDirectories(config);

// Aggregate specific directories
File reportFile = aggregator.aggregateSelectedBenchmarks(config, 
    Arrays.asList("benchmark_2024-01-15", "benchmark_2024-01-16"));

// OR
File reportFile = aggregator.aggregateSelectedBenchmarks(config, available);

// Aggregate and view in browser
aggregator.aggregateAndShowReportInBrowser(config);

// For GUI/BenchmarkAggregatorFrame integration (existing pattern)
aggregator.aggregateFrameSelectedBenchmarks(selectedResults, nameMapping);
```

## Dependencies

- No new external dependencies added
- Uses existing Timefold benchmark framework components
- Desktop integration uses standard Java AWT (optional functionality)

## Note

If this PR is approved, the documentation will need to be updated as the following: the latest version of the documentation will be updated with the new methods and usage examples, the prior versions of the documentation will keep [this version](https://github.com/TimefoldAI/timefold-solver/pull/1774).


benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
```diff
@@ -40,11 +48,145 @@ public void setBenchmarkReportConfig(BenchmarkReportConfig benchmarkReportConfig
...
+
+        // Configure the aggregator instance
+        this.setBenchmarkDirectory(benchmarkDirectory);
+        this.setBenchmarkReportConfig(benchmarkConfig.getBenchmarkReportConfig());
+
+        // Perform the aggregation - returns HTML report file
+        return this.aggregate(allSingleBenchmarkResults, solverBenchmarkResultNameMap);
+    }
+
+    public File aggregateFrameSelectedBenchmarks(List<SingleBenchmarkResult> selectedSingleBenchmarkResults,
```
triceo: IMO this method needs a better name. Maybe something that tells that it deals with the UI? "Frame" doesn't say much, in my opinion.

(I know it had been like this already, but if we're modernizing this, let's do it right.)

PatrickDiallo23:
@triceo Thank you for your review! Regarding the Unit Tests, does Timefold's best practices include using Assertions or Mockito? Are you using given,when,then pattern? What practice would you advice me to follow? Also, regarding the method name I was thinking about `aggregateUiSelectedBenchmarks`. What do you think?

triceo:
@PatrickDiallo23 We use AssertJ, which has a well-defined API for writing even the most complex assertions.

Wrt. the method name - `aggregateSelectedBenchmarksInUi`?

PatrickDiallo23:
> AssertJ,

@triceo I have a quick question: the unit tests are almost done, the only issue I have is setting the benchmarkDirectory field in the xml to perform the aggregation test.

If I set it to `target/benchmarkTest/output`, then I can only run the tests separately after I ran once `mvn clean install` (because, of course, the target is not yet generated)
If I set it to `data` or `local/benchmarkReport`, the aggregation tests run before other benchmark tests so it cannot find any benchmark results.
I tried to add some mock benchmark results in `benchmark/src/test/resources/ai/timefold/solver/benchmark/impl/result/aggregator` but it cannot find the directory.

Do you have any advice on how should I approach this thing? Thanks!

triceo:
It would be an antipattern to rely in the tests on something that a previous test may or may not have created. Therefore your first two options should indeed be out of the question.

With JUnit 5, you can have your test work in a [random temp directory](https://www.baeldung.com/junit-5-temporary-directory). In a `@Before` method, I'd create stuff in that directory; either by copying from `resources`, or through straight generating.

Storing stuff in `src/test/resources` is the correct approach. It is strange that a resource would not be found there, if it is there. You want to use [this method](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getResourceAsStream-java.lang.String-) to locate those files. The root is inside `src/test/resources`, and therefore path `a` should resolve to `src/test/resources/a`.

Besides that, not much help I can give you. If it can't find those files, figure out why; you're likely pointing to the wrong root.

PatrickDiallo23:
Thank you for your advice! I managed to solve the issue. I updated the PR. When you have time, please review it. Thanks!

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
```diff
@@ -40,11 +48,145 @@ public void setBenchmarkReportConfig(BenchmarkReportConfig benchmarkReportConfig
...
+
+        // Configure the aggregator instance
+        this.setBenchmarkDirectory(benchmarkDirectory);
+        this.setBenchmarkReportConfig(benchmarkConfig.getBenchmarkReportConfig());
+
+        // Perform the aggregation with selected results - returns HTML report file
+        return this.aggregate(selectedSingleBenchmarkResults, solverBenchmarkResultNameMap);
+    }
+
+    public List<String> getAvailableBenchmarkDirectories(PlannerBenchmarkConfig benchmarkConfig) {
```
triceo, line 143: This method is only used in test—please keep it out of `src/main/java`.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
```diff
@@ -40,11 +48,145 @@ public void setBenchmarkReportConfig(BenchmarkReportConfig benchmarkReportConfig
...
+        }
+
+        File htmlOverviewFile = this.benchmarkReport.getHtmlOverviewFile();
+        Desktop desktop = Desktop.isDesktopSupported() ? Desktop.getDesktop() : null;
+        if (desktop != null && desktop.isSupported(Action.BROWSE)) {
+            try {
+                desktop.browse(htmlOverviewFile.getAbsoluteFile().toURI());
+            } catch (IOException e) {
+                throw new IllegalStateException(
+                        "Failed showing htmlOverviewFile (" + htmlOverviewFile + ") in the default browser.", e);
```
triceo: The solver recently started using a different pattern for exceptions, which improves their readability if there are more arguments.

                        "Failed showing htmlOverviewFile (%s) in the default browser.".formatted(e)) 

Please use the new pattern and apply it consistently.

benchmark/src/test/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregatorTest.java
```diff
@@ -0,0 +1,258 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+class BenchmarkAggregatorTest {
+
+    private BenchmarkAggregator benchmarkAggregator;
+
+    @BeforeEach
+    void setUp() {
+        benchmarkAggregator = new BenchmarkAggregator();
+    }
```
triceo, line 96: Arguably unnecessary. This will do the same thing:

    private final BenchmarkAggregator benchmarkAggregator = new BenchmarkAggregator();

benchmark/src/test/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregatorTest.java
```diff
@@ -0,0 +1,258 @@
...
+        PlannerBenchmarkConfig config = new PlannerBenchmarkConfig();
+        config.setBenchmarkDirectory(tempDir.toFile());
+
+        List<String> directories = benchmarkAggregator.getAvailableBenchmarkDirectories(config);
+
+        assertThat(directories).isEmpty();
+    }
+
+    @Test
+    void getAvailableBenchmarkDirectories_withValidDirectories() throws IOException {
```
triceo: The exception is not thrown, please remove the declaration and apply consistently throughout the file.

benchmark/src/test/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregatorTest.java
```diff
@@ -0,0 +1,258 @@
...
+        PlannerBenchmarkConfig config = PlannerBenchmarkConfig
+                .createFromXmlResource("ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml");
+
+        List<String> directories = benchmarkAggregator.getAvailableBenchmarkDirectories(config);
+
+        assertThat(directories)
+                .hasSize(4)
+                .containsExactlyInAnyOrder("2025-09-05_141529", "2025-09-05_141529_1", "2025-09-05_141530",
```
triceo: I'm not sure what I think about including these directories in the resources, and referencing them from there.

- First of all, reaching directly into the project is generally discouraged; you should only work with what is available on the classpath. (Which will be everything in `src/test/resources`, but without that prefix.) Using files in temp directories you create yourself is also fair game.
- Second, this is a lot of data and boilerplate to store in Git purely for the purposes of checking that it exists. (Including the `website/` directories, which is just a set of identical files.) 

The cleanest way would probably be to run the benchmark in a temp directory once or twice during `@BeforeAll`, with some very short terminations, and then check the directories from there.

More importantly though - the method to check for available directories, that is only used in tests. Why are we testing a method which is only used in tests? Arguably, this can all be removed without losing any test coverage.

What am I missing?

benchmark/src/test/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregatorTest.java
```diff
@@ -0,0 +1,258 @@
...
+    @Test
+    void aggregateAndShowReportInBrowser_withValidResults() {
+        PlannerBenchmarkConfig config = PlannerBenchmarkConfig
+                .createFromXmlResource("ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml");
+
+        File htmlReportFile = benchmarkAggregator.aggregateAndShowReportInBrowser(config);
+
+        assertThat(htmlReportFile).isNotNull();
+        assertThat(htmlReportFile.exists()).isTrue();
+        assertThat(htmlReportFile.getName()).endsWith(".html");
```
triceo, line 277: This is another test which requires copies of the benchmark reports in resources. We want to check that the benchmarker understand the reports _today_, not some reports from long past. 

(Even though we do not intend to break those reports, we do not commit to backwards compatibility there.)

IMO this is exactly the test that requires running several very short benchmarks (milliseconds) and aggregating them on the fly. An end-to-end test, without relying on report copies in resources.

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plannerBenchmark>
+    <benchmarkDirectory>src/test/resources/ai/timefold/solver/benchmark/impl/result/aggregator</benchmarkDirectory>
```
triceo: This file should only be a template. It should be post-processed in the test to point to a temp directory.

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
```diff
@@ -0,0 +1,24 @@
...
+    <warmUpSecondsSpentLimit>0</warmUpSecondsSpentLimit>
+    <benchmarkReport>
+        <locale>en_US</locale>
+    </benchmarkReport>
+    <solverBenchmark>
+        <problemBenchmarks>
+            <solutionFileIOClass>ai.timefold.solver.persistence.common.api.domain.solution.RigidTestdataSolutionFileIO</solutionFileIOClass>
+            <inputSolutionFile>target/test/benchmarkTest/input.xml</inputSolutionFile>
```
triceo: During the post-processing, we can have this point to a temp file as well.
Pointing into `target` is an anti-pattern and makes the test fragile.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
```diff
@@ -40,11 +48,145 @@ public void setBenchmarkReportConfig(BenchmarkReportConfig benchmarkReportConfig
...
+        // Perform the aggregation with selected results - returns HTML report file
+        return this.aggregate(selectedSingleBenchmarkResults, solverBenchmarkResultNameMap);
+    }
+
+    public List<String> getAvailableBenchmarkDirectories(PlannerBenchmarkConfig benchmarkConfig) {
+        File benchmarkDirectory = benchmarkConfig.getBenchmarkDirectory();
+        List<String> directories = new ArrayList<>();
+
+        if (!benchmarkDirectory.exists() || !benchmarkDirectory.isDirectory()) {
+            return directories;
```
triceo: You can return `Collections.emptyList()` here.
Which allows you to only allocate `directories` later - moving line 144 to after this block.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
PatrickDiallo23: I also added that method because in a real app it would be easier to retrieve the available directories for aggregation than create your own logic for it or copy paste the names. 

At least in my case, I used this kind of feature in my research app and I am thinking that it would be useful for other devs.

benchmark/src/test/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregatorTest.java
PatrickDiallo23: @triceo  Regarding the method to check for available directories, I mentioned in a previous comment that this method can used to retrieved the available benchmarks results to do the aggregation. Maybe I don't want to aggregate all the benchmarks, but a subset of them. And for this, I want to have a very simplified process meaning that I don't want to paste the directories names or write myself the logic to retrieve the available directories.

Regarding this comment, I will try to create some benchmarks during @BeforeAll and use them to validate the tests.

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
PatrickDiallo23: Do you mean to point to a temp directory that contains all the benchmark reports and that was created before during @BeforeAll, right?

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
PatrickDiallo23: > Do you mean to point to a temp directory that contains all the benchmark reports and that was created before during @BeforeAll, right?

The same thing applies here as well, right?

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
triceo: Yes.

benchmark/src/test/resources/ai/timefold/solver/benchmark/api/testdataBenchmarkConfigWithReport.xml
triceo: Yes.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
triceo: Fair enough. Let's keep the method.

triceo:
One more thing I just realized - with the method now moved to the codebase, we can update the section in the docs to remove the code from there, replacing it with a snippet to call the method that newly exists.

PatrickDiallo23:
> One more thing I just realized - with the method now moved to the codebase, we can update the section in the docs to remove the code from there, replacing it with a snippet to call the method that newly exists.

Do you want me to update that section in this PR? Or should I create another PR after we solve this? Another thing would be if I update the documentation that means those methods will be available in the latest Timefold version, right? If so, there is a way to announce people that the older version should still use the initial approach that I proposed in a previous version (PR)?

PatrickDiallo23:
In the meantime, I was able to solve your requests. I hope that this approach is better and everything works properly. Locally it works fine in my case. Also, I am waiting with interest for your response regarding the documentation. Thank you for your suppport and reviews!

triceo:
> Do you want me to update that section in this PR? 

Yes.

> If so, there is a way to announce people that the older version should still use the initial approach that I proposed in a previous version (PR)?

There will be a release later today, worst case tomorrow. If we merge this before then, we do not need to worry about this.

benchmark/src/main/java/ai/timefold/solver/benchmark/impl/aggregator/BenchmarkAggregator.java
```diff
@@ -1,14 +1,18 @@
 package ai.timefold.solver.benchmark.impl.aggregator;
 
+import java.awt.Desktop;
+import java.awt.Desktop.Action;
 import java.io.File;
+import java.io.IOException;
 import java.time.OffsetDateTime;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
triceo: We do not use asterisk imports. Please configure your IDE to not use them.

triceo (approved):
LGTM after final comment resolved + docs updated.

PatrickDiallo23:
@triceo I provided the final (I hope) changes. Please review them.

triceo (approved):
Had to fix source code formatting (`mvn clean install` would have done it for you), other than that LGTM.

Will merge later today, thank you!

PatrickDiallo23:
Thank you very much for your support! Sorry that I forgot to run mvn clean install, but glad that everything is okay.

PR #1778, opened 2025-09-03
chore: make @ShadowVariablesInconsistent optional

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.IdentityHashMap;
+
+import ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor;
+import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public class ConsistencyTracker<Solution_> {
+    private final IdentityHashMap<Object, Boolean> entityToIsInconsistentMap = new IdentityHashMap<>();
```
triceo: I wonder... can this be a set? Not contains => false?
(I realize that set is effectively a map of object to boolean, but that doesn't mean we need to expose it.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,42 @@
...
+    private final HashMap<Class<?>, EntityConsistencyState<Solution_>> entityClassToConsistencyStateMap = new HashMap<>();
+
+    public EntityConsistencyState<Solution_>
+            getDeclarativeEntityConsistencyState(EntityDescriptor<Solution_> entityDescriptor) {
+        return entityClassToConsistencyStateMap.computeIfAbsent(entityDescriptor.getEntityClass(),
+                ignored -> new EntityConsistencyState<>(entityDescriptor, entityToIsInconsistentMap));
+    }
+
+    public void initializeForEntities(SolutionDescriptor<Solution_> solutionDescriptor,
```
triceo: I don't like this pattern. It exposes the object in an uninitialized state, only to initialize it later. It is unnecessary.

Let's create a static constructor and hide this in there. If we can avoid exposing half-built objects, let's.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -47,31 +47,51 @@ public DefaultShadowVariableSessionFactory(
     }
 
     public static <Solution_> VariableReferenceGraph buildGraph(
+            ConsistencyTracker<Solution_> consistencyTracker,
```
triceo: The argument lists to these methods are getting out of hand; readability suffers. Some builder pattern would help, and the private methods arguably don't need to be static. That way, we don't need to pass the same state to all of them.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/FixedVariableReferenceGraph.java
```diff
@@ -23,10 +23,26 @@ public FixedVariableReferenceGraph(VariableReferenceGraphBuilder<Solution_> oute
 
         // Now that we know the topological order of nodes, add
         // each node to changed.
+        changedVariableNotifier = outerGraph.changedVariableNotifier;
```
triceo, line 26: I see a bit of an inconsistency in our approach here.

In the `ConsistencyTracker`, we split the initialization into a separate method, so that we don't clutter the constructor with it.
Here, we have no problems with cluttering the constructor with initialization.

Maybe we pick on style or another, and stick with it consistently?
Personally, the static constructor is IMO the best way - I never liked exposing the actual constructors to the outside.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/SingleDirectionalParentVariableReferenceGraph.java
```diff
@@ -41,9 +42,9 @@ public SingleDirectionalParentVariableReferenceGraph(
         this.changedVariableNotifier = changedVariableNotifier;
         var shadowEntities = Arrays.stream(entities).filter(monitoredEntityClass::isInstance)
                 .sorted(topologicalOrderComparator).toArray();
-        var inconsistentDescriptor =
-                sortedDeclarativeShadowVariableDescriptors.get(0).getEntityDescriptor()
-                        .getShadowVariablesInconsistentDescriptor();
+        var entityConsistencyState =
```
triceo, line 45: Same comment on the constructor as above.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -234,7 +240,9 @@ public void resetWorkingSolution() {
                     scoreDirector.getSolutionDescriptor(),
                     scoreDirector,
                     shadowVariableGraphCreator);
-            shadowVariableSession = shadowVariableSessionFactory.forSolution(scoreDirector.getWorkingSolution());
+            shadowVariableSession =
+                    shadowVariableSessionFactory.forSolution(consistencyTracker, scoreDirector.getWorkingSolution());
+            shadowVariableSession.updateVariables();
```
triceo: Is there a situation where this update would not be necessary, or even harmful?
This is possibly the same situation as above - we expose something partially uninitialized, and expect the user to remember to initialize it for us. These situations should be avoided.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -54,7 +54,8 @@ public void clearShadowVariablesListenerQueue() {
 
     @Override
     public void setWorkingSolutionWithoutUpdatingShadows(Solution_ workingSolution) {
-        session = scoreDirectorFactory.newSession(workingSolution, constraintMatchPolicy, derived);
+        session = scoreDirectorFactory.newSession(workingSolution, variableListenerSupport.getConsistencyTracker(),
```
triceo, line 76: In Move Streams, I solved this ever-expanding argument list by introducing `SessionContext` - a simple wrapper record that holds the information that may need to get propagated throughout the codebase.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -95,6 +101,28 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
 
     private <A> @NonNull UniConstraintStream<A> forEachForCriteria(@NonNull Class<A> sourceClass,
             ForEachFilteringCriteria criteria) {
+        // Required for node sharing, since using a lambda will create different instances
```
triceo: The method is already complex enough even without long inner type declarations. Let's move it outside of the method. (Would keep it in the class though.)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -145,11 +174,21 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
 
     @Override
     public @NonNull <A> UniConstraintStream<A> from(@NonNull Class<A> fromClass) {
+        // Required for node sharing, since using a lambda will create different instances
+        record PredicateSupplier<Solution_, A>(
```
triceo: Dtto.

triceo (commented):
The concept is good. Requesting some changes to implementation, good to go after that.

triceo:
Also, we may have a problem in one of the tests, something is running forever:

image

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityConsistencyState.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.IdentityHashMap;
+
+import ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public final class EntityConsistencyState<Solution_> {
```
triceo: Maybe we make this generic in `Entity_`? I don't like the `Object` everywhere.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
Christopher-Chianelli: No, it a three-state Boolean: not present = unknown, true = is inconsistent, false = is consistent

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
Christopher-Chianelli: Doing this calculation here is harmful, since it is called in `resetShadowVariablesWithoutUpdatingShadows`; it was done so consistency infomation is available for initial score calculation; probably should somehow move it closer to the `inital score calculation` logic. (It is harmful due to `ConstraintVerifier`)

Christopher-Chianelli:
> Also, we may have a problem in one of the tests, something is running forever:
> image

That job encountered an internal server error and failed to attach to any machine, IIRC

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
triceo: My point is the same as with the constructor above. We are exposing something half-initialized, which only makes sense if that half-initialization is necessary somewhere else. If it's not, the `updateVariables()` method should be folded into the constructor.

Christopher-Chianelli: The "half-initialized" is necessary somewhere else - `ConstraintVerifier` (which does not update shadows by default); hence the `updateVariables` needs to be moved even further away (hence the slightly different behaviour between `given` and `givenSolution` currently).

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/FixedVariableReferenceGraph.java
Christopher-Chianelli: ConsistencyTracker is actually supposed to be initialized by the graph in its first update call (as the graph know what is consistent and what is not); 
`initializeForEntities` is a exclusively for `ConstraintVerifier` which does not create a graph (see `fireAndForget`), so it create it own graph from the given entities.


`FixedVariableReferenceGraph` and `SingleDirectionalParentVariableReferenceGraph` are special since their nodes/entities are always guaranteed to be consistent, thus they mark entities as consistent in their constructor so we don't do pointless work in `updateChanged`.

`DefaultVariableReferenceGraph`/`AffectedEntityUpdater` has dynamic consistency so it does all work in `updateChanged`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -151,11 +149,11 @@ private boolean updateEntityShadowVariables(GraphNode<Solution_> entityVariable,
     }
 
     private boolean updateLoopedStatusOfEntity(Object entity, int entityId,
```
triceo: This method can be made generic in <Entity_>; `Object` carries no semantics, `Entity_` makes it clear which `Object`s go together.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,54 @@
...
+                // set inconsistency of entities, which ConstraintVerifier does not want
+                // if the user specified a specific inconsistency value.
+                new GraphStructure.GraphStructureAndDirection(GraphStructure.ARBITRARY, null, null),
+                new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor, ChangedVariableNotifier.empty(),
+                        entityList.toArray()).withConsistencyTracker(this));
+        graph.setUnknownInconsistencyValues();
+        isFrozen = true;
+    }
+
+    public boolean isFrozen() {
```
triceo, line 77: Deserves explanatory Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,54 @@
...
+            new HashMap<>();
+    private boolean isFrozen;
+
+    public EntityConsistencyState<Solution_, Object>
+            getDeclarativeEntityConsistencyState(EntityDescriptor<Solution_> entityDescriptor) {
+        return entityClassToConsistencyStateMap.computeIfAbsent(entityDescriptor.getEntityClass(),
+                ignored -> new EntityConsistencyState<>(entityDescriptor, entityToIsInconsistentMap));
+    }
+
+    public void setUnknownConsistencyFromEntityShadowVariablesInconsistent(SolutionDescriptor<Solution_> solutionDescriptor,
```
triceo: The concept of "unknown inconsistency" ought to be explained in Javadoc. Either here, or on `Graph.setUnknownInconsistencyValues()`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,54 @@
...
+            Object[] entityOrFacts) {
+        var entityList = new ArrayList<>(entityOrFacts.length);
```
triceo: Arguably doing this work via a stream can turn this into an array directly, without having the create intermediate lists.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -0,0 +1,54 @@
...
+        graph.setUnknownInconsistencyValues();
+        isFrozen = true;
+    }
+
+    public boolean isFrozen() {
+        return isFrozen;
+    }
+
+    @Override
+    public String toString() {
```
triceo: This class can be a `record`. In this case, it would help IMO, because much is boilerplate here.

This is predicated on the assumption that `setUnknownConsistencyFromEntityShadowVariablesInconsistent` can be turned into a constructor. Otherwise this comment is moot, and therefore the `isFrozen` made final.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -46,46 +46,72 @@ public DefaultShadowVariableSessionFactory(
         this.graphCreator = graphCreator;
     }
 
-    public static <Solution_> VariableReferenceGraph buildGraph(
+    public record GraphDescriptor<Solution_>(ConsistencyTracker<Solution_> consistencyTracker,
```
triceo, line 49: I like this! Although I'd have moved it to a separate file. This one is already long enough to digest in one piece.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -46,46 +46,72 @@ public DefaultShadowVariableSessionFactory(
...
+    }
+
+    public static <Solution_> VariableReferenceGraph buildGraph(GraphDescriptor<Solution_> graphDescriptor) {
+        var graphStructureAndDirection = GraphStructure.determineGraphStructure(graphDescriptor.solutionDescriptor(),
+                graphDescriptor.entities());
         LOGGER.trace("Shadow variable graph structure: {}", graphStructureAndDirection);
+        return buildGraphForStructureAndDirection(graphStructureAndDirection, graphDescriptor);
+    }
+
+    public static <Solution_> VariableReferenceGraph buildGraphForStructureAndDirection(
```
triceo: Only called from the same package, therefore `public` is not necessary.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EmptyVariableReferenceGraph.java
```diff
@@ -26,4 +26,9 @@ public String toString() {
         return "{}";
     }
 
+    @Override
+    public void setUnknownInconsistencyValues() {
```
triceo: Let's try to keep boilerplate methods (such as `toString`) last.
Also, I wonder if this method should have been made `default` on the interface, avoiding the need for such overrides.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityConsistencyState.java
```diff
@@ -0,0 +1,81 @@
...
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public final class EntityConsistencyState<Solution_, Entity_> {
+    private final IdentityHashMap<Entity_, Boolean> entityToIsInconsistentMap;
+
+    @Nullable
+    private final ExternalizedShadowVariableInconsistentProcessor<Solution_> externalizedShadowVariableInconsistentProcessor;
+
+    private final VariableDescriptor<Solution_> randomDeclarativeVariableDescriptor;
```
triceo: Considering how it's used, I suggest calling it `arbitrary` instead of `random`.

Also, I wonder... is there any way how consistency state would change without any declarative variable changing? If not, why would it ever be necessary to specifically trigger this? Consistency information is already set when the DSV value supplier is called. Therefore the update of consistency information will be guaranteed as a side-effect.

If this code has an actual purpose, it should arguably be documented in the comments.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -52,9 +53,27 @@ public void clearShadowVariablesListenerQueue() {
...
 
+    /**
+     * The function is exclusively available for the Bavet score director, and its use must be approached with caution.
+     * The primary purpose of this method is
+     * to enable the {@code ConstraintVerifier}
+     * to update the consistency status of entities in the solution without updating shadows.
+     * <p>
+     * This must be done before setWorkingSolutionWithoutUpdatingShadows, which inserts the entities.
+     */
+    public void updateConsistencyFromSolution(Solution_ solution) {
```
triceo, line 65: I see the point of this for Bavet - but doesn't Incremental also have the same problem? It also loads the problem from a starting solution, and works with before/after calls.

triceo: Although as I write this, I realize that `ConstraintVerifier` is, by definition, only applicable to CS.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -92,20 +98,41 @@ public <Stream_ extends BavetAbstractConstraintStream<Solution_>> Stream_ share(
...
+                return criteria == that.criteria
+                        && Objects.equals(entityDescriptor.getEntityClass(), that.entityDescriptor.getEntityClass());
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(entityDescriptor.getEntityClass(), criteria);
+        }
```
triceo: As the record doesn't use any arrays, reimplementing equals/hashCode should not be necessary. This is not performance-sensitive code, so micro-optimizations that may be possible in custom impls do not apply here either.

If the goal is to compare entity descriptors by class, then I suggest we add equals to entity descriptor; that is where this check logically belongs. Two entity descriptors for the same class are a practical impossibility; they _should_ equal.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/ForEachFilteringCriteria.java
```diff
@@ -1,28 +1,33 @@
...
         this.entityDescriptorToPredicateFunction = entityDescriptorToPredicateFunction;
     }
 
     @Nullable
     @SuppressWarnings("unchecked")
-    public <A> Predicate<A> getFilterForEntityDescriptor(EntityDescriptor<?> entityDescriptor) {
-        return (Predicate<A>) entityDescriptorToPredicateFunction.apply(entityDescriptor);
+    public <A> Predicate<A> getFilterForEntityDescriptor(ConsistencyTracker<?> consistencyTracker,
+            EntityDescriptor<?> entityDescriptor) {
+        return (Predicate<A>) entityDescriptorToPredicateFunction.apply(consistencyTracker, entityDescriptor);
```
triceo, line 31: For node-sharing, is this guaranteed to return the same instance every time?

Looking at where this leads, it'll be `entityConsistencyState::isEntityConsistent` - and that will be a different object every time.

For peace of mind and the future generations, there should be a test guaranteeing this behavior.

triceo (commented):
Significantly better!
Leaving what I hope to be final comments before merging.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/ForEachFilteringCriteria.java
Christopher-Chianelli: We already have node sharing tests; this predicate is cached inside `ConstraintNodeBuildHelper` since it returns different instances (as it depends on `ConsistencyTracker`). `ForEach` also wraps the predicate inside `ForEachFilteringCriteriaPredicateFunction` that implements equality/hashCode based on entity class + criteria.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
Christopher-Chianelli: Yep; solves a similar but different issue as "clearShadowVariablesListenerQueue", where consistency needs to be updated as `ConstraintVerifier` bypasses the variable listener API.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/EntityConsistencyState.java
Christopher-Chianelli: `null` is a valid value for a declarative variable, so it could be the case all declarative variables on an entity remained `null` because all their suppliers returned `null`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
Christopher-Chianelli: I rather keep it in the same file, since it is strongly coupled to this file/only used within the file.

triceo (approved):

PR #1779, opened 2025-09-03
chore: avoid removing iterators prematurely

zepfred:
This PR addresses an issue where a random iterator can be removed prematurely if the `hasNext` call returns false. The new approach returns an invalid move when the maximum number of attempts to generate a valid move is not reached.


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
```diff
@@ -26,15 +31,42 @@
  * The decorator returns a list of reachable values for a specific value.
  * It enables the creation of a filtering tier when using entity-provided value ranges,
  * ensuring only valid and reachable values are returned.
+ * A value is considered reachable to another value if both exist within their respective entity value ranges.
+ * <p>
+ * The decorator can only be applied to list variables.
+ * <p>
+ * <code>
  * <p>
```
triceo: Tags do not belong into preformatted code blocks. They will not be rendered as tags, they will be rendered literally as `<p>`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
```diff
@@ -205,40 +235,29 @@ private void initialize() {
...
+            return entityList != null && !entityList.isEmpty();
         }
 
         @Override
-        protected Object createUpcomingSelection() {
-            initialize();
+        public Object next() {
             if (entityList.isEmpty()) {
-                return noUpcomingSelection();
+                throw new NoSuchElementException();
```
triceo, line 260: It's unusual for selectors to throw exceptions, effectively killing the solver.
Why is this necessary and shouldn't there at least be a message describing the problem?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
```diff
@@ -4,35 +4,64 @@
...
+ * <p>
+ * The decorator can only be applied to list variables.
+ * <p>
+ * <code>
+ * <p>
  * e1 = entity_range[v1, v2, v3]
+ * <p>
  * e2 = entity_range[v1, v4]
- *
+ * <p>
```
triceo: Same comment.

triceo:
This needs a counterpart PR to Enterprise.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
zepfred: The IntelliJ doc preview does not show that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
zepfred: This will occur if the `next` call is made even after `hasNext` returns false. The logic from `UpcomingSelectionIterator` follows this approach.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java
triceo: Error:  /home/runner/work/timefold-solver/timefold-solver/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/decorator/FilteringValueRangeSelector.java:39: 
            error: block element not allowed within inline element <code>: p

zepfred:
The CI failure is related to the recent DSV changes.

PR #1794, opened 2025-09-09
test: add @PlanningId to concurrent testdata to allow multithreaded testing

Christopher-Chianelli:


triceo (approved):
Do we need this testdata in this repo? If not, let's move to enterprise. Otherwise LGTM.

Christopher-Chianelli:
> Do we need this testdata in this repo? If not, let's move to enterprise. Otherwise LGTM.

Yes; there are community tests that use `TestdataConcurrent`

PR #1795, opened 2025-09-09
perf: use a compact nonnegative int counter map to minimize memory usage for edge counts

Christopher-Chianelli:


triceo (approved):
Universally positive results.

PR #1802, opened 2025-09-10
chore: improve swap move selector approach

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -179,12 +171,20 @@ public Iterator<Object> iterator() {
...
+            // The value of 10 was chosen after some experiments with different values.
             // So we won't spend excessive time trying to generate a single move for the current selection.
             // If we are unable to generate a move, the move iterator can still be used in later iterations.
+            var maxBailoutSize = 10;
+            if (hasMoveFilter) {
+                // When a move filter is enabled,
+                // attempting to find a reachable value multiple times may be ineffective,
+                // as it could still be excluded by the move filter,
+                // leading to performance degradation.
+                maxBailoutSize = 1;
```
triceo: I don't understand why this should perform better. IMO it should perform exactly the same.

With bailout size == 1, the returned move will not be doable, therefore skipped by the solver. And the solver will iterate these non-doable moves until it randomly finds one that is doable.

In other words - exactly what would have happened with a larger bailout size. Only with the difference that larger bailout size would not generate any non-doable moves and only iterate internally, possibly being a bit more efficient.

zepfred: If a move filter is present, the solver might spend time locating a reachable entity, only to be filtered out by the move filter. I've run several experiments with the bailout size, and the value of 10 appears to improve quality without significantly impacting move speed.

Nonetheless, I changed the implementation to only fix the issue related to not using the correct replayed value during the iteration.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
```diff
@@ -445,10 +445,10 @@ void randomWithEntityValueRangeAndFiltering() {
         // and that causes the B destination to become an invalid destination
         assertCodesOfNeverEndingMoveSelector(moveSelector,
                 "1 {A[1]->A[1]}",
-                "3 {B[0]->A[0]}",
-                "3 {B[0]->A[0]}",
-                "3 {B[0]->A[2]}",
-                "3 {B[0]->A[1]}");
+                "3 {B[0]->B[0]}",
+                "3 {B[0]->B[0]}",
+                "3 {B[0]->B[0]}",
```
triceo: That is a lot of pointless moves generated.
I wonder if we should change the test to actually check that something productive is created.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/config/SolverProperty.java
```diff
@@ -52,6 +52,7 @@ public enum SolverProperty {
             value -> ConstraintStreamImplType.valueOf(value.toString())),
     CONSTRAINT_STREAM_AUTOMATIC_NODE_SHARING("constraint-stream-automatic-node-sharing",
             SolverProperties::setConstraintStreamAutomaticNodeSharing, value -> Boolean.valueOf(value.toString())),
+    RANDOM_SEED("random-seed", SolverProperties::setRandomSeed, value -> Long.parseLong(value.toString())),
```
triceo, line 55: Technically, these changes should have come in a different pull request. They are unrelated to move selectors.

triceo (approved):
Small comments, then LGTM.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
triceo: How sure are we of these supposed quality improvements?
We are always interested in those - but are they _real_? I have seen plenty of hit-and-miss situations with these PRs by now.

Specifically, the large bailouts exist so that a random number is never wasted. Example: if we have randomly selected an entity, we want to make sure the selection is used and therefore we will iterate over _a lot_ of values until we actually decide that there is nothing to do and we bail out. This is done so that entities aren't skipped. If we have low bailouts, it could happen that some entities would never be randomly selected, because their values would always bail out too early.

zepfred: I ran several experiments that show me the move speed is highly affected by the high bailout size, as the move still can be filtered out by the custom filter. The data suggest that the chosen bailout size improves the speed and quality of solutions, resulting in better medium scores. However, I have used only one medium dataset to validate that.

I understand that the entity can still be selected because the move iterator is never removed from the active iterators. The order in which we can select it has definitely changed, but I'm not so sure if the probability has changed.

One advantage of the list var logic is that we use the reachable values to iterate over the reachable values only. For basic vars, we iterate over all entities, whether reachable or not. I attempted to create a structure  `ReachableEntities` similar to `ReachableValues`, but it proved to be very resource-intensive, so I abandoned that idea. We observed during the list variable validation that using `ReachableValues` reduces move speed, but it is worthwhile as better solutions are produced.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/config/SolverProperty.java
zepfred: Okay. I'll remove the related changes.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
zepfred: The results are this way because the arrangement of the entity range is too restrictive. A small change in the value range of `A` completely alters the results.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -99,6 +99,10 @@ public boolean isValueReachable(Object origin, @Nullable Object otherValue) {
         return originItemValue.valueMap.containsKey(Objects.requireNonNull(otherValue));
     }
 
+    public boolean isAcceptsNullValue() {
```
triceo: Strange name. Why not just `acceptsNullValue`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -451,4 +544,28 @@ public Object next() {
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return !values.isEmpty();
+        }
+
+        @Override
+        public Object next() {
+            if (values.isEmpty()) {
```
triceo: Let's refer to `hasNext()` instead.

Issue #1429, opened 2025-02-27
ScoreAnalysis.summary() alternative that shows all constraints/matches etc


ge0ffrey:
Currently it only shows the top 3 or so.

For debugging purposes (User Experience), it's often handy to be able to show them all, or the top 20.
Not all cases have 700mb of matches.

Proposal A)

`ScoreAnalysis.summary(int topLimit)`

to get them all, use summary(Integer.MAX_VALUE)

simontiffert:
I was having this issue, too. 

My workaround to get the output as string with more constraint matches and indictments:
```        
ScoreExplanation<Schedule, HardMediumSoftScore> explanation = solutionManager.explain(solution);
String explanationString = DefaultScoreExplanation.explainScore(explanation.getScore(), explanation.getConstraintMatchTotalMap().values(), explanation.getIndictmentMap().values(), 200, 200);
```

PatrickDiallo23:
Do you mind if I also take a look on this issue if it hasn't already been implemented? Thanks!

triceo:
@PatrickDiallo23 Contributions are welcome!

PR #1807, opened 2025-09-12
feat: add ScoreAnalysis.summarize(int topLimit) method to customize constraint/matches display limit

PatrickDiallo23:
## Summary

This PR adds `ScoreAnalysis.summarize(int topLimit)` method as an alternative to `ScoreAnalysis.summarize()` alternative that shows all constraints/matches  (#1429). 

The method provides a concise summary of constraint matches, limiting the number of matches displayed per constraint to the specified topLimit (that must be > 0). If the number of matches exceeds the limit, the summary includes an ellipsis and a count of the remaining matches.

## Example Behavior

For `topLimit = 2`, the summary will display up to 2 matches per constraint and indicate the remaining count if more matches exist.
For` topLimit = Integer.MAX_VALUE`, all matches are displayed without truncation.

## Tests Added

- **Valid Limits**: Verifies correct behavior for various topLimit values, including edge cases like Integer.MAX_VALUE.
- **Invalid Limits**: Ensures exceptions are thrown for non-positive topLimit values.
- **Ellipsis Verification**: Confirms the summary correctly appends the ellipsis and remaining match count when the limit is exceeded.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -150,7 +150,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - uses: graalvm/setup-graalvm@7a1da54cb7fdef4ea19f6ecdfa9ecf59dc5a48fe # v1
+      - uses: graalvm/setup-graalvm@aba6a077d71fbfc02138d7470c4ad6e7f85bd2a9 # v1
         with:
           java-version: ${{matrix.java-version}}
           distribution: 'graalvm-community'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,8 +63,8 @@ jobs:
       # But unless something catastrophic happened, PR checks on source branch already ensured that all tests pass.
       - name: Set release version and build release
         run: |
-          mvn -Dfull versions:set -DnewVersion=${{ github.event.inputs.version }}
-          mvn -Dfull deploy -DskipTests=${{ github.event.inputs.dryRun }} -DaltDeploymentRepository=local::default::file://`pwd`/target/staging-deploy
+          ./mvnw -Dfull versions:set -DnewVersion=${{ github.event.inputs.version }}
+          ./mvnw -Dfull deploy -DskipTests=${{ github.event.inputs.dryRun }} -DaltDeploymentRepository=local::default::file://`pwd`/target/staging-deploy
           cp docs/target/antora-template.yml docs/src/antora.yml
           git add docs/src/antora.yml
           find . -name 'pom.xml' | xargs git add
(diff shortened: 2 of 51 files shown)


core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
```diff
@@ -300,6 +300,64 @@ Explanation of score (%s):
...
+     * Provides a summary of the solution's score analysis using the {@link #summarize(int)} method
+     * in a similar way to {@link #summarize()}.
+     * It is possible to specify the maximum number of constraint matches to display per constraint by passing
+     * the desired limit as an argument.
+     *
+     * @param topLimit maximum number of constraint matches to show per constraint.
+     *        Use {@link Integer#MAX_VALUE} to show all matches.
+     */
+    @SuppressWarnings("java:S3457")
+    public @NonNull String summarize(int topLimit) {
```
triceo, line 275: Let's not duplicate all of this code.
Surely the existing code can be extracted and generalized so that both the old method and this new method can reuse it.

The old method applies a pre-configured limit. It should not be too hard to implement that method with this method.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ConstraintAnalysis.java
```diff
@@ -234,6 +234,55 @@ Explanation of score (%s):
...
+     * Return a diagnostic text that explains part of the score quality using the {@link #summarize(int)} method
+     * in a similar way to {@link #summarize()}.
+     * It is possible to specify the maximum number of constraint matches to display per constraint by passing
+     * the desired limit as an argument.
+     *
+     * @param topLimit maximum number of constraint matches to display per constraint
+     *        Use {@link Integer#MAX_VALUE} to show all matches.
+     */
+
+    public @NonNull String summarize(int topLimit) {
```
triceo, line 216: Dtto.

triceo (changes requested):
Thanks for the contribution!
Will likely be good to go once the code duplication is eliminated.

PatrickDiallo23:
@triceo I refactored the code. Please take a look.

PatrickDiallo23:
I could also just let only the implementation of the summarize(int topLimit) since it can also cover the default one. What do you think?

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ConstraintAnalysis.java
```diff
@@ -200,6 +200,28 @@ public String constraintPackage() {
...
+        if (topLimit < 1) {
+            throw new IllegalArgumentException("The topLimit (" + topLimit + ") must be at least 1.");
```
triceo: We now use a new convention for exception messages with arguments.

    throw new IllegalArgumentException("The topLimit (%d) must be at least 1.".formatted(topLimit));

Provides better readability for long exception messages.
(Not so much for these short ones, but conventions are conventions.)

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
```diff
@@ -259,6 +259,28 @@ public Collection<ConstraintAnalysis<Score_>> constraintAnalyses() {
...
+        if (topLimit < 1) {
+            throw new IllegalArgumentException("topLimit (" + topLimit + ") must be positive.");
```
triceo: Same comment on exception message formatting.

triceo (approved):
LGTM when comments resolved!

PatrickDiallo23:
@triceo done

triceo:
Thanks for the contribution, @PatrickDiallo23!

Issue #1254, opened 2024-12-05
SolverConfigOverride should support setting the time spend directly (similar to SolverConfig itself)


ge0ffrey:
Before

```
solverFactory.buildSolver(new SolverConfigOverride<VehicleRoutePlan>()
                .withTerminationConfig(new TerminationConfig().withSpentLimit(...));
```

After

```
solverFactory.buildSolver(new SolverConfigOverride<VehicleRoutePlan>()
                .withTerminationSpentLimit(...)));
```

Same for unimproved time

PatrickDiallo23:
In this case, wouldn't it be better for **SolverConfigOverride** class to support all `with methods` from **SolverConfig** class to override more configs? Or should the focus be only on time spend logic?

triceo:
Overriding solver config means that the solver factory cannot pre-compute things, and therefore much more needs to happen at solver start, worsening the cold start time. For that reason, we only allow overrides for things that:

- We need to build at solver start anyway,
- somebody could actually need to override at solver start.

Termination fits both criteria. Different termination conditions for different kinds of runs (for example long run v. short run) make sense. Maybe things like environment mode (for regular runs v. asserted runs). But if anyone wants to override - say - phase configuration, or late acceptance settings, or constraint provider, then they are effectively creating a different solver config anyway, and might as well do it outright.

PatrickDiallo23:
I understand. Then, I will provide a solution that includes only termination related methods. Thanks for clarifying!

PR #1808, opened 2025-09-13
feat: add SolverConfigOverride support for setting the time spend directly

PatrickDiallo23:
## Summary

This PR aims to add support for setting the time spend directly, without relying on manually creating a `TerminationConfig` object (#1254).

Added `withTerminationUnimprovedSpentLimit` and `withTerminationSpentLimit` methods in the `SolverConfigOverride` class. Also, I added 2 tests to ensure code coverage and a proper functionality.

## Example

```
solverFactory.buildSolver(new SolverConfigOverride<VehicleRoutePlan>()
                .withTerminationSpentLimit(...)));

# OR

solverFactory.buildSolver(new SolverConfigOverride<VehicleRoutePlan>()
                .withTerminationUnimprovedSpentLimit(...)));
```

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
@@ -11,46 +11,17 @@ import org.jspecify.annotations.NonNull;
 /**
  * Includes settings to override default {@link ai.timefold.solver.core.api.solver.Solver} configuration.
  * <p>
- * This class provides an API to override solver termination settings. You can:
+ * This class provides an API to override solver termination settings. The following options are available:
  * <ul>
  * <li>Use {@link #withTerminationConfig(TerminationConfig)} to set a complete termination configuration</li>
  * <li>Use {@link #withTerminationSpentLimit(Duration)} and/or {@link #withTerminationUnimprovedSpentLimit(Duration)}
  * to set specific time limits independently or in combination</li>
  * </ul>
  * <p>
- * <strong>Important ordering constraint:</strong> If you use {@link #withTerminationConfig(TerminationConfig)},
+ * <strong>Important ordering constraint:</strong> If {@link #withTerminationConfig(TerminationConfig)} is used,
  * it must be called before any specific termination methods like {@link #withTerminationSpentLimit(Duration)}
  * or {@link #withTerminationUnimprovedSpentLimit(Duration)}. This prevents accidental override of previously
  * set specific configurations.
- * <p>
- * Valid usage patterns:
- * 
- * <pre>{@code
- * // Pattern 1: Use specific termination methods only
- * new SolverConfigOverride<MySolution>()
- *         .withTerminationSpentLimit(Duration.ofMinutes(5))
- *         .withTerminationUnimprovedSpentLimit(Duration.ofMinutes(2));
- *
- * // Pattern 2: Start with base config, then add specific settings
- * new SolverConfigOverride<MySolution>()
- *         .withTerminationConfig(new TerminationConfig())
- *         .withTerminationSpentLimit(Duration.ofMinutes(5));
- *
- * // Pattern 3: Use complete termination config with all settings
- * new SolverConfigOverride<MySolution>()
- *         .withTerminationConfig(new TerminationConfig()
- *                 .withSpentLimit(Duration.ofMinutes(5))
- *                 .withUnimprovedSpentLimit(Duration.ofMinutes(2)));
- * }</pre>
- * <p>
- * Invalid usage pattern (will throw IllegalStateException):
- * 
- * <pre>{@code
- * // DON'T DO THIS - withTerminationConfig() after specific methods
- * new SolverConfigOverride<MySolution>()
- *         .withTerminationSpentLimit(Duration.ofMinutes(5))
- *         .withTerminationConfig(new TerminationConfig()); // Will throw exception
- * }</pre>
  *
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
@@ -67,12 +38,7 @@ public final class SolverConfigOverride<Solution_> {
      * Sets the solver {@link TerminationConfig}, providing a base configuration that can be further
      * customized with specific termination methods.
      * <p>
-     * <strong>Important:</strong> This method must be called before any specific termination methods
-     * like {@link #withTerminationSpentLimit(Duration)} or {@link #withTerminationUnimprovedSpentLimit(Duration)}.
-     * Calling this method after specific termination settings have been applied will throw an exception
-     * to prevent accidental override of those settings.
-     * <p>
-     * After calling this method, you can chain additional specific termination methods to further
+     * After calling this method, additional specific termination methods can be chained to further
      * customize the configuration:
      * 
      * <pre>{@code
@@ -82,15 +48,30 @@ public final class SolverConfigOverride<Solution_> {
      *         .withTerminationUnimprovedSpentLimit(Duration.ofMinutes(2));
      * }</pre>
      *
+     * <p>
+     * <strong>Important:</strong> This method must be called before any specific termination methods
+     * like {@link #withTerminationSpentLimit(Duration)} or {@link #withTerminationUnimprovedSpentLimit(Duration)}.
+     * <p>
+     * Calling this method after specific termination settings have been applied will throw an exception
+     * to prevent accidental override of those settings.
+     *
+     *
+     * <pre>{@code
+     * new SolverConfigOverride<MySolution>()
+     *         .withTerminationSpentLimit(Duration.ofMinutes(5))
+     *         .withTerminationConfig(new TerminationConfig()); // Will throw exception
+     * }</pre>
+     *
      * @param terminationConfig allows overriding the default termination config of {@link Solver}
-     * @return this SolverConfigOverride instance for method chaining
+     * @return this
      * @throws IllegalStateException if specific termination settings have already been applied
      */
     public @NonNull SolverConfigOverride<Solution_> withTerminationConfig(@NonNull TerminationConfig terminationConfig) {
         if (hasSpecificTerminationSettings) {
-            throw new IllegalStateException("Cannot set terminationConfig after specific termination settings "
-                    + "(withTerminationSpentLimit or withTerminationUnimprovedSpentLimit) have been applied. "
-                    + "Call withTerminationConfig() first, or use the TerminationConfig builder methods directly.");
+            throw new IllegalStateException("""
+                    Cannot set terminationConfig after specific termination settings
+                    (withTerminationSpentLimit or withTerminationUnimprovedSpentLimit) have been applied.
+                    Maybe call withTerminationConfig() first, or use the TerminationConfig builder methods directly.""");
         }
         this.terminationConfig =
                 Objects.requireNonNull(terminationConfig, "Invalid terminationConfig (null) given to SolverConfigOverride.");
@@ -126,7 +107,7 @@ public final class SolverConfigOverride<Solution_> {
      * }</pre>
      *
      * @param spentLimit the maximum duration the solver is allowed to run
-     * @return this SolverConfigOverride instance for method chaining
+     * @return this
      */
     public @NonNull SolverConfigOverride<Solution_> withTerminationSpentLimit(@NonNull Duration spentLimit) {
         if (this.terminationConfig == null) {
@@ -171,7 +152,7 @@ public final class SolverConfigOverride<Solution_> {
      * }</pre>
      *
      * @param unimprovedSpentLimit the maximum duration the solver is allowed to run without improvement
-     * @return this SolverConfigOverride instance for method chaining
+     * @return this
      */
     public @NonNull SolverConfigOverride<Solution_>
             withTerminationUnimprovedSpentLimit(@NonNull Duration unimprovedSpentLimit) {


core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -31,4 +33,36 @@ public TerminationConfig getTerminationConfig() {
                 Objects.requireNonNull(terminationConfig, "Invalid terminationConfig (null) given to SolverConfigOverride.");
         return this;
     }
+
+    /**
+     * Sets a time limit for the solver to run, overriding the default termination configuration.
```
triceo: IMO the method Javadocs here should describe their mutual interactions.

What happens if I call `withTerminationConfig` before `withTerminationSpentLimit`?
What happens if I call it after?
What happens if I combine `withTerminationSpentLimit` and `withTerminationUnimprovedSpentLimit`?

Think about the answers to those questions. Are these interactions what you would intuitively expect to happen? Are these interactions something that the user would want?

Curious to hear your answers.

triceo (commented):
Thanks for the contribution!
I've asked some questions in the comments inline, let's have a conversation about this design.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
PatrickDiallo23: From my point of view, when I provided this solution, I thought that `withTerminationConfig`, `withTerminationSpentLimit`, `withTerminationUnimprovedSpentLimit` are methods that can be used separately. 

As a dev I would use the first method if I want to provide a comprehensive and a more complex Termination Configuration. For the other 2 methods, I would use them if I want to provide a simple and a quickly config. In my cases, I used most of the time the second method to override the spent limit based on the end user needs.

I never used combination of these methods, but even if someone wants to do combine them I would say that:
1) To combine `withTerminationSpentLimit` and `withTerminationUnimprovedSpentLimit` would be safe (I assume) because we check for the instance of the Termination Config and it will only be set the needed parameter
2) To call `withTerminationConfig` before `withTerminationSpentLimit` would be the best choice because we want to first create the Termination Config then override some variables if needed. The other option, would be just weird IMHO.

Please let me know what do you think.

triceo: When designing APIs, don't just think about what you want to do - think about what the API _allows_ people to do. 

> The other option, would be just weird IMHO.

Exactly. Yet we allow them to do this - and believe me, if something is physically possible, somebody somewhere will do it. Should we document this behavior? Should we make it impossible by throwing an exception? Or is the problem so big that the new API no longer makes sense because of it?

You're designing the API, you should consider these problems, and pick the best answer. It's not just a matter of introducing the methods - you always have to consider how they work together. So - how _should_ this API behave?

PatrickDiallo23: I haven't thought like this before. Quite challenging indeed. Initially I tried to keep the same functionality as in `SolverConfig` class, but it seems that it wasn't a good idea to only reproduce it in the `SolverConfigOverride` class. I will try to provide an answer and please let me know if it is a feasible and good direction for Timefold.

So, I would do the following:

1) Allow users to combine `withTerminationSpentLimit` and `withTerminationUnimprovedSpentLimit` with proper documentation in the comments (and also in the Timefold documentation if needed)
2) Since `withTerminationConfig` contains most of the methods related to Termination, I would prohibit (by throwing an exception and proper documenting the usage of the method) to chain this method with `withTerminationSpentLimit` like this: `withConfigOverride(new SolverConfigOverride<Timetable>() .withTerminationConfig(new TerminationConfig()).withTerminationSpentLimit(problem.getDuration())` because we can do it like this: `withConfigOverride(new SolverConfigOverride<Timetable>() .withTerminationConfig(new TerminationConfig().withTerminationSpentLimit(problem.getDuration()))` (subtle difference but I think it matters).
3) Allow users to use the methods independently (only use `withTerminationUnimprovedSpentLimit` for example) if they want.

I don't know if it makes sense, but please let me know if my approach is good and if I should consider other scenarios. Thanks!

triceo: Wrt. 1 and 3 - I agree.

Wrt. your second point, I think it will be easier if we allow this:

    withConfigOverride(
        new SolverConfigOverride<Timetable>()
            .withTerminationConfig(...)
            .withTerminationSpentLimit(problem.getDuration())

They can provide whatever starting configuration they want. However, we should not allow this:

    withConfigOverride(
        new SolverConfigOverride<Timetable>()
            .withTerminationSpentLimit(problem.getDuration())
            .withTerminationConfig(...)

As that would potentially override what they previously said.

Side note: With these improvements on the way, I question if `withTerminationConfig(...)` should have been created in the first place. It allows for the most flexibility, but also creates this slightly confusing API where calling method X possibly invalidates Y, but not vice versa. Let this be a lesson on thinking forward when designing APIs.

PatrickDiallo23:
@triceo I updated the API as we discussed. I tried to provide a comprehensive documentation so that it will be easier for developers to understand how to use the methods. Please let me know if I used a good approach and if I should also add other unit tests (and where: in the other modified file or I should create SolverConfigOverrideTest class)?

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
+ * new SolverConfigOverride<MySolution>()
+ *         .withTerminationConfig(new TerminationConfig()
+ *                 .withSpentLimit(Duration.ofMinutes(5))
+ *                 .withUnimprovedSpentLimit(Duration.ofMinutes(2)));
+ * }</pre>
+ * <p>
+ * Invalid usage pattern (will throw IllegalStateException):
+ * 
+ * <pre>{@code
+ * // DON'T DO THIS - withTerminationConfig() after specific methods
```
triceo: I'd remove this line. It's enough to say that this pattern is invalid, and we generally don't use capitals to shout at users.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
+ * // Pattern 3: Use complete termination config with all settings
+ * new SolverConfigOverride<MySolution>()
+ *         .withTerminationConfig(new TerminationConfig()
+ *                 .withSpentLimit(Duration.ofMinutes(5))
+ *                 .withUnimprovedSpentLimit(Duration.ofMinutes(2)));
+ * }</pre>
+ * <p>
+ * Invalid usage pattern (will throw IllegalStateException):
+ * 
+ * <pre>{@code
```
triceo: Does this pattern actually work? (I mean "does it render correctly?").
IMO just `<pre>` should work, `@code` seems excessive.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
-     * Sets the solver {@link TerminationConfig}.
+     * Sets the solver {@link TerminationConfig}, providing a base configuration that can be further
+     * customized with specific termination methods.
+     * <p>
+     * <strong>Important:</strong> This method must be called before any specific termination methods
+     * like {@link #withTerminationSpentLimit(Duration)} or {@link #withTerminationUnimprovedSpentLimit(Duration)}.
+     * Calling this method after specific termination settings have been applied will throw an exception
+     * to prevent accidental override of those settings.
+     * <p>
+     * After calling this method, you can chain additional specific termination methods to further
```
triceo:
```suggestion
     * After calling this method, additional specific termination methods can be chained to further
```

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
+ * <li>Use {@link #withTerminationSpentLimit(Duration)} and/or {@link #withTerminationUnimprovedSpentLimit(Duration)}
+ * to set specific time limits independently or in combination</li>
+ * </ul>
+ * <p>
+ * <strong>Important ordering constraint:</strong> If you use {@link #withTerminationConfig(TerminationConfig)},
+ * it must be called before any specific termination methods like {@link #withTerminationSpentLimit(Duration)}
+ * or {@link #withTerminationUnimprovedSpentLimit(Duration)}. This prevents accidental override of previously
+ * set specific configurations.
+ * <p>
+ * Valid usage patterns:
```
triceo: I wonder if we need the examples here, if we already mention them near the methods. One of the two places seems unnecessary.

I think we should remove these usage examples from the methods.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
+     * <pre>{@code
+     * new SolverConfigOverride<MySolution>()
+     *         .withTerminationConfig(new TerminationConfig())
+     *         .withTerminationSpentLimit(Duration.ofMinutes(5))
+     *         .withTerminationUnimprovedSpentLimit(Duration.ofMinutes(2));
+     * }</pre>
      *
      * @param terminationConfig allows overriding the default termination config of {@link Solver}
-     * @return this
+     * @return this SolverConfigOverride instance for method chaining
```
triceo: Typically we just do:

```suggestion
     * @return this
```

The type is obvious from the Javadoc, and the chaining bit was already mentioned.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
  * Includes settings to override default {@link ai.timefold.solver.core.api.solver.Solver} configuration.
+ * <p>
+ * This class provides an API to override solver termination settings. You can:
+ * <ul>
+ * <li>Use {@link #withTerminationConfig(TerminationConfig)} to set a complete termination configuration</li>
+ * <li>Use {@link #withTerminationSpentLimit(Duration)} and/or {@link #withTerminationUnimprovedSpentLimit(Duration)}
+ * to set specific time limits independently or in combination</li>
+ * </ul>
+ * <p>
+ * <strong>Important ordering constraint:</strong> If you use {@link #withTerminationConfig(TerminationConfig)},
```
triceo: We generally don't address the user directly. ("You")

```suggestion
 * <strong>Important ordering constraint:</strong> If {@link #withTerminationConfig(TerminationConfig)} is used,
```

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
```diff
@@ -4,65 +4,182 @@
...
-     * @return this
+     * @return this SolverConfigOverride instance for method chaining
+     * @throws IllegalStateException if specific termination settings have already been applied
      */
     public @NonNull SolverConfigOverride<Solution_> withTerminationConfig(@NonNull TerminationConfig terminationConfig) {
+        if (hasSpecificTerminationSettings) {
+            throw new IllegalStateException("Cannot set terminationConfig after specific termination settings "
+                    + "(withTerminationSpentLimit or withTerminationUnimprovedSpentLimit) have been applied. "
+                    + "Call withTerminationConfig() first, or use the TerminationConfig builder methods directly.");
+        }
```
triceo, line 75: You can use a multi-line comment. The last statement typically ends up on its own line.


```suggestion
        if (hasSpecificTerminationSettings) {
            throw new IllegalStateException("""
                     Cannot set terminationConfig after specific termination settings (withTerminationSpentLimit or withTerminationUnimprovedSpentLimit) have been applied.
                     Maybe call withTerminationConfig() first, or use the TerminationConfig builder methods directly.""");
        }
```

This suggestion is likely improperly formatted, I did it manually. Please run source formatting before pushing the changes.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
PatrickDiallo23: In my case it is working properly like this:

```
// Coding
// Comment
```

If I delete this then is would look like this:

// Coding
// Comment

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
PatrickDiallo23: I am not sure if I understood correctly: do you mean to keep the usage examples only at the class level, right?

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
triceo: OK. If the Javadoc compiles like this, I'm good.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverConfigOverride.java
triceo: Sorry, my post is confusing. I mean to keep the examples at the method level only.

PatrickDiallo23:
@triceo I made the changes. I hope everything is better now.

triceo:
Thank you, Patrick!

PR #1811, opened 2025-09-15
feat: Constraint Streams memoizedFilter()

triceo:


triceo:
@larsbeck I'll wait with merging for your feedback. Please don't let this wait long.

triceo:
Will be replaced by a better solution.

PR #1813, opened 2025-09-15
chore: Migrate internal variable listeners to new API

Christopher-Chianelli:
New API summary

- No before/after entity added/removed events; instead, resetWorkingSolution initializes the entities.

- ProblemChanges addEntity/removeEntity now reset the working solution; this was needed anyways, since declarative shadow variables assume no entities are added/removed.

- Single class, InnerVariableListener, with a beforeChange/afterChange method. InnerVariableListener is generic in the event type, which is either a BasicVariableChangeEvent or ListVariableChangeEvent.

- User VariableListeners are wrapped inside a LegacyCustomShadowVariable...VariableListener that delegates to the variable listener.


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/custom/LegacyCustomShadowVariableBasicVariableListener.java
```diff
@@ -0,0 +1,78 @@
...
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public record LegacyCustomShadowVariableBasicVariableListener<Solution_>(
+        Class<?>[] sourceEntityClasses,
+        VariableListener<Solution_, Object> customVariableListener)
+        implements
+            InnerVariableListener<Solution_, BasicVariableChangeEvent<Object>> {
+
+    private boolean handlesSourceEntity(Object sourceEntity) {
```
triceo: This bothers me. Var listeners are supposed to be fast, and this being called in every before/after is a problem.

- If we make this check in before, why do this in after as well and not reuse the result instead?
- And if we only trigger the custom var listener if this check passes, why create/trigger the before method at all?

It seems to me as if the listener logic should have excluded calling these listeners even before the code gets to here.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/custom/LegacyCustomShadowVariableListVariableListener.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public ChangeEventType listenedEventType() {
+        return ChangeEventType.LIST;
+    }
+
+    @Override
+    public void beforeChange(InnerScoreDirector<Solution_, ?> scoreDirector,
+            ListVariableChangeEvent<Entity_, Element_> event) {
+        if (event instanceof ListElementsChangeEvent<Entity_, Element_> changeEvent) {
```
triceo: Similar comment as with the basic listener. If we know this is not the case, why are we even triggering the code? IMO this should have been handled earlier.

core/src/main/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirector.java
```diff
@@ -22,19 +22,19 @@ public DefaultProblemChangeDirector(InnerScoreDirector<Solution_, ?> scoreDirect
     public <Entity> void addEntity(@NonNull Entity entity, @NonNull Consumer<Entity> entityConsumer) {
         Objects.requireNonNull(entity, () -> "Entity (" + entity + ") cannot be null.");
         Objects.requireNonNull(entityConsumer, () -> "Entity consumer (" + entityConsumer + ") cannot be null.");
-        scoreDirector.beforeEntityAdded(entity);
         entityConsumer.accept(entity);
-        scoreDirector.afterEntityAdded(entity);
+        // Rebuild the shadow variable graph and update all shadows
+        scoreDirector.setWorkingSolution(scoreDirector.getWorkingSolution());
```
triceo: setWorkingSolution is super expensive. I wonder if we should instead just store the information that it needs to happen, and do it later. Think of code that adds/removes multiple entities - it will be very inefficient.

(I don't recall if that's already been the case before this PR. Maybe it was. But perhaps we have a way to improve this now.)

Also, if we're already doing a solution reset, doing all the other before/after method calls is pointless.

core/src/main/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirector.java
```diff
@@ -22,19 +22,19 @@ public DefaultProblemChangeDirector(InnerScoreDirector<Solution_, ?> scoreDirect
...
     @Override
     public <Entity> void removeEntity(@NonNull Entity entity, @NonNull Consumer<Entity> entityConsumer) {
         Objects.requireNonNull(entity, () -> "Entity (" + entity + ") cannot be null.");
         Objects.requireNonNull(entityConsumer, () -> "Entity consumer (" + entityConsumer + ") cannot be null.");
         var workingEntity = lookUpWorkingObjectOrFail(entity);
-        scoreDirector.beforeEntityRemoved(workingEntity);
         entityConsumer.accept(workingEntity);
-        scoreDirector.afterEntityRemoved(workingEntity);
+        // Rebuild the shadow variable graph and update all shadows
+        scoreDirector.setWorkingSolution(scoreDirector.getWorkingSolution());
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/ShadowVariableDescriptor.java
```diff
@@ -53,7 +52,7 @@ public void setGlobalShadowOrder(int globalShadowOrder) {
      */
     public abstract List<VariableDescriptor<Solution_>> getSourceVariableDescriptorList();
 
-    public abstract Collection<Class<? extends AbstractVariableListener>> getVariableListenerClasses();
+    public abstract Collection<String> getVariableListenerClassNames();
```
triceo: I wonder if this change is necessary.

It seems to me much more useful to have the class directly, as opposed to having a name of it. I'd prefer reverting this change.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/anchor/AnchorVariableListener.java
```diff
@@ -1,18 +1,21 @@
...
-import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullMarked;
 
 /**
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
-public class AnchorVariableListener<Solution_> implements VariableListener<Solution_, Object>, AnchorVariableSupply {
+@NullMarked
+public class AnchorVariableListener<Solution_, Entity_, Value_> implements
+        InnerVariableListener<Solution_, BasicVariableChangeEvent<Entity_>>, AnchorVariableSupply {
```
triceo: I wonder if there should be a super-type for basic vars (and another for list vars), which would allow to share some of these methods between different implementations.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/violation/ShadowVariableSnapshot.java
```diff
@@ -32,9 +32,7 @@ void validate(Consumer<String> violationMessageConsumer) {
                     + ")'s corrupted value (" + originalValue + ") changed to uncorrupted value (" + newValue
                     + ") after all variable listeners were triggered without changes to the genuine variables.\n"
                     + "      Maybe one of the listeners ("
-                    + shadowVariableDescriptor.getVariableListenerClasses().stream()
-                            .map(Class::getSimpleName)
-                            .collect(Collectors.toList())
+                    + shadowVariableDescriptor.getVariableListenerClassNames()
```
triceo: If we're touching these exceptions, let's convert them to the new (much more readable) style of

    """
        Something %s. 
        Maybe ...""".formatted(...);

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ChangeEvent.java
```diff
@@ -0,0 +1,4 @@
+package ai.timefold.solver.core.impl.domain.variable;
+
+public interface ChangeEvent {
```
triceo: Can we not seal this?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ChangeEventType.java
```diff
@@ -0,0 +1,6 @@
+package ai.timefold.solver.core.impl.domain.variable;
+
+public enum ChangeEventType {
```
triceo, line 3: If we only have two event types and they are sealed, I wonder if we need this enum.
Instead, we could just reference the two `ChangeEvent` subclasses.
Seems like duplication.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolutionManager.java
```diff
@@ -76,7 +77,11 @@ private <Result_> Result_ callScoreDirector(Solution_ solution,
             if (isShadowVariableUpdateEnabled) {
                 scoreDirector.setWorkingSolution(nonNullSolution);
             } else {
+                // Some notificables resetWorkingSolution update shadow variables
```
triceo: notifiables

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupply.java
```diff
@@ -86,13 +88,15 @@ public interface ListVariableStateSupply<Solution_> extends
      * @param element never null
      * @return null if the element is the first element in the list
      */
+    @Nullable
     Object getPreviousElement(Object element);
```
triceo: If we already have the entity/value generics, why not use them here?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupply.java
```diff
@@ -86,13 +88,15 @@ public interface ListVariableStateSupply<Solution_> extends
...
 
     /**
      *
      * @param element never null
      * @return null if the element is the last element in the list
      */
+    @Nullable
     Object getNextElement(Object element);
```
triceo: Dtto.

triceo (commented):
Leaving comments on the chosen design, mostly.

I appreciate the bold decision to drop the before/after entity events. It creates some additional considerations though, specifically around performance. resetWorkingSolution is super-heavy, and as many entities are added/removed in bulk, triggering the reset for each is going to present a problem.

Speaking of performance, the new listeners are now conditional; only executed if a certain condition is met. My bet is that this will represent another place where we'll see a performance drop - did you do any benchmarks? Before we merge this, we have to. Var listeners are used heavily throughout the solver, and in some cases there present a considerable portion of the CPU time; we can not afford regressions here.

It also seems we need an enterprise counterpart to this. Finally, please rebase this on latest main. Especially after this is merged later this week: https://github.com/TimefoldAI/timefold-solver/pull/1802.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/custom/LegacyCustomShadowVariableListVariableListener.java
Christopher-Chianelli: That because `ListVariableChangeEvent` is a sealed interface with two implementations (`ElementUnassigned` and `ListElementsChangeEvent`); Required since ListVariableListener handles both sub list change and element unassigned with different parameters and methods. In Java 21, it can happen with a switch.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ChangeEventType.java
Christopher-Chianelli: Needed since we cannot read the generics of instances.

core/src/test/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirectorTest.java
```diff
@@ -22,20 +22,20 @@ class DefaultProblemChangeDirectorTest {
...
     void complexProblemChange_correctlyNotifiesScoreDirector() {
-        final TestdataLavishEntityGroup entityGroupOne = new TestdataLavishEntityGroup("entityGroupOne");
-        final TestdataLavishValueGroup valueGroupOne = new TestdataLavishValueGroup("valueGroupOne");
-        final TestdataLavishEntity addedEntity = new TestdataLavishEntity("newly added entity", entityGroupOne);
-        final TestdataLavishEntity removedEntity = new TestdataLavishEntity("entity to remove", entityGroupOne);
-        final TestdataLavishValue addedFact = new TestdataLavishValue("newly added fact", valueGroupOne);
-        final TestdataLavishValue removedFact = new TestdataLavishValue("fact to remove", valueGroupOne);
-        final TestdataLavishEntity changedEntity = new TestdataLavishEntity("changed entity", entityGroupOne);
-        final TestdataLavishValue changedEntityValue = new TestdataLavishValue("changed entity value", valueGroupOne);
+        final var entityGroupOne = new TestdataLavishEntityGroup("entityGroupOne");
```
triceo: `final` should not be necessary here; these variables are all effectively final anyway.

core/src/test/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirectorTest.java
```diff
@@ -81,4 +81,125 @@ void complexProblemChange_correctlyNotifiesScoreDirector() {
         verify(scoreDirectorMock, times(1)).beforeProblemFactRemoved(removedFact);
         verify(scoreDirectorMock, times(1)).afterProblemFactRemoved(removedFact);
     }
+
+    @Test
+    void verify_noResetSolutionIfNoEntitiesAddedOrRemoved() {
+        final var entityGroupOne = new TestdataLavishEntityGroup("entityGroupOne");
```
triceo: Same in this method.

core/src/test/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirectorTest.java
```diff
@@ -81,4 +81,125 @@ void complexProblemChange_correctlyNotifiesScoreDirector() {
...
+        var testdataSolution = TestdataLavishSolution.generateSolution();
+        problemChange.doChange(testdataSolution, defaultProblemChangeDirector);
+
+        verify(scoreDirectorMock, times(0)).setWorkingSolution(any());
+        verify(scoreDirectorMock, times(1)).triggerVariableListeners();
+    }
+
+    @Test
+    void verify_ResetSolutionIfEntitiesAddedOrRemoved() {
+        final var entityGroupOne = new TestdataLavishEntityGroup("entityGroupOne");
```
triceo: And here.

core/src/test/java/ai/timefold/solver/core/impl/solver/change/DefaultProblemChangeDirectorTest.java
```diff
@@ -81,4 +81,125 @@ void complexProblemChange_correctlyNotifiesScoreDirector() {
...
+        testdataSolution = TestdataLavishSolution.generateSolution();
+        removeProblemChange.doChange(testdataSolution, defaultProblemChangeDirector);
+
+        verify(scoreDirectorMock, times(2)).setWorkingSolution(any());
+        verify(scoreDirectorMock, times(0)).triggerVariableListeners();
+    }
+
+    @Test
+    void verify_ResetSolutionOnceIfEntitiesAddedOrRemovedThenVariablesChanged() {
+        final var entityGroupOne = new TestdataLavishEntityGroup("entityGroupOne");
```
triceo: And finally here.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -64,6 +65,46 @@ public void initialize(InnerScoreDirector<Solution_, ?> scoreDirector, int initi
...
+        if (!scoreDirector.expectShadowVariablesInCorrectState() && (externalizedIndexProcessor != null ||
+                externalizedInverseProcessor != null ||
+                externalizedPreviousElementProcessor != null ||
+                externalizedNextElementProcessor != null)) {
+            // If the elements have any shadows, set them to null if no entity has their values
+            // We do not want to do this eagerly, since variable listener events are not triggered.
+            scoreDirector.getSolutionDescriptor().visitEntitiesByEntityClass(
+                    scoreDirector.getWorkingSolution(), sourceVariableDescriptor.getElementType(),
+                    element -> {
+                        MutableInt found = new MutableInt(0);
```
triceo: Since you only use values 0 and 1, this could be `AtomicBoolean`.
JDK devs even recognized that that's what people often want to use these atomic classes for, and added `setPlain(...)` on them - a way to treat them just as regular non-atomic mutable carriers.

triceo: Also, `var`?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -64,6 +65,46 @@ public void initialize(InnerScoreDirector<Solution_, ?> scoreDirector, int initi
...
+                        scoreDirector.getSolutionDescriptor().visitEntitiesByEntityClass(
+                                scoreDirector.getWorkingSolution(),
+                                sourceVariableDescriptor.getEntityDescriptor().getEntityClass(),
+                                entity -> {
+                                    for (var entityElement : sourceVariableDescriptor.getValue(entity)) {
```
triceo: This seems inefficient. We continue iterating even though `found` is already `1`. This is caused by the unfortunate internal iteration of `visitEntitiesByEntityClass`, which is not a good fit here. I recommend refactoring to standard external iteration, which can then be correctly shortcut.

triceo (approved):
LGTM after the minor comments are resolved.
I'm assuming that the actual deprecations and docs changes (if any) are going to come in another PR?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
Christopher-Chianelli: We return true though when found is set, which from the Javadoc of `visitEntitiesByEntityClass` stops iteration.
The outer `visitEntitiesByEntityClass` need to visit all elements, since we don't know if any of them are stale (and each iteration of the outer `visitEntitiesByEntityClass` resets found to false). The inner `visitEntitiesByEntityClass` can exit early.

Christopher-Chianelli:
@triceo correct; deprecation and docs will be in a separate PR.

Christopher-Chianelli:
It appears `SolverManagerTest.solveWithTerminationUnimprovedSpentLimit` might be a flaky test (that, or a cosmic ray caused a bit flip in a runner).

PR #1827, opened 2025-09-22
ci: upgrade to JDK 25

triceo:
When this works, upgrade Enterprise and Quickstarts too.


triceo:
Mac version is still not available:
https://github.com/adoptium/temurin/issues/96

(I confirm that on my own M3 MBP.)

zepfred (approved):

PR #1828, opened 2025-09-22
chore: clean up naming and accessibility in Move Streams

triceo:


core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -23,6 +17,49 @@ public interface UniDataStream<Solution_, A> extends DataStream<Solution_> {
...
+    /**
+     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}, with the array being empty.
+     */
+    @SuppressWarnings("unchecked")
+    default <B> BiDataStream<Solution_, A, B> join(UniDataStream<Solution_, B> otherStream) {
+        return join(otherStream, new BiDataJoiner[0]);
+    }
+
+    /**
+     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}
```
zepfred:
```suggestion
     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}.
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -23,6 +17,49 @@ public interface UniDataStream<Solution_, A> extends DataStream<Solution_> {
...
+     */
+    @SuppressWarnings("unchecked")
+    default <B> BiDataStream<Solution_, A, B> join(UniDataStream<Solution_, B> otherStream, BiDataJoiner<A, B> joiner) {
+        return join(otherStream, new BiDataJoiner[] { joiner });
+    }
+
+    /**
+     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}
```
zepfred:
```suggestion
     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}.
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -23,6 +17,49 @@ public interface UniDataStream<Solution_, A> extends DataStream<Solution_> {
...
+     */
+    @SuppressWarnings("unchecked")
+    default <B> BiDataStream<Solution_, A, B> join(UniDataStream<Solution_, B> otherStream, BiDataJoiner<A, B> joiner1,
+            BiDataJoiner<A, B> joiner2) {
+        return join(otherStream, new BiDataJoiner[] { joiner1, joiner2 });
+    }
+
+    /**
+     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}
```
zepfred:
```suggestion
     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}.
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/UniDataStream.java
```diff
@@ -23,6 +17,49 @@ public interface UniDataStream<Solution_, A> extends DataStream<Solution_> {
...
+     */
+    @SuppressWarnings("unchecked")
+    default <B> BiDataStream<Solution_, A, B> join(UniDataStream<Solution_, B> otherStream, BiDataJoiner<A, B> joiner1,
+            BiDataJoiner<A, B> joiner2, BiDataJoiner<A, B> joiner3) {
+        return join(otherStream, new BiDataJoiner[] { joiner1, joiner2, joiner3 });
+    }
+
+    /**
+     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}
```
zepfred:
```suggestion
     * As defined by {@link #join(UniDataStream, BiDataJoiner[])}.
```

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/MoveStreamFactory.java
```diff
@@ -35,33 +35,33 @@ public interface MoveStreamFactory<Solution_> {
...
      * Start a {@link DataStream} of all instances of the sourceClass
      * that are known as {@link ProblemFactCollectionProperty problem facts}
      * or {@link PlanningEntity planning entities}.
      * If the sourceClass is a genuine or shadow entity,
      * it returns instances regardless of their pinning status.
-     * Otherwise as defined by {@link #enumerate(Class, boolean)}.
+     * Otherwise as defined by {@link #forEach(Class, boolean)}.
      */
-    <A> UniDataStream<Solution_, A> enumerateIncludingPinned(Class<A> sourceClass, boolean includeNull);
+    <A> UniDataStream<Solution_, A> forEachUnfiltered(Class<A> sourceClass, boolean includeNull);
```
zepfred, line 51: I wonder if we should keep the term "IncludingPinned." It would make sense without needing to read the documentation.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/generic/Moves.java
```diff
@@ -0,0 +1,96 @@
...
+import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningListVariableMetaModel;
+import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningVariableMetaModel;
+import ai.timefold.solver.core.preview.api.domain.metamodel.PositionInList;
+import ai.timefold.solver.core.preview.api.move.Move;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public final class Moves {
```
zepfred, line 16: The name seems incomplete. Perhaps we should call it `MoveOperation`,  `MoveUtil`, etc.

zepfred: This class is very helpful. I sometimes get confused about importing the correct move class.

triceo: Short doesn't mean incomplete. :-)

We have Joiners, ConstraintCollectors.
JDK has Collections, Files, ...

IMO this name is good.

core/src/main/java/ai/timefold/solver/core/impl/move/streams/maybeapi/stream/MoveStreamFactory.java
triceo: `forEachUnfiltered` is the bare minimum that needs to exist regardless of what is actually being removed by `forEach`. Any other methods will be introduced when we see if we need them.

PR #1830, opened 2025-09-24
chore: deprecate variable listeners in Javadoc

### LINKED ISSUE #1524: Deprecate Variable Listeners
Convert all internal uses of Variable Listeners to the new mechanism, then deprecate Variable Listeners for removal.

Christopher-Chianelli:
Note: variable listeners were already deprecated in user docs.


triceo (approved):
Very happy to see this! LGTM when green.

(Please provide the PR to fix the flaky test. If the test is for something that's not in public API, then I suggest we remove both.)

PR #1831, opened 2025-09-25
test: remove race condition by making both branches return same value in SolverManagerTest

Christopher-Chianelli:
There was a race condition in SolverManagerTest#solveWithTerminationUnimprovedSpentLimit. It calls UnimprovedTimeMillisSpentTermination#calculateTimeGradient, which returns 0.0 if the current phase does not send best solution events, or solve duration since last best solution/limit if it does. The test was relying on finishing before the Solver got to LocalSearchPhase, which is unreliable. To fix it, we mock the clocks the Solver uses, so it always see solve duration since last best solution as 0.


triceo (approved):
Since we already test the effects of the override in another test in this class, I'd say we should limit this test to just making sure that the `with...` method correctly sets the termination; effectively, that the getter will return whatever the wither did.

LGTM after that, thanks for looking into this.

PR #1833, opened 2025-09-25
feat: enable sorting for CH with list variable

zepfred:


zepfred:
@triceo I noticed other factories used by configurations and annotations that are not located in the public API. Maybe, we should handle that in a separate PR as well.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,31 @@
+package ai.timefold.solver.core.api.domain.common;
+
+import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.impl.heuristic.move.Move;
+import ai.timefold.solver.core.impl.heuristic.selector.Selector;
+
+/**
+ * Creates a weight to decide the order of a collections of selections
+ * (a selection is a {@link PlanningEntity}, a planningValue, a {@link Move} or a {@link Selector}).
```
triceo, line 17: Where do we weight selectors?
(And if selectors are not public API, should this be mentioning them?)

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * The selections are then sorted by their weight,
+ * normally ascending unless it's configured descending.
```
triceo, line 19: Configured how, where?

core/src/main/java/ai/timefold/solver/core/api/domain/common/SorterWeightFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * <p>
+ * Implementations are expected to be stateless.
+ * The solver may choose to reuse instances.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the selection type
+ */
+@FunctionalInterface
+public interface SorterWeightFactory<Solution_, T> {
```
triceo: `@NullMarked`?

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningListVariable.java
```diff
@@ -55,5 +57,37 @@
...
 
-    // TODO value comparison: https://issues.redhat.com/browse/PLANNER-2542
+    /**
+     * Allows a collection of planning values for this variable to be sorted by strength.
+     * A strengthWeight estimates how strong a planning value is.
+     * Some algorithms benefit from planning on weaker planning values first or from focusing on them.
+     * <p>
+     * The {@link Comparator} should sort in ascending strength.
+     * For example: sorting 3 computers on strength based on their RAM capacity:
+     * Computer B (1GB RAM), Computer A (2GB RAM), Computer C (7GB RAM),
```
triceo: Not a typical use for a list variable, is it?
Let's do VRP instead.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/SelectionSorterWeightFactory.java
```diff
@@ -18,14 +19,8 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @param <T> the selection type
  */
+@Deprecated(forRemoval = true, since = "1.27.0")
```
triceo: I believe that when something is deprecated, Javadoc requires a `@deprecated` tag as well.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -35,7 +38,36 @@ public DestinationSelector<Solution_> buildDestinationSelector(HeuristicConfigPo
...
+            entitySelectorConfig.setSorterManner(configPolicy.getEntitySorterManner());
+        }
+        if (entityValueRangeRecorderId != null && entitySelectorConfig.getSelectionOrder() != null
+                && entitySelectorConfig.getSelectionOrder() == SelectionOrder.SORTED) {
+            // Sorting entities is not permitted when using an entity value range,
+            // as the list of reachable entities is only generated after selecting a value.
+            // This process prevents sorting and caching at the phase level.
+            throw new IllegalStateException("""
+                    The destination selector cannot to sort the entity list when an entity value range is used.
+                    Maybe remove the setting "EntitySorterManner" from the phase config.
```
triceo:
```suggestion
                    The destination selector cannot sort the entity list when an entity value range is used.
                    Maybe remove "EntitySorterManner" from the phase config.
```

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -35,7 +38,36 @@ public DestinationSelector<Solution_> buildDestinationSelector(HeuristicConfigPo
...
+                && entitySelectorConfig.getSorterManner() == null) {
+            entitySelectorConfig.setCacheType(SelectionCacheType.PHASE);
+            entitySelectorConfig.setSelectionOrder(SelectionOrder.SORTED);
+            entitySelectorConfig.setSorterManner(configPolicy.getEntitySorterManner());
+        }
+        if (entityValueRangeRecorderId != null && entitySelectorConfig.getSelectionOrder() != null
+                && entitySelectorConfig.getSelectionOrder() == SelectionOrder.SORTED) {
+            // Sorting entities is not permitted when using an entity value range,
+            // as the list of reachable entities is only generated after selecting a value.
+            // This process prevents sorting and caching at the phase level.
```
triceo: See, this makes no sense.
We generally don't allow entities to be added or removed while the solver is running. Therefore once the entity list is sorted, it will never change after that. Therefore it _should_ be possible to use value ranges.

Also... this whole code is a bit strange. First we _decide_ to set cache to PHASE, and _then_ we fail because we don't support that? Then why are we setting the PHASE cache in the first place?

triceo (commented):
As always, the different selector combinations are causing me to ask questions - see inline.

Also, please provide a migration recipe for the move of the interface to the public API; the uses of the old interface should be replaced by the new one.

Finally, the change needs to show up in the upgrade recipe in our docs.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
zepfred: The sorter is created by the `WeightFactorySelectionSorter`, which is utilized by `ValueSelectorFactory` and `EntitySelectorFactory` when creating the sorter nodes `SortingValueSelector` and `SortingEntitySelector`, respectively.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
zepfred: The `sortManner` field in the value and entity selector configuration specifies how the data will be sorted. There is also a setting called `constructionHeuristicType` that automatically configures how entities and values are sorted.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
zepfred: > We generally don't allow entities to be added or removed while the solver is running. Therefore once the entity list is sorted, it will never change after that. Therefore it should be possible to use value ranges.

We cannot sort the list at the phase level when using the selector framework because using entity-range for the list variable creates the node `FilteringEntityByValueSelector`. This node needs the left value selector to pick a value first, so only reachable entities are fetched. That's a limitation when using the new filtering nodes.

> Also... this whole code is a bit strange. First we decide to set cache to PHASE, and then we fail because we don't support that? Then why are we setting the PHASE cache in the first place?

I can relocate the check elsewhere because this is a new fail-fast that wasn't present earlier. You're right about having too many selector configurations, and this is one of them.

zepfred: Looking at the logic from `ValueSelectorFactory`, it is clear that we can sort the data at the step level when using entity-range. I've decided to go with a similar approach.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
triceo: Don't tell me, tell the users. :-)

zepfred:
> Finally, the change needs to show up in the upgrade recipe in our docs.

I have included a recipe for automatic migration to the new interface. Do we need to add anything to our docs?

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,46 @@
...
+@NullMarked
+@FunctionalInterface
+public interface SorterFactory<Solution_, T> {
+
+    /**
+     * @param solution never null, the {@link PlanningSolution} to which the selection belongs or applies to
+     * @param selection never null, a {@link PlanningEntity}, a planningValue, a {@link Move} or a {@link Selector}
+     * @return never null, for example a {@link Integer}, {@link Double} or a more complex {@link Comparable}
+     */
+    Comparable createSorter(Solution_ solution, T selection);
```
triceo: Shouldn't this be `Comparable<T>`?

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -79,17 +79,17 @@ interface NullDifficultyComparator extends Comparator {
...
-     * The {@link SelectionSorterWeightFactory} alternative for {@link #difficultyComparatorClass()}.
+     * The {@link SorterFactory} alternative for {@link #difficultyComparatorClass()}.
      * <p>
      * Do not use together with {@link #difficultyComparatorClass()}.
      *
      * @return {@link NullDifficultyWeightFactory} when it is null (workaround for annotation limitation)
      * @see #difficultyComparatorClass()
      */
-    Class<? extends SelectionSorterWeightFactory> difficultyWeightFactoryClass() default NullDifficultyWeightFactory.class;
+    Class<? extends SorterFactory> difficultyWeightFactoryClass() default NullDifficultyWeightFactory.class;
```
triceo: I know this PR is not done yet, so just so we don't forget: this needs to be deprecated and replaced by the comparator factory. (Also, should the SorterFactory be renamed to ComparatorFactory?)

core/src/build/revapi-differences.json
triceo, line 1: Please remember these changes and remove them before the PR is final.
These changes will be completely different when the PR is finished.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningVariable.java
```diff
@@ -78,27 +79,64 @@
...
+     * Do not use together with {@link #comparatorFactoryClass()}.
+     *
+     * @return {@link NullComparator} when it is null (workaround for annotation limitation)
+     * @see #comparatorFactoryClass()
+     */
+    Class<? extends Comparator> comparatorClass() default NullComparator.class;
+
     /** Workaround for annotation limitation in {@link #strengthComparatorClass()}. */
-    interface NullStrengthComparator extends Comparator {
+    interface NullStrengthComparator extends NullComparator {
```
triceo: Does this need to exist? Why not just `NullComparator`?

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningVariable.java
```diff
@@ -78,27 +79,64 @@
...
+     * @see #comparatorClass()
+     */
+    Class<? extends SorterFactory> comparatorFactoryClass() default NullComparatorFactory.class;
+
     /** Workaround for annotation limitation in {@link #strengthWeightFactoryClass()}. */
-    interface NullStrengthWeightFactory extends SelectionSorterWeightFactory {
+    interface NullStrengthWeightFactory extends NullComparatorFactory {
+    }
+
+    interface NullComparatorFactory<Solution_, T> extends SelectionSorterWeightFactory<Solution_, T> {
```
triceo: This is confusing. How can `NullComparatorFactory` be extending `SelectionSorterWeightFactory`? Shouldn't it extend `ComparatorFactory`?

core/src/main/java/ai/timefold/solver/core/api/domain/common/SorterFactory.java
```diff
@@ -0,0 +1,46 @@
...
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the selection type
+ * 
+ * @see ValueSelectorConfig
+ * @see EntitySelectorConfig
+ * @see ConstructionHeuristicPhaseConfig
+ */
+@NullMarked
+@FunctionalInterface
+public interface SorterFactory<Solution_, T> {
```
triceo: I wonder... shouldn't this be `ComparatorFactory`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
```diff
@@ -35,7 +37,23 @@ public DestinationSelector<Solution_> buildDestinationSelector(HeuristicConfigPo
...
+                && configPolicy.getEntitySorterManner() != NONE;
+        var entityDescriptor = deduceEntityDescriptor(configPolicy, entitySelectorConfig.getEntityClass());
+        var hasDifficultySorter = entityDescriptor.getDecreasingDifficultySorter() != null;
+        if (hasSortManner && hasDifficultySorter && entitySelectorConfig.getSorterManner() == null) {
+            if (entityValueRangeRecorderId == null) {
+                // Solution-range model
+                entitySelectorConfig.setCacheType(SelectionCacheType.PHASE);
+            } else {
+                // Entity-range model requires the sorting to be done in each step
+                entitySelectorConfig.setCacheType(SelectionCacheType.STEP);
```
triceo, line 52: The range never changes, why would the sort have to change?

core/src/test/java/ai/timefold/solver/core/testdomain/common/TestSortableFactory.java
```diff
@@ -0,0 +1,17 @@
+package ai.timefold.solver.core.testdomain.common;
+
+import ai.timefold.solver.core.api.domain.common.SorterFactory;
+import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionSorterWeightFactory;
+
+public class TestSortableFactory
+        implements SelectionSorterWeightFactory<Object, TestSortableObject>, SorterFactory<Object, TestSortableObject> {
```
triceo: Do we need to test legacy interfaces? I wouldn't.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -1600,15 +1600,15 @@ The solver may choose to reuse them in different contexts.
 
 
 [#sortedSelectionBySelectionSorterWeightFactory]
```
triceo: ID no longer valid.
That said, maybe we want the old IDs to still work; add both ids?

triceo:
As I familiarize myself with the code, I added comments for things I noticed. It's likely that this PR is not yet finished, so you're probably already planning to resolve some of these. That said, the PR is not a draft, therefore I am actively looking at it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
zepfred: ~~Makes sense. This change would apply only for value selectors.~~

zepfred: The entity list might change when using entity value-range because the list of reachable entities can differ. Therefore, we need to choose a different cache type, or the list of values could become inconsistent.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
zepfred: ~~No, it could be any class that implements `Comparable`. We have tests that return an Integer.~~

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningVariable.java
zepfred: If any user is using it, then it wouldn't be completely backward compatible to rename the class. This setting is valid:
`strengthComparatorClass = PlanningVariable.NullStrengthComparator.class,`

zepfred:
@triceo we include the property `sorterWeightFactoryClass` in `EntitySelectorConfig`, `ValueSelectorConfig`, and `MoveSelectorConfig`. Should we deprecate it and introduce a new property called `sorterComparatorFactoryClass`?

The property `sorterWeightFactoryClass` is already of type `ComparatorFactory`.

triceo:
> @triceo we include the property `sorterWeightFactoryClass` in `EntitySelectorConfig`, `ValueSelectorConfig`, and `MoveSelectorConfig`. Should we deprecate it and introduce a new property called `sorterComparatorFactoryClass`?

Yes, the old names should be deprecated. Code which uses them should still work, but the code which combines the old names and the new names should fail fast - it's either all the old names, or all the new names.

Arguably `sorterComparatorFactoryClass` - why not just `ComparatorFactoryClass`? I'm not sure.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
zepfred: There is no reference to this ID, so it can be safely renamed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/DestinationSelectorFactory.java
triceo: The comment should say that, because it's not at all obvious.

core/src/test/java/ai/timefold/solver/core/testdomain/common/TestSortableFactory.java
zepfred: The goal is to have test coverage.

zepfred:
> > @triceo we include the property `sorterWeightFactoryClass` in `EntitySelectorConfig`, `ValueSelectorConfig`, and `MoveSelectorConfig`. Should we deprecate it and introduce a new property called `sorterComparatorFactoryClass`?
> 
> Yes, the old names should be deprecated. Code which uses them should still work, but the code which combines the old names and the new names should fail fast - it's either all the old names, or all the new names.
> 
> Arguably `sorterComparatorFactoryClass` - why not just `ComparatorFactoryClass`? I'm not sure.

We already have `sorterComparatorClass` of type `Comparator`

benchmark/src/main/resources/benchmark.xsd
```diff
@@ -1085,6 +1088,9 @@
           <xs:element minOccurs="0" name="sorterWeightFactoryClass" type="xs:string"/>
                               
           
+          <xs:element minOccurs="0" name="sorterComparatorFactoryClass" type="xs:string"/>
```
triceo: Two questions:

- If there is a sorterComparatorFactoryClass, shouldn't there also be a sorterComparatorClass?
- Why the "sorter" prefix? Why not just "comparatorFactoryClass"?

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,46 @@
...
+@NullMarked
+@FunctionalInterface
+public interface ComparatorFactory<Solution_, T> {
+
+    /**
+     * @param solution never null, the {@link PlanningSolution} to which the selection belongs or applies to
+     * @param selection never null, a {@link PlanningEntity}, a planningValue, a {@link Move} or a {@link Selector}
+     * @return never null, for example a {@link Integer}, {@link Double} or a more complex {@link Comparable}
+     */
+    Comparable createSorter(Solution_ solution, T selection);
```
triceo: `Comparable` has a generic type. Let's not do raw types in public API anymore.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,46 @@
...
+ * @see EntitySelectorConfig
+ * @see ConstructionHeuristicPhaseConfig
+ */
+@NullMarked
+@FunctionalInterface
+public interface ComparatorFactory<Solution_, T> {
+
+    /**
+     * @param solution never null, the {@link PlanningSolution} to which the selection belongs or applies to
+     * @param selection never null, a {@link PlanningEntity}, a planningValue, a {@link Move} or a {@link Selector}
```
triceo: Do we need to reference the selectors? They're not public API, and I'm not even sure why a user would want to sort selectors.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -69,27 +69,74 @@ interface NullPinningFilter extends PinningFilter {
...
+
+    /**
+     * The {@link ComparatorFactory} alternative for {@link #comparatorClass()}.
+     * <p>
+     * Do not use together with {@link #comparatorClass()}.
+     *
+     * @return {@link NullComparatorFactory} when it is null (workaround for annotation limitation)
+     * @see #comparatorClass()
+     */
+    Class<? extends ComparatorFactory> comparatorFactoryClass() default NullComparatorFactory.class;
```
triceo: This is the correct name. So why does the XSD and Revapi show the ones with the "sorter" prefix?

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -69,27 +69,74 @@ interface NullPinningFilter extends PinningFilter {
      * <p>
      * Do not use together with {@link #difficultyWeightFactoryClass()}.
      *
+     * @deprecated Deprecated in favor of {@link #comparatorClass()}.
+     *
      * @return {@link NullDifficultyComparator} when it is null (workaround for annotation limitation)
      * @see #difficultyWeightFactoryClass()
      */
+    @Deprecated(forRemoval = true, since = "1.28.0")
     Class<? extends Comparator> difficultyComparatorClass() default NullDifficultyComparator.class;
```
triceo, line 113: Let's move the deprecated stuff to the end. We want to see the active stuff first.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningVariable.java
```diff
@@ -78,27 +78,74 @@
      * <p>
      * Do not use together with {@link #strengthWeightFactoryClass()}.
      *
+     * @deprecated Deprecated in favor of {@link #comparatorClass()}.
+     *
      * @return {@link NullStrengthComparator} when it is null (workaround for annotation limitation)
      * @see #strengthWeightFactoryClass()
      */
+    @Deprecated(forRemoval = true, since = "1.28.0")
     Class<? extends Comparator> strengthComparatorClass() default NullStrengthComparator.class;
```
triceo, line 122: Please move deprecated stuff to the end.

core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/ConstructionHeuristicType.java
```diff
@@ -57,44 +57,22 @@ public enum ConstructionHeuristicType {
...
-        }
+        return switch (this) {
+            case FIRST_FIT, WEAKEST_FIT, STRONGEST_FIT -> EntitySorterManner.NONE;
+            case FIRST_FIT_DECREASING, WEAKEST_FIT_DECREASING, STRONGEST_FIT_DECREASING -> EntitySorterManner.DESCENDING;
+            case ALLOCATE_ENTITY_FROM_QUEUE, ALLOCATE_TO_VALUE_FROM_QUEUE, CHEAPEST_INSERTION, ALLOCATE_FROM_POOL ->
+                EntitySorterManner.DESCENDING_IF_AVAILABLE;
+        };
     }
 
     public @NonNull ValueSorterManner getDefaultValueSorterManner() {
```
triceo: Let's replace `@NonNull` references with `@NullMarked` - less boilerplate.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
```diff
@@ -34,6 +34,7 @@
         "sorterManner",
         "sorterComparatorClass",
         "sorterWeightFactoryClass",
+        "sorterComparatorFactoryClass",
```
triceo: Ah, this is where it comes from! Let's standardize and shorten it to "comparatorClass" and "comparatorFactoryClass".

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
```diff
@@ -246,12 +269,23 @@ public void setSelectedCountLimit(@Nullable Long selectedCountLimit) {
         return this;
     }
 
+    /**
+     * @deprecated Deprecated in favor of {@link #withSorterComparatorFactoryClass(Class)}
+     * @param weightFactoryClass the factory class
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
     public @NonNull EntitySelectorConfig
```
triceo: Same comment on `@NonNull`. `@NullMarked` is a preferred alternative and if we're already touching the code, let's no longer introduce `@NonNull` if we can do `@NullMarked`.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -68,6 +68,7 @@
         "filterClass",
         "sorterComparatorClass",
         "sorterWeightFactoryClass",
+        "sorterComparatorFactoryClass",
```
triceo: Same comment on naming.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/MoveSelectorConfig.java
```diff
@@ -201,12 +224,23 @@ public void setFixedProbabilityWeight(@Nullable Double fixedProbabilityWeight) {
...
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
     public @NonNull Config_ withSorterWeightFactoryClass(
-            @NonNull Class<? extends SelectionSorterWeightFactory> sorterWeightFactoryClass) {
+            @NonNull Class<? extends ComparatorFactory> sorterWeightFactoryClass) {
         this.sorterWeightFactoryClass = sorterWeightFactoryClass;
         return (Config_) this;
     }
 
+    public @NonNull Config_
```
triceo: Same comment on `@NonNull`.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/value/ValueSelectorConfig.java
```diff
@@ -35,6 +35,7 @@
         "sorterManner",
         "sorterComparatorClass",
         "sorterWeightFactoryClass",
+        "sorterComparatorFactoryClass",
```
triceo: Naming.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/value/ValueSelectorConfig.java
```diff
@@ -257,12 +280,23 @@ public void setSelectedCountLimit(@Nullable Long selectedCountLimit) {
...
     }
 
+    /**
+     * @deprecated Deprecated in favor of {@link #withSorterComparatorFactoryClass(Class)}
+     * @param weightFactoryClass the factory class
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
     public @NonNull ValueSelectorConfig
-            withSorterWeightFactoryClass(@NonNull Class<? extends SelectionSorterWeightFactory> weightFactoryClass) {
+            withSorterWeightFactoryClass(@NonNull Class<? extends ComparatorFactory> weightFactoryClass) {
```
triceo: `@NonNull`

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/FactorySelectionSorter.java
```diff
@@ -71,13 +72,13 @@ public boolean equals(Object other) {
             return true;
         if (other == null || getClass() != other.getClass())
             return false;
-        WeightFactorySelectionSorter<?, ?> that = (WeightFactorySelectionSorter<?, ?>) other;
-        return Objects.equals(selectionSorterWeightFactory, that.selectionSorterWeightFactory)
-                && Objects.equals(appliedWeightComparator, that.appliedWeightComparator);
+        var that = (FactorySelectionSorter<?, ?>) other;
```
triceo: Arguably we have a better pattern for `equals` now:

    return (other instanceof ... that) && Objects.equal(...) &&  Objects.equal(...);

Effectively, it becomes a one-liner.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/SelectionSorterWeightFactory.java
```diff
@@ -15,17 +18,22 @@
...
+ * @deprecated Deprecated in favor of {@link ComparatorFactory}.
+ * 
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  * @param <T> the selection type
  */
-@FunctionalInterface
-public interface SelectionSorterWeightFactory<Solution_, T> {
+@Deprecated(forRemoval = true, since = "1.28.0")
+@NullMarked
+public interface SelectionSorterWeightFactory<Solution_, T> extends ComparatorFactory<Solution_, T> {
```
triceo, line 28: No need to improve deprecated interfaces. Keep it as it was.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
```diff
@@ -182,6 +184,28 @@ protected boolean isBaseInherentlyCached() {
         return true;
     }
 
+    private static String determineSorterComparatorFactoryPropertyName(EntitySelectorConfig entitySelectorConfig) {
```
triceo: "Sorter" again.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
```diff
@@ -182,6 +184,28 @@ protected boolean isBaseInherentlyCached() {
...
+            throw new IllegalArgumentException(
+                    "The entitySelectorConfig (%s) cannot have a %s (%s) and %s (%s) at the same time.".formatted(
+                            entitySelectorConfig, "sorterWeightFactoryClass", weightFactoryClass,
+                            "sorterComparatorFactoryClass", comparatorFactoryClass));
+        }
+        return weightFactoryClass != null ? "sorterWeightFactoryClass" : "sorterComparatorFactoryClass";
+    }
+
+    private static Class<? extends ComparatorFactory>
+            determineSorterComparatorFactoryClass(EntitySelectorConfig entitySelectorConfig) {
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/EntitySelectorFactory.java
```diff
@@ -251,29 +275,35 @@ private EntitySelector<Solution_> applyFiltering(EntitySelector<Solution_> entit
     protected void validateSorting(SelectionOrder resolvedSelectionOrder) {
         var sorterManner = config.getSorterManner();
         var sorterComparatorClass = config.getSorterComparatorClass();
-        var sorterWeightFactoryClass = config.getSorterWeightFactoryClass();
+        var sorterComparatorFactoryPropertyName = determineSorterComparatorFactoryPropertyName(config);
+        var sorterComparatorFactoryClass = determineSorterComparatorFactoryClass(config);
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/AbstractMoveSelectorFactory.java
```diff
@@ -149,36 +171,37 @@ private MoveSelector<Solution_> applyFiltering(MoveSelector<Solution_> moveSelec
     }
 
     protected void validateSorting(SelectionOrder resolvedSelectionOrder) {
-        if ((config.getSorterComparatorClass() != null || config.getSorterWeightFactoryClass() != null
+        var sorterComparatorFactoryPropertyName = determineSorterComparatorFactoryPropertyName(config);
+        var sorterComparatorFactoryClass = determineSorterComparatorFactoryClass(config);
```
triceo: "Sorter"

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/AbstractMoveSelectorFactory.java
```diff
@@ -187,24 +210,26 @@ protected MoveSelector<Solution_> applySorting(SelectionCacheType resolvedCacheT
         if (resolvedSelectionOrder == SelectionOrder.SORTED) {
             SelectionSorter<Solution_, Move<Solution_>> sorter;
             var sorterComparatorClass = config.getSorterComparatorClass();
-            var sorterWeightFactoryClass = config.getSorterWeightFactoryClass();
+            var sorterComparatorFactoryPropertyName = determineSorterComparatorFactoryPropertyName(config);
+            var sorterComparatorFactoryClass = determineSorterComparatorFactoryClass(config);
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -216,6 +216,28 @@ protected boolean determineBaseRandomSelection(GenuineVariableDescriptor<Solutio
         };
     }
 
+    private static String determineSorterComparatorFactoryPropertyName(ValueSelectorConfig valueSelectorConfig) {
```
triceo: "Sorter"

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -216,6 +216,28 @@ protected boolean determineBaseRandomSelection(GenuineVariableDescriptor<Solutio
...
+        if (weightFactoryClass != null && comparatorFactoryClass != null) {
+            throw new IllegalArgumentException(
+                    "The valueSelectorConfig (%s) cannot have a %s (%s) and %s (%s) at the same time.".formatted(
+                            valueSelectorConfig, "sorterWeightFactoryClass", weightFactoryClass,
+                            "sorterComparatorFactoryClass", comparatorFactoryClass));
+        }
+        return weightFactoryClass != null ? "sorterWeightFactoryClass" : "sorterComparatorFactoryClass";
+    }
+
+    private static Class<? extends ComparatorFactory>
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -272,29 +294,34 @@ protected ValueSelector<Solution_> applyInitializedChainedValueFilter(HeuristicC
     protected void validateSorting(SelectionOrder resolvedSelectionOrder) {
         var sorterManner = config.getSorterManner();
         var sorterComparatorClass = config.getSorterComparatorClass();
-        var sorterWeightFactoryClass = config.getSorterWeightFactoryClass();
+        var sorterComparatorFactoryPropertyName = determineSorterComparatorFactoryPropertyName(config);
+        var sorterComparatorFactoryClass = determineSorterComparatorFactoryClass(config);
```
triceo: Dtto.

core/src/test/java/ai/timefold/solver/core/testdomain/valuerange/sort/comparator/oldapproach/TestdataOldSortableEntityProvidingEntity.java
```diff
@@ -0,0 +1,53 @@
+package ai.timefold.solver.core.testdomain.valuerange.sort.comparator.oldapproach;
```
triceo: I don't like the naming "newapproach" and "oldapproach". Once we remove deprecated stuff, there will only be "newapproach" and future generations will ask what the "oldapproach" was. Do we even need a different package for this?

migration/src/main/resources/META-INF/rewrite/ToLatest.yml
```diff
@@ -34,5 +34,96 @@ recipeList:
   - ai.timefold.solver.migration.v8.AsConstraintRecipe
   - ai.timefold.solver.migration.v8.RemoveConstraintPackageRecipe
   - ai.timefold.solver.migration.v8.SolutionManagerRecommendAssignmentRecipe
+  - org.openrewrite.java.ChangeMethodName:
```
triceo: I'd put this into a single migration recipe, and subsequently add tests for that recipe. This is very fragile, easily broken.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -1587,7 +1588,7 @@ You'll also need to configure it (unless it's annotated on the domain model and
     <entitySelector>
       <cacheType>PHASE</cacheType>
       <selectionOrder>SORTED</selectionOrder>
-      <sorterComparatorClass>...VisitDifficultyComparator</sorterComparatorClass>
+      <sorterComparatorClass>...VisitComparator</sorterComparatorClass>
```
triceo: Remember to rename this throughout the docs as well.

docs/src/modules/ROOT/pages/optimization-algorithms/overview.adoc
```diff
@@ -1599,16 +1600,16 @@ The solver may choose to reuse them in different contexts.
 ====
 
 
-[#sortedSelectionBySelectionSorterWeightFactory]
-===== Sorted selection by `SelectionSorterWeightFactory`
+[#sortedSelectionByComparatorFactory]
```
triceo, line 1603: Arguably we want to keep the old IDs as well, or else the old URLs will break.
I vaguely recall that you can have two IDs for the same section; but maybe I'm wrong, in which case what you do here is fine.

migration/src/main/resources/META-INF/rewrite/ToLatest.yml
triceo: You can perhaps even test the migration script on ESS? ESS uses the old API for sorting, if I remember correctly.

benchmark/src/main/resources/benchmark.xsd
zepfred: > If there is a sorterComparatorFactoryClass, shouldn't there also be a sorterComparatorClass?

The configuration already includes `sorterComparatorClass`. That's the reason I chose the name `sorterComparatorFactoryClass` to keep the names consistent.

> Why the "sorter" prefix? Why not just "comparatorFactoryClass"?

I can't explain why we added the prefix in the past, but using `sorterComparatorFactoryClass` aligns with the field `sorterComparatorClass`.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
zepfred: The model annotations do not include the "sorter" prefix, but the selector and move configurations do. The XSD files are generated from these config classes, which is why the prefix is included.

zepfred: IMO, the prefix makes things clear in the configuration classes, and I believe we should keep them. However, model annotations do not include the prefix. Should we add it there?

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
zepfred: I'm unsure about removing the prefix because it clearly indicates that it's related to sorting.

benchmark/src/main/resources/benchmark.xsd
triceo: I'd rather we deprecate `sorterComparatorClass` in favor of `comparatorClass`. That way, the naming here will be consistent with the naming on the annotations.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
zepfred: I understand your point. The issue is that `SelectionSorterWeightFactory` extends this class, and the generic types in `ComparatorFactory` cannot be changed or it will break user code. Updating the method definition like that will work, but the child classes will still get a warning.

`<V extends Comparable<V>> V createSorter(Solution_ solution, T selection);`

The following code is possible, but we will still receive a unchecked overriding warning:

```
        ComparatorFactory<TestdataSolution, TestdataEntity> weightFactory = new ComparatorFactory<>() {
            @Override
            public Integer createSorter(TestdataSolution solution, TestdataEntity selection) {
                return Integer
                        .valueOf(selection.getCode().charAt(0));
            }
        };
```

Another problem is that we can't use lambda anymore:

`ComparatorFactory<TestdataSolution, TestdataEntity> weightFactory = (solution, selection) -> Integer
                .valueOf(selection.getCode().charAt(0));`

triceo: Then the solution would be for `SelectionSorterWeightFactory` to not extend this class. The solver would have to create internal adaptors, but that's arguably not a big deal.

The new code should be as clean as possible. The deprecated code should not make the new code worse.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,47 @@
...
+ * @param <T> the selection type
+ * @param <V> the returning type
+ *
+ * @see ValueSelectorConfig
+ * @see EntitySelectorConfig
+ * @see ConstructionHeuristicPhaseConfig
+ */
+@NullMarked
+@FunctionalInterface
+public interface ComparatorFactory<Solution_, T, V extends Comparable<V>> {
```
triceo: `XFactory` should create `X`, therefore this should be `ComparableFactory`. Otherwise it is confusing.

But I wonder... why doesn't this return a `Comparator`? The interface could then be simpler:

    interface ComparatorFactory<Solution_, T> {


        Comparator<T> createComparator(Solution_ solution); 

    }

Note that this requires no third generic argument, making implementations easier on the user.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * @param solution never null, the {@link PlanningSolution} to which the selection belongs or applies to
+     * @param selection never null, a {@link PlanningEntity}, a planningValue, a {@link Move} or a {@link Selector}
+     * @return never null, for example a {@link Integer}, {@link Double} or a more complex {@link Comparable}
+     */
+    V createSorter(Solution_ solution, T selection);
```
triceo: Regardless of how the interface declaration looks, the method name is not correct. `createComparator` or `createComparable`.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -69,27 +101,44 @@ interface NullPinningFilter extends PinningFilter {
...
+     *
+     * @deprecated Deprecated in favor of {@link NullComparator}.
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
+    interface NullDifficultyComparator<T> extends NullComparator<T> {
     }
 
     /**
-     * The {@link SelectionSorterWeightFactory} alternative for {@link #difficultyComparatorClass()}.
+     * The {@link ComparatorFactory} alternative for {@link #difficultyComparatorClass()}.
```
triceo: This Javadoc is not correct.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/PlanningVariable.java
```diff
@@ -78,27 +110,43 @@
...
+     *
+     * @deprecated Deprecated in favor of {@link NullComparator}.
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
+    interface NullStrengthComparator<T> extends NullComparator<T> {
     }
 
     /**
-     * The {@link SelectionSorterWeightFactory} alternative for {@link #strengthComparatorClass()}.
+     * The {@link ComparatorFactory} alternative for {@link #strengthComparatorClass()}.
```
triceo: Also incorrect.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -263,35 +266,74 @@ private void processMovable(PlanningEntity entityAnnotation) {
...
+        if (selectedComparatorClass != null && selectedComparatorFactoryClass != null) {
+            throw new IllegalStateException("The entityClass (%s) cannot have a %s (%s) and a %s (%s) at the same time."
+                    .formatted(entityClass, selectedComparatorPropertyName, selectedComparatorClass.getName(),
+                            selectedComparatorFactoryPropertyName, selectedComparatorFactoryClass.getName()));
+        }
+        if (selectedComparatorClass != null) {
+            var comparator = ConfigUtils.newInstance(this::toString, selectedComparatorPropertyName, selectedComparatorClass);
+            descendingSorter = new ComparatorSelectionSorter<>(comparator, SelectionSorterOrder.DESCENDING);
+        }
+        if (selectedComparatorFactoryClass != null) {
```
triceo: Shouldn't this be `else if`? Both conditions cannot be true at the same time, but this code makes it seem like they can.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/BasicVariableDescriptor.java
```diff
@@ -41,8 +43,61 @@ protected void processPropertyAnnotations(DescriptorPolicy descriptorPolicy) {
...
+        }
+        // Selected settings
+        var selectedComparatorPropertyName = "comparatorClass";
+        var selectedComparatorClass = comparatorClass;
+        var selectedComparatorFactoryPropertyName = "comparatorFactoryClass";
+        Class<?> selectedComparatorFactoryClass = comparatorFactoryClass;
+        if (strengthComparatorClass != null) {
+            selectedComparatorPropertyName = "strengthComparatorClass";
+            selectedComparatorClass = strengthComparatorClass;
+        }
```
triceo, line 94: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/GenuineVariableDescriptor.java
```diff
@@ -155,36 +157,39 @@ private ValueRangeDescriptor<Solution_> buildValueRangeDescriptor(DescriptorPoli
...
                     SelectionSorterOrder.DESCENDING);
         }
-        if (strengthWeightFactoryClass != null) {
-            SelectionSorterWeightFactory<Solution_, Object> strengthWeightFactory = newInstance(this::toString,
-                    "strengthWeightFactoryClass", strengthWeightFactoryClass);
-            increasingStrengthSorter = new WeightFactorySelectionSorter<>(strengthWeightFactory,
-                    SelectionSorterOrder.ASCENDING);
-            decreasingStrengthSorter = new WeightFactorySelectionSorter<>(strengthWeightFactory,
-                    SelectionSorterOrder.DESCENDING);
+        if (comparatorFactoryClass != null) {
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/ComparatorFactoryAdapter.java
```diff
@@ -0,0 +1,17 @@
...
+public class ComparatorFactoryAdapter<Solution_, T, V extends Comparable<V>> implements ComparatorFactory<Solution_, T, V> {
+
+    private final SelectionSorterWeightFactory<Solution_, T> sorterWeightFactory;
+
+    public ComparatorFactoryAdapter(SelectionSorterWeightFactory<Solution_, T> sorterWeightFactory) {
+        this.sorterWeightFactory = sorterWeightFactory;
+    }
+
+    @Override
+    public V createSorter(Solution_ solution, T selection) {
```
triceo: Same comment as with the `ComparatorFactory` - doesn't create a sorter, so it shouldn't be called as if it does.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/FactorySelectionSorter.java
```diff
@@ -0,0 +1,82 @@
...
+                this.appliedComparator = Collections.reverseOrder();
+                break;
+            default:
+                throw new IllegalStateException(
+                        "The selectionSorterOrder (%s) is not implemented.".formatted(selectionSorterOrder));
+        }
+    }
+
+    @Override
+    public void sort(ScoreDirector<Solution_> scoreDirector, List<T> selectionList) {
```
triceo: Seems unnecessary. It propagates the `ScoreDirector` type to an entirely unrelated package of classes, only for the calling code to avoid saying `getWorkingSolution()`. IMO should be removed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/FactorySelectionSorter.java
```diff
@@ -0,0 +1,82 @@
...
+        sort(scoreDirector.getWorkingSolution(), selectionList);
+    }
+
+    /**
+     * @param solution never null, the {@link PlanningSolution} to which the selections belong or apply to
+     * @param selectionList never null, a {@link List}
+     *        of {@link PlanningEntity}, planningValue, {@link Move} or {@link Selector}
+     */
+    public void sort(Solution_ solution, List<T> selectionList) {
```
triceo: This should IMO be the main (only?) method on the `SelectionSorter` interface.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/MoveSelectorFactoryTest.java
```diff
@@ -201,15 +202,58 @@ void applySorting_withoutAnySortingClass() {
 
     @Test
     void applySorting_withSorterComparatorClass() {
-        final MoveSelector<TestdataSolution> baseMoveSelector = SelectorTestUtils.mockMoveSelector();
-        DummyMoveSelectorConfig moveSelectorConfig = new DummyMoveSelectorConfig();
-        moveSelectorConfig.setSorterOrder(SelectionSorterOrder.ASCENDING);
-        moveSelectorConfig.setSorterComparatorClass(DummyComparator.class);
+        // Old setting
+        {
```
triceo: Please don't do this.
Different things deserve different test methods, if not even different test classes.
You will appreciate this in a few months' time, when we do Solver 2.0 and you're going to have to remove all the old code.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -372,35 +372,36 @@ As Timefold Solver will mutate the <<planningVariable,genuine>> or <<shadowVaria
 As Java's `enum` and `record` types are immutable by design, these can not be used as planning entities.
 
 
-[#planningEntityDifficulty]
-=== Planning entity difficulty
+[#planningEntitySorting]
+=== [[planningEntityDifficulty]]Planning entity sorting
```
triceo, line 376: I've never seen this syntax before. Have you tried both anchors in the deployed documentation? Does it work?

migration/src/main/java/ai/timefold/solver/migration/v8/SortingMigrationRecipe.java
```diff
@@ -0,0 +1,129 @@
...
+
+import ai.timefold.solver.migration.AbstractRecipe;
+
+import org.openrewrite.Recipe;
+import org.openrewrite.java.ChangeAnnotationAttributeName;
+import org.openrewrite.java.ChangeMethodName;
+import org.openrewrite.java.ChangeType;
+import org.openrewrite.java.ReplaceConstantWithAnotherConstant;
+
+public class SortingMigrationRecipe extends AbstractRecipe {
```
triceo, line 13: Great! Now we can't break this even if we tried really hard.

migration/src/test/java/ai/timefold/solver/migration/v8/SortingMigrationRecipeTest.java
```diff
@@ -0,0 +1,152 @@
...
+class SortingMigrationRecipeTest implements RewriteTest {
+
+    @Override
+    public void defaults(RecipeSpec spec) {
+        spec.recipe(new SortingMigrationRecipe())
+                .parser(AbstractRecipe.JAVA_PARSER);
+    }
+
+    @Test
+    void constraintMethods() {
```
triceo: Method name seems incorrect for what it does.

core/src/main/java/ai/timefold/solver/core/api/domain/common/ComparatorFactory.java
zepfred: That's a good point. I think that if we change the type to `Comparator`, I could make `SelectionSorterWeightFactory` extend `ComparatorFactory` again and revert all related changes.

zepfred: The idea has proven to be good. However, we now have two properties that accept a Comparator, but one can access the current solution when creating it.

zepfred: I'm unsure if this will lead to more confusion.

triceo: Hmmm. I wonder why the previous implementation looked like this. It's a very non-obvious design. There must have been a reason and we could possibly be losing some corner case that would no longer be possible. 

Good point about the similarities between `comparatorClass` and `comparatorFactoryClass`. I see a clear distinction - `comparatorClass` is a comparator which doesn't understand anything; it is a static comparator for the most simple use cases. `comparatorFactoryClass` allows you to tailor the comparison logic to what's happening inside the solution, making it more suitable towards our models. I bet no model will use `comparatorClass`, but quickstarts would.

zepfred: The more I think about using only `Comparator`, the more I like it. It simplifies the code base, and I can't think of a use case where we couldn't solve it with `Comparator`. Regardless, this would only be a requirement for new code, since the old `Comparable` approach could still be used with `SelectionSorterWeightFactory`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/ComparatorFactoryAdapter.java
zepfred: This class has been removed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/FactorySelectionSorter.java
zepfred: This class has been modified, and this method no longer exists.

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
zepfred: Yep. You can try this with the current documentation deployment. 

https://5d19adf8.timefold-docs.pages.dev/timefold-solver/latest/using-timefold-solver/modeling-planning-problems#planningEntityDifficulty
https://5d19adf8.timefold-docs.pages.dev/timefold-solver/latest/using-timefold-solver/modeling-planning-problems#planningValueStrength

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/BasicVariableDescriptor.java
zepfred: This cannot be changed, as the use of both properties will be checked later.

triceo:
Anything you still want to do, @zepfred?
Since you're gone the rest of the week, I want to be sure before I merge it.

PR #1834, opened 2025-09-26
chore: rename Move Streams to Neighborhoods

triceo:
There are no changes in logic, just moving things around and changing names.


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningVariableDiff.java
```diff
@@ -27,7 +28,10 @@ public String toString() {
                   Old value: %s
                   New value: %s
                 """
-                .formatted(variableMetaModel.isGenuine() ? "genuine" : "shadow", variableMetaModel.name(),
+                .formatted(
```
zepfred, line 31: Should we add a base method to return the description and override it in child classes?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/Neighborhood.java
```diff
@@ -0,0 +1,8 @@
+package ai.timefold.solver.core.impl.neighborhood.maybeapi;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface Neighborhood {
```
zepfred, line 6: Let's add a comment to this class.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/DefaultPlanningVariableDiff.java
triceo: The method would have to exist on the public API. This is not something that a user needs, therefore I do not clutter the public API with this.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/Neighborhood.java
triceo: Good point. Before the API is published, all Javadoc will be reviewed and docs written - there are more places than just this where content is missing.

PR #1836, opened 2025-09-26
fix: fail fast when referenced entities are not included in ConstraintVerifier.given or updateShadowVariables

Christopher-Chianelli:
Let a, b be entities, and let a.start = b.end + 1. Previously, if you were to do

    constraintVerifier.verifyThat(...).given(a)

or

    SolutionManager.updateShadowVariables(Solution.class, a)

an non-descriptive IllegalArgumentException would be thrown, since a references a non-given entity b, which is unknown to the ShadowVariableGraph.

Now, we discover all entities referenced by the given entities, and how they are referenced, and include them in the exception message.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
@@ -136,7 +136,8 @@ public final class ShadowVariableUpdateHelper<Solution_> {
             return new InternalShadowVariableSession<>(solutionDescriptor,
                     DefaultShadowVariableSessionFactory.buildGraph(
                             new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
-                                    ChangedVariableNotifier.empty(), entities).withDiscoveredReferencedEntities()));
+                                    ChangedVariableNotifier.empty(), entities)
+                                    .assertingNoReferencedMissingEntities()));
         }
 
         /**
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
@@ -136,7 +136,7 @@ public final class ShadowVariableUpdateHelper<Solution_> {
             return new InternalShadowVariableSession<>(solutionDescriptor,
                     DefaultShadowVariableSessionFactory.buildGraph(
                             new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
-                                    ChangedVariableNotifier.empty(), entities)));
+                                    ChangedVariableNotifier.empty(), entities).withDiscoveredReferencedEntities()));
         }
 
         /**
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
@@ -59,7 +59,7 @@ public record ConsistencyTracker<Solution_>(
                         new GraphStructure.GraphStructureAndDirection(GraphStructure.ARBITRARY, null, null),
                         new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
                                 ChangedVariableNotifier.empty(),
-                                entities).withConsistencyTracker(this));
+                                entities).withConsistencyTracker(this).withDiscoveredReferencedEntities());
 
         // Graph will either be DefaultVariableReferenceGraph or EmptyVariableReferenceGraph
         // If it is empty, we don't need to do anything.
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
@@ -58,8 +58,9 @@ public record ConsistencyTracker<Solution_>(
                         // if the user specified a specific inconsistency value.
                         new GraphStructure.GraphStructureAndDirection(GraphStructure.ARBITRARY, null, null),
                         new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
-                                ChangedVariableNotifier.empty(),
-                                entities).withConsistencyTracker(this).withDiscoveredReferencedEntities());
+                                ChangedVariableNotifier.empty(), entities)
+                                .withConsistencyTracker(this)
+                                .assertingNoReferencedMissingEntities());
 
         // Graph will either be DefaultVariableReferenceGraph or EmptyVariableReferenceGraph
         // If it is empty, we don't need to do anything.
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
@@ -79,7 +79,7 @@ public class DefaultShadowVariableSessionFactory<Solution_> {
                     RootVariableSource<?, ?> referredVariableSource) {
                 String getMessage() {
                     return """
-                            The entity's (%s) shadow variable (%s) refers to a non-given entity (%s)
+                            The entity's (%s) shadow variable (%s) refers to a declarative shadow variable on a non-given entity (%s)
                             variable via the source path (%s).
                             """.formatted(sourceEntity, referringShadowVariable.getVariableName(),
                             missingReferredEntity,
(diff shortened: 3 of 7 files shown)


triceo:
IMO failing fast with a good explanation is the better approach here.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -59,7 +59,7 @@ void setUnknownConsistencyFromEntityShadowVariablesInconsistent(SolutionDescript
                         new GraphStructure.GraphStructureAndDirection(GraphStructure.ARBITRARY, null, null),
                         new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
                                 ChangedVariableNotifier.empty(),
-                                entities).withConsistencyTracker(this));
+                                entities).withConsistencyTracker(this).withDiscoveredReferencedEntities());
```
triceo:
```suggestion
new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
ChangedVariableNotifier.empty(), entities)
    .withConsistencyTracker(this)
    .withDiscoveredReferencedEntities());
```

For improved readability.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
                     variableReferenceGraphBuilder, entities, graphCreator);
         }
 
+        public GraphDescriptor<Solution_> withDiscoveredReferencedEntities() {
+            record MissingEntity(Object sourceEntity,
```
triceo: Even though this is perfectly legal, I'd still prefer if types were not tested into methods. The method is complex enough as it is.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+                }
+            } while (anyChanged.getPlain());
+
+            throw new IllegalArgumentException("""
+                    Found referenced entities that were not given:
+                    %s
+                    When ConstraintVerifier.verifyThat().given(...) or
+                    %s.updateShadowVariables(solutionClass, ...) is used,
+                    all referenced entities must be passed in as arguments.
+                    Maybe add the missing entities as arguments?
```
triceo, line 129: IMO the list of referenced entities should be somehow clearly distinguished from the rest of the message. Maybe some whitespace before and after?

Also, the missing entity message itself is quite verbose, considering that it will possibly be repeated many times.

Finally, the `limit(5)` suggests that the list should end with `and N more...` if some items are omitted.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+                    return System.identityHashCode(missingReferredEntity);
+                }
+            }
+            var entitySet = CollectionUtils.newIdentityHashSet(entities.length);
+            entitySet.addAll(Arrays.asList(entities));
+            var missingEntitySet = new HashSet<MissingEntity>();
+
+            var declarativeShadowDescriptors = solutionDescriptor.getDeclarativeShadowVariableDescriptors();
+
+            for (var entity : entities) {
```
triceo: 8 levels of nesting is clearly too much.
The conditions in the `if` statements can be inverted; early `continue` can eliminate two levels of nesting. This also helps with understanding this entire code structure, because once `continue` happens, I know I no longer have to think about that context.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+                for (var shadowDescriptor : declarativeShadowDescriptors) {
+                    if (shadowDescriptor.getEntityDescriptor().getEntityClass().isAssignableFrom(entity.getClass())) {
```
triceo: This can possibly simplified if the `solutionDescriptor.getDeclarativeShadowVariableDescriptors()` allows to query per-entity. Better keep that code there, where it can be reused, and not here, where the code is already complex enough.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+                        for (var source : shadowDescriptor.getSources()) {
+                            for (var variableSourceReference : source.variableSourceReferences()) {
+                                if (variableSourceReference.isDeclarative()) {
```
triceo: Arguably could be exposed as a method on `shadowDescriptor`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+            // so visit them too
+            AtomicBoolean anyChanged = new AtomicBoolean(false);
+            do {
+                anyChanged.setPlain(false);
+                for (var entity : missingEntitySet.stream().map(MissingEntity::missingReferredEntity).toList()) {
+                    for (var shadowDescriptor : declarativeShadowDescriptors) {
+                        if (shadowDescriptor.getEntityDescriptor().getEntityClass().isAssignableFrom(entity.getClass())) {
+                            for (var source : shadowDescriptor.getSources()) {
+                                for (var variableSourceReference : source.variableSourceReferences()) {
+                                    if (variableSourceReference.isDeclarative()) {
```
triceo: Same comments as above; this whole nesting structure can be significantly simplified by moving the query methods to the descriptors.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -68,6 +72,105 @@ public GraphDescriptor<Solution_> withConsistencyTracker(ConsistencyTracker<Solu
...
+                // No new entities were discovered; reuse the original descriptor
+                return this;
+            }
+
+            // The newly discovered entities might also reference unknown entities,
+            // so visit them too
+            AtomicBoolean anyChanged = new AtomicBoolean(false);
+            do {
+                anyChanged.setPlain(false);
+                for (var entity : missingEntitySet.stream().map(MissingEntity::missingReferredEntity).toList()) {
```
triceo: Instead of creating the list, I'd just do a `forEach` and map inside of the lambda.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableUpdateHelper.java
```diff
@@ -136,7 +136,7 @@ public static <Solution_> InternalShadowVariableSession<Solution_> build(
             return new InternalShadowVariableSession<>(solutionDescriptor,
                     DefaultShadowVariableSessionFactory.buildGraph(
                             new DefaultShadowVariableSessionFactory.GraphDescriptor<>(solutionDescriptor,
-                                    ChangedVariableNotifier.empty(), entities)));
+                                    ChangedVariableNotifier.empty(), entities).withDiscoveredReferencedEntities()));
```
triceo: Same comment on formatting the builder pattern.

triceo (commented):
IMO the logic could and should be cleaned up.

Also, considering how much iteration there will potentially be, I wonder if it makes sense to try to catch all the missing entities. Maybe we fail on the first one found? Tests deserve to perform well too.

But thinking about this further, if no such entities will be found, then the iteration will be performed in its entirety, and therefore the best case is also the slowest case. This is clearly not ideal.

The original code threw `IllegalArgumentException` - what if we only fail in that place, but with a better exception? Can we provide a decent exception there, will at least some useful information?

Christopher-Chianelli:
@triceo do note this code is only run on `ConstraintVerifier.verifyThat().given()` and `SolutionManager.updateShadowVariables()`, as such, the number of entities is extremely limited (like, 5), and it does not impact solving.

triceo:
> @triceo do note this code is only run on `ConstraintVerifier.verifyThat().given()` and `SolutionManager.updateShadowVariables()`, as such, the number of entities is extremely limited (like, 5), and it does not impact solving.

Yes, I figured. Two points though:

- Tests deserve to be fast too.
- It is a fair assumption that the number of entities will be limited. In our tests, that seems to be the case. Yet every time I see a new user codebase, I am surprised by how many of my assumptions were wrong.

The excessive nesting needs to be fixed. The fact that there are several nested loops can possibly be forgiven, if we add a comment in the code describing that the assumption is that the user will not provide an excessive amount of entities in their tests.

triceo (approved):

Issue #1805, opened 2025-09-12
docs should replace solutionManager.explain() with .analyse()


ge0ffrey:
The "Explainable AI" doc chapter:
https://docs.timefold.ai/timefold-solver/latest/constraints-and-score/understanding-the-score
starts with .explain() instead of .analyse().
The former should be avoided. It will be deprecated as soon as we figure out the JSON-friendly Indictment replacement.

<img width="802" height="206" alt="Image" src="https://github.com/user-attachments/assets/c8bbb67a-ec1c-47ff-823b-81efe1c40e64" />

PatrickDiallo23:
Would it be a good idea to add a "good first issue" label if the resolution would consist a simple change? I think this is a good opportunity for those who want to contribute to other projects but they don't know where to start.

triceo:
@PatrickDiallo23 Good point. Added.

PatrickDiallo23:
If nobody wants to do this issue and you need it in the next release, can you ping me do to the change? Thanks!

triceo:
@PatrickDiallo23 Feel free to take this one when you find the time.

PR #1846, opened 2025-09-30
docs: replace solutionManager.explain() with .analyse()

PatrickDiallo23:
This PR replaces the `explain()` method with the `analyse()` one since the former method will be deprecated in the future versions of Timefold (#1805).

### DIFF
--- a/docs/src/modules/ROOT/pages/constraints-and-score/understanding-the-score.adoc
+++ b/docs/src/modules/ROOT/pages/constraints-and-score/understanding-the-score.adoc
@@ -17,7 +17,8 @@ Java::
 [source,java,options="nowrap"]
 ----
 SolutionManager<VehicleRoutePlan, HardSoftLongScore> solutionManager = SolutionManager.create(solverFactory);
-System.out.println(solutionManager.analyze(vehicleRoutePlan));
+ScoreAnalysis<HardSoftLongScore> scoreAnalysis = solutionManager.analyze(vehicleRoutePlan);
+System.out.println(scoreAnalysis.summarize());
 ----
 ====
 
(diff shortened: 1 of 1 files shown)


triceo (changes requested):
Thanks for the contribution! Unfortunately, this change is not equivalent to the original.

Please review the Javadoc of `analyze()` and `ScoreAnalysis`, and update the code snippet to actually do what the documentation says it will.

docs/src/modules/ROOT/pages/constraints-and-score/understanding-the-score.adoc
```diff
@@ -273,7 +273,7 @@ Java::
 [source,java,options="nowrap"]
 ----
 SolutionManager<VehicleRoutePlan, HardSoftLongScore> solutionManager = SolutionManager.create(solverFactory);
-ScoreExplanation<VehicleRoutePlan, HardSoftLongScore> scoreExplanation = solutionManager.explain(vehicleRoutePlan);
+ScoreExplanation<VehicleRoutePlan, HardSoftLongScore> scoreExplanation = solutionManager.analyze(vehicleRoutePlan);
```
triceo: I suggest you run the code before you make any documentation changes.
This code will not compile.

In this case, it was correct to keep the `explain` method here, because score analysis doesn't support indictments and this section deals exclusively with indictments.

PatrickDiallo23: @triceo So the only change was in the introduction of the page, right? I thought that you wanted to replace the `explain()` method with a more comprehensive variant (with more details or functionalities) named `analyze()`. This is what I saw at the first glance in the codebase. Sorry for the confusion.

triceo: There are two parts to this:

- We want explain() to be replaced by analyze(), true.
- But the second part uses explain() for something that analyze() cannot do, and therefore that change was not correct.

The change to the introduction needs to be tweaked as well though, because the way you wrote it here wouldn't print out what you think it would. (Therefore my recommendation to read the Javadoc and/or look at the implementation.)

PatrickDiallo23: I provided a new variant. Please take a look.

PR #1849, opened 2025-10-01
fix: process deep-cloned entities in a queue for GizmoSolutionCloner

Christopher-Chianelli:
- Previously, when a class is deep-cloned, a method that constructs and fully initializes a clone of that class is called. This has the potential to cause a StackOverflowException when a deep-cloned class refers to many other deep-cloned classes, such as in chained models.

- Now, when a field is deeped cloned, it adds the field initializer to a cloneQueue, and that cloneQueue can only be processed for the first entity being cloned. As such, stack-height does not grow indefinitely.

- Perform various cleanups in the GizmoSolutionCloner

- Change method `isInvalid` to `getIsInvalid` so the GizmoSolutionCloner recongize it as the getter of the corresponding field.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
@@ -10,6 +10,7 @@ import java.lang.reflect.WildcardType;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -21,11 +22,11 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -37,15 +38,13 @@ import ai.timefold.solver.core.impl.domain.solution.cloner.DeepCloningUtils;
 import ai.timefold.solver.core.impl.domain.solution.cloner.FieldAccessingSolutionCloner;
 import ai.timefold.solver.core.impl.domain.solution.cloner.PlanningCloneable;
 import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
-import ai.timefold.solver.core.impl.util.MutableReference;
+import ai.timefold.solver.core.impl.util.Pair;
 
 import io.quarkus.gizmo.AssignableResultHandle;
-import io.quarkus.gizmo.BranchResult;
 import io.quarkus.gizmo.BytecodeCreator;
 import io.quarkus.gizmo.ClassCreator;
 import io.quarkus.gizmo.ClassOutput;
 import io.quarkus.gizmo.FieldDescriptor;
-import io.quarkus.gizmo.MethodCreator;
 import io.quarkus.gizmo.MethodDescriptor;
 import io.quarkus.gizmo.ResultHandle;
 
@@ -59,6 +58,30 @@ public class GizmoSolutionClonerImplementor {
     private static final String FALLBACK_CLONER = "fallbackCloner";
     public static final boolean DEBUG = false;
 
+    protected record ClonerDescriptor(SolutionDescriptor<?> solutionDescriptor,
+            Map<Class<?>, GizmoSolutionOrEntityDescriptor> memoizedSolutionOrEntityDescriptorMap,
+            SortedSet<Class<?>> deepClonedClassesSortedSet,
+            ClassCreator classCreator) {
+    }
+
+    protected record ClonerMethodDescriptor(GizmoSolutionOrEntityDescriptor entityDescriptor,
+            BytecodeCreator bytecodeCreator,
+            ResultHandle createdCloneMap,
+            boolean isBottom,
+            ResultHandle cloneQueue) {
+        public ClonerMethodDescriptor withBytecodeCreator(BytecodeCreator bytecodeCreator) {
+            return new ClonerMethodDescriptor(entityDescriptor, bytecodeCreator, createdCloneMap, isBottom, cloneQueue);
+        }
+
+        public ClonerMethodDescriptor withCreatedCloneMap(ResultHandle createdCloneMap) {
+            return new ClonerMethodDescriptor(entityDescriptor, bytecodeCreator, createdCloneMap, isBottom, cloneQueue);
+        }
+
+        public ClonerMethodDescriptor withCloneQueue(ResultHandle cloneQueue) {
+            return new ClonerMethodDescriptor(entityDescriptor, bytecodeCreator, createdCloneMap, isBottom, cloneQueue);
+        }
+    }
+
     /**
      * Return a comparator that sorts classes into instanceof check order.
      * In particular, if x is a subclass of y, then x will appear earlier
@@ -69,7 +92,7 @@ public class GizmoSolutionClonerImplementor {
      *         x &lt; y if x is assignable from y.
      */
     public static Comparator<Class<?>> getInstanceOfComparator(Set<Class<?>> deepClonedClassSet) {
-        Map<Class<?>, Integer> classToSubclassLevel = new HashMap<>();
+        var classToSubclassLevel = new HashMap<Class<?>, Integer>();
         deepClonedClassSet
                 .forEach(clazz -> {
                     if (deepClonedClassSet.stream()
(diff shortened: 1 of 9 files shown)


core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerTest.java
```diff
@@ -40,7 +41,7 @@ void debuggingDisabled() {
     @Override
     protected <Solution_> SolutionCloner<Solution_> createSolutionCloner(SolutionDescriptor<Solution_> solutionDescriptor) {
         var className = GizmoSolutionClonerFactory.getGeneratedClassName(solutionDescriptor);
-        var classBytecodeHolder = new MutableReference<byte[]>(null);
+        var classBytecodeHolder = new ArrayList<Pair<String, byte[]>>();
```
triceo: I think `bytecodeHolder` just became complex enough to warrant its own type.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerTest.java
```diff
@@ -77,12 +77,13 @@ public String getName() {
...
-                if (className.equals(name)) {
-                    return defineClass(name, byteCode, 0, byteCode.length);
-                } else {
-                    // Not a Gizmo generated class; load from context class loader
-                    return Thread.currentThread().getContextClassLoader().loadClass(name);
+                for (var entry : classBytecodeHolder) {
+                    if (entry.key().equals(name)) {
+                        return defineClass(name, entry.value(), 0, entry.value().length);
+                    }
                 }
```
triceo: This complexity can be hidden in the new type, exposed as a single well-named method.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
```diff
@@ -324,6 +330,52 @@ protected void cloneExtendedSolution() {
...
+    protected void cloneDeeplyNestedSolution() {
+        var solutionDescriptor = TestdataDependencySolution.buildSolutionDescriptor();
+        var cloner = createSolutionCloner(solutionDescriptor);
+
+        var startTime = LocalDate.of(2000, 1, 1).atTime(LocalTime.MIDNIGHT);
+        var solutionEntity = new TestdataDependencyEntity(LocalDate.of(2000, 1, 1).atTime(LocalTime.MIDNIGHT));
+        var solutionValue = new TestdataDependencyValue("root", Duration.ofHours(1L));
+
+        var outer = solutionValue;
+        var NESTED_COUNT = 10_000;
```
triceo: Any chance we could ensure that the size is just enough to trigger the stack overflow error? This test exists to check for that, so it would be unfortunate if - for whatever reason - the count were less than what is required to trigger the error on the specific machine/JVM running the test.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -59,6 +58,30 @@ public class GizmoSolutionClonerImplementor {
     private static final String FALLBACK_CLONER = "fallbackCloner";
     public static final boolean DEBUG = false;
 
+    protected record ClonerDescriptor(SolutionDescriptor<?> solutionDescriptor,
```
triceo: By convention, we typically put inner types at the end of the class.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -78,22 +101,22 @@ public static Comparator<Class<?>> getInstanceOfComparator(Set<Class<?>> deepClo
...
             // (for instance, Tiger extends Cat (1) implements Animal (0))
             isChanged = false;
             for (Class<?> clazz : deepClonedClassSet) {
-                Optional<Integer> maxParentSubclassLevel = classToSubclassLevel.keySet().stream()
+                var maxParentSubclassLevel = classToSubclassLevel.keySet().stream()
                         .filter(otherClazz -> otherClazz != clazz && otherClazz.isAssignableFrom(clazz))
                         .map(classToSubclassLevel::get)
                         .max(Integer::compare);
 
                 if (maxParentSubclassLevel.isPresent()) {
```
triceo: Arguably you can do `ifPresent` on the result of the stream pipeline directly. Save yourself the if statement, and the variable declaration.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -137,51 +160,49 @@ public static void defineClonerFor(Supplier<GizmoSolutionClonerImplementor> impl
...
-                    deepCloneClassesThatAreNotSolutionSortedSet);
+        var clonerDescriptor = new ClonerDescriptor(solutionDescriptor, memoizedSolutionOrEntityDescriptorMap,
+                deepCloneClassesThatAreNotSolutionSortedSet,
+                classCreator);
+
+        implementor.createFields(clonerDescriptor);
+        implementor.createConstructor(clonerDescriptor);
+        implementor.createSetSolutionDescriptor(clonerDescriptor);
+        implementor.createCloneSolution(clonerDescriptor);
+        implementor.createCloneSolutionRun(clonerDescriptor, solutionClassSet, instanceOfComparator);
```
triceo: Is there ever a situation where these methods would not be called together?
If not, I suggest they be merged in a single method. (Under the hood, they can still be separate, just not accessible to the public; I'd like to avoid situation where the object is incompletely initialized, if it can be avoided.)

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -269,121 +291,134 @@ protected void createSetSolutionDescriptor(ClassCreator classCreator, SolutionDe
...
+        var thisObj = methodCreator.getMethodParam(0);
+        var solutionNullBranchResult = methodCreator.ifNull(thisObj);
+        try (var solutionIsNullBranch = solutionNullBranchResult.trueBranch()) {
+            solutionIsNullBranch.returnValue(thisObj); // thisObj is null
+        }
 
-            isSubclassBranch.invokeInterfaceMethod(
-                    MethodDescriptor.ofMethod(Map.class, "put", Object.class, Object.class, Object.class),
-                    createdCloneMap, thisObj, clone);
+        try (var solutionIsNotNullBranch = solutionNullBranchResult.falseBranch()) {
```
triceo, line 283: This is a very long branch. I suggest breaking it down into smaller, well-named methods.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -269,121 +291,134 @@ protected void createSetSolutionDescriptor(ClassCreator classCreator, SolutionDe
...
+                    var fieldValue = gizmoMemberDescriptor.readMemberValue(isSubclassBranch, thisObj);
+                    var cloneValue = isSubclassBranch.createVariable(deeplyClonedField.getType());
+                    writeDeepCloneInstructions(clonerDescriptor,
+                            new ClonerMethodDescriptor(solutionSubclassDescriptor, isSubclassBranch, createdCloneMap,
+                                    true, isSubclassBranch.newInstance(MethodDescriptor.ofConstructor(ArrayDeque.class))),
+                            deeplyClonedField,
+                            gizmoMemberDescriptor, fieldValue, cloneValue);
+
+                    if (!gizmoMemberDescriptor.writeMemberValue(isSubclassBranch, clone, cloneValue)) {
+                        throw new IllegalStateException("The member (" + gizmoMemberDescriptor.getName() + ") of class (" +
```
triceo: Let's use the new pattern with `formatted(...)`.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -269,121 +291,134 @@ protected void createSetSolutionDescriptor(ClassCreator classCreator, SolutionDe
...
-            currentBranch = isSubclassBranchResult.falseBranch();
+                currentBranch = isSubclassBranchResult.falseBranch();
+            }
+            var errorBuilder = currentBranch.newInstance(MethodDescriptor.ofConstructor(StringBuilder.class, String.class),
+                    currentBranch.load("Failed to create clone: encountered ("));
+            var APPEND =
+                    MethodDescriptor.ofMethod(StringBuilder.class, "append", StringBuilder.class, Object.class);
+
+            currentBranch.invokeVirtualMethod(APPEND, errorBuilder, thisObjClass);
+            currentBranch.invokeVirtualMethod(APPEND, errorBuilder, currentBranch.load(") which is not a known subclass of " +
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
```diff
@@ -937,123 +1002,190 @@ protected String getEntityHelperMethodName(Class<?> entityClass) {
...
         }
-        noCloneBranch.invokeInterfaceMethod(
-                MethodDescriptor.ofMethod(Map.class, "put", Object.class, Object.class, Object.class),
-                cloneMap, toClone, cloneObj);
 
-        for (GizmoMemberDescriptor shallowlyClonedField : entityDescriptor.getShallowClonedMemberDescriptors()) {
-            writeShallowCloneInstructions(entityDescriptor, noCloneBranch, shallowlyClonedField, toClone, cloneObj, cloneMap,
-                    deepClonedClassesSortedSet);
-        }
+        try (var originalNoCloneBranch = hasCloneBranchResult.falseBranch()) {
```
triceo, line 1052: Another very complex branch deserves to be broken down.

triceo (commented):
Since we're already improving the code, I'm making some further improvement suggestions of my own. Reading this code is IMO harder than it could be.

Also, please make sure you work with up-to-date `main`, because CI is currently failing due to that.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
Christopher-Chianelli: There no method to determine the maximum number of frames on the stack (and each frame could have a different size in memory). 10_000 causes a StackOverflow on my machine with the old code.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/cloner/gizmo/GizmoSolutionClonerImplementor.java
Christopher-Chianelli: This does not initialize an object; it literally creating the fields/constructors/methods on the class object. I prefer to have them separate, since Quarkus overrides them and separation of responsibilities.

triceo (approved):
LGTM. Please review the Sonar comments.

PR #1850, opened 2025-10-01
fix: don't test the filtering joiner if a tuple is already retracting

triceo:
The included tests describe the problematic situation.
The comments in the code explain the solution.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
@@ -1,13 +1,13 @@
 package ai.timefold.solver.core.impl.bavet.common;
 
+import java.util.function.Consumer;
+
 import ai.timefold.solver.core.impl.bavet.common.tuple.AbstractTuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleState;
 import ai.timefold.solver.core.impl.bavet.common.tuple.UniTuple;
 import ai.timefold.solver.core.impl.util.ElementAwareList;
 import ai.timefold.solver.core.impl.util.ElementAwareListEntry;
-
-import java.util.function.Consumer;
 
 /**
  * This class has two direct children: {@link AbstractIndexedJoinNode} and {@link AbstractUnindexedJoinNode}.


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
```diff
@@ -115,6 +115,19 @@ protected final void innerUpdateRight(UniTuple<Right_> rightTuple, Consumer<Cons
 
     private void processOutTupleUpdate(LeftTuple_ leftTuple, UniTuple<Right_> rightTuple, ElementAwareList<OutTuple_> outList,
             ElementAwareList<OutTuple_> outTupleList, int outputStoreIndexOutEntry) {
+        if (!leftTuple.state.isActive()) {
```
Christopher-Chianelli, line 118: Do you need to also check rightTuple is active, or does that get checked somewhere else or is guarateed to always be active?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -136,6 +136,20 @@ protected void updateCounterFromLeft(LeftTuple_ leftTuple, UniTuple<Right_> righ
 
     protected void updateCounterFromRight(UniTuple<Right_> rightTuple, ExistsCounter<LeftTuple_> counter,
             ElementAwareList<FilteringTracker<LeftTuple_>> rightTrackerList) {
+        var leftTuple = counter.leftTuple;
+        if (!leftTuple.state.isActive()) {
```
Christopher-Chianelli, line 140: Do you need something similar on line 129 where you update the counter from left?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
triceo: I was not able to reproduce the same problem with updates coming from the left. Maybe the problem is there, but without confirmation, I'm not inclined to do something about it preemptively.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
triceo: See answer to previous comment.

PR #1853, opened 2025-10-03
feat: allow caching the results of an unfiltered UniStream (WIP)

### LINKED ISSUE #1806: Add `filterStatic` (vs. `filter`) to ConstraintStreams
We often write statements like: `factory.forEach(Shift.class).filter(someCondition)` where `someCondition` only uses data that never changes (static).
Whenever a `shift` changes in the example above, the filter is evaluated. Running this evaluation often (over and over) is costly.

Instead of evaluating the filter every time a `shift` changes, I propose to cache the filtered list of `shifts`.

We have worked around this issue by introducing a new type, for which we precompute the filtering and use that instead of the original `Shift.class`.

Another positiv effect of providing something like `filterStatic` and describing the difference to `filter` may be that it may make the behavior of `filter` even more clear.

Christopher-Chianelli:
- Introduces ConstraintFactory.staticData(Uni), which creates an independent NodeNetwork from the provided Uni stream.

- Introduces TupleSourceRoot that represent top level tuple producers (the ForEach nodes now implement this, as well as a new StaticDataUniNode).

- The StaticDataUniNode caches the results of its internal NodeNetwork and maps the tuples to new tuples compatiable with the external NodeNetwork

- Inserts/retracts invalidate the cache and cause the output tuples to be updated.

- Updates uses the cache and do not notify the internal NodeNetwork.


core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -180,11 +180,17 @@ public void afterProblemFactAdded(Object problemFact) {
         super.afterProblemFactAdded(problemFact);
     }
 
-    // public void beforeProblemPropertyChanged(Object problemFactOrEntity) // Do nothing
+    @Override
+    public void beforeProblemPropertyChanged(Object problemFactOrEntity) {
+        // Since this is called when a fact (not a variable changes),
```
triceo:
```suggestion
        // Since this is called when a fact (not a variable) changes,
```

triceo:
This is surprisingly simple! Some high-level comments; leaving naming out of the picture for now:

> Introduces ConstraintFactory.staticData(Uni), which creates an independent NodeNetwork from the provided Uni stream.

I'm not a big fan of this. The programming model doesn't look very nice IMO.

    factory.staticData(factory.forEach()...)
        .filter(...)

Something like this would be more stream-y:

    factory.forEach().
        ...
        .asStatic()
        .filter(...)

This brings its own problems. But both proposals share the same issue - chaining static streams. Nothing (other than run-time fail-fasts) prevents you from doing `staticData(staticData(...).filter(...))`. IMO we should decide (and enforce) that a static stream can only be built once, and everything after that is dynamic. And ideally, this would happen at type level, not at runtime.

This was the main idea of `filterStatic()` - because we can actually guarantee, at type level, that once any "non-static" method is called, a "static" method can never be called again. (The reverse doesn't work, unless we want to duplicate the entire stream API, which we don't.) IMO we shouldn't rule out `filterStatic()` + `joinStatic()` - under the hood, it can be implemented like in this PR, but the public API IMO need not offer this absolute flexibility to turn _anything_ into static data.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/RecordingTupleNode.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void retract(Tuple_ tuple) {
+        // Do nothing
+    }
+
+    @Override
+    public Propagator getPropagator() {
+        return new StaticPropagationQueue<>(new TupleLifecycle<>() {
```
triceo: So this propagates to nowhere? The whole node seems to exist only to record the tuples; arguably, the code should be refactored to not require this to be a node, and just use the recorder directly.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/BavetRootNode.java
```diff
@@ -0,0 +1,51 @@
+package ai.timefold.solver.core.impl.bavet.common;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface TupleSourceRoot<A> {
+    void insert(A a);
```
triceo: I am not sure why this used to be `@Nullable`, but I remember clearly that I was adding it there for a reason. 

I recommend adding null checks, running the entire CI incl. quickstarts, and if the null checks don't trigger, then I say it's safe for this to be non-null.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/StaticDataUniNode.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Override
+    public void insert(Object a) {
+        if (tupleMap.containsKey(a)) {
+            return;
+        }
+        invalidateCache();
+        tupleMap.put(a, new ArrayList<>());
+        insertIntoNodeNetwork(a);
+        recalculateTuples();
```
triceo: Doing this on every insert/retract seems excessively expensive. There will be a large burst of inserts at the beginning; they ought to be cached and processed in a batch, probably before propagation starts.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/StaticDataUniNode.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Override
+    public void update(UniTuple<A> tuple) {
+        // Do nothing; this is a source node
+    }
+
+    @Override
+    public void retract(UniTuple<A> tuple) {
+        // Do nothing; this is a source node
+    }
```
triceo: Arguably if these methods aren't needed, then this shouldn't implement `TupleLifecycle`. If it doesn't look like a duck and doesn't quack like a duck, it shouldn't be called a duck.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/uni/BavetStaticDataUniConstraintStream.java
```diff
@@ -0,0 +1,121 @@
...
+import ai.timefold.solver.core.impl.score.buildin.SimpleScoreDefinition;
+import ai.timefold.solver.core.impl.score.constraint.ConstraintMatchPolicy;
+import ai.timefold.solver.core.impl.score.stream.bavet.BavetConstraintFactory;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.BavetConstraintStreamBinaryOperation;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.ConstraintNodeBuildHelper;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.bridge.BavetAftBridgeUniConstraintStream;
+import ai.timefold.solver.core.impl.score.stream.common.RetrievalSemantics;
+import ai.timefold.solver.core.impl.score.stream.common.inliner.AbstractScoreInliner;
+
+public class BavetStaticDataUniConstraintStream<Solution_, A> extends BavetAbstractUniConstraintStream<Solution_, A>
```
triceo: I know that we said we don't want to deal with joins yet. But considering that the same mechanism would be used to support joins as well, I think we deserve to see how this would look for a bi-stream.

(Also, considering that this suddenly looks like much less work than originally estimated, maybe we actually _do_ take joins in scope?)

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -192,6 +192,17 @@ public interface ConstraintFactory {
...
+    // ************************************************************************
+    // staticData
+    //************************************************************************
+
+    /**
+     * Computes and caches the tuples that would be produced by the given stream.
+     * As this is cached, it is vital the stream does not reference any variables
+     * (genuine or otherwise).
+     */
+    <A> @NonNull UniConstraintStream<A> staticData(UniConstraintStream<A> stream);
```
triceo: What happens when this stream is used both statically and non-statically?
(Think node-sharing.)

triceo:
To bring an alternative API idea:

     factory.forEachXYZ(Something.class, filter)

This would produce a `StaticUniStream` which extends `UniStream`, and only adds one method - `joinXYZ()`; this join only allows to join other static streams and returns `StaticBiStream` - allowing possibly for a static bi join (and similarly for tri).

In essence, this brings a very simple, clear API:

- We support static filters and joins (maybe also ifExists, which is essentially a join).
- Static operations come first.
- Once you go dynamic, you can never go back.
- This is enforced at the type level, preventing impossible situations from being modelled.
- Node-sharing is trivial; nothing changes in the status quo.

It does have a downside - it doesn't allow arbitrary static streams, such as groupBy, map, flatten etc. Right now, nobody is asking for it, and I think it is a decent trade-off to get the benefits; but we can possibly ask other people for their opinion on this.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
Christopher-Chianelli: The output stream can be node shared if the input streams are identical (not implemented currently since this is a POC). The input stream will not be node shared; its an independent network.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -192,6 +192,18 @@ public interface ConstraintFactory {
...
+    // staticData
+    //************************************************************************
+
+    /**
+     * Computes and caches the tuples that would be produced by the given stream.
+     * As this is cached, it is vital the stream does not reference any variables
+     * (genuine or otherwise).
+     */
+    <Stream_ extends ConstraintStream> @NonNull Stream_
+            staticData(StaticDataSupplier<@NonNull Stream_> staticDataSupplier);
```
triceo: TODO: naming

core/src/main/java/ai/timefold/solver/core/api/score/stream/StaticDataFactory.java
```diff
@@ -0,0 +1,7 @@
+package ai.timefold.solver.core.api.score.stream;
+
+import ai.timefold.solver.core.api.score.stream.uni.UniConstraintStream;
+
+public interface StaticDataFactory {
```
triceo: TODO: naming

core/src/main/java/ai/timefold/solver/core/api/score/stream/StaticDataSupplier.java
```diff
@@ -0,0 +1,8 @@
+package ai.timefold.solver.core.api.score.stream;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface StaticDataSupplier<Stream_ extends ConstraintStream> {
```
triceo: TODO: naming

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractStaticDataNode.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Override
+    public final void insert(Object a) {
+        if (tupleMap.containsKey(a)) {
+            return;
+        }
+        invalidateCache();
+        tupleMap.put(a, new ArrayList<>());
+        insertIntoNodeNetwork(a);
+        recalculateTuples();
```
triceo: Alternative implementation suggestion:

Inserts/updates/retracts go to a map.
If an update lands and an insert is already there, the update is ignored.
If an update lands twice, we only keep the first one.
if a retract lands and an insert is already there, the insert is removed and the retract is ignored.

All the operations are then pushed all at once, before propagation starts.
This way, we minimize the amount of operations that need to be performed, and we also group them in time, nothing happens before propagation; which would likely help with sizing collections etc.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractStaticDataNode.java
```diff
@@ -0,0 +1,159 @@
...
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public abstract class AbstractStaticDataNode<Tuple_ extends AbstractTuple> extends AbstractNode
+        implements TupleSourceRoot<Object> {
+    private final StaticPropagationQueue<Tuple_> propagationQueue;
+    private final Map<Object, List<Tuple_>> tupleMap = new IdentityHashMap<>(1000);
+    private final NodeNetwork nodeNetwork;
```
triceo: I'd find a better name here, something like `nestedNodeNetwork`, `independentNodeNetwork`... To signify that this differs from the main node network.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractStaticDataNode.java
```diff
@@ -0,0 +1,159 @@
...
+                .forEach(node -> ((AbstractForEachUniNode) node).retract(toRetract));
+    }
+
+    private void invalidateCache() {
+        tupleMap.values().stream().flatMap(List::stream).forEach(this::retractExisting);
+        recordingTupleNode.getTupleRecorder().reset();
+    }
+
+    private void recalculateTuples() {
+        // Settle all the inserts/retracts that happen
```
triceo: We typically settle retracts before updates, and updates before inserts—this is to handle liveness of data. If you first do inserts, then you will inevitably be joining with data that will be retracted later, wasting CPU time.

This is part of the reasoning for my suggestion to only recalculate at propagation time, when you already know all the operations that are going to happen in the current cycle.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractStaticDataNode.java
```diff
@@ -0,0 +1,159 @@
...
+        var recorder = recordingTupleNode.getTupleRecorder();
+        for (var mappedTupleEntry : tupleMap.entrySet()) {
+            mappedTupleEntry.getValue().clear();
+            var invalidated = mappedTupleEntry.getKey();
+            recorder.recordingInto(mappedTupleEntry.getValue(), this::remapTuple, () -> {
+                // Do an update on the object and settle the network; this will update precisely the
+                // tuples mapped to this node, which will then be recorded
+                nodeNetwork.getTupleSourceRootNodes(invalidated.getClass())
+                        .forEach(node -> ((AbstractForEachUniNode) node).update(invalidated));
+                nodeNetwork.settle();
```
triceo: Why would you settle on every update? Is that necessary for anything?
I'd rather settle after all operations had been queued.

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/PrecomputeQuadNode.java
```diff
@@ -0,0 +1,29 @@
...
+            RecordingTupleLifecycle<QuadTuple<A, B, C, D>> recordingTupleNode,
+            int outputStoreSize,
+            TupleLifecycle<QuadTuple<A, B, C, D>> nextNodesTupleLifecycle,
+            Class<?>[] sourceClasses) {
+        super(nodeNetwork, recordingTupleNode, nextNodesTupleLifecycle, sourceClasses);
+        this.outputStoreSize = outputStoreSize;
+    }
+
+    @Override
+    protected QuadTuple<A, B, C, D> remapTuple(QuadTuple<A, B, C, D> tuple) {
```
triceo, line 25: Why is this necessary? Wouldn't just retracting and inserting the tuple be enough?
Tuples are expensive to create, the `outputStoreSize` can be very large for complex streams; best avoided if possible.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/TupleSourceRoot.java
```diff
@@ -0,0 +1,51 @@
+package ai.timefold.solver.core.impl.bavet.common;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public interface TupleSourceRoot<A> {
```
triceo: Maybe just `RootNode`?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/BavetStaticDataBuildHelper.java
```diff
@@ -0,0 +1,81 @@
...
+import ai.timefold.solver.core.impl.score.buildin.SimpleScoreDefinition;
+import ai.timefold.solver.core.impl.score.constraint.ConstraintMatchPolicy;
+import ai.timefold.solver.core.impl.score.stream.common.inliner.AbstractScoreInliner;
+
+public final class BavetStaticDataBuildHelper<Tuple_ extends AbstractTuple> {
+    private final NodeNetwork nodeNetwork;
+    private final RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle;
+    private final Class<?>[] sourceClasses;
+
+    public <Solution_> BavetStaticDataBuildHelper(BavetAbstractConstraintStream<Solution_> staticConstraintStream) {
```
triceo: Some of this logic is very similar with the other build helper.
Consider sharing code.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/RetrievalSemantics.java
```diff
@@ -29,6 +30,15 @@ public enum RetrievalSemantics {
      * Applies when the stream comes off of a {@link ConstraintFactory#forEach(Class)} family of methods.
      */
     STANDARD,
+
+    /**
+     * Joins and conditional propagation always include entities with null planning variables,
+     * regardless of whether their planning variables allow unassigned values.
+     * <p>
+     * Applies when the stream comes off of a {@link StaticDataFactory#forEachUnfiltered(Class)} (Class)} family of methods.
```
triceo: The second `(Class)}` is there in error.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -192,6 +192,18 @@ public interface ConstraintFactory {
...
+    // ************************************************************************
+    // staticData
+    //************************************************************************
+
+    /**
+     * Computes and caches the tuples that would be produced by the given stream.
+     * As this is cached, it is vital the stream does not reference any variables
+     * (genuine or otherwise).
+     */
+    <Stream_ extends ConstraintStream> @NonNull Stream_
```
triceo: Why is the supplier necessary? Together with the data factory, this introduces concepts that are a bit boilerplate-y. (Knowing Geoff, I'm pretty sure this will be his first question when we show this to him.)

Other suggestions:

    factory.precomputed().forEach(...)

Clearly has the problem that it becomes unclear when the static part would end. Would require something like `.endPrecomputed()`, or perhaps

    factory.forEach(...)
        ...
        .precomputeAndThen()
        ...;

My motivation: consider the code from one of the tests; it's not at all obvious where the precomputed part ends:

    factory.staticData(data -> data.forEachUnfiltered(TestdataLavishValue.class)
                        .join(TestdataLavishEntity.class)
                        .filter((value, entity) -> entity.getEntityGroup() == entityGroup
                                && value.getValueGroup() == valueGroup))
                        .filter((value, entity) -> entity.getValue() == value1)
                        .penalize(SimpleScore.ONE)
                        .asConstraint(TEST_CONSTRAINT_NAME)

IMO there needs to be a very clear boundary, because this boundary marks where semantics change; in one part, no variables are allowed. Past the boundary, variables are fine.

triceo (commented):
Submitting more comments as I get familiar with the changes.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
triceo: The input stream may need to be shared though - why build the network twice if the user wants to reuse things?

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
Christopher-Chianelli: That because the intended way to use the API is with a method reference:
```java
UniConstraintStream<Example> myStaticData(StaticDataFactory factory) {
    return factory.forEachUnfiltered(Example.class)
                 .filter(example -> example.flag);
}

Constraint constraintUsingExample(ConstraintFactory factory) {
    return factory.staticData(this::myStaticData)
                             .filter(example -> example.variable == 1)
                             .penalize(SimpleScore.ONE)
                             .asConstraint("Example with dynamic and static data");
}
```

Here, it is extremely clear what is precomputed and what is not: everything in `myStaticData` is precomputed, and everything in `constraintUsingExample` is not precomputed. And since there no `precompute` method on the stream, you cannot accidentally precompute something that uses variable (unless you do it inside the method that return the static data). And we get the entire CS API for free (and remove the useless `forEach` and `forEachIncludingUnassigned`).

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/PrecomputeQuadNode.java
Christopher-Chianelli: So... are you telling me `outputStoreSize` is useless? What if the precomputed stream is simpler then then dynamic stream, and thus, have a smaller `outputStoreSize`?

triceo: No. What I'm saying is that, when you do retract+insert, you will be able to reuse the tuple incl. its output store. (Which will be emptied after the retract propagates.)

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
triceo: > That because the intended way to use the API is with a method reference:

When you sell it like this, it suddenly looks better. A good idea. But I'm thinking that we can probably have the same, while avoiding all the new types:

    forEachStatic(...) // StaticUniStream (extends UniStream)
        ...            // StaticUniStream (extends UniStream)
        .xxx()         // UniStream; new API to switch from static to dynamic
        ...            // UniStream
        .penalize()

The downside is that you could create a complete penalizing constraint that is completely static. The only solution to that would be a fail-fast.

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/PrecomputeQuadNode.java
Christopher-Chianelli: Smells of premature optimization; the store is a final array and cannot be resized, and this "optimization" relies on the precomputed stream having a larger or same sized store as the dynamic stream. And remapping is only perform once: when the stream is precomputed (i.e. nothing done on update).

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
Christopher-Chianelli: `StaticUniStream` does not work without duplicating the entire ConstraintStream API. By extending UniStream, `filter`, `join`, etc. returns `UniStream`/`BiStream` etc. and not `StaticUniStream`. And we cannot add a generic parameter to UniStream without breaking API (and even that not a complete solution, since generics cannot have generics, so we run into issues with map and join).

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractStaticDataNode.java
Christopher-Chianelli: It so we know which tuples map to which input. If we queue all the "fake" updates then settle, we have no way of knowing which input cause what update.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
triceo: Fair point.

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/PrecomputeQuadNode.java
triceo: Nevermind; I understand the purpose of this better now. What I want cannot be done.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/TupleRecorder.java
```diff
@@ -0,0 +1,36 @@
...
+public final class TupleRecorder<Tuple_ extends AbstractTuple> {
+    List<Tuple_> recordedTupleList;
+    UnaryOperator<Tuple_> mapper;
+    IdentityHashMap<Tuple_, Tuple_> inputTupleToOutputTuple = new IdentityHashMap<>();
+
+    public void reset() {
+        inputTupleToOutputTuple.clear();
+    }
+
+    public void recordingInto(List<Tuple_> recordedTupleList, UnaryOperator<Tuple_> mapper,
```
triceo: Wouldn't it make more sense to create a new instance at this point?
Looks like we're changing the entire state of this class, therefore keeping the original instance seems pointless.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/AbstractForEachUniNode.java
```diff
@@ -85,6 +97,11 @@ protected void retractExisting(@Nullable A a, UniTuple<A> tuple) {
         }
     }
 
+    @Override
+    public final void settle() {
```
triceo: This exposes a confusing API.

Why is there a `settle` method, and also a `propagator` which allows external settling? Let's try to find a different solution; clear APIs have only 1 way of doing things.

core/src/main/java/ai/timefold/solver/core/impl/bavet/NodeNetwork.java
```diff
@@ -32,14 +32,21 @@ public int layerCount() {
         return layeredNodes.length;
     }
 
-    public Stream<AbstractForEachUniNode<?>> getForEachNodes(Class<?> factClass) {
+    public Stream<BavetRootNode<?>> getAllTupleSourceRootNodes() {
```
triceo: If the type changed, the method name should change as well. In this case, probably `getRootNodes()` would suffice.

core/src/main/java/ai/timefold/solver/core/impl/bavet/NodeNetwork.java
```diff
@@ -32,14 +32,21 @@ public int layerCount() {
...
         // The latter is for FromSolution nodes.
         return declaredClassToNodeMap.entrySet()
                 .stream()
-                .filter(entry -> factClass == PlanningSolution.class || entry.getKey().isAssignableFrom(factClass))
-                .map(Map.Entry::getValue)
-                .flatMap(List::stream);
+                .flatMap(entry -> entry.getValue().stream());
+    }
+
+    public Stream<BavetRootNode<?>> getTupleSourceRootNodes(Class<?> factClass) {
```
triceo: Dtto.

triceo:
Good progress! What do you still need to do here?

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/AbstractForEachUniNode.java
Christopher-Chianelli: There a strong possibility of a performance regression happening if we introduce a new PropagationQueue type, since then the hot-loop of settle inside the NodeNetwork will have three valid interfaces of Propagator, and JVM method dispatch generate less efficient code when a call site can refer to three or move implementations of a class.

Christopher-Chianelli:
Implementation should be done; need more tests and a finalized design.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/BavetRootNode.java
Christopher-Chianelli: I think it is for neighboorhoods; all the failing tests (when a null check is added to AbstractForEach) are in `ai.timefold.solver.core.impl.neighborhood.*`. (`ChangeMoveDefinition`, `ListChangeMoveDefinitionTest`,  `UniEnumeratingStreamTest`  and `solveWithNeighborhoodsListVar`).

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -1582,9 +1583,19 @@ <C> TriConstraintStream<A, B, C> concat(@NonNull TriConstraintStream<A, B, C> ot
             @NonNull Function<A, B> paddingFunction) {
         var firstStream = this;
         var remapped = firstStream.map(ConstantLambdaUtils.biPickFirst());
-        var secondStream = getConstraintFactory().forEach(otherClass)
-                .ifNotExists(remapped, Joiners.equal());
-        return firstStream.concat(secondStream, paddingFunction);
+
+        if (firstStream instanceof AbstractConstraintStream<?> abstractConstraintStream) {
```
triceo, line 1587: When would a `BiConstraintStream` not be an instance of `AbstractConstraintStream`? What else would it be an instance of?

core/src/test/java/ai/timefold/solver/core/impl/score/stream/common/ConstraintStreamFunctionalTest.java
```diff
@@ -174,6 +174,36 @@ default void expandToQuad() {
...
+
+    void precompute_flattenLast();
+
+    void precompute_map();
+
+    void precompute_concat();
+
+    void precompute_distinct();
+
+    void precompute_complement();
```
triceo: These classes are already long enough that I would not put these methods in there. They are different enough (dealing with precomputed streams) to get their own class hierarchy.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetConstraintFactory.java
```diff
@@ -179,6 +202,39 @@ public Predicate<A> apply(@NonNull ConstraintNodeBuildHelper<Solution_, ?> helpe
         }
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public @NonNull <Stream_ extends @NonNull ConstraintStream> Stream_
```
triceo: As we update existing code, I suggest that (as a general rule), we replace `@NonNull` with `@NullMarked`. Most things are non-null these days, and this just creates _a lot_ of boilerplate.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/uni/BavetPrecomputeUniConstraintStream.java
```diff
@@ -0,0 +1,51 @@
...
+        staticConstraintStream.getChildStreamList().add(recordingStaticConstraintStream);
+    }
+
+    public void setAftBridge(BavetAftBridgeUniConstraintStream<Solution_, A> aftStream) {
+        this.aftStream = aftStream;
+    }
+
+    @Override
+    public <Score_ extends Score<Score_>> void buildNode(ConstraintNodeBuildHelper<Solution_, Score_> buildHelper) {
+        var staticDataBuildHelper = new BavetPrecomputeBuildHelper<UniTuple<A>>(recordingStaticConstraintStream);
```
triceo: We no longer use "static data" to describe this feature.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/uni/BavetPrecomputeUniConstraintStream.java
```diff
@@ -0,0 +1,51 @@
...
+import ai.timefold.solver.core.impl.bavet.uni.PrecomputeUniNode;
+import ai.timefold.solver.core.impl.score.stream.bavet.BavetConstraintFactory;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.BavetPrecomputeBuildHelper;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.ConstraintNodeBuildHelper;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.bridge.BavetAftBridgeUniConstraintStream;
+import ai.timefold.solver.core.impl.score.stream.common.RetrievalSemantics;
+
+public class BavetPrecomputeUniConstraintStream<Solution_, A> extends BavetAbstractUniConstraintStream<Solution_, A>
+        implements TupleSource {
+    private final BavetAbstractConstraintStream<Solution_> recordingStaticConstraintStream;
```
triceo: `StaticConstraintStream` is no more.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Override
+    public final void insert(Object a) {
+        // do not remove a retract of the same fact (a fact was updated)
+        queuedInsertSet.add(a);
+    }
+
+    @Override
+    public final void update(Object a) {
+        queuedUpdateSet.add(a);
```
triceo: I wonder... should we prevent the update, if we see that `a` is already inserting? In that case, we only need to track the insert.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
```diff
@@ -0,0 +1,184 @@
...
+        } else {
+            propagationQueue.update(tuple);
+        }
+    }
+
+    @Override
+    public final void retract(Object a) {
+        // remove an insert then retract (a fact was inserted but retracted before settling)
+        // do not remove a retract then insert (a fact was updated)
+        if (!queuedInsertSet.remove(a)) {
```
triceo: Possibly incomplete? If we see a retract, we arguably want to remove any update as well.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
```diff
@@ -0,0 +1,184 @@
...
+    private void invalidateCache() {
+        objectToOutputTuplesMap.values().stream().flatMap(List::stream).forEach(this::retractExisting);
+        inputTupleToOutputTupleMap.clear();
+    }
+
+    private void recalculateTuples() {
+        for (var mappedTupleEntry : objectToOutputTuplesMap.entrySet()) {
+            mappedTupleEntry.getValue().clear();
+            var invalidated = mappedTupleEntry.getKey();
+            recordingTupleNode.startRecording(
```
triceo: Arguably this API could be better. `startRecording` (`recorded`?) could return something `AutoCloseable` (perhaps even the `Recorder` instance which gets created out of sight), turning this into a try-with-resources. That way, it becomes much harder to forget to stop recording, and the `stopRecording` call doesn't even need to exist.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
```diff
@@ -0,0 +1,184 @@
...
+            for (var updatedTuple : objectToOutputTuplesMap.get(updatedObject)) {
+                updateExisting(updatedObject, updatedTuple);
+            }
+        }
+        queuedUpdateSet.clear();
+    }
+
+    private void insertNew(Tuple_ tuple) {
+        var state = tuple.state;
+        if (state != TupleState.CREATING) {
```
triceo: Why would the tuple not be `CREATING` if it's being inserted?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/BavetRootNode.java
triceo: In that case, the arguments need to remain `@Nullable`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/AbstractForEachUniNode.java
triceo: I understand that, but it needs to be weighed against the complexity and understandability of the code. If possible, there should be one way of doing things.

If our problem is that `Propagator` would no longer be bimorphic, then arguably nothing prevents us from refactoring the code such that it always goes through `settle()`. But doesn't that just shift the megamorphic issue to a different place? If we fundamentally have 3 different things to implement, then the optimization doesn't apply to us anyway, and other considerations will win.

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
Christopher-Chianelli: It a public, non-sealed API. A user could in theory implement their own Uni/Bi/Tri/Quad constraint stream. The previous code would have work with said user-implemented constraint stream, hence not failing fast when encountering a user implementation.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
Christopher-Chianelli: Done in settle (removeAll is faster then N removes)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
Christopher-Chianelli: Debating that, although I do know some performance optimizations are turned off inside try bodies

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
Christopher-Chianelli: Done in settle (removeAll is faster than N removes)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractPrecomputeNode.java
Christopher-Chianelli: Usage is ` objectToOutputTuplesMap.values().stream().flatMap(List::stream).forEach(this::insertNew);`; as multiple output tuple lists can share the same output tuple (ex: for a BiStream, the lists for both A and B will have the tuple (A,B)), uniqueness is not guaranteed, and this is probably faster than doing an identity distinct.

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
triceo: I would rather fail fast in that situation; we do not expect people to do this, or rather we actively don't want them to.

core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/AbstractForEachUniNode.java
Christopher-Chianelli: If it were be in `Propagator`, the megamorphic call site would be in the hottest and most frequently looped section of ConstraintStream (called many times per calculation). Where it is now, it is only called once per calculation. And given that  Propagator being bimorphic is explictly called out inside our developer Javadoc (see AbstractNode), I figured the performance was measured to determine method dispatch was taking significant runtime.

triceo: We should be able to find out easily; if this is a performance issue, the score director benchmarks will be super-sensitive to it.

Christopher-Chianelli: The score director benchmarks will not measure it until `precompute` is actually used in the benchmarks (Java cares about what happens at runtime, not about was classes are on the classpath).

triceo: Sounds like a reason to go through the examples and find one that would benefit from `precompute`. Off the top of my head, nurse rostering had some crazy joins in there.

Also, if this is only a problem if precompute is actually used, then this does not cause a regression; we can merge the cleaner version, and if we later find out that we have a performance issue, we address it separately.

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -1582,9 +1583,21 @@ <C> TriConstraintStream<A, B, C> concat(@NonNull TriConstraintStream<A, B, C> ot
...
+            var secondStream = switch (abstractConstraintStream.getRetrievalSemantics()) {
+                case STANDARD, LEGACY -> getConstraintFactory().forEach(otherClass);
```
triceo, line 1589: The only way how the code could get here through `LEGACY` is if the user somehow used `from` instead of `forEach` -  this has been deprecated for so long that I'd rather throw an exception here. Also, it may even be impossible state, because in this particular situation (`complement`) it is us who specify what operation will be used.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -190,7 +191,19 @@ public interface ConstraintFactory {
...
+    <A> BiConstraintStream<A, A> forEachUniquePair(Class<A> sourceClass, BiJoiner<A, A>... joiners);
+
+    // ************************************************************************
+    // staticData
+    //************************************************************************
+
+    /**
+     * Computes and caches the tuples that would be produced by the given stream.
+     * As this is cached, it is vital the stream does not reference any variables
+     * (genuine or otherwise).
```
triceo: I'd mention that if variables are used at any point (maybe even include an example of an offending filter), this will result in score corruptions.

Also, for absolute clarity, we may want to mention that whatever _follows_ the `precompute(...)` can and should include variables.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintFactory.java
```diff
@@ -402,6 +415,6 @@ public interface ConstraintFactory {
      */
     @Deprecated(forRemoval = true)
     @NonNull
```
triceo: Let's review the class for all leftover uses of `@NonNull` now that it's `@NullMarked`.

core/src/main/java/ai/timefold/solver/core/api/score/stream/PrecomputeFactory.java
```diff
@@ -0,0 +1,7 @@
+package ai.timefold.solver.core.api.score.stream;
+
+import ai.timefold.solver.core.api.score.stream.uni.UniConstraintStream;
+
+public interface PrecomputeFactory {
```
triceo, line 13: Since this is public API, it deserves at least the minimal Javadoc.

core/src/main/java/ai/timefold/solver/core/api/score/stream/PrecomputeFactory.java
```diff
@@ -0,0 +1,7 @@
...
+    <A> UniConstraintStream<A> forEachUnfiltered(Class<A> sourceClass);
```
triceo, line 43: Especially this method, which should IMO reiterate that no variables must be used here.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/common/bi/AbstractBiConstraintStreamPrecomputeTest.java
```diff
@@ -0,0 +1,395 @@
...
+
+public abstract class AbstractBiConstraintStreamPrecomputeTest extends AbstractConstraintStreamTest
+        implements ConstraintStreamPrecomputeTest {
+    protected AbstractBiConstraintStreamPrecomputeTest(ConstraintStreamImplSupport implSupport) {
+        super(implSupport);
+    }
+
+    @Override
+    @TestTemplate
+    public void precompute_filter_0_changed() {
```
triceo: For the sake of brevity, I'd remove the `precompute_` prefix; the class name already fulfills that role.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverMultipleSolverAutoConfigurationTest.java
```diff
@@ -297,6 +298,7 @@ void solve() {
     }
 
     @Test
+    @ResourceLock("yaml")
```
triceo: Ah, I see! I fixed several of the same issue, and I clearly missed one spot.
I fixed it in other places by merging the two test methods into one; has the same effect as your solution, but will less lines of code and arguably no worse to read. Let's apply for consistency, please.

triceo (approved):
I love this! Small comments.

Please submit a second PR with docs updates. I'll ask Tom to review that one.

core/src/main/java/ai/timefold/solver/core/api/score/stream/PrecomputeFactory.java
triceo: Also, as I was going through our examples, I immediately ran into the need for `forEachUnfilteredUniquePair` :-)

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
Christopher-Chianelli: Not impossible state, and would be breaking existing behaviour; i.e.

```java
constraintFactory.from(Shift.class)
    .groupBy(Shift::getEmployee, count())
    .complement(Employee.class, employee -> 0)
```
works currently, and would trigger the proposed exception.

PR #1863, opened 2025-10-07
chore: Neighborhoods get forEachAssignablePosition()

triceo:


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListChangeMoveDefinition.java
```diff
@@ -39,50 +38,15 @@ public class ListChangeMoveDefinition<Solution_, Entity_, Value_>
...
-                    var valueCount = solutionView.countValues(variableMetaModel, entity);
-                    if (value == null || valueCount == 0) { // This will trigger assignment of the value at the end of the list.
-                        return ElementPosition.of(entity, valueCount);
-                    } else { // This will trigger assignment of the value immediately before this value.
-                        return solutionView.getPositionOf(variableMetaModel, value);
-                    }
-                })
-                .distinct();
-        // Finally the stream of these positions is joined with the stream of all existing values,
+        var entityValuePairs = moveStreamFactory.forEachAssignablePosition(variableMetaModel);
```
zepfred, line 48: It looks much cleaner now!

PR #1867, opened 2025-10-09
fix: avoid NPE caused by entity filter

zepfred:
This PR includes tests for an issue identified in version `1.26`, which has already been fixed.


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
```diff
@@ -479,7 +479,7 @@ public Object next() {
                 if (isReachable(currentReplayedEntity, next, oneSideValidation)) {
                     return next;
                 }
-            } while (bailoutSize > 0);
+            } while (bailoutSize > 0 && entityIterator.hasNext());
```
triceo: So we do need this change?
Why?

zepfred: No, we don't.

PR #1869, opened 2025-10-10
docs: update performance tips

larsbeck:
This pull request updates documentation and workflow files. The most significant changes include renaming and updating references to the Antora configuration file, refining documentation about constraint usage, and adding new performance tips for score calculation.

**Antora configuration and workflow updates:**

* Renamed `docs/src/antora-template.yml` to `docs/src/antora.yml` and updated comments to clarify its role in the release process.
* Updated workflow scripts in `.github/workflows/pull_request_secure.yml` and `.github/workflows/release.yml` to reference the new `antora.yml` file instead of `antora-template.yml`, ensuring the correct file is used during builds and releases. [[1]](diffhunk://#diff-6b00f696bd20785d1ff5b641f11c39c03fcb31f989db5ba79fcefed7d27b3965L248) [[2]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L68-R68)
* Updated references in documentation to point to `antora.yml` instead of `antora-template.yml`.

**Documentation improvements:**

* Replaced the "Pointless constraints" section with a clearer explanation of how to handle unused constraints using zero constraint weight, improving guidance for users on optimizing performance.
* Added new sections on indexing, hashing, looping, automatic node sharing (Enterprise Edition), benchmarking, and validation.

### DIFF
--- a/docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
+++ b/docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
@@ -304,7 +304,8 @@ The code on the hot path of your application needs to be as fast as possible.
 * If you need to do a lookup in a collection to find an element, ideally you are indexing into an array.
 * The second best option is to use a `HashMap` or `HashSet`.
 * Avoid using `java.util.Stream` or any other form of explicit iteration in constraints, as that is much slower.
-[#automaticNodeSharing]
+
+[#enableAutomaticNodeSharing]
 == Enable automatic node sharing (Enterprise Edition only)
 If you are using the xref:enterprise-edition/enterprise-edition.adoc[Enterprise Edition], you should xref:enterprise-edition/enterprise-edition.adoc#automaticNodeSharing[enable automatic node sharing] as it can significantly speed up score calculation.
 


docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -76,25 +76,9 @@ then turn them into xref:using-timefold-solver/modeling-planning-problems.adoc#c
...
-just return out of the constraint as soon as you can.
-There is no need to dynamically change your score calculation based on the dataset.
-====
-
-In xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams],
-if you set the xref:constraints-and-score/constraint-configuration.adoc#definingAndOverridingConstraintWeights[constraint weight] to zero,
-the constraint will be disabled and have no performance impact at all.
+== Unused constraints
 
+If a constraint is only relevant when certain data is present on the input (for example if you want a capacity constraint to only be evaluated when capacity information is present), set the xref:constraints-and-score/constraint-configuration.adoc#definingAndOverridingConstraintWeights[constraint weight] to zero. In xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams], this will result in the constraint not being added to the node network and therefore have no performance impact at all.
```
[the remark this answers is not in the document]
triceo: Copilot has a point here. In addition, we typically deal with long sentences like so:

    If a constraint is only relevant when certain data is present on the input 
    (for example if you want a capacity constraint to only be evaluated when capacity information is present), 
    set the xref:constraints-and-score/constraint-configuration.adoc#definingAndOverridingConstraintWeights[constraint weight] to zero. 
    In xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams], 
    this will result in the constraint not being added to the node network and therefore have no performance impact at all.

Notice the line breaks, which make the text easier to read while editing.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -307,6 +291,23 @@ leading to much more iteration every time a lesson changes.
 
 For that reason, it is generally recommended putting Joiners based on enum fields or boolean fields last.
 
+[#indexingHashingLooping]
+== Indexing, Hashing, Looping
+The code on the hot path of your application needs to be as fast as possible. When you use xref:using-timefold-solver/modeling-planning-problems.adoc#shadowVariable[Shadow Variables] you will provide a `supplier` method that computes the value of the shadow variable. This method will be called very often during score calculation, so it needs to be fast. If you need to do a lookup in a collection to find an element, ideally you are indexing into an array. The second best option is to use a `HashMap` or `HashSet`. Avoid using `java.util.Stream` or any other form of explicit iteration in constraints as that is much slower.
+
+[#automaticNodeSharing]
+== Enable automatic node sharing (Enterprise Edition only)
+If you are using the xref:enterprise-edition/enterprise-edition.adoc[Enterprise Edition], you should xref:enterprise-edition/enterprise-edition.adoc#automaticNodeSharing[enable automatic node sharing] as it can significantly speed up score calculation.
```
triceo, line 310: Doesn't enterprise enable this automatically?

larsbeck: We set it manually in every model's `application.properties`. [This suggests](https://github.com/TimefoldAI/timefold-solver-enterprise/blob/8c621e92e695173bd58506ce18ff53a0fad43e06/quarkus/deployment/src/main/java/ai/timefold/solver/enterprise/quarkus/deployment/TimefoldEnterpriseProcessor.java#L26-L42) that it is disabled by default.

larsbeck:
> Thank you for the improvements! Please go through the text and add line breaks as suggested.
> 
> Also, the build is currently failing due to errors in documentation. After all those are resolved, LGTM.

If I understand correctly, the build is failing because it runs the old version of the workflow which relies on the (now renamed) `antora-template.yml` to be present. I can re-add (duplicate) that file to make it pass and remove it in a subsequent PR. Or we disable this check for this PR. Let me know what you prefer @triceo .

triceo:
@larsbeck You're probably right about the old version of the workflow. I'll merge this and if it's still broken, I'll figure it out then. But there is a problem in the docs:

    Error:  asciidoctor: WARN: enterprise-edition/enterprise-edition.adoc: line 773: id assigned to section already in use: automaticNodeSharing

PR #1876, opened 2025-10-14
ci: address flaky tests

triceo:
Caches the Spring Boot version from Maven Central.
Improves reliability of barrage test by increasing timeouts and moving it out of the way of other parallel tests.
Addresses Spring Boot context errors when running in parallel by splitting the uberclass into smaller, self-contained parts.


spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverConstraintAutoConfigurationTest.java
```diff
@@ -0,0 +1,173 @@
+package ai.timefold.solver.spring.boot.autoconfigure;
```
zepfred, line 1: Do you think we can apply the same approach to the Quarkus tests?

triceo: I was trying to parallelize the Quarkus test coverage, but Quarkus unfortunately doesn't support it. So I did not touch those tests at all.

PR #1885, opened 2025-10-21
perf: optimize memory usage of precompute

Christopher-Chianelli:
Also settle the node network in Bavet when
the working solution is set to a new instance so
time taken by precomputes are not considered by
terminations.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
@@ -136,7 +136,7 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
             // settle the inner node network, so the inserts/retracts do not interfere
             // with the recording of the first object's tuples
             internalNodeNetwork.settle();
-            recalculateTuples(internalNodeNetwork, recordingTupleLifecycle);
+            recalculateTuples(internalNodeNetwork, objectClassToRootNodes, recordingTupleLifecycle);
 
             propagationQueue.propagateRetracts();
         }
@@ -195,7 +195,9 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
         objectToOutputTuplesMap.clear();
     }
 
-    private void recalculateTuples(NodeNetwork internalNodeNetwork, RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle) {
+    private void recalculateTuples(NodeNetwork internalNodeNetwork,
+            Map<Class<?>, List<BavetRootNode<?>>> classToRootNodeList,
+            RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle) {
         var internalTupleToOutputTupleMap = new IdentityHashMap<Tuple_, Tuple_>(seenEntitySet.size());
         for (var invalidated : seenEntitySet) {
             var mappedTuples = new ArrayList<Tuple_>();
@@ -203,7 +205,7 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
                     new TupleRecorder<>(mappedTuples, internalTupleToOutputTupleMapper, internalTupleToOutputTupleMap))) {
                 // Do a fake update on the object and settle the network; this will update precisely the
                 // tuples mapped to this node, which will then be recorded
-                internalNodeNetwork.getRootNodesAcceptingType(invalidated.getClass())
+                classToRootNodeList.get(invalidated.getClass())
                         .forEach(node -> ((BavetRootNode<Object>) node).update(invalidated));
                 internalNodeNetwork.settle();
             }
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
@@ -6,6 +6,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
 import ai.timefold.solver.core.impl.bavet.NodeNetwork;
@@ -13,6 +14,7 @@ import ai.timefold.solver.core.impl.bavet.common.tuple.AbstractTuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.RecordingTupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleState;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.BavetPrecomputeBuildHelper;
 import ai.timefold.solver.core.impl.util.CollectionUtils;
 
 import org.jspecify.annotations.NullMarked;
@@ -33,23 +35,18 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
     private final Set<Object> updateQueue;
     private final Set<Object> insertQueue;
 
-    private final NodeNetwork internalNodeNetwork;
-    private final RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle;
+    private final Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier;
     private final UnaryOperator<Tuple_> internalTupleToOutputTupleMapper;
-    private final Map<Tuple_, Tuple_> internalTupleToOutputTupleMap;
     private final Map<Object, List<Tuple_>> objectToOutputTuplesMap;
 
     private final StaticPropagationQueue<Tuple_> propagationQueue;
 
     public RecordAndReplayPropagator(
-            NodeNetwork internalNodeNetwork,
-            RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle,
+            Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier,
             UnaryOperator<Tuple_> internalTupleToOutputTupleMapper,
             TupleLifecycle<Tuple_> nextNodesTupleLifecycle, int size) {
-        this.internalNodeNetwork = internalNodeNetwork;
-        this.recordingTupleLifecycle = recordingTupleLifecycle;
+        this.precomputeBuildHelperSupplier = precomputeBuildHelperSupplier;
         this.internalTupleToOutputTupleMapper = internalTupleToOutputTupleMapper;
-        this.internalTupleToOutputTupleMap = CollectionUtils.newIdentityHashMap(size);
         this.objectToOutputTuplesMap = CollectionUtils.newIdentityHashMap(size);
 
         // Guesstimate that updates are dominant.
@@ -61,11 +58,10 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
     }
 
     public RecordAndReplayPropagator(
-            NodeNetwork internalNodeNetwork,
-            RecordingTupleLifecycle<Tuple_> recordingTupleLifecycle,
+            Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier,
             UnaryOperator<Tuple_> internalTupleToOutputTupleMapper,
             TupleLifecycle<Tuple_> nextNodesTupleLifecycle) {
-        this(internalNodeNetwork, recordingTupleLifecycle, internalTupleToOutputTupleMapper, nextNodesTupleLifecycle, 1000);
+        this(precomputeBuildHelperSupplier, internalTupleToOutputTupleMapper, nextNodesTupleLifecycle, 1000);
     }
 
     public void insert(Object object) {
@@ -88,6 +84,10 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
     @Override
     public void propagateRetracts() {
         if (!retractQueue.isEmpty() || !insertQueue.isEmpty()) {
+            var precomputeBuildHelper = precomputeBuildHelperSupplier.get();
+            var internalNodeNetwork = precomputeBuildHelper.getNodeNetwork();
+            var recordingTupleLifecycle = precomputeBuildHelper.getRecordingTupleLifecycle();
+
             updateQueue.removeAll(retractQueue);
             updateQueue.removeAll(insertQueue);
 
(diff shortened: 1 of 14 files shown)


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -89,31 +90,46 @@ public void propagateRetracts() {
...
+            allObjectSet.forEach(this::insertIntoInternalNodeNetwork);
+
             // settle the inner node network, so the inserts/retracts do not interfere
             // with the recording of the first object's tuples
             internalNodeNetwork.settle();
-            recalculateTuples();
+            recalculateTuples(allObjectSet);
+
+            // Free the memory now the internal node network is no longer needed
+            allObjectSet.forEach(this::retractFromInternalNodeNetwork);
```
triceo: IMO the more efficient solution is to throw away the network when no longer needed. 

- You will save all `allObjectSet` collection operations (which won't come cheap), and you will also save all the collections that would be left over in the node network. Sure, they would be empty, but the underlying arrays (in queues, joiners, ...) would still be sized and occupying quite a bit of memory. Also, arguably, solving a memory issue by introducing another massive collection is counterproductive, at least temporarily until the collection is thrown away.
- The retracts themselves are quite a lot of work, which we can possibly avoid.
- Rebuilding the network when needed will be a cost for sure, but we don't actually need it very often. (When does this actually happen? Problem fact changes?) And it is questionable whether rebuilding the network is as expensive as clearing it.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -282,6 +282,15 @@ public final void setWorkingSolution(Solution_ workingSolution) {
...
         forceTriggerVariableListeners();
         expectShadowVariablesInCorrectState = originalShouldAssert;
+        afterSetWorkingSolution();
+    }
+
+    /**
+     * Note: by default does nothing. Subclasses should override this if they
+     * need to compute something after shadow variables are set.
+     */
+    public void afterSetWorkingSolution() {
```
triceo: Since this is being called from within the score director, we can make this `protected`. Let's not expose methods we don't need to.

triceo (commented):
Not approving yet, since I think that the node network reset issue can possibly lead to large changes to this PR.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -87,33 +84,51 @@ public void retract(Object object) {
...
 
-            retractQueue.forEach(this::retractFromInternalNodeNetwork);
-            insertQueue.forEach(this::insertIntoInternalNodeNetwork);
             retractQueue.clear();
             insertQueue.clear();
 
+            for (var object : allObjectSet) {
+                internalNodeNetwork.getRootNodesAcceptingType(object.getClass())
+                        .forEach(node -> ((BavetRootNode<Object>) node).insert(object));
+            }
```
triceo: This is a major source of inefficiency.
The determination of which nodes accept the type happens over and over, every time we process one of the hundreds of thousands of objects.

If we already have to aggregate all the objects, then we might as well aggregate them into a map of sets, where the key is the class. In that way, the overhead of node processing will be much less. It may prevent us from properly sizing the sets, but I consider that only a small downside.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -87,33 +84,51 @@ public void retract(Object object) {
...
             updateQueue.removeAll(insertQueue);
+
+            var allObjectSet = CollectionUtils
+                    .newIdentityHashSet(insertQueue.size() + objectToOutputTuplesMap.size());
+
             // Do not remove queued retracts from inserts; if a fact property
             // change, there will be both a retract and insert for that fact
+
+            allObjectSet.addAll(objectToOutputTuplesMap.keySet());
+            allObjectSet.removeAll(retractQueue);
```
triceo: I am not so sure this is as efficient as we think.

In order to be able to do a bulk `addAll`, we have to do `removeAll` with some other collection. IMO a much simpler "add if not contains" in a loop would result in overall less `add` operations, with the amount of `contains` remaining the same (since `removeAll` does something similar anyway).

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -87,33 +84,51 @@ public void retract(Object object) {
     @Override
     public void propagateRetracts() {
```
triceo, line 95: In the JFR recording you sent, this method takes up easily 50 % of the flame graph. Stands to reason that optimizing this method would bring the biggest benefits, until something else becomes the dominator.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -144,36 +190,28 @@ private void retractIfPresent(Tuple_ tuple) {
...
+        var internalTupleToOutputTupleMap = new IdentityHashMap<Tuple_, Tuple_>(seenEntitySet.size());
+        for (var invalidated : seenEntitySet) {
+            var mappedTuples = new ArrayList<Tuple_>();
             try (var unusedActiveRecordingLifecycle = recordingTupleLifecycle.recordInto(
-                    new TupleRecorder<>(mappedTupleEntry.getValue(), internalTupleToOutputTupleMapper,
-                            (IdentityHashMap<Tuple_, Tuple_>) internalTupleToOutputTupleMap))) {
+                    new TupleRecorder<>(mappedTuples, internalTupleToOutputTupleMapper, internalTupleToOutputTupleMap))) {
                 // Do a fake update on the object and settle the network; this will update precisely the
                 // tuples mapped to this node, which will then be recorded
                 internalNodeNetwork.getRootNodesAcceptingType(invalidated.getClass())
```
triceo: We're caching this above, why not here?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/RecordingTupleLifecycle.java
```diff
@@ -23,7 +23,13 @@ public void close() {
     @Override
     public void insert(Tuple_ tuple) {
         if (tupleRecorder != null) {
-            throw new IllegalStateException("Impossible state: tuple %s was inserted during recording".formatted(tuple));
+            throw new IllegalStateException("""
```
triceo, line 26: I assume we're addressing this in a follow-up PR?

triceo (approved):
LGTM when comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/RecordingTupleLifecycle.java
Christopher-Chianelli: This is basically unique to flattenLast, since it the only operation I am aware of that can do inserts on an update (primary, when different instances are detected. For example:

```java
record ShiftLocationPair(Shift shift, Location location) {}

precomputeFactory.forEachUnfiltered(Shift.class)
     .flattenLast(shift -> shift.getLocations()
             .stream()
             .map(location -> new ShiftLocationPair(shift, location))
             .toList()
     )
```

triceo: I know, but I assumed that we were going to solve this in a follow-up PR. Is that not the case? 

IMO the solution you proposed - the distinct approach - is not acceptable. Are you saying there isn't any other one?

Christopher-Chianelli: As for how address it, we either treat the entire precompute as distinct (so we can identify tuple by their contents), or we treat it as an illegal argument to precompute. To see the particular problem, imagine we have a very bad flattenLast:

```java
constraintFactory.forEachUnfiltered(Shift.class)
     .flattenLast(shift -> List.of(new Random().nextInt()))
```

The list contents differs, so each update produces new tuples.

We could try to make flattenLast try to identify the tuple via iteration order in precompute, but that relies on iteration order being consistent, and flattenLast can be used with any iterable, including sets.

triceo: I've been thinking for a long time that `flattenLast` needs a redesign. The implementation is horrible - it works, but arguably it could be better. And this particular problem could/should be fixed there.

If we once produced a tuple for element X, and we see element X in the new collection, we should not be triggering a new tuple. If we originally produced 2 tuples for element X and now we only see one element X, then we of course need to retract one. Whether tuples are triggered new, or the existing ones updated, is an implementation detail - maybe we should reconsider how `flattenLast` does things.

Wrt. your example - I agree that it is bad, and it is bad for more reason than one. This cannot be precomputed, because _this is not static_. Random is ultimately a variable, and therefore doesn't meet the basic requirements. Let's discuss a different example, one that can actually happen. The idea is that the collection will always produce _the same thing_ - otherwise it is not static.

Christopher-Chianelli: Issue is, with mapping functions, we can get equal but different references. For example:
```java
record ShiftHolder(Shift shift) {}
constraintFactory.forEachUnfiltered(Shift.class)
     .flattenLast(shift -> List.of(new ShiftHolder(shift)))
```

Currently, this does a retract and an insert on an update (because different ShiftHolder instances are returned).

triceo: This seems to be a combination of two issues:

- This is the user not respecting the contract; if it's not the same thing, then the stream is not static.
- But this is _partially_ our fault - as explained above. `flattenLast` _could_ send updates instead of insert+retract, if we make it so.

Christopher-Chianelli: FWIW, `flattenLast` already do updates if the reference in the collection are the same (i.e.
```java
constraintFactory.forEachUnfiltered(Shift.class)
     .flattenLast(shift -> List.of(shift))
```
)

If we do updates even when references are not the same, we depend on iteration order, which would break things like this:

```java
constraintFactory.forEachUnfiltered(Shift.class)
     .flattenLast(shift -> {
        // Fact has no equals or hashCode, so identity hash code is used
        var a = new Fact(shift.getA());
        var b = new Fact(shift.getB());
        var set = new HashSet<Fact>();
        set.add(a);
        set.add(b):
        return set;
})
```

(i.e. one iteration, a might be first, another, b might be first).

triceo: The iteration order shouldn't matter though - that's why I'm saying `flattenLast` is currently wrong.

Assume that originally we inserted X, X, Y, X.
Now at update, we see X, Y, X.

What this tells me is that we need to retract X once. (Doesn't matter which, they are all the same X.) The other 2 Xs and the Y need to be updated.

I don't see why the iteration order should matter here. And if it does, then arguably `precompute` requires `LinkedHashSet` or `List`.

Christopher-Chianelli: 1. Do note the X and Y we are talking about here are NEW INSTANCES of equal value. Iteration order that we are discussing is to be used as a proxy for what should be equivalent. i.e. we are talking about flattenLast(x -> List.of(new Holder(x)), and not flattenLast(x -> List.of(x))

If we inserted `X, X, Y, X` and then it was `X, Y, X` on update, then the stream being precomputed returned different values.

The issue is we need to identify shared tuples; for example:

```java
precomputeFactory.forEachUnfilteredUniquePair(Shift.class)
    .map(Pair::of)
    .flattenLast(pair -> List.of(pair));
```

with two shifts A and B, there is one output tuple (A, B), and an update on either A or B updates that output tuple, and it should only penalize by 1, not two. We use `update` when precomputing to discover what tuples A/B directly affects.

Assume on insert,  flattenLast returned X, Y (stored in tuples Tx and Ty respectively), Tx pass a filter, Ty does not. When the node network is updated to query A tuples, flattenLast returned X, Y, so A get Tx as its tuple. When the node network is updated to query B tuples, flattenLast returned Y, X, so B get Ty as its tuple. And thus a penalty of 2 instead of 1, since A and B were both suppose to get Tx.

Christopher-Chianelli: As for how flattenLast currently operates on update, it keep track of how many times it encountered a duplicate reference (i.e. by id/pointer), having a separate tuple for each duplicate. If more duplicates/a new item are encountered, new tuples are created. If less duplicates are encountered, some tuples are retracted.

triceo: So, talking about this specifically:

    precomputeFactory.forEachUnfilteredUniquePair(Shift.class)
        .map(Pair::of)
        .flattenLast(pair -> List.of(pair));

We receive a `Pair(a, b)`, such that:

- `a_then.equals(a_now)` is true,
- `b_then.equals(b_now)` is true,
- therefore `pair_then.equals(pair_now)` is true.

`flattenLast` sees that the pairs are equal, and therefore should trigger an update, not an insert+retract. If that is already the case, I don't understand what we're talking about. If that is not the case, then `flattenLast` ought to be fixed.

Christopher-Chianelli: But we see `pair_then != pair_now`; `flattenLast` uses reference equality, not object equality.

triceo: And that's the problem. :-) It shouldn't. Much like `groupBy` doesn't.

Christopher-Chianelli: Relevant section: the choice to use reference equality instead of object equality was deliberate to avoid depending on user equality: https://github.com/TimefoldAI/timefold-solver/blob/9eb30e870385c8b3e58ab39e902d99256ad521c5/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenLastNode.java#L94-L103

PR #1900, opened 2025-10-28
chore: improve sorting operation

zepfred:
This PR enables sorting the value ranges at the `ValueRangeManager` level. 

The sorting feature is likely useful during the CH phase, and the new approach can address the sorting requirement for the existing CH configurations. However, the existing selector sorting nodes remain available, as they are required for range filtering or nearby selectors. Also, these corner cases may only make sense for the LS phases.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
@@ -198,7 +198,7 @@ public final class ReachableValues<Entity_, Value_> {
         List<Value_> getRandomAccessValueList(List<ReachableItemValue<Entity_, Value_>> allValues,
                 @Nullable ValueRangeSorter<Value_> valueRangeSorter) {
             var valuesList = new ArrayIndexedList<>(extractAllIndexes(valueBitSet), allValues,
-                    v -> (Value_) v.value);
+                    v -> v.value);
             if (valueRangeSorter != null) {
                 valueRangeSorter.sort(valuesList);
             }
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
@@ -335,7 +335,7 @@ final class ValueRangeState<Solution_, Entity_, Value_> {
             loadEntityValueRange(i, valueIndexMap, valueRange, reachableValueList);
         }
         var sorterAdapter = sorter != null ? SelectionSorterAdapter.of(cachedWorkingSolution, sorter) : null;
-        return new ReachableValues(entityIndexItem, valueIndexItem, expectedTypeOfValue, sorterAdapter,
+        return new ReachableValues<>(entityIndexItem, valueIndexItem, expectedTypeOfValue, sorterAdapter,
                 variableDescriptor.getValueRangeDescriptor().acceptsNullInValueRange());
     }
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByEntitySelector.java
@@ -455,7 +455,7 @@ public final class FilteringEntityByEntitySelector<Solution_> extends AbstractDe
             if (value != null) {
                 this.oneSideValidation = true;
                 var entityList = reachableValues.extractEntitiesAsList(value);
-                this.entityIterator = new RandomListIterator(entityList, workingRandom);
+                this.entityIterator = new RandomListIterator<>(entityList, workingRandom);
                 this.iterationBailoutSize = entityList.size();
             } else {
                 this.oneSideValidation = false;


zepfred:
@triceo The PR is ready for review and will rerun all previous model validations, just like we did when adding the sorting feature.

zepfred:
There is an issue with the quickstart compilation job, and I will take a look.

core/src/main/java/ai/timefold/solver/core/api/domain/valuerange/ValueRange.java
```diff
@@ -45,6 +46,14 @@ public interface ValueRange<T> {
      */
     boolean contains(@Nullable T value);
 
+    /**
+     * The sorting operation copies the current value range and sorts it using the provided sorter.
+     *
+     * @param sorter never null, the value range sorter
+     * @return A new instance of the value range, with the data sorted.
+     */
+    ValueRange<T> sort(ValueRangeSorter<T> sorter);
```
triceo: I do not think this should exist. `ValueRange` is public API, `ValueRangeSorter` is not. You cannot reference non-public API from public API.

Also, looking at what the method actually does in implementations, there is no reason why we couldn't do this without exposing this method. We do not expect nor want people to implement their own `ValueRange` implementations, therefore we are in full control over what happens. (Essentially, we can treat `ValueRange` as a sealed type, even though it is not one.)

I do not want `ValueRange` to evolve or change. I want to remove it in Solver 2.0 - nobody uses it.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
```diff
@@ -116,4 +135,9 @@ public <Value_> ValueRangeCache<Value_> buildCache(Collection<Value_> collection
 
     }
 
+    private enum CacheType {
```
triceo: `CacheType` and `Builder` can be merged. 
Both are enums, both mean the exact same thing.
This is just a needless obfuscation.

Alternatively, `cacheType` can become a boolean.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/SelectionSetSorter.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Decides the order of a {@link Set} of selection values.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ * @param <T> the selection type
+ */
+@NullMarked
+@FunctionalInterface
+public interface SelectionSetSorter<Solution_, T> {
```
triceo: Not sure why this needs to exist.
Why not have `SelectionSorter` accept and return a `Collection`?
Strangely, `SelectionSorter` doesn't return _anything_...

Also, we have the comparator factory class now. Why do we even have this type anymore?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
```diff
@@ -107,24 +114,22 @@ public Iterator<Object> endingIterator(Object entity) {
...
-        return randomSelection == that.randomSelection && Objects.equals(valueRangeDescriptor, that.valueRangeDescriptor);
+        return randomSelection == that.randomSelection && Objects.equals(valueRangeDescriptor, that.valueRangeDescriptor)
+                && Objects.equals(selectionSorter, that.selectionSorter);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(valueRangeDescriptor, randomSelection);
+        return Objects.hash(valueRangeDescriptor, selectionSorter, randomSelection);
     }
```
triceo, line 125: Please keep the order of fields consistent between equals and hashcode.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -122,10 +122,19 @@ public ValueSelector<Solution_> buildValueSelector(HeuristicConfigPolicy<Solutio
...
-                        randomSelection);
         var instanceCache = configPolicy.getClassInstanceCache();
+        /*
+         * When a filtering value range or nearby are used,
+         * we opt to sort the data using a sorting node selector instead of sorting at the value range level.
+         * This choice is required
+         * because the FilteringValueRangeSelector or the related Nearby selector does not respect the sort order
+         * provided by the child selectors.
+         */
+        var canSortAtValueRangeLevel = entityValueRangeRecorderId == null && nearbySelectionConfig == null;
```
triceo: So... let's assume I have provided the nearby distance meter, and nothing else. (No selector configuration.) Does nearby use value ranges?

If not, then I wonder why we're really doing this. The idea of value ranges is that there would be _one_ place where all of this happens, but if something as critical as nearby still ignores value ranges, then arguably the idea failed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -260,8 +271,45 @@ private static Class<? extends ComparatorFactory> determineComparatorFactoryClas
         }
     }
 
+    private SelectionSorter<Solution_, Object> determineSorter(GenuineVariableDescriptor<Solution_> variableDescriptor,
+            SelectionOrder resolvedSelectionOrder, ClassInstanceCache instanceCache) {
+        SelectionSorter<Solution_, Object> sorter = null;
+        if (resolvedSelectionOrder == ai.timefold.solver.core.config.heuristic.selector.common.SelectionOrder.SORTED) {
```
triceo: Please remove the FQN.

triceo: Also, the condition can be inverted so that literally the entire method isn't nested in a block.

triceo:
> There is an issue with the quickstart compilation job, and I will take a look.

The Vert.x exception is a red herring. Your real problem is here:

    025-11-04 07:29:04,008 ERROR [org.acm.pro.res.ProjectJobSchedulingResource] (pool-8-thread-1) Failed solving jobId (52a73dc5-6458-4eda-bb33-61a1b2a59fd4).: java.lang.UnsupportedOperationException
	at ai.timefold.solver.core.impl.domain.valuerange.AbstractCountableValueRange.sort(AbstractCountableValueRange.java:44)

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -376,26 +394,43 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
                 fromEntityMap.computeIfAbsent(entity,
-                        e -> new CountableValueRange[solutionDescriptor.getValueRangeDescriptorCount()]);
-        var valueRange = valueRangeList[valueRangeDescriptor.getOrdinal()];
-        if (valueRange == null) { // Avoid computeIfAbsent on the hot path; creates capturing lambda instances.
+                        e -> new CountableValueRangeItem[solutionDescriptor.getValueRangeDescriptorCount()]);
+        var item = valueRangeList[valueRangeDescriptor.getOrdinal()];
+        var valueRange = item != null ? (CountableValueRange<T>) item.countableValueRange() : null;
+        var valueRangeSorter = item != null ? item.sorter() : null;
+        // Avoid computeIfAbsent on the hot path; creates capturing lambda instances.
+        // We read and sort the data again if needed
```
triceo: What computeIfAbsent? I don't see any map here.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -376,26 +394,43 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+        if (valueRange == null || (sorter != null && !Objects.equals(valueRangeSorter, sorter))) {
```
triceo: This is weird. Why would the same value range ever be sorted with a different sorter? Why does the code even make this possible?

core/src/main/java/ai/timefold/solver/core/api/domain/valuerange/ValueRange.java
zepfred: We need `ValueRangeSorter` because `SelectionSorter` requires the solution. Since the operation occurs within the value range implementations, the solution is not accessible. The implementation `SelectionSorterAdapter` addresses that.

Yep, we don't need to expose the sorting operation in the public API.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/decorator/SelectionSetSorter.java
zepfred: > Not sure why this needs to exist.

The interface was created to handle the set sorting because `SelectionSorter` only sorts lists. I thought of two solutions to delete this class.
1. Add a default method that accepts a `Set`, creates an `ArrayList` from it, sorts it, and returns a `Set` using `Set.copyOf`. This implementation is simpler but less efficient. However, I'm unsure if it will cause any speed regression.
2. Add a new method that accepts a `Set` and allows child classes to implement it. Since the child classes have access to the comparator, we can use the `TreeSet` without copying the data twice.

> Why not have SelectionSorter accept and return a Collection?

It might work, and I'll give it a try. This also means we'll need to refactor all places that expect an in-place sorting operation. That adds a third option:

3. Update the `SelectionSorter` contract to accept and return a `Collection`.

> Also, we have the comparator factory class now. Why do we even have this type anymore?

I don't get what you're asking.

zepfred: > Why not have SelectionSorter accept and return a Collection?

We can't do that because some selectors depend on the collection type being a `List`. Operations such as `listIterator()` cannot be performed with the base `Collection` type.

triceo: My overall point is that the `SelectionSorter` interface is weird. We now have a `ComparatorFactory`, or we can just use a `Comparator`. They can be used with any collection, not just with a list, or with a set.

It seems that `SelectionSorter` - because of its weird interface - creates a lot of complexity and non-standard code.

zepfred: I have updated the `SelectionSorter` contract with the new method. This makes the implementation clearer. Indeed, the `Collection` could be the only type, but we rely on the list in other parts of the solver.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/FromEntityPropertyValueSelector.java
zepfred: The code was generated by IntelliJ.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: Some comments from previous implementations.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: The phase initialization logic can call operations like `countOnSolution` or `countOnEntity`, which do not consider sorting and initialize the value range without a sorter. Later, the phase will request the value range with a sorting requirement. The condition makes it possible to apply sorting when needed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
zepfred: > So... let's assume I have provided the nearby distance meter, and nothing else. (No selector configuration.) Does nearby use value ranges?

I'm unsure if this is even possible. The point is, if there is a nearby configuration, then it means we will create nodes that ignore the sorting provided by the child selectors.

> If not, then I wonder why we're really doing this. The idea of value ranges is that there would be one place where all of this happens, but if something as critical as nearby still ignores value ranges, then arguably the idea failed.

I see your point and will work on a solution. That means we will need to pass the sorter to the range filtering and nearby nodes.

triceo:
I'm thinking we're approaching this all wrong.

We're building this hierarchy of selectors, decorating with additional behaviors. But value ranges shouldn't work like that. Value ranges are meant to be the _root_ selector. When sorting is necessary, it should be applied to the _root_ selector directly, not via some decorator later. The root iteration should already happen in the sorted order.

How do values come into the value range selector? Who puts them there? That is where we need to start - and that source of data needs to be the value range manager. Which will already have everything sorted. And then we can start building all the other functionality on top of that - the filter decorator etc., all of that can be layered on top of what's already sorted.

This is a large refactor, and maybe we don't want to do it right now. But this is the way to make sure that value range manager is the only source of truth for data. And consequently, it is the only way how we can safely produce value ranges which change with their entity variables.

zepfred:
> I'm thinking we're approaching this all wrong.
> 
> We're building this hierarchy of selectors, decorating with additional behaviors. But value ranges shouldn't work like that. Value ranges are meant to be the _root_ selector. When sorting is necessary, it should be applied to the _root_ selector directly, not via some decorator later. The root iteration should already happen in the sorted order.
> 

That's the current approach. The base method `buildBaseValueSelector` creates the root node, which uses the sorter and fetches the sorted value range when needed. I'm currently working to remove the step `applySorting` when filtering range nodes or nearby selectors are used.

> How do values come into the value range selector? Who puts them there? That is where we need to start - and that source of data needs to be the value range manager. Which will already have everything sorted. And then we can start building all the other functionality on top of that - the filter decorator etc., all of that can be layered on top of what's already sorted.
> 

Yes, that's correct. However, the range filtering and the nearby nodes ignore the root selector sort order because they use different structures to fetch the values: reachable values and the distance matrix.

> This is a large refactor, and maybe we don't want to do it right now. But this is the way to make sure that value range manager is the only source of truth for data. And consequently, it is the only way how we can safely produce value ranges which change with their entity variables.

I have a solution to maintain the `ValueRangeManager` as the single source of truth for the reachable values structure and have not yet looked into the nearby selector.

zepfred:
@triceo I'll begin reviewing the necessary changes to the enterprise repository, and `ValueSelectorFactory` will also be modified to pass the sorter to the inner nodes.

This change is needed because both `FilteringValueRangeSelector` and the nearby nodes use a different structure to compute the values list, and they might not follow the sort order from the root nodes.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -124,17 +146,40 @@ public ReachableItemValue(Object value, int entityListSize, int valueListSize) {
...
                 randomAccessValueList.add(value);
             }
         }
+
+        private void checkSorting(@Nullable ValueRangeSorter<V> valueRangeSorter) {
+            if (valueRangeSorter != null && !sorted) {
+                var sortedList = valueRangeSorter.sort(randomAccessValueList);
+                randomAccessValueList.clear();
+                randomAccessValueList.addAll(sortedList);
+                sorted = true;
```
triceo: This is inefficient. First we create a sorted list. Then we clear the other list, and add all elements in order. 

IMO the sorting has to be designed differently:
- The list first has to be copied.
- And then sorted in place.

Sorting a copy may sometimes be necessary. But often, sorting in place is possible, and much more efficient.

triceo: Also, in this particular case, if `randomAccessValueList` were not `final`, you could have avoided all the clearing and adding as well.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/SortingEntitySelector.java
```diff
@@ -60,7 +60,8 @@ public void constructCache(SolverScope<Solution_> solverScope) {
             return;
         }
         super.constructCache(solverScope);
-        sorter.sort(solverScope.getScoreDirector(), cachedEntityList);
+        // We need to update the cachedEntityList since the sorter will copy the data and return a sorted list
+        cachedEntityList = sorter.sort(solverScope.getScoreDirector().getWorkingSolution(), cachedEntityList);
```
triceo: This is one place where the original code was being more efficient, because it was sorting in place.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/decorator/SortingMoveSelector.java
```diff
@@ -25,7 +25,8 @@ public SortingMoveSelector(MoveSelector<Solution_> childMoveSelector, SelectionC
     @Override
     public void constructCache(SolverScope<Solution_> solverScope) {
         super.constructCache(solverScope);
-        sorter.sort(solverScope.getScoreDirector(), cachedMoveList);
+        // We need to update the cachedMoveList since the sorter will copy the data and return a sorted list
+        cachedMoveList = sorter.sort(solverScope.getScoreDirector().getWorkingSolution(), cachedMoveList);
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/IterableFromSolutionPropertyValueSelector.java
```diff
@@ -153,19 +161,16 @@ private void checkCachedEntityListIsDirty() {
...
-        return randomSelection == that.randomSelection &&
-                Objects.equals(valueRangeDescriptor, that.valueRangeDescriptor) && minimumCacheType == that.minimumCacheType;
+        return randomSelection == that.randomSelection && Objects.equals(valueRangeDescriptor, that.valueRangeDescriptor)
+                && Objects.equals(selectionSorter, that.selectionSorter) && minimumCacheType == that.minimumCacheType;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(valueRangeDescriptor, minimumCacheType, randomSelection);
+        return Objects.hash(valueRangeDescriptor, selectionSorter, minimumCacheType, randomSelection);
```
triceo: Let's always make sure that the order of fields in equals is the same as in hashcode.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -359,11 +358,41 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_> valueRangeDescriptor, Solution_ solution,
+            @Nullable SelectionSorter<Solution_, T> sorter) {
+        var item = fromSolution[valueRangeDescriptor.getOrdinal()];
+        var valueRange =
+                item != null ? (CountableValueRange<T>) item.countableValueRange() : null;
+        var valueRangeSorter = item != null ? item.sorter() : null;
+        // The phase initialization logic can call operations like countOnSolution or countOnEntity,
+        // which do not consider sorting and initialize the value range without a sorter.
+        // Therefore, we return the range if there is no sorter or applied sorter is the same as the given sorter
```
triceo: If we know that there will be a sorter in the future (which we do know), why not just sort this the first time it's called, and avoid all of this complexity?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -103,19 +110,34 @@ public boolean acceptsNullValue() {
...
+
+    public ReachableValues<E, V> copy(@Nullable ValueRangeSorter<V> valueRangeSorter) {
+        Map<V, ReachableItemValue<E, V>> newValues = ConfigUtils.isGenericTypeImmutable(valueClass)
+                ? new HashMap<>(values.size())
+                : new IdentityHashMap<>(values.size());
+        for (Map.Entry<V, ReachableItemValue<E, V>> entry : values.entrySet()) {
+            newValues.put(entry.getKey(), entry.getValue().copy());
+        }
+        return new ReachableValues<>(newValues, valueClass, valueRangeSorter, acceptsNullValue);
+    }
```
triceo: This is incredibly resource-hungry. Not only do we copy the entire map. We also call `copy` on the values, which will in turn copy all the nested collections! This operation has the potential of completely ruining the performance of anything where it's called. These collections will have thousands of elements.

Consider the following:

- We know that there will be a sorter, because we know the solver config. Therefore `ValueRangeManager` can be created already with the final version of the sorter.
- Value ranges do not change, therefore sorting is a one-time operation. Once sorted, the values will not change, and therefore neither does their order. (This will eventually change, but we will deal with that then; for now, value ranges are static.)
- For operations such as count, it does not matter if values are sorted or not. Therefore they may as well be sorted from the beginning.

From these three facts, it follows that all of this should be created, immediately sorted in-place, and never touched after that. Which would also significantly simplify the code, because the needless state changes create a lot of complexity.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/IterableFromSolutionPropertyValueSelector.java
zepfred: IntelliJ generated that code. Perhaps comparing the primitive fields first represents some kind of optimization.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
zepfred: Good point!

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/IterableFromSolutionPropertyValueSelector.java
triceo: It does. But then it should also come first in the hashcode.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
```diff
@@ -89,6 +89,12 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
         this.childEntitySelector.solvingStarted(solverScope);
     }
 
+    @Override
+    public void solvingEnded(SolverScope<Solution_> solverScope) {
```
triceo: `FilteringEntityByEntitySelector` doesn't seem to do this. Why? And why was it necessary here, now?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelector.java
```diff
@@ -78,6 +79,19 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+        super.solvingEnded(solverScope);
+        this.listVariableStateSupply = null;
+    }
+
+    @Override
+    public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
+        super.phaseEnded(phaseScope);
+        this.solverScope = null;
+        this.workingRandom = null;
+    }
```
triceo, line 93: We are suddenly adding this to all the selectors, and I have to ask... why now? Why was it not a problem ever before, in the 10+ years that the selector code has existed? 

This smells. We are introducing new behavior, IMO without fully understanding why it worked perfectly fine without it.

core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
```diff
@@ -130,6 +130,10 @@ public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
         if (assertPhaseScoreFromScratch) {
             var score = phaseScope.getSolverScope().calculateScore();
             try {
+                // Before making the assertion, we reset the value manager to release resources,
+                // as the assertion will need to reallocate the solution range,
+                // and it may lead to OOM errors.
+                phaseScope.getScoreDirector().getValueRangeManager().reset(null);
```
triceo, line 136: And what happens _after_ the assertion is finished? Do we suddenly need to recompute everything again?

This is a performance killer. Effectively, this is a decision for assert modes to be so slow that they are effectively unusable. (And they are already slow now.)

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -449,6 +449,7 @@ public void close() {
         if (listVariableStateSupply != null) {
             getSupplyManager().cancel(listVariableStateSupply.getSourceVariableDescriptor().getStateDemand());
         }
+        valueRangeManager.reset(null);
```
triceo: I don't think this should be necessary. When ScoreDirector instance is thrown away, this instance will be thrown away as well. GC will handle this.

This is just extra code that accomplishes the default behavior.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+    public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_> valueRangeDescriptor, Solution_ solution,
+            @Nullable SelectionSorter<Solution_, T> sorter) {
+        var item = fromSolution[valueRangeDescriptor.getOrdinal()];
+        // No item, we set the left side by default
+        if (item == null) {
+            var valueRange = fetchValueRangeFromSolution(valueRangeDescriptor, solution, sorter);
+            fromSolution[valueRangeDescriptor.getOrdinal()] = new ValueRangeItem<>(null, valueRange, sorter, null, null);
+            Class<?> valueClass = findValueClass(valueRange);
+            var valueIndexMap = buildIndexMap(valueRange.createOriginalIterator(), (int) valueRange.getSize(),
+                    ConfigUtils.isGenericTypeImmutable(valueClass));
```
triceo: Arguably you don't need this extra variable, because you're casting it on the next line anyway. This line does not give any extra information.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+            return valueRange;
+        } else {
+            throw new IllegalStateException(
+                    "Impossible state: the value range (%s) with sorter (%s) does not align with the existing ascending (%s) and descending (%s) sorters."
+                            .formatted(valueRangeDescriptor, sorter, leftSorter, rightSorter));
+        }
+    }
+
+    private Map<Object, Integer> getIndexMapFromSolution(ValueRangeDescriptor<Solution_> valueRangeDescriptor) {
+        getFromSolution(valueRangeDescriptor);
```
triceo: This is not particularly nice. This is a method that does something, doesn't say what it does, and doesn't return anything.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+            throw new UnsupportedOperationException("""
+                    Impossible state: value range (%s) on planning solution (%s) is not countable.
+                    Maybe replace %s with %s."""
+                    .formatted(valueRangeDescriptor, solution, DoubleValueRange.class.getSimpleName(),
+                            BigDecimalValueRange.class.getSimpleName()));
+        } else if (valueRangeDescriptor.acceptsNullInValueRange()) {
+            valueRange = new NullAllowingCountableValueRange<>(countableValueRange);
+        } else {
+            valueRange = countableValueRange;
+        }
```
triceo: I'd extract this to a separate method. That way, you get to return the value range directly, avoiding the need for line 445 and at the same time making smaller, self-contained methods.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
     @SuppressWarnings("unchecked")
     public <T> CountableValueRange<T> getFromEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor, Object entity) {
+        return getFromEntity(valueRangeDescriptor, entity, null);
+    }
+
+    /**
+     * @throws IllegalStateException if called before {@link #reset(Object)} is called
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public <T> CountableValueRange<T> getFromEntity(ValueRangeDescriptor<Solution_> valueRangeDescriptor, Object entity,
```
triceo: This method is very long. Surely there are self-contained pieces that could be split off into other methods and called from here.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+                        "Impossible state: value range (%s) on planning entity (%s) is not sortable."
+                                .formatted(valueRangeDescriptor, entity));
+            }
+            var sorterAdapter = SelectionSorterAdapter.of(cachedWorkingSolution, sorter);
+            valueRange = (CountableValueRange<T>) sortableValueRange.sort(sorterAdapter);
+        }
+        return valueRange;
+    }
+
+    private static <T> BitSet getBitSetValueRange(CountableValueRange<T> valueRange, Map<Object, Integer> valueIndexMap) {
```
triceo: This method will be very expensive. It happens for every single value range, it iterates over it (possibly 20k items) and for each item it will do a hash lookup. If there is one place that would benefit from micro-optimization, it is this one. And then when the lookup is done, there will be a conversion from `Integer` to `int`.

My bet is that this is going to jump out in profiling, and by far.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +365,255 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+     * @param entity the entity
+     * @param valueRange the entity value range
+     * @return returns an existing entity with a matching value range, or returns null if it does not exist.
+     */
+    private @Nullable <T> Object findEntityBitSetMatch(ValueRangeDescriptor<Solution_> valueRangeDescriptor, Object entity,
+            int entitySize, CountableValueRange<T> valueRange) {
+        // We create a BitSet from the range values
+        // and check if another identical range already exists to prevent duplication
+        var valueIndexMap = getIndexMapFromSolution(valueRangeDescriptor);
+        var valueRangeBitSet = getBitSetValueRange(valueRange, valueIndexMap);
```
triceo: I wonder... are there situations where we could avoid this?

We know the size of the collection; if the sizes differ, they are different collections. We also know the identity of the collection; if they are the same object, they are by definition duplicates and we don't need to compare bitsets.

Any way how we could use that information for our benefit?

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -509,22 +766,32 @@ private static <T> void loadEntityValueRange(int entityIndex, Map<Object, Intege
...
-            }
+            // We unset the current value index to import only the values that are reachable
+            allValuesBitSet.clear(valueIndex);
+            item.addValues(allValuesBitSet);
+            // We set it again to restore the state for the next value
+            allValuesBitSet.set(valueIndex);
         }
     }
 
     public void reset(@Nullable Solution_ workingSolution) {
```
triceo, line 209: I question the need for this.
This is a very expensive thing. All that array filling, all that iteration.
Only to do the same thing that the GC would do as well, for almost free.

Don't reset, throw away instead. Unless you've proven that this is actually necessary, and better than the alternative.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -61,10 +65,11 @@
 public final class ValueRangeManager<Solution_> {
```
triceo, line 42: This class has become extremely complex and barely readable. It needs refactoring, breaking down into smaller logical units, and composing them back up.

Consider this code, purely as one example:

        var fromEntity = fromEntityBitSet[valueRangeDescriptor.getOrdinal()];
        if (fromEntity == null) {
            fromEntity = new HashMap<>(entitySize);
            fromEntityBitSet[valueRangeDescriptor.getOrdinal()] = fromEntity;
            fromEntity.put(valueRangeBitSet, entity);
            return null;
        }

I know what it does. It's not rocket science. But by being part of another method, the purpose of this code and of that method is being conflated and confused. The meaning is lost among the mechanics of moving things from place to place.

Once you start splitting these things into separate methods, suddenly logical components will start jumping out. You will start seeing how things organize into classes that can expose semantics and hide implementation details. And when you follow this process consistently, you will find that `ValueRangeManager` becomes a simple class, with a clear purpose and obvious implementation; it will not only work well, but also _read_ well.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -87,7 +87,7 @@ public <Score_ extends Score<Score_>> void processWorkingSolutionDuringStep(Abst
         stepScope.setBestScoreImproved(bestScoreImproved);
         if (bestScoreImproved) {
             phaseScope.setBestSolutionStepIndex(stepScope.getStepIndex());
-            var newBestSolution = stepScope.createOrGetClonedSolution();
+            var newBestSolution = stepScope.cloneWorkingSolution();
```
triceo, line 90: Why are we changing this?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/entity/decorator/FilteringEntityByValueSelector.java
zepfred: There's no need to run it, indeed.

zepfred: On the other hand, the `stepStarted` event should be triggered.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelector.java
zepfred: `ListRuinRecreateMoveSelector` isn't that old. The purpose of these changes is to release allocated resources and prevent memory leaks. I reviewed all nodes to identify non-final fields whose references we haven't cleared by the end of the phase and the solver.

For example, the CH repository `PlacerBasedMoveRepository` was not clearing the `placementIterator` reference, causing related resources such as value range caches and reachable value structures to never be reclaimed after the CH phase.

Also, these are private instances that are not used once they are cleared.

core/src/main/java/ai/timefold/solver/core/impl/phase/AbstractPhase.java
zepfred: This function is called only during the `phaseEnded` event and will not be recalculated for the current phase. During my analysis of the issue, I found that the solver was unable to calculate the score from scratch, as the derived score must also allocate resources with `ValueRangeManager`.

This change aimed to allow the solver to transition to the LS phase without encountering an OOM error. At the time of the modification, the root cause, a memory leak, had not yet been resolved. However, this could still be useful if the available memory is insufficient to keep two score directors in memory.

It is important to note that the ranges will not be recalculated. This is because we change the working solution at the end of each phase, which will clear the range cache and update the list variable state when necessary. The current logic requires recalculating the range cache or the reachable values structure each time a new phase begins.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelector.java
triceo: See, this is the strange part to me.

Why would we clear _the inside_ of something, when _the outside_ should just get garbage collected? We should remove a reference to the thing, and everything inside the thing will get collected automatically.

Specifically in the case of `PlacerBasedMoveRepository` - why does that even survive the CH phase? Something is keeping it alive, and that is what we should be solving. What you're doing here is curing the symptoms, not the disease.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
zepfred: There is no need to keep a reference to the cloned solution in the step scope, as the only place calling this logic is `BestSolutionRecaller`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: I see your point. It was an attempt to free resources while searching for memory leaks. I'm restoring it to its original state and will run another benchmark once all comments are addressed.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: I believe we can use this information to improve hashtable lookup performance.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
zepfred: I'm unsure how we can improve this. Remember that the index map, which includes all planning values and the ordinal IDs, is not available when the descriptors are created or when we extract the data.

The class `CompositeCountableValueRange` could benefit from access to the map of values and ordinal IDs, but the current logic doesn't support passing this information, and the solution to enable that seems worse to me. On the other hand, other range caches like `ListValueRange` need to read all the data to compute the `BitSet`.

Please let me know if you have any suggestions for improving it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelector.java
zepfred: Let's revisit `PlacerBasedMoveRepository`. The CH phase instance remains with the AbstractSolver until the process completes. This means that `DefaultConstructionHeuristicPhase` contains a final field `moveRepository` that is not cleared. Since the final field references an iterator (`placementIterator`) that may also hold references to `ReachableValues`, the range cache and all internal reachable values structures will not be released, even if the `ValueRangeManager` is reset.

zepfred: You can confirm what I'm telling you at `AbstractSolver#runPhases`.

triceo: I believe you. :-) But my point is - by removing the phase instance once it's no longer needed, you solve the root cause.

Whether we want to do it is another thing. But it is cleaner than resetting things and hoping that everyone always remembers to do it.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +366,292 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+        if (item.leftSorter() == null) {
+            // The current left sorter is null and need to be updated
+            var newItem = buildEntityValueRangeItem(valueRangeDescriptor, entity, sorter, false);
+            valueRangeList[valueRangeDescriptor.getOrdinal()] = new ValueRangeItem<>(entity,
+                    // Left item from the newly created item
+                    newItem.leftItem(),
+                    newItem.leftSorter(),
+                    // Current right item
+                    item.rightItem(),
+                    item.rightSorter());
```
triceo: Notice: This is code duplication. It is also not important at this point in time for me to know that this happens.

If you hide this in the `ValueRangeItem` constructor, this method immediately becomes easier to read. Hide complexity, separate concerns.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -358,59 +366,292 @@ public <T> CountableValueRange<T> getFromSolution(ValueRangeDescriptor<Solution_
...
+        } else if (item.rightItem() == null) {
+            // The new item stores the sorted right item at the left position
+            var newItem = buildEntityValueRangeItem(valueRangeDescriptor, entity, sorter, false);
+            valueRangeList[valueRangeDescriptor.getOrdinal()] = new ValueRangeItem<>(entity,
+                    // Current left item
+                    item.leftItem(),
+                    item.leftSorter(),
+                    // Right item from the newly created item
+                    newItem.leftItem(),
+                    newItem.leftSorter());
```
triceo: This is slightly different than above, so we cannot simply share the code. But what we can do is to create two static constructors on `ValueRangeItem` which will hide this away.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -85,8 +91,10 @@ public static <Solution_> ValueRangeManager<Solution_> of(SolutionDescriptor<Sol
      */
     public ValueRangeManager(SolutionDescriptor<Solution_> solutionDescriptor) {
         this.solutionDescriptor = Objects.requireNonNull(solutionDescriptor);
-        this.fromSolution = new CountableValueRange[solutionDescriptor.getValueRangeDescriptorCount()];
-        this.reachableValues = new ReachableValues[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.fromSolution = new ValueRangeItem[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.fromSolutionValueIndexMap = new Map[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.reachableValues = new ValueRangeItem[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.fromEntityBitSet = new HashMap[solutionDescriptor.getValueRangeDescriptorCount()];
```
triceo: This field is now being used from exactly one method. You know what that tells me?

The field and the method can be extracted to a separate type. Hide the complexity, expose a well-named type with a well-named method. Nobody needs to know that there is an array, and hashmaps etc. etc.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -85,8 +91,10 @@ public static <Solution_> ValueRangeManager<Solution_> of(SolutionDescriptor<Sol
      */
     public ValueRangeManager(SolutionDescriptor<Solution_> solutionDescriptor) {
         this.solutionDescriptor = Objects.requireNonNull(solutionDescriptor);
-        this.fromSolution = new CountableValueRange[solutionDescriptor.getValueRangeDescriptorCount()];
-        this.reachableValues = new ReachableValues[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.fromSolution = new ValueRangeItem[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.fromSolutionValueIndexMap = new Map[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.reachableValues = new ValueRangeItem[solutionDescriptor.getValueRangeDescriptorCount()];
```
triceo: All code dealing with this field is also hidden in just a few methods. They do not deal with much of the other state. This tells me they can be extracted into another class. Hide complexity, separate concerns, avoid massive classes.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -85,8 +91,10 @@ public static <Solution_> ValueRangeManager<Solution_> of(SolutionDescriptor<Sol
      */
     public ValueRangeManager(SolutionDescriptor<Solution_> solutionDescriptor) {
```
triceo, line 65: All these arrays are based on value range descriptors.
Which means this class is effectively N instances within 1 class; 1 for each value range descriptor.

Arguably, if you refactor the logic so that each VRM works with exactly one VRD, the code will also get much simpler. (You will just have to have one "router" class to delegate requests to these instances based on the specific VRD used.) Something like `valueRangeManager.getFor(valueRangeDescriptor)`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/ListRuinRecreateMoveSelector.java
zepfred: Removing the phases after the process finishes is already done. The point is to keep data from finished phases while another is still running.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeManager.java
```diff
@@ -85,8 +61,22 @@ public static <Solution_> ValueRangeManager<Solution_> of(SolutionDescriptor<Sol
...
         this.solutionDescriptor = Objects.requireNonNull(solutionDescriptor);
-        this.fromSolution = new CountableValueRange[solutionDescriptor.getValueRangeDescriptorCount()];
-        this.reachableValues = new ReachableValues[solutionDescriptor.getValueRangeDescriptorCount()];
+        this.state = new ValueRangeDescriptorState[solutionDescriptor.getValueRangeDescriptorCount()];
```
triceo: Arguably you can make an optimization here.

What is the most common use case? A single value range! (Barely anyone has two or more.)

If you know that the count is `1`, store the value range into its own field X, leave the array to null. Introduce a method which reads the correct state based on whether the X is null or not; make the rest of the code use this method.

That way you get to have a more efficient storage, but the rest of the code will never know.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeDescriptorState.java
```diff
@@ -0,0 +1,459 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
+import ai.timefold.solver.core.impl.heuristic.selector.common.ReachableValues;
+import ai.timefold.solver.core.impl.heuristic.selector.common.ReachableValues.ReachableValuesIndex;
+import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionSorter;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+final class ValueRangeDescriptorState<Solution_, T> {
```
triceo: Arguably, this is not value range descriptor state. Value range descriptor does not have a state. A value range does.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,459 @@
...
+        var allValuesBitSet = new BitSet(reachableValueList.size());
+        // We create a bitset containing all possible values from the range to optimize operations
+        for (var i = 0; i < valueRange.getSize(); i++) {
+            var value = valueRange.get(i);
+            if (value == null) {
+                continue;
+            }
+            allValuesBitSet.set(valueIndexMap.get(value));
+        }
+        for (var i = 0; i < valueRange.getSize(); i++) {
```
triceo: It should be possible to merge these two loops into one. Considering that this method is very important, this will be beneficial.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,459 @@
...
+            return new ValueRangeItem<>(null, leftItem, leftSorter, null, null);
+        }
+
+        public static <Solution_, T> ValueRangeItem<Solution_, T> of(@Nullable T leftItem,
+                @Nullable SelectionSorter<Solution_, ?> leftSorter, @Nullable T rightItem,
+                @Nullable SelectionSorter<Solution_, ?> rightSorter) {
+            return new ValueRangeItem<>(null, leftItem, leftSorter, rightItem, rightSorter);
+        }
+
+        public static <Solution_, T> ValueRangeItem<Solution_, T> of(@Nullable Object entity, @Nullable T leftItem,
```
triceo: This is already an improvement, but doesn't really help much with readability. There is a bunch of arguments, all of them are nullable, how do I know which of these methods does what? `of` is not a good enough name – use the name of the method to explain what it actually does; why or when do I want to call this one, and not the `5` other `of`s?

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,459 @@
...
+     */
+    private record BitSetItem(int size, int hash, BitSet item) {
+
+        public static <T> BitSetItem of(CountableValueRange<T> valueRange, BitSet item) {
+            return new BitSetItem((int) valueRange.getSize(), item.hashCode(), item);
+        }
+
+        @Override
+        public int hashCode() {
+            return hash;
```
triceo, line 421: This is a nice idea.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeDescriptorState.java
```diff
@@ -0,0 +1,459 @@
...
+            }
+            indexMap.put(value, idx++);
+        }
+        return indexMap;
+    }
+
+    private List<ReachableValues.ReachableItemValue> initReachableValueList(CountableValueRange<T> valueRange,
+            int entityListSize) {
+        var size = (int) valueRange.getSize();
+        var valueList = new ArrayList<ReachableValues.ReachableItemValue>(size);
```
triceo: Do you need an array list here? Does it need to be mutable?
I'd consider turning the value range to an iterator, that iterator to a stream, and terminating that stream with `toList()`.

This will give you an immutable list at the end. As the author of the code, you have options of making your intentions known – if you don't want people to modify your stuff, don't allow them. 

It sends a strong signal that you intended for this collection to never be modified; therefore if someone ever tries to modify it, it will force them to think why you made it impossible, and that thought process will require deeper understanding and  hopefully prevent them from making a mistake.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeDescriptorState.java
```diff
@@ -0,0 +1,459 @@
...
+                continue;
+            }
+            valueClass = value.getClass();
+            break;
+        }
+        return valueClass;
+    }
+
+    private static Map<Object, Integer> buildIndexMap(Iterator<@Nullable Object> allValues, int size, boolean isImmutable) {
+        Map<Object, Integer> indexMap = isImmutable ? new HashMap<>(size) : new IdentityHashMap<>(size);
```
triceo: This is, unfortunately, not the correct way to size a map.
See `CollectionUtils`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeDescriptorState.java
```diff
@@ -0,0 +1,459 @@
...
+            var entity = entityList.get(i);
+            var valueRange = getFromEntity(entity, entityList.size(), null);
+            loadEntityValueRange(i, valueIndexMap, valueRange, reachableValueList);
+        }
+        var sorterAdapter = sorter != null ? SelectionSorterAdapter.of(cachedWorkingSolution, sorter) : null;
+        return new ReachableValues(entityIndexItem, valueIndexItem, valueClass, sorterAdapter,
+                variableDescriptor.getValueRangeDescriptor().acceptsNullInValueRange());
+    }
+
+    private @Nullable Class<?> findValueClass(CountableValueRange<T> valueRange) {
```
triceo: What is a "value class" here?
In Java, a value class is a class without an identity. Did you mean that? If not, I suggest you find a better name. Perhaps just `getTypeOfValue`. (Not "value type", because that has the same problem as "value class".)

Also, the code you have here is not correct – assume that `T` is `Number`. Then this collection can mix `Integer`, `Double`, `BigDecimal`. And your code will only get one of them. None of them are correct – you want to find out `Number`.

You should be able to read that information from the value range descriptor. If not, read the generic type from the class information, and make the method available on the value range descriptor. But I'm 90 % sure it's already there.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
zepfred: I don't think so. Before this change, the complexity was O(n^2) because of an inner for loop, and now it is O(2n). We need the `BitSet` to set all reachable values from a given value without an inner for loop.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeDescriptorState.java
zepfred: We need a list for random access, and it can also be immutable.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -22,23 +25,25 @@
 @NullMarked
 public final class ReachableValues {
```
triceo: This type should get a generic parameter (or parameters) describing what the `Object` is.
Is the `Object` in place X the same thing as the `Object` in place `Y`? Strategic use of generics will make that crystal clear.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -157,59 +186,75 @@ boolean containsValue(int valueIndex) {
...
+            if (valueRangeSorter != null) {
+                valueRangeSorter.sort(valuesList);
             }
-            return onDemandRandomAccessValueList;
+            return (List<Object>) valuesList;
         }
     }
 
+    @NullMarked
+    public record ReachableValuesIndex<T>(Map<Object, Integer> indexMap, List<T> allItems) {
```
triceo: Arguably same here. What is the object here? Is it a value? Is it an entity? Can it be both? The generic should say that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -121,21 +136,30 @@ public boolean acceptsNullValue() {
...
+
+    public void clear() {
+        firstCachedObject = null;
+        secondCachedObject = null;
+        this.onDemandRandomAccessEntity = new List[valuesIndex.allItems().size()];
+        this.onDemandRandomAccessValue = new List[valuesIndex.allItems().size()];
     }
 
     @NullMarked
     public static final class ReachableItemValue {
```
triceo: And here too. What is the `Object`? Do all the methods which use `Object` mean the same type? I have a suspicion that sometimes `Object` means "entity" and sometimes it means "value". The generics should say that.

Every time you see `Object`, you should think: would a generic type help describe what this is?

Sometimes it's impossible to say what the type is on the class level. That's when we can declare the generic on the method level. But there should typically be no reason to use `Object`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,427 @@
...
+import ai.timefold.solver.core.impl.heuristic.selector.common.ReachableValues.ReachableValuesIndex;
+import ai.timefold.solver.core.impl.heuristic.selector.common.decorator.SelectionSorter;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.impl.util.MutableInt;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+final class ValueRangeState<Solution_, T> {
```
triceo: `T` deserves a better name. What is it - a value? An entity? Both?
You can say all of that. In this case, it looks like `T` is actually `Value_`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,427 @@
...
+    private final ValueRangeDescriptor<Solution_> valueRangeDescriptor;
+    private final Solution_ cachedWorkingSolution;
+
+    // Solution related fields
+    private @Nullable ValueRangeItem<Solution_, CountableValueRange<T>> fromSolutionItem;
+    private @Nullable Map<Object, Integer> fromSolutionValueIndexMap;
+
+    // Entity related fields
+    private @Nullable Map<Object, ValueRangeItem<Solution_, CountableValueRange<T>>> fromEntityMap;
+    private @Nullable Map<BitSetItem, Object> fromEntityBitSet;
```
triceo: See all this use of `Object`? The code would be so much easier to understand if we used generics instead.

If it's not possible on the variable level, then at least on the method level.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    /**
+     * The record holds a reference to {@link BitSet},
+     * a precomputed hash to avoid recalculating it every time since the BitSet should be immutable.
+     * It also includes the size of the BitSet to skip unnecessary comparisons when searching it in the map.
+     */
+    private record BitSetItem(int size, int hash, BitSet item) {
+
+        public static <T> BitSetItem of(CountableValueRange<T> valueRange, BitSet item) {
```
triceo: In this case though, `T` brings absolutely no value. `CountableValueRange<?>` does the exact same job here.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -0,0 +1,427 @@
...
+        while (valueIndex >= 0) {
+            var item = reachableValueList.get(valueIndex);
+            item.addEntity(entityIndex);
+            // We unset the current value index to import only the values that are reachable
+            item.addValuesExcept(allValuesBitSet, valueIndex);
+            valueIndex = allValuesBitSet.nextSetBit(valueIndex + 1);
+        }
+    }
+
+    private record ValueRangeItem<Solution_, T>(@Nullable Object entity, @Nullable T leftItem,
```
triceo: `T` again.

We generally use `T` for types which are completely generic. For example - `List<T>`, because the list would accept _anything_.

In this case, we know that `T` is a value. So make it say `Value_`.
Make your code self-documenting.

PR #1901, opened 2025-10-29
feat: add constraint stream profiling

Christopher-Chianelli:
Profiling constraints is notorishing difficult, since each component of a constraint are converted to nodes, some of which are shared. As such, a JVM method profile is basically unreadable and does not represent how much time is actually spent for each constraint.

To aid in profiling, an optional constraintStreamProfilingMode configuration was added. If set to a value other than NONE, it wraps each tuple lifecycle node inside a ProfilingTupleLifecycle, which will measure how long each lifecycle executes. The ProfilingTupleLifecycle find out what constraint is responsible for creating that lifecycle by getting snapshot of the stack traces from its constraint stream's creator (when a constraint stream is shared, their stack traces are merged into the same set).

At the end of solving, a profiling summary is then produced in the INFO log. The details differ depending on the profiling mode:

- In the BY_METHOD profiling mode, (className, methodName) is used as the key

- In the BY_LINE profiling mode, (className, methodName, lineNumber) is used as the key.

The methods/lines are printed in descending order of time percentage spent. The sum of time percentage spent may be over 100%, since methods/lines can share time spent with other methods/lines.

timefold.solver.constraint-stream-profiling-mode was added as a property to Quarkus and Spring Boot to configure profiling (defaults to NONE).


benchmark/src/main/resources/benchmark.xsd
```diff
@@ -458,6 +458,9 @@
           <xs:element minOccurs="0" name="constraintStreamAutomaticNodeSharing" type="xs:boolean"/>
                               
           
+          <xs:element minOccurs="0" name="constraintStreamProfilingMode" type="tns:constraintProfilingMode"/>
```
triceo: In the interest of not asking the user to make niche choices, I'd remove this. See below.

core/src/main/java/ai/timefold/solver/core/config/score/director/ConstraintProfilingMode.java
```diff
@@ -0,0 +1,22 @@
+package ai.timefold.solver.core.config.score.director;
+
+/**
+ * Controls how constraints are profiled.
+ * Enabling profiling have a minor performance impact.
+ */
+public enum ConstraintProfilingMode {
```
triceo: IMO this is needless complexity. Do not profile by method or line - profile _by constraint_. That is what people actually want to see; the list of constraints and how much time they take to run.

Also, the names of constraints typically do not change, while method names and lines do as code is refactored over time. People who use bytecode obfuscation would get absolutely no information out of this, and I wonder what happens with automatic node sharing etc., where we mutate code heavily without giving access to that new source code. 

All in all, this information is too unreliable to present it to the user. The only reliable source is the constraint name; the rest, if it needs to exist, should remain hidden.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListSwapMoveProviderTest.java
```diff
@@ -125,7 +126,7 @@ private <Solution_> Iterable<Move<Solution_>> createMoveIterable(MoveDefinition<
         var firstEntityClass = solutionDescriptor.getMetaModel().genuineEntities().get(0).type();
         var constraintProvider = new TestingConstraintProvider(firstEntityClass);
         var scoreDirectorFactory = new BavetConstraintStreamScoreDirectorFactory<>(solutionDescriptor, constraintProvider,
-                EnvironmentMode.TRACKED_FULL_ASSERT);
+                EnvironmentMode.TRACKED_FULL_ASSERT, ConstraintProfilingMode.NONE);
```
triceo: Let's have an overload which doesn't require this to be specified.
It is boilerplate.

triceo (commented):
I like this in principle! When it comes to the implementation, I have concerns. See specifically the comment regarding profiling mode.

I also see this hooks itself into tuple lifecycle. But I don't see it tracking propagation overhead anywhere. Arguably it should.

The logging aspect is OK, but I am missing any methods to consume this information at runtime. How about Micrometer metrics? Or custom JFR events? Thinking out loud here: if it triggered an event (micrometer / JFR) every time it finished a single measurement, these could be aggregated at the end as well as queried in-flight. More often than not, we will be run in the platform - people will want a way of monitoring their solvers.

I question the name "profiling"; this doesn't use a profiler, and doesn't work in a way in which profilers work. The name should be changed in order to not invite confusing comparisons with actual profiling.

core/src/main/java/ai/timefold/solver/core/config/score/director/ConstraintProfilingMode.java
Christopher-Chianelli: This does not look at lambdas, it looks at the `.filter`\`.join`/etc. code, which is not modified by node sharing (i.e. it looks at the filter call, not what within the filter call).
Arguably, having method name + line number is useful; it tells you what part of a constraint is taking the most runtime (i.e. is it a filter, join, or groupBy).

Christopher-Chianelli:
In regards to it tracking  propagation overhead, it did in a previous implementation (by wrapping all the Propogators). However, this is a form of double counting; the Propogators fundamentally call lifecycles, which are profiled. Additionally, Propogators correspond to parents of nodes instead of nodes themselves, which can cause absurdities such as `forEach` taking a large percentage of a profile (since it is a parent to a `join`/`filter`).

core/src/main/java/ai/timefold/solver/core/config/score/director/ConstraintProfilingMode.java
triceo: The line number still changes with automatic node sharing - new fields are added at the top of the class, moving others down. (Although how does it even work with bytecode generation, when there is no source code? It seems that this metadata has to be fundamentally unreliable.) 

Even with the changes that the Java compiler does when it produces bytecode, line numbers are still unreliable. Code can be added here and there. The idea of trying to do this per building block is IMO noble, but falls apart too easily. I don't think it's necessary for this concept to succeed - when the users see the expensive constraint, that'll already be plenty information to proceed.

triceo:
> However, this is a form of double counting; the Propogators fundamentally call lifecycles, which are profiled. 

Not sure about that. These lifecycles count for a lot, true. But propagation also deals with the queues, and iterates a lot. IMO it's not necessarily a double count.

> Additionally, Propogators correspond to parents of nodes instead of nodes themselves, which can cause absurdities such as `forEach` taking a large percentage of a profile (since it is a parent to a `join`/`filter`).

Which is another good argument that this should not profile per-method or per-line, but per-constraint. Doesn't matter which part takes such a long amount of time - it's the constraint that's what matters.

core/src/main/java/ai/timefold/solver/core/config/score/director/ConstraintProfilingMode.java
Christopher-Chianelli: Line numbers are recorded in bytecode; so they line numbers in stack traces do not change if that bytecode is modified (provided you don't delete/change the metadata instruction that records the line number).

Christopher-Chianelli: As for how line numbers look in bytecode:
```
// Some previous instructions
IADD
INVOKE_VIRTUAL
Label LINE_START
LINENUMBER 10 LINE_START
// Some instructions after
NOP
INVOKE_INTERFACE
FADD
Label ANOTHER_LINE
LINENUMBER 11 ANOTHER_LINE
```
In this example, line 10 correspond to `NOP, INVOKE_INTERFACE, FADD`.

benchmark/src/main/resources/benchmark.xsd
```diff
@@ -2897,6 +2900,30 @@
   </xs:simpleType>
       
   
+  <xs:simpleType name="constraintProfilingMode">
```
triceo: Having read your previous comments, I still believe this is needless complexity, but I have a compromise proposal.

We remove this setting - there is no need for people to choose between these options. More importantly, there is no reason why these should be _mutually exclusive_. No, if I were doing this profiling, I'd want all the information I can get, as opposed to running the thing several times with different configs.

So, here's my proposal:

- The setting goes away. It becomes a boolean - enabled or disabled.
- Under the hood, we keep two ways of profiling; by constraint, and then by node. Both are active at the same time.
- When printing the outputs, we print a breakdown. We start with how much time is taken by the constraint, and then one level down, we break it down per node.
- By line and by method goes away. This is an advanced feature and, much like with node network visualization, it requires some understanding of the underlying internals. If we deal with nodes only, it will be both correct and consistent with the visualization.

benchmark/src/main/resources/benchmark.xsd
triceo: I think we need a better name than "profiling".
It's confusing the terms - there is no profiler here, there is no profile, IMO this name does more harm than good.

Christopher-Chianelli: There is a `profile` though; it a flat profile: https://sourceware.org/binutils/docs/gprof/Flat-Profile.html

Christopher-Chianelli:
This is what a profile looks like now:

```
Constraint Profiling Summary
  Work shared by org.acme.schooltimetabling.domain/Room conflict, org.acme.schooltimetabling.domain/Student group conflict, org.acme.schooltimetabling.domain/Student group subject variety, org.acme.schooltimetabling.domain/Teacher conflict, org.acme.schooltimetabling.domain/Teacher room stability, org.acme.schooltimetabling.domain/Teacher time efficiency (79.28%, 10464105459ns)
  Work shared by org.acme.schooltimetabling.domain/Teacher time efficiency (10.08%, 1329891726ns)
  Work shared by org.acme.schooltimetabling.domain/Teacher room stability (6.29%, 829846622ns)
  Work shared by org.acme.schooltimetabling.domain/Student group subject variety (3.67%, 483839225ns)
  Work shared by org.acme.schooltimetabling.domain/Student group conflict (0.37%, 49107293ns)
  Work shared by org.acme.schooltimetabling.domain/Room conflict (0.16%, 21585301ns)
  Work shared by org.acme.schooltimetabling.domain/Teacher conflict (0.15%, 20162643ns)

Constraint Profiling Breakdown
  Work shared by org.acme.schooltimetabling.domain/Room conflict, org.acme.schooltimetabling.domain/Student group conflict, org.acme.schooltimetabling.domain/Student group subject variety, org.acme.schooltimetabling.domain/Teacher conflict, org.acme.schooltimetabling.domain/Teacher room stability, org.acme.schooltimetabling.domain/Teacher time efficiency
    JOIN (72.98%, 7637089549ns)
      By Operation
        UPDATE (99.88%, 7628054546ns)
        INSERT (0.08%, 6372938ns)
        RETRACT (0.03%, 2662065ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33 (24.54%, 1874118595ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56 (21.19%, 1618086279ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46 (19.82%, 1513463898ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77 (15.28%, 1166571547ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92 (11.88%, 907564341ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66 (7.30%, 557284889ns)
    PROPAGATION (27.02%, 2827015910ns)
      By Operation
        UPDATE (67.12%, 1897462674ns)
        INSERT (20.41%, 576988801ns)
        RETRACT (12.47%, 352564435ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77 (68.98%, 1950132747ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92 (58.98%, 1667348577ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66 (57.89%, 1636653143ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56 (54.55%, 1542047166ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33 (53.67%, 1517384171ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46 (52.94%, 1496483996ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:76 (49.40%, 1396606778ns)
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:91 (49.40%, 1396606778ns)

  Work shared by org.acme.schooltimetabling.domain/Teacher time efficiency
    FILTER (84.37%, 1122032509ns)
      By Operation
        UPDATE (51.17%, 574090252ns)
        RETRACT (27.04%, 303395964ns)
        INSERT (21.79%, 244546293ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:79 (100.00%, 1122032509ns)
    SCORING (15.63%, 207859217ns)
      By Operation
        RETRACT (72.95%, 151623406ns)
        UPDATE (15.62%, 32462652ns)
        INSERT (11.44%, 23773159ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:84 (100.00%, 207859217ns)

  Work shared by org.acme.schooltimetabling.domain/Teacher room stability
    FILTER (76.13%, 631772012ns)
      By Operation
        UPDATE (99.92%, 631243027ns)
        INSERT (0.06%, 395960ns)
        RETRACT (0.02%, 133025ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:68 (100.00%, 631772012ns)
    SCORING (23.87%, 198074610ns)
      By Operation
        UPDATE (55.41%, 109753784ns)
        RETRACT (44.48%, 88096546ns)
        INSERT (0.11%, 224280ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:69 (100.00%, 198074610ns)

  Work shared by org.acme.schooltimetabling.domain/Student group subject variety
    FILTER (87.20%, 421889685ns)
      By Operation
        UPDATE (51.68%, 218025826ns)
        RETRACT (26.59%, 112178665ns)
        INSERT (21.73%, 91685194ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:96 (100.00%, 421889685ns)
    SCORING (12.80%, 61949540ns)
      By Operation
        RETRACT (91.26%, 56536197ns)
        INSERT (4.39%, 2719814ns)
        UPDATE (4.35%, 2693529ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:101 (100.00%, 61949540ns)

  Work shared by org.acme.schooltimetabling.domain/Student group conflict
    SCORING (100.00%, 49107293ns)
      By Operation
        INSERT (62.38%, 30635185ns)
        RETRACT (37.12%, 18226439ns)
        UPDATE (0.50%, 245669ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:59 (100.00%, 49107293ns)

  Work shared by org.acme.schooltimetabling.domain/Room conflict
    SCORING (100.00%, 21585301ns)
      By Operation
        INSERT (60.99%, 13165893ns)
        RETRACT (38.74%, 8361684ns)
        UPDATE (0.27%, 57724ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:39 (100.00%, 21585301ns)

  Work shared by org.acme.schooltimetabling.domain/Teacher conflict
    SCORING (100.00%, 20162643ns)
      By Operation
        INSERT (63.46%, 12795555ns)
        RETRACT (36.09%, 7275850ns)
        UPDATE (0.45%, 91238ns)
      By Location
        org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:49 (100.00%, 20162643ns)
```
By location percentage may add up to over 100% due to node sharing; the rest should always add up to 100%.

triceo:
> Constraint Profiling Summary
>   Work shared by org.acme.schooltimetabling.domain/Room conflict, org.acme.schooltimetabling.domain/Student group conflict, org.acme.schooltimetabling.domain/Student group subject variety, org.acme.schooltimetabling.domain/Teacher conflict, org.acme.schooltimetabling.domain/Teacher room stability, org.acme.schooltimetabling.domain/Teacher time efficiency (79.28%, 10464105459ns)

If we remove the package from this, suddenly this will be better readable. The packages are deprecated anyway. When they're gone though, it'll probably be useful to surround the constraint name with quotes.

What does "work shared by" mean? I'd understand it if it were only for multiple constraints, but the same is used below when only one constraint is listed.

`10464105459ns` is surely not the ideal way to represent this number. `10.464 seconds` seems better.

Wouldn't a breakdown by node/operation make sense here as well? Overall for the entire constraint provider, if 80 % is just joins, that's useful information.

> By location percentage may add up to over 100% due to node sharing; the rest should always add up to 100%.

I really don't like that it's there at all. It makes most of the output, you drown in the text, and apparently you also need to understand that the numbers there mean something else than everywhere else. This output doesn't help, only makes things more difficult.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/LeftTupleLifecycleImpl.java
```diff
@@ -10,6 +12,11 @@ record LeftTupleLifecycleImpl<Tuple_ extends Tuple>(LeftTupleLifecycle<Tuple_> l
         Objects.requireNonNull(leftTupleLifecycle);
     }
 
+    @Override
+    public LifecycleKind lifecycleKind() {
+        return LifecycleKind.JOIN;
```
triceo: Not correct; this is used for any node which has left and right inputs.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/RightTupleLifecycleImpl.java
```diff
@@ -10,6 +12,11 @@ record RightTupleLifecycleImpl<Tuple_ extends Tuple>(RightTupleLifecycle<Tuple_>
         Objects.requireNonNull(rightTupleLifecycle);
     }
 
+    @Override
+    public LifecycleKind lifecycleKind() {
+        return LifecycleKind.JOIN;
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
```diff
@@ -17,23 +19,35 @@
 import ai.timefold.solver.core.impl.bavet.common.tuple.RightTupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.Tuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
+import ai.timefold.solver.core.impl.score.stream.bavet.common.Scorer;
 
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
 public abstract class AbstractNodeBuildHelper<Stream_ extends BavetStream> {
```
triceo, line 32: IMO another case in point for how locations make things more difficult, yet the value is questionable.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/LifecycleKind.java
```diff
@@ -0,0 +1,15 @@
+package ai.timefold.solver.core.impl.bavet.common;
+
+public enum LifecycleKind {
```
triceo: This needs documentation. What's the point? Are all handed the same? Are all from constraint streams? Wouldn't it be better to look forward through the lifecycle to the actual node? (Currently not possible, but _could be_.)

This is brittle, and effectively duplicates node logic. Also, the fact that ifExists is represented by join (and therefore not measurable) is unfortunate. IMO needs to be addressed.

triceo (commented):
The output could use being shorter and clearer.
The implementation around `LifecycleKind` needs rethinking.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/LifecycleKind.java
Christopher-Chianelli: It measurable; the location of an `ifExists` would be different than the location of a `join`. This is largely only there to support the "break down by node type". A break down by location is more useful to a user, who will be able to lookup the class file and line number.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
Christopher-Chianelli: How is this comment related to `@NullMarked`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/LeftTupleLifecycleImpl.java
Christopher-Chianelli: This is for us; users would rather have the location than the node type. Although `concat`, `ifExists` and `join` are different operations, they all fundamentally depends on a "joining" conceptually. Would you rather it be called `LEFT`/`RIGHT`?

Christopher-Chianelli:
> What does "work shared by" mean? I'd understand it if it were only for multiple constraints, but the same is used below when only one constraint is listed.

Work performed exclusively by the listed group. When more than one constraint is in the group, that means they are sharing some node, and thus removing a constraint does not remove the work it does in the group (unless all constraints in the group are removed). When it is alone, it means the work done exclusively by a given constraint (where removing the constraint guarantees removing that work).

> 10464105459ns is surely not the ideal way to represent this number. 10.464 seconds seems better.

Acknowledged, but it is the raw captured data. The issue is if we try to humanize the number, then we need to ask questions such as do we convert it to minutes above a certain duration? Then hours? And then there the potential of losing accuracy and making comparisons between constraints harder (as you need to convert them to a common unit).

> Wouldn't a breakdown by node/operation make sense here as well? Overall for the entire constraint provider, if 80 % is just joins, that's useful information.

Useful infomation for who? This information also shows up in the location list.

>> By location percentage may add up to over 100% due to node sharing; the rest should always add up to 100%.

> I really don't like that it's there at all. It makes most of the output, you drown in the text, and apparently you also need to understand that the numbers there mean something else than everywhere else. This output doesn't help, only makes things more difficult.

This is the only text that is actually actionable by the user; it tells them what is actually hot in their code, and they don't need to do a string/constant lookup.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/LeftTupleLifecycleImpl.java
triceo: No, I'd rather it said CONCAT, IF_EXISTS and JOIN.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNodeBuildHelper.java
triceo: I added the comment on the line with the class declaration, to imply it is related to the class and the changes in it.

Christopher-Chianelli: The location data is literally how we are able to link constraints to their nodes.

triceo:
> When it is alone, it means the work done exclusively by a given constraint (where removing the constraint guarantees removing that work).

Arguably if this case used different words, I wouldn't have to ask the question.
Different cases require different descriptions.

> Acknowledged, but it is the raw captured data. The issue is if we try to humanize the number, then we need to ask questions such as do we convert it to minutes above a certain duration? Then hours? And then there the potential of losing accuracy and making comparisons between constraints harder (as you need to convert them to a common unit).

You already picked a common unit, just the worst one. :-) I see how this could possibly be hours, days. Or nanos. Practically, it won't be either. It will not be much less (who stops the solver after a millisecond?), and it will not be much more (who needs hours to establish these numbers?).

IMO if we use seconds as a baseline, that's a reasonable estimate. Saying "5:25:00" (hms) also isn't too hard, if we decide we need to. The "problem" comes when the time is much smaller than a second - thankfully, it will not be.

> Useful information for who? This information also shows up in the location list.

You probably noticed that I don't like the location list.

> This is the only text that is actually actionable by the user; it tells them what is actually hot in their code, and they don't need to do a string/constant lookup.

Since it doesn't cross-link to code, you're copy-pasting anyway. IMO constraint names work much better; those will always be the same. With line numbers, you're making many assumptions; that the code over there is the same as the one you're looking at, that the code was compiled to include those line numbers, that no crazy tooling modified that code. (I love Lombok, which always points me to places that don't really exist.)

I know that this is one way how you can report node sharing. IMO the downsides outweigh that; this will not be reliable, and only leads to questions.

triceo:
(I mistakenly edited your comment instead of adding mine. My apologies, I tried to restore the original content as faithfully as I could.)

Christopher-Chianelli:
Locations are necessary; a constraint can use the same operation multiple times; an example taken from our docs:

```java
return cf.forEach(Shift.class)
         .groupBy(Shift::getEmployee, ConstraintCollector.sum(Shift::getDuration))
         .complement(Employee.class, employee -> 0)
         .groupBy(loadBalance(...));
```

Here, `GROUP_BY` will have two different nodes; which one is taking more runtime? The `sum` or `loadBalance`? You cannot tell that with only a single number, you need two.

Christopher-Chianelli:
Now looks like this (shared locations are shown as sets, so all percentages add up to 100):

```
Constraint Profiling Summary
"Teacher time efficiency" (23.07%, 2.65 seconds)
"Room conflict" (15.33%, 1.76 seconds)
"Student group conflict" (13.98%, 1.61 seconds)
"Teacher conflict" (12.53%, 1.44 seconds)
"Student group subject variety" (12.48%, 1.43 seconds)
"Teacher room stability" (12.18%, 1.40 seconds)
Work shared by "Room conflict", "Student group conflict", "Student group subject variety", "Teacher conflict", "Teacher room stability", "Teacher time efficiency" (10.43%, 1.20 seconds)

Constraint Profiling Breakdown
"Teacher time efficiency"
    FOR_EACH (39.35%, 1.04 seconds)
      By Operation
        UPDATE (99.87%, 1.04 seconds)
        INSERT (0.09%, 0.00 seconds)
        RETRACT (0.04%, 0.00 seconds)
      By Node
        FOR_EACH Node 6 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77 (53.42%, 0.56 seconds)
        FOR_EACH Node 7 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77 (46.58%, 0.49 seconds)
    FILTER (36.07%, 0.96 seconds)
      By Operation
        UPDATE (49.05%, 0.47 seconds)
        RETRACT (28.03%, 0.27 seconds)
        INSERT (22.92%, 0.22 seconds)
      By Node
        FILTER Node 5 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:79 (100.00%, 0.96 seconds)
    JOIN (17.82%, 0.47 seconds)
      By Operation
        INSERT (41.07%, 0.19 seconds)
        UPDATE (32.92%, 0.16 seconds)
        RETRACT (26.01%, 0.12 seconds)
      By Node
        JOIN Propagator 26 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77 (100.00%, 0.47 seconds)
    SCORING (6.76%, 0.18 seconds)
      By Operation
        RETRACT (72.97%, 0.13 seconds)
        UPDATE (15.55%, 0.03 seconds)
        INSERT (11.48%, 0.02 seconds)
      By Node
        SCORING Node 4 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:84 (100.00%, 0.18 seconds)

"Room conflict"
    FOR_EACH (92.97%, 1.64 seconds)
      By Operation
        UPDATE (99.83%, 1.63 seconds)
        INSERT (0.12%, 0.00 seconds)
        RETRACT (0.05%, 0.00 seconds)
      By Node
        FOR_EACH Node 20 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33 (50.94%, 0.83 seconds)
        FOR_EACH Node 19 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33 (49.06%, 0.80 seconds)
    JOIN (5.93%, 0.10 seconds)
      By Operation
        INSERT (43.83%, 0.05 seconds)
        RETRACT (31.69%, 0.03 seconds)
        UPDATE (24.48%, 0.03 seconds)
      By Node
        JOIN Propagator 22 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33 (100.00%, 0.10 seconds)
    SCORING (1.10%, 0.02 seconds)
      By Operation
        INSERT (61.19%, 0.01 seconds)
        RETRACT (38.55%, 0.01 seconds)
        UPDATE (0.26%, 0.00 seconds)
      By Node
        SCORING Node 18 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:39 (100.00%, 0.02 seconds)

"Student group conflict"
    FOR_EACH (89.57%, 1.44 seconds)
      By Operation
        UPDATE (99.91%, 1.44 seconds)
        INSERT (0.06%, 0.00 seconds)
        RETRACT (0.03%, 0.00 seconds)
      By Node
        FOR_EACH Node 14 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56 (51.18%, 0.74 seconds)
        FOR_EACH Node 13 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56 (48.82%, 0.70 seconds)
    JOIN (7.71%, 0.12 seconds)
      By Operation
        INSERT (46.49%, 0.06 seconds)
        RETRACT (35.04%, 0.04 seconds)
        UPDATE (18.47%, 0.02 seconds)
      By Node
        JOIN Propagator 24 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56 (100.00%, 0.12 seconds)
    SCORING (2.71%, 0.04 seconds)
      By Operation
        INSERT (61.88%, 0.03 seconds)
        RETRACT (37.61%, 0.02 seconds)
        UPDATE (0.52%, 0.00 seconds)
      By Node
        SCORING Node 12 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:59 (100.00%, 0.04 seconds)

"Teacher conflict"
    FOR_EACH (92.73%, 1.33 seconds)
      By Operation
        UPDATE (99.90%, 1.33 seconds)
        INSERT (0.07%, 0.00 seconds)
        RETRACT (0.03%, 0.00 seconds)
      By Node
        FOR_EACH Node 17 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46 (50.85%, 0.68 seconds)
        FOR_EACH Node 16 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46 (49.15%, 0.66 seconds)
    JOIN (6.01%, 0.09 seconds)
      By Operation
        INSERT (44.31%, 0.04 seconds)
        RETRACT (28.82%, 0.02 seconds)
        UPDATE (26.87%, 0.02 seconds)
      By Node
        JOIN Propagator 23 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46 (100.00%, 0.09 seconds)
    SCORING (1.26%, 0.02 seconds)
      By Operation
        INSERT (62.64%, 0.01 seconds)
        RETRACT (36.92%, 0.01 seconds)
        UPDATE (0.44%, 0.00 seconds)
      By Node
        SCORING Node 15 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:49 (100.00%, 0.02 seconds)

"Student group subject variety"
    FOR_EACH (54.98%, 0.79 seconds)
      By Operation
        UPDATE (99.83%, 0.79 seconds)
        INSERT (0.12%, 0.00 seconds)
        RETRACT (0.04%, 0.00 seconds)
      By Node
        FOR_EACH Node 2 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92 (51.87%, 0.41 seconds)
        FOR_EACH Node 3 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92 (48.13%, 0.38 seconds)
    FILTER (25.31%, 0.36 seconds)
      By Operation
        UPDATE (49.77%, 0.18 seconds)
        RETRACT (27.22%, 0.10 seconds)
        INSERT (23.01%, 0.08 seconds)
      By Node
        FILTER Node 1 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:96 (100.00%, 0.36 seconds)
    JOIN (15.96%, 0.23 seconds)
      By Operation
        INSERT (40.98%, 0.09 seconds)
        UPDATE (34.45%, 0.08 seconds)
        RETRACT (24.57%, 0.06 seconds)
      By Node
        JOIN Propagator 27 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92 (100.00%, 0.23 seconds)
    SCORING (3.75%, 0.05 seconds)
      By Operation
        RETRACT (90.80%, 0.05 seconds)
        UPDATE (4.64%, 0.00 seconds)
        INSERT (4.56%, 0.00 seconds)
      By Node
        SCORING Node 0 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:101 (100.00%, 0.05 seconds)

"Teacher room stability"
    FILTER (39.11%, 0.55 seconds)
      By Operation
        UPDATE (99.91%, 0.55 seconds)
        INSERT (0.07%, 0.00 seconds)
        RETRACT (0.02%, 0.00 seconds)
      By Node
        FILTER Node 9 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:68 (100.00%, 0.55 seconds)
    FOR_EACH (33.84%, 0.47 seconds)
      By Operation
        UPDATE (99.67%, 0.47 seconds)
        INSERT (0.23%, 0.00 seconds)
        RETRACT (0.09%, 0.00 seconds)
      By Node
        FOR_EACH Node 10 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66 (55.79%, 0.26 seconds)
        FOR_EACH Node 11 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66 (44.21%, 0.21 seconds)
    JOIN (14.69%, 0.21 seconds)
      By Operation
        UPDATE (82.97%, 0.17 seconds)
        INSERT (11.82%, 0.02 seconds)
        RETRACT (5.21%, 0.01 seconds)
      By Node
        JOIN Propagator 25 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66 (100.00%, 0.21 seconds)
    SCORING (12.35%, 0.17 seconds)
      By Operation
        UPDATE (56.97%, 0.10 seconds)
        RETRACT (42.91%, 0.07 seconds)
        INSERT (0.12%, 0.00 seconds)
      By Node
        SCORING Node 8 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:69 (100.00%, 0.17 seconds)

Work shared by "Room conflict", "Student group conflict", "Student group subject variety", "Teacher conflict", "Teacher room stability", "Teacher time efficiency"
    FOR_EACH (100.00%, 1.20 seconds)
      By Operation
        UPDATE (96.23%, 1.15 seconds)
        INSERT (2.62%, 0.03 seconds)
        RETRACT (1.14%, 0.01 seconds)
      By Node
        FOR_EACH Propagator 21 shared at locations [org.acme.schooltimetabling.solver.TimetableConstraintProvider#roomConflict:33, org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupConflict:56, org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:91, org.acme.schooltimetabling.solver.TimetableConstraintProvider#studentGroupSubjectVariety:92, org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherConflict:46, org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherRoomStability:66, org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:76, org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:77] (100.00%, 1.20 seconds)
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractNode.java
```diff
@@ -1,14 +1,23 @@
...
+
 /**
  * @see PropagationQueue Description of the propagation mechanism.
  */
+@NullMarked
 public abstract class AbstractNode {
 
     private long id;
     private long layerIndex = -1;
+    private @Nullable SortedSet<ConstraintNodeLocation> locationSet;
```
triceo, line 24: I'd mention this is nullable because it only applies to profiling and therefore is disabled most of the time.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/BavetAbstractConstraintStream.java
```diff
@@ -22,22 +24,55 @@ public abstract class BavetAbstractConstraintStream<Solution_>
...
     protected BavetAbstractConstraintStream(BavetConstraintFactory<Solution_> constraintFactory,
             RetrievalSemantics retrievalSemantics) {
         super(retrievalSemantics);
         this.constraintFactory = constraintFactory;
         this.parent = null;
+        this.streamLocationSet = new TreeSet<>();
+        streamLocationSet.add(determineStreamLocation());
+    }
+
+    public abstract StreamKind getStreamKind();
```
triceo: IMO this is actually worse than it was before - the number of duplications is significantly higher.

Is there any particular reason why we can't do `tupleLifecycle instanceof BavetAbstractMapNode` etc.? This would centralize the logic and rely on what we already know.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/BavetAbstractConstraintStream.java
```diff
@@ -22,22 +24,55 @@ public abstract class BavetAbstractConstraintStream<Solution_>
         extends AbstractConstraintStream<Solution_>
         implements BavetStream {
 
+    private static final String BAVET_IMPL_PACKAGE = "ai.timefold.solver.core.impl.bavet";
+    private static final String BAVET_SCORE_IMPL_PACKAGE = "ai.timefold.solver.core.impl.score.stream";
+    private static final String BAVET_SCORE_API_PACKAGE = "ai.timefold.solver.core.api.score.stream";
```
triceo: Not excited about this. This will not change if we ever move Bavet around, and things will silently stop working. (Which could happen.) Any chance we could infer this from the classes we have?

(Another way how the locations are introducing brittleness into our codebase.)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/ConstraintProfile.java
```diff
@@ -0,0 +1,134 @@
...
+        selfDuration += duration;
+        breakdownByLocationDuration.add(duration);
+        streamKindDuration.add(duration);
+        breakdownByOperationDuration.add(duration);
+    }
+
+    private String formatDuration(long nanos) {
+        var NANOS_IN_SECOND = 1_000_000_000L;
+        var seconds = (double) nanos / NANOS_IN_SECOND;
+        return "%.2f seconds".formatted(seconds);
```
triceo: I'd prefer `Duration.toSeconds()` and `toMillisPart()`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/ConstraintProfile.java
```diff
@@ -0,0 +1,134 @@
...
+                formatDuration(selfDuration));
+    }
+
+    public String getBreakdown() {
+        var out = new StringBuilder();
+        var sortedStreamKindList = new ArrayList<>(breakdownByStreamKind.keySet());
+        sortedStreamKindList.sort(Comparator.<StreamKind> comparingLong(kind -> breakdownByStreamKind.get(kind).longValue())
+                .reversed());
+
+        out.append(getProfileId()).append("\n");
```
triceo: Let's think of our Windows friends and provide the platform-specific sequence for this please.
(Also everywhere else.)

It should be as easy as replacing `\n` with `%n`.

triceo (commented):
Some more comments, but we're getting there. The output is much more readable now!

Needs documentation. In there, it specifically needs to explain what it measures and how (wall clock / CPU time / ...), what is the expected impact on CS performance, and how we expect people to use this. (Long runs? Production?) It should also break down an example output, describe its components and how to understand the information given. Come to think of, it should even mention _where_ the information will be printed and how to make sure that it is. (Logging config.)

I'm still not wild about the world "profiling". Is it profiling if there's not a profiler? No Java agents, nothing resembling the usual profiling tools for the JVM? Could lead to confusion of terms. What are our other options for naming?

triceo (approved):
LGTM once documented.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -855,6 +855,7 @@ To use automatic node sharing outside Quarkus, your `ConstraintProvider` class m
 Debugging breakpoints put inside your constraints will not be respected, because the `ConstraintProvider` class will be transformed when this feature is enabled.
 ====
 
+[#nodeSharing]
```
TomCools, line 858: Hmm, i expected more in the `enterprise-edition.adoc` file. Is this a standard feature? Why not an enterprise exclusive one?

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -315,6 +315,145 @@ Whatever you do, benchmark on a large and diverse set of inputs.
...
++
+[source,text]
+----
+    FILTER (47.80%, 3.06 seconds)
+      ...
+      By Node
+        FILTER Node 5 defined at location org.acme.schooltimetabling.solver.TimetableConstraintProvider#teacherTimeEfficiency:79 (100.00%, 3.06 seconds)
+----
+
+xref:enterprise-edition/enterprise-edition.adoc#nodeSharing[When node sharing] is used, the time spent will be reported with "Work shared by":
```
TomCools:
```suggestion
xref:enterprise-edition/enterprise-edition.adoc#nodeSharing[When node sharing (a enterprise edition feature)] is used, the time spent will be reported with "Work shared by":
```

People don't see the hyperlink, so i think it adds value to mention it explicitly.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -315,6 +315,145 @@ Whatever you do, benchmark on a large and diverse set of inputs.
...
+* How much time spent in score calculation is spent by each constraint (relative percentage and total seconds) in descending order.
++
+[source,text]
+----
+"Teacher time efficiency" (27.98%, 7.05 seconds)
+----
+
+* A breakdown for each constraint (in descending order of time spent):
+
+** The time spent (relative percentage of time spent by constraint and total seconds), by each constraint stream building block (such as `join` or `filter`).
```
TomCools:
```suggestion
** Time spent by each constraint stream building block (e.g., join, filter), shown as % and total seconds.
```

The double "time spent" mention makes it harder to read.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -315,6 +315,145 @@ Whatever you do, benchmark on a large and diverse set of inputs.
...
+[source,text]
+----
+    FILTER (47.80%, 3.06 seconds)
+      By Operation
+        UPDATE (50.81%, 1.08 seconds)
+        RETRACT (31.00%, 1.01 seconds)
+        INSERT (18.19%, 0.07 seconds)
+----
++
+NOTE: This info is only relevant to the Timefold Solver team and can be safely ignored.
```
TomCools: I would move this NOTE up before the code example. 
Currently: 
- Give context
- Explainer
- Tell people they can ignore it

Desired: 
- Give short amount of context
- Tell people they can ignore it
- Explainer

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -315,6 +315,145 @@ Whatever you do, benchmark on a large and diverse set of inputs.
...
++
+[source,text]
+----
+"Teacher time efficiency"
+    FILTER (47.80%, 3.06 seconds)
+----
+
+** Each constraint stream building block is further broken down to:
+
+*** The time spent (relative percentage of time spent by building block and total seconds) of each propagation operation.
```
TomCools:
```suggestion
*** Time spent by each constraint stream building block, shown as % and total seconds of each propagation operation.
```

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
```diff
@@ -315,6 +315,145 @@ Whatever you do, benchmark on a large and diverse set of inputs.
...
+    FILTER (47.80%, 3.06 seconds)
+      By Operation
+        UPDATE (50.81%, 1.08 seconds)
+        RETRACT (31.00%, 1.01 seconds)
+        INSERT (18.19%, 0.07 seconds)
+----
++
+NOTE: This info is only relevant to the Timefold Solver team and can be safely ignored.
+
+*** The time spent (relative percentage of time spent by building block and total seconds) of each defined node and the location(s) the nodes were defined at.
```
TomCools:
```suggestion
*** Time spent by each constraint stream building block, shown as % and total seconds of each each defined node and the location(s) the nodes were defined at.
```

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
triceo: We didn't consider this, but it's totally doable.

docs/src/modules/ROOT/pages/integration/config-properties.adoc
```diff
@@ -28,6 +28,9 @@ The random seed to be used in the solving process.
 Enable xref:using-timefold-solver/running-the-solver.adoc#environmentMode[runtime assertions] to detect common bugs in your
 implementation during development.
 
+{property_prefix}timefold.solver.{solver_name_prefix}constraint-stream-profiling-enabled::
+Enable xref:constraints-and-score/performance.adoc#constraintProfiling[constraint profiling] to identify the constraints taking the most time in score calculation and thus are worth optimizing.
```
TomCools:
```suggestion
Enable xref:constraints-and-score/performance.adoc#constraintProfiling[constraint profiling] to identify the constraints taking the most time in score calculation and thus might be worth optimizing.
```

Some constraints are just intrinsically harder. It's not because they take more time that they are "always" worth optimizing.

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
Christopher-Chianelli: Node sharing is not an enterprise edition feature. Automatic node sharing is. The difference is automatic node sharing rewrites the class to share functions via static constants (so node sharing can be done more).

triceo (approved):

docs/src/modules/ROOT/pages/constraints-and-score/performance.adoc
TomCools: Right, forgot that nuance for a second.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/ConstraintNodeProfileId.java
```diff
@@ -0,0 +1,92 @@
...
+    public String toString() {
+        if (qualifier == Qualifier.NONE) {
+            return "%s %d".formatted(streamKind, key);
+        } else {
+            var qualifierString = switch (qualifier) {
+                case NODE -> "Node";
+                case LEFT_INPUT -> "Left Input";
+                case RIGHT_INPUT -> "Right Input";
+                default -> throw new IllegalStateException("Impossible state: unknown qualifier (%s)."
+                        .formatted(qualifier));
```
Christopher-Chianelli, line 52: Since Qualifier is an enum, and the switch yields values, it will be a compile error if the switch is not exhaustive; the default can be removed.
```suggestion
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirectorFactory.java
```diff
@@ -36,9 +37,12 @@ public final class BavetConstraintStreamScoreDirectorFactory<Solution_, Score_ e
         }
         var constraintProviderClass = getConstraintProviderClass(config, providedConstraintProviderClass);
         var constraintProvider = ConfigUtils.newInstance(config, "constraintProviderClass", constraintProviderClass);
+        Boolean profilingMode =
+                Objects.requireNonNullElse(config.getConstraintStreamProfilingEnabled(), false);
         ConfigUtils.applyCustomProperties(constraintProvider, "constraintProviderClass",
                 config.getConstraintProviderCustomProperties(), "constraintProviderCustomProperties");
-        return new BavetConstraintStreamScoreDirectorFactory<>(solutionDescriptor, constraintProvider, environmentMode);
+        return new BavetConstraintStreamScoreDirectorFactory<>(solutionDescriptor, constraintProvider, environmentMode,
+                profilingMode);
```
Christopher-Chianelli:
```suggestion
        var profilingEnabled =
                Objects.requireNonNullElse(config.getConstraintStreamProfilingEnabled(), false);
ConfigUtils.applyCustomProperties(constraintProvider, "constraintProviderClass",
config.getConstraintProviderCustomProperties(), "constraintProviderCustomProperties");
        return new BavetConstraintStreamScoreDirectorFactory<>(solutionDescriptor, constraintProvider, environmentMode,
                profilingEnabled);
```

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -127,6 +127,14 @@ public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> ge
...
         var randomFactory = buildRandomFactory(environmentMode);
         var previewFeaturesEnabled = solverConfig.getEnablePreviewFeatureSet();
+        Boolean profilingMode = solverConfig.getScoreDirectorFactoryConfig().getConstraintStreamProfilingEnabled();
+
+        if (moveThreadCount != null &&
+                profilingMode != null &&
+                profilingMode) {
+            throw new IllegalStateException("Multithreaded solving is unsupported with constraintStreamProfilingEnabled (%s)."
+                    .formatted(profilingMode));
+        }
```
Christopher-Chianelli:
```suggestion
        var profilingEnabled = solverConfig.getScoreDirectorFactoryConfig().getConstraintStreamProfilingEnabled();

        if (moveThreadCount != null &&
                profilingEnabled != null &&
                profilingEnabled) {
            throw new IllegalStateException("Multithreaded solving is unsupported with constraintStreamProfilingEnabled (%s)."
                    .formatted(profilingEnabled));
        }
```

Christopher-Chianelli (commented):
Changes LGTM (Cannot approve as I am the author of the PR)

triceo (approved):

PR #1903, opened 2025-10-30
chore: improve Neighborhoods design for scalability

triceo:
Previously, Neighborhoods depended on joins on the left-hand side - these joins are, however, possibly massive. With large datasets, joining 100k entities with themselves would result in billions of entries per move. This is clearly not acceptable and would never scale.

The new approach only uses joins within `pick()`, and these joins will be enumerated on-demand. Therefore the memory use of this mechanism will be minimized. (Although possibly still significant, especially for random selection.)

The extra benefit of this approach is that it actually respects the order of entities and facts that was given by the user.

Further improvements to scalability are likely still possible, but will be introduced when the situation calls for it. (Which I expect it will.)

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/UniqueRandomSequence.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/UniqueRandomSequence.java
@@ -19,7 +19,7 @@ public sealed interface UniqueRandomSequence<T>
 
     /**
      * Picks a random element from the list which has not already been removed.
-     * Once an element of the list is removed either via {@link #remove(Random)} or {@link #remove(int)},
+     * Once an element of the list is removed either via {@link #remove(Random)},
      * it will never be returned again by this method.
      *
      * @param workingRandom the random number generator to use


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiOriginalMoveIterator.java
```diff
@@ -0,0 +1,88 @@
...
+ * It goes through all tuples coming from the left, and for each left tuple,
+ * it goes through all tuples coming from the right that match the joiner and filter.
+ * It is more efficient than {@link BiRandomMoveIterator},
+ * because it does not need to enumerate any list of tuples up-front
+ * and all access happens on-demand.
+ */
+@NullMarked
+final class BiOriginalMoveIterator<Solution_, A, B> implements Iterator<Move<Solution_>> {
+
+    private static final UniTuple EMPTY_TUPLE = new UniTuple<>(null, 0);
```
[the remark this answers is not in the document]
triceo: Technically correct, but doing so results in its own warnings; I pick this as the least hassle.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
```diff
@@ -17,46 +22,60 @@ public class ListSwapMoveDefinition<Solution_, Entity_, Value_>
...
-                .filter((solutionView, leftPosition, rightPosition) -> solutionView.isValueInRange(variableMetaModel,
-                        rightPosition.entity(), leftPosition.value())
-                        && solutionView.isValueInRange(variableMetaModel, leftPosition.entity(), rightPosition.value()));
-        // Finally pick the moves.
-        return moveStreamFactory.pick(result)
+                .filter((solutionView, value) -> solutionView.getPositionOf(variableMetaModel, value) instanceof PositionInList)
+                .map((solutionView, value) -> new FullElementPosition<>(value,
+                        solutionView.getPositionOf(variableMetaModel, value).ensureAssigned(), planningIdGetter));
+        // TODO this requires everything that is ever swapped to implement @PlanningID; likely not acceptable
+        return moveStreamFactory.pick(assignedValueStream)
```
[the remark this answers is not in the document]
triceo: Wrong.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/joiner/DefaultBiEnumeratingJoiner.java
```diff
@@ -14,7 +14,7 @@
 
 @SuppressWarnings({ "unchecked", "rawtypes" })
 @NullMarked
-public final class DefaultBiEnumeratingJoiner<A, B> extends AbstractDataJoiner<B> implements BiEnumeratingJoiner<A, B> {
+public final class DefaultBiEnumeratingJoiner<A, B> extends AbstractJoiner<B> implements BiEnumeratingJoiner<A, B> {
```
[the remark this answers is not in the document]
triceo: Not sure what this is even saying. Ignoring.

triceo:
Taking this back to a draft, as I thought of one more improvement.

triceo:
Converted to draft again, following up on a conversation with Geoff.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComposingList.java
```diff
@@ -0,0 +1,107 @@
...
+ * If an element from any particular sublist is never {@link #get(int) accessed},
+ * it need not be stored in memory.
+ *
+ * <p>
+ * Note that this implementation does not support modifications of the list, or any of its sublists.
+ *
+ * @param <E> the type of elements in this list
+ */
+@NullMarked
+final class ComposingList<E> extends AbstractList<E> {
```
zepfred, line 23: Nice!

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
```diff
@@ -21,13 +23,12 @@ public ChangeMoveDefinition(PlanningVariableMetaModel<Solution_, Entity_, Value_
 
     @Override
     public MoveStream<Solution_> build(MoveStreamFactory<Solution_> moveStreamFactory) {
-        var enumeratingStream =
-                moveStreamFactory.forEachEntityValuePair(variableMetaModel)
-                        .filter((solutionView, entity, value) -> {
```
zepfred, line 26: The reasoning behind using this node sharing functions is clear to me, but I’m not sure if it should be part of the API. Why not move these default functions outside of the API to a helper class?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
```diff
@@ -21,13 +23,12 @@ public ChangeMoveDefinition(PlanningVariableMetaModel<Solution_, Entity_, Value_
...
-                            Value_ currentValue = solutionView.getValue(variableMetaModel, Objects.requireNonNull(entity));
-                            return !Objects.equals(currentValue, value);
-                        });
-        return moveStreamFactory.pick(enumeratingStream)
+        var nodeSharingSupportFunctions =
+                ((DefaultMoveStreamFactory<Solution_>) moveStreamFactory).getNodeSharingSupportFunctions(variableMetaModel);
```
zepfred, line 27: Are the functions from `NodeSharingSupportFunctions` going to be reused? If not, we can add them to `ChangeMoveDefinition`, similar to how we integrated `ListChangeMoveDefinition::isValidChange?`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListChangeMoveDefinition.java
```diff
@@ -68,7 +64,7 @@ public MoveStream<Solution_> build(MoveStreamFactory<Solution_> moveStreamFactor
                 });
     }
 
-    private boolean isValidChange(SolutionView<Solution_> solutionView, Value_ value, ElementPosition targetPosition) {
+    private boolean isValidChange(SolutionView<Solution_> solutionView, ElementPosition targetPosition, Value_ value) {
```
zepfred, line 68: I understand that this is compatible for node sharing, correct?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
```diff
@@ -1,63 +1,84 @@
...
-        // Eliminate pairs that cannot be swapped due to value range restrictions.
-        var result = validAssignedValueUniquePairStream
-                .filter((solutionView, leftPosition, rightPosition) -> solutionView.isValueInRange(variableMetaModel,
-                        rightPosition.entity(), leftPosition.value())
-                        && solutionView.isValueInRange(variableMetaModel, leftPosition.entity(), rightPosition.value()));
-        // Finally pick the moves.
-        return moveStreamFactory.pick(result)
-                .asMove((solutionView, leftPosition, rightPosition) -> Moves.swap(leftPosition.elementPosition,
-                        rightPosition.elementPosition, variableMetaModel));
+                .filter((solutionView, value) -> solutionView.getPositionOf(variableMetaModel, value) instanceof PositionInList)
```
zepfred, line 33: Should we move this to a separate function, or will it also be compatible with node sharing?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
```diff
@@ -1,63 +1,84 @@
...
+                .map((solutionView, value) -> new FullElementPosition<>(value,
+                        solutionView.getPositionOf(variableMetaModel, value).ensureAssigned(), planningIdGetter));
+        if (planningIdGetter == null) { // If the user hasn't defined a planning ID, we will follow a slower path.
```
zepfred, line 36: Does "slow path" mean enumerating both a <-> b and b ↔ a?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
```diff
@@ -1,63 +1,84 @@
...
-                PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
-                SolutionView<Solution_> solutionView, Value_ value) {
-            var assignedElement = solutionView.getPositionOf(variableMetaModel, value).ensureAssigned();
-            return new FullElementPosition<>(value, assignedElement);
+    private boolean isValidSwap(SolutionView<Solution_> solutionView, FullElementPosition<Entity_, Value_> leftPosition,
+            FullElementPosition<Entity_, Value_> rightPosition) {
+        if (Objects.equals(leftPosition, rightPosition)) {
+            return false;
         }
+        return solutionView.isValueInRange(variableMetaModel, rightPosition.entity(), leftPosition.value())
```
zepfred, line 60: Idea for discussion:

I've been reflecting on our recent discussion about value ranges and something I considered while working on the cascading shadow variable listener. What if we utilize the value range manager or another solver component to generate and store ordinal IDs and its list positions for planning entities and planning values?

If we accomplish that, validation of the value range would be quicker and potentially beneficial for DSV. It seems there was a recent message regarding one of your ideas about that, and I feel neighborhoods could be an opportunity to try to work on it.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/Moves.java
```diff
@@ -36,13 +35,12 @@ public static <Solution_, Entity_, Value_> Move<Solution_> change(Entity_ entity
...
-        return new SwapMove<>(Collections.singletonList(variableMetaModel), leftEntity, rightEntity);
+        return swap(leftEntity, rightEntity, Collections.singletonList(variableMetaModel));
     }
 
-    @SafeVarargs
     public static <Solution_, Entity_> Move<Solution_> swap(Entity_ leftEntity, Entity_ rightEntity,
-            PlanningVariableMetaModel<Solution_, Entity_, Object>... variableMetaModels) {
-        return new SwapMove<>(Arrays.asList(variableMetaModels), leftEntity, rightEntity);
+            List<PlanningVariableMetaModel<Solution_, Entity_, Object>> variableMetaModelList) {
+        return new SwapMove<>(variableMetaModelList, leftEntity, rightEntity);
```
zepfred, line 44: The utility method includes the planning entities first and the model next, whereas the move class does not follow this order. Should we align both classes to match the parameter order?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/SwapMoveDefinition.java
```diff
@@ -60,39 +66,59 @@
...
+    }
+
+    private Move<Solution_> buildMove(SolutionView<Solution_> solutionView, Entity_ a, Entity_ b) {
+        return Moves.swap(a, b, variableMetaModelList);
+    }
+
+    private <A> @Nullable DefaultBiEnumeratingJoiner<A, A> buildLessThanId(Class<A> sourceClass) {
+        SolutionDescriptor<Solution_> solutionDescriptor =
+                ((DefaultPlanningSolutionMetaModel<Solution_>) entityMetaModel.solution()).solutionDescriptor();
+        MemberAccessor planningIdMemberAccessor = solutionDescriptor.getPlanningIdAccessor(sourceClass);
```
zepfred: This logic can be extracted to be reused.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/MoveStreamFactory.java
```diff
@@ -89,6 +74,4 @@ public interface MoveStreamFactory<Solution_> {
 
     <A> UniSamplingStream<Solution_, A> pick(UniEnumeratingStream<Solution_, A> enumeratingStream);
 
-    <A, B> BiSamplingStream<Solution_, A, B> pick(BiEnumeratingStream<Solution_, A, B> enumeratingStream);
```
zepfred, line 92: The API looks simpler now.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiOriginalMoveIterator.java
```diff
@@ -0,0 +1,96 @@
...
+    private final UniLeftDatasetInstance<Solution_, A> leftDatasetInstance;
+    private final UniRightDatasetInstance<Solution_, A, B> rightDatasetInstance;
+
+    // Fields required for iteration.
+    private @Nullable Move<Solution_> nextMove;
+    private @Nullable Iterator<UniTuple<A>> leftTupleIterator;
+    private @Nullable Iterator<UniTuple<B>> rightTupleIterator;
+
+    @SuppressWarnings("unchecked")
+    private UniTuple<A> leftTuple = EMPTY_TUPLE;
```
zepfred, line 39: It is not clear why we need to initialize this variable.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiRandomMoveIterator.java
```diff
@@ -0,0 +1,137 @@
...
+        if (nextMove != null) {
+            return true;
+        }
+
+        while (!leftTupleSequence.isEmpty()) {
+            var leftElement = leftTupleSequence.pick(workingRandom);
+            var leftTuple = leftElement.value();
+            var rightTupleSequence = rightTupleSequenceMap.computeIfAbsent(leftTuple, this::computeRightSequence);
+            try {
+                var bTuple = rightTupleSequence.remove(workingRandom);
```
zepfred, line 124: When `rightTupleSequence` is exhausted after removing the last element, we should check if it is empty before trying to fetch and remove the next element.

`UniqueRandomSequence` appears to be a finite sequence as items are removed. Why not include an `isEmpty` method?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiRandomMoveIterator.java
```diff
@@ -0,0 +1,137 @@
...
+                var leftFact = leftTuple.factA;
+                var rightFact = bTuple.factA;
+                nextMove = context.buildMove(leftFact, rightFact);
+            } catch (NoSuchElementException e) {
+                // We cannot guarantee that the right sequence is empty, because we do not check filtering eagerly.
+                // Therefore we can run into a situation where there are no more elements passing the filter,
+                // even though the sequence is not technically empty.
+                // We only find this out at runtime.
+                leftTupleSequence.remove(leftElement.index());
```
zepfred: Throwing exceptions has a cost in the JVM, and I wonder if we should just return null instead.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/AbstractLeftDatasetInstance.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.function.Predicate;
+
+import ai.timefold.solver.core.impl.bavet.common.tuple.AbstractTuple;
+import ai.timefold.solver.core.impl.util.ElementAwareArrayList;
+import ai.timefold.solver.core.impl.util.ListEntry;
+
+import org.jspecify.annotations.NullMarked;
+
+@NullMarked
+public abstract class AbstractLeftDatasetInstance<Solution_, Tuple_ extends AbstractTuple>
```
zepfred, line 12: I'm not sure if I like the class name. The name does not convey the class's goal, and the words "left" and "right" seem meaningless.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/DefaultUniqueRandomSequence.java
```diff
@@ -0,0 +1,119 @@
...
+            // use the closest index to avoid skewing the probability
+            index = determineActiveIndex(workingRandom, index);
+            if (index < 0 || index >= length) {
+                throw new NoSuchElementException();
+            }
+        }
+        return index;
+    }
+
+    private int determineActiveIndex(Random workingRandom, int randomIndex) {
```
zepfred, line 77: Cool!

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/DefaultUniqueRandomSequence.java
```diff
@@ -0,0 +1,119 @@
...
+        // if the distance is equal, randomly choose between them,
+        // otherwise return the one that is closer to the random index
+        if (nextIndexDistance == previousIndexDistance) {
+            return workingRandom.nextBoolean() ? nextClearIndex : previousClearIndex;
+        }
+        return nextIndexDistance < previousIndexDistance ? nextClearIndex : previousClearIndex;
+    }
+
+    @Override
+    public T remove(Random workingRandom) {
```
zepfred, line 93: What occurs when there are no more items left to remove?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/FilteredUniqueRandomSequence.java
```diff
@@ -0,0 +1,76 @@
...
+    public int pickIndex(Random workingRandom) {
+        if (delegate.isEmpty()) {
+            throw new NoSuchElementException("No more elements to pick from.");
+        }
+        var nonRemovedElement = delegate.pick(workingRandom);
+        var originalRandomIndex = nonRemovedElement.index();
+
+        var actualValueIndex = originalRandomIndex;
+        var value = nonRemovedElement.value();
+        while (!filter.test(value)) {
```
zepfred, line 53: This condition seems to be incorrect. Why are we not checking if `delegate` is empty?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/FilteredUniqueRandomSequence.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public T remove(Random workingRandom) {
+        return remove(pickIndex(workingRandom));
+    }
+
+    @Override
+    public T remove(int index) {
+        var element = delegate.remove(index);
```
zepfred: What occurs if the `delegate` is empty?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/uni/AbstractUniEnumeratingStream.java
```diff
@@ -151,8 +151,14 @@ public AbstractUniEnumeratingStream<Solution_, A> distinct() {
         return groupBy(ConstantLambdaUtils.identity());
     }
 
-    public UniDataset<Solution_, A> createDataset() {
-        var stream = shareAndAddChild(new TerminalUniEnumeratingStream<>(enumeratingStreamFactory, this));
+    public UniLeftDataset<Solution_, A> createLeftDataset() {
```
zepfred, line 154: Again the name using "left" and "right" words provides no clue about its goal.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
triceo: I don't know what you mean. The class is not part of the public API.

If you mean that this should not be part of the `DefaultMoveStreamFactory`, I disagree. `MoveStreamFactory` is the main unit of this API, and all move streams created by this factory need to share these functions. When a new factory is created, new functions need to be created. Therefore it makes sense for these functions to be strongly linked to `MoveStreamFactory`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
triceo: Their only reason for existence is so that they can be shared. It is a good point that `isValidChange` should probably be among those functions too.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListChangeMoveDefinition.java
triceo: Moving this to shared functions.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
triceo: That's only part of it. The reason why it's slow is the lack of a join - this will be a filter over unrestrained values, possibly trying and eliminating many before finding one.

There is no avoiding this. If there is no identity, we cannot use operations which require identity.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
triceo: I think it's inevitable, but it also doesn't need to be connected to Neighborhoods. If we make this happen, other places in the solver will benefit too. IMO this improvement will be delivered independently.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiOriginalMoveIterator.java
triceo: If we didn't, the variable would have to be `@Nullable`. And there is no need for that.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/AbstractLeftDatasetInstance.java
triceo: "Left" and "Right" here are, in fact, very well defined. It's the left and right side in a join - in "A+B", "A" is left and "B" is right. When we support tri streams, there will be "AB" as left and "C" as right. This is consistent with the naming we already use in CS, and also in SQL etc.

As far as the name "DatasetInstance" is concerned... "Dataset" is certainly not perfect, it doesn't say much; that said, no better name emerged. When you accept that "Dataset" is the name, then "DatasetInstance" flows directly from it.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
triceo: Actually now I realize that during pick, there is no node sharing anymore, and therefore sharing these functions is pointless.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ListSwapMoveDefinition.java
triceo: Node sharing here would make sense, but we are using an internal type and therefore cannot expose the functionality as a whole.

Overall, this should not be a problem now. If/when we ever run into this as an issue, we'll find a solution.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiRandomMoveIterator.java
triceo: It does have a cost (only when the exception is actually thrown), but that cost is not too big. If we ever see this as a problem, let's look into this, but for now, I think it will be fine.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/DefaultUniqueRandomSequence.java
triceo: An exception is thrown. Isn't that clear enough? What is the motivation behind this question?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/FilteredUniqueRandomSequence.java
triceo: I slightly modified the logic to make it flow clearer.

triceo:
Comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
zepfred: I mean `DefaultMoveStreamFactory`, including the function `getNodeSharingSupportFunctions`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
zepfred: What do you mean by pointless'?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
zepfred: Okay. It needs to be shared, but now I wonder if this is too hidden. Do we expect our users to use the node sharing feature? If so, they will need to understand implementation details to cast to `DefaultMoveStreamFactory`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/AbstractLeftDatasetInstance.java
zepfred: I recall some DBMS using "outer" and "inner" terms to describe the left and right nodes, respectively.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/DefaultUniqueRandomSequence.java
zepfred: It relates to my main comment about avoiding throwing exceptions.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiRandomMoveIterator.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private void pickNextMove(UniqueRandomSequence.SequenceElement<UniTuple<A>> leftElement) {
+        var leftTuple = leftElement.value();
+        var rightTupleSequence = rightTupleSequenceMap.computeIfAbsent(leftTuple, this::computeRightSequence);
+        var remove = false;
+        if (rightTupleSequence.isEmpty()) {
+            remove = true;
+        } else {
+            try {
+                var bTuple = rightTupleSequence.remove(workingRandom);
```
zepfred, line 124: It looks better. However, we still might throw `NoSuchElementException` at this call. Would it be better to just return null?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
triceo: No, users are not expected to use this. Users are on their own - just like they are in CS.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/maybeapi/move/ChangeMoveDefinition.java
triceo: There are two parts to a move stream - one part is cached. (That's the CS part.) The other part is fully expanded, that is the pick part. In other words:

    .pick(thisIsCached, joiners)
    .pick(thisIsCached2, joiners2)

`thisIsCached` and `thisIsCached2` works exactly the same as CS. Every tuple is expanded in memory. But the two picks work differently! Instead of creating a full cartesian product of the two caches, it will just randomly choose one of each. This is what allows Neighborhoods to scale. And it also explains why there is no node sharing here - at this point, there are no nodes anymore. Joiners are, more-or-less, a smart filter of (A, B) pairs.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/move/BiRandomMoveIterator.java
triceo: I don't want to return null. I'm not sure if `null` may be a valid value to return. Time will tell.

PR #1905, opened 2025-10-30
docs: add docs for precompute

Christopher-Chianelli:


docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
```diff
@@ -1622,6 +1622,82 @@ This stream now contains all `Employee` instances, both with and without shifts
...
+Constraint noOverlappingShifts(ConstraintFactory constraintFactory) {
+    return constraintFactory.precompute(this::overlappingShifts)
+        .filter((shift1, shift2, overlap) -> shift1.getEmployee() != null && shift1.getEmployee() == shift2.getEmployee())
+        .penalize(HardSoftScore.ONE_HARD, (shift1, shift2, overlap) -> overlap)
+        .asConstraint("No overlapping shifts");
+}
+----
+====
+
+[IMPORTANT]
```
triceo: I don't think this deserves an `IMPORTANT` - it's just a normal description of the feature. The next bit - the score corruption - that is IMO correctly important; it's a typical caveat that people will run into.

Don't overdo it with highlighting - if everything is highlighted, nothing is.

triceo (approved):
I added some more structure to the text.
LGTM now.

PR #1908, opened 2025-10-31
fix: make flattenLast use object equality instead of identity

Christopher-Chianelli:


core/src/test/java/ai/timefold/solver/core/impl/score/stream/common/uni/AbstractUniConstraintStreamTest.java
```diff
@@ -2414,34 +2415,49 @@ public void expandToQuad() {
     @Override
     @TestTemplate
     public void flattenLastWithDuplicates() {
```
triceo, line 2417: Why was this change to a test necessary?
I'd like to understand the extent to which we're actually potentially breaking people.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenLastNode.java
```diff
@@ -169,4 +130,36 @@ public Propagator getPropagator() {
...
+                outTupleList.add(inserted);
+                insertConsumer.accept(inserted);
+            } else {
+                updateConsumer.accept(outTupleList.get(listIndex));
+            }
+        }
+
+        void retract(Consumer<OutTuple_> retractedConsumer) {
+            for (var i = newCount.intValue(); i < outTupleList.size(); i++) {
+                retractedConsumer.accept(outTupleList.get(i));
```
triceo: If the other consumers are called `insertConsumer` and `updateConsumer`, shouldn't this be called `retractConsumer`? The other two do not use past tense.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenLastNode.java
```diff
@@ -41,117 +46,73 @@ public final void insert(InTuple_ tuple) {
...
+                propagationQueue::update);
     }
 
     protected abstract OutTuple_ createTuple(InTuple_ originalTuple, FlattenedItem_ item);
 
     @Override
     public final void update(InTuple_ tuple) {
-        List<OutTuple_> outTupleList = tuple.getStore(flattenLastStoreIndex);
-        if (outTupleList == null) {
+        Map<Object, FlattenLastMapEntry<FlattenedItem_, OutTuple_>> outTupleMap = tuple.getStore(flattenLastStoreIndex);
```
triceo: The full type of this map is crazy enough that I'd create a new interface for it as a shorthand.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenLastNode.java
```diff
@@ -2,15 +2,20 @@
...
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import ai.timefold.solver.core.impl.bavet.common.tuple.AbstractTuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleState;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.impl.util.MutableInt;
 
 public abstract class AbstractFlattenLastNode<InTuple_ extends AbstractTuple, OutTuple_ extends AbstractTuple, EffectiveItem_, FlattenedItem_>
```
triceo, line 20: Nice simplification of the code!

triceo (commented):
Considering that we are making this change to improve the behavior of `precompute` and remove that one pesky exception from there, I am missing those changes in here - or are you planning another PR for that?

Also - technically, we are breaking backwards compatibility here. (Even though, in this case, acceptably.) We should write a note in the upgrade recipe, describing the change so that people know what to expect, and how to fix it.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/common/uni/AbstractUniConstraintStreamTest.java
Christopher-Chianelli: Because I noticed we did not actually test update; only retract and insert.

triceo (approved):
Made small changes to the upgrade recipe.
LGTM.

Issue #1907, opened 2025-10-31
Add noPenalty()  and  noReward() assertions


greyhairredbear:
Add `noPenalty()` (as alias for `penalizesBy(0)`) and `noReward()` (as alias for `rewardsWith(0)`) functions to `SingleConstraintAssertion`.

greyhairredbear:
Update: This issue will be completed by adding a unified `hasNoImpact()` assertion. See the discussion on #1909 for further information

PR #1909, opened 2025-11-01
feat: Add noPenalty() and noReward() assertions

PatrickDiallo23:
This PR adds noPenalty() and noReward() assertion functions (#1907).

### DIFF
--- a/test/src/main/java/ai/timefold/solver/test/impl/score/stream/AbstractSingleConstraintAssertion.java
+++ b/test/src/main/java/ai/timefold/solver/test/impl/score/stream/AbstractSingleConstraintAssertion.java
@@ -742,9 +742,9 @@ public abstract sealed class AbstractSingleConstraintAssertion<Solution_, Score_
         var expectation = message != null ? message : "Broken expectation.";
         return """
                 %s
-                Constraint: %s
-                Expected: no impact
-                Actual impact: %s (%s)
+                        Constraint: %s
+                          Expected: no impact
+                     Actual impact: %s (%s)
 
                   %s""".formatted(
                 expectation,


triceo:
Hello @PatrickDiallo23!

The issue description could have used a bit more thought. Specifically:

- What is the difference between `noPenalty()` and `noReward()`? Do both need to exist, or just one?
- Is either of those a good name? Considering the fluent structure of the API, something like `verifyThat(...).given(...).hasNoImpact()` would likely be much smoother.

WDYT?

PatrickDiallo23:
> Hello @PatrickDiallo23!
> 
> The issue description could have used a bit more thought. Specifically:
> 
> * What is the difference between `noPenalty()` and `noReward()`? Do both need to exist, or just one?
> * Is either of those a good name? Considering the fluent structure of the API, something like `verifyThat(...).given(...).hasNoImpact()` would likely be much smoother.
> 
> WDYT?

Hi @triceo,

When I did the PR I already assumed that this was the direction you agreed (with 2 methods).
From my point of view, since you will be using the method(s) when you are testing the constraints, I would go for the mentioned ‘hasNoImpact()’ to keep things more simpler. If you agree going with the method name you have suggested, I will make my change.

triceo:
@PatrickDiallo23 Some issues are filed quickly in order not to forget, and a good samaritan picks them up before they can be refined. :-) In this case, I think that my proposal will work better than what is written in that issue.

PatrickDiallo23:
> @PatrickDiallo23 Some issues are filed quickly in order not to forget, and a good samaritan picks them up before they can be refined. :-) In this case, I think that my proposal will work better than what is written in that issue.

@triceo, while I was working on this change, I realized something:

When you are using hasNoImpact() as a single method, you are basically saying that for the given facts there are no penalties nor rewards. If we are going by this logic, it is very hard to have some facts that have exactly no impact at all. For example, you could call hasNoImpact for some facts while expecting 0 rewards, but the test could still fail because there are some penalties. I hope this makes sense.

I could try to keep using a single method, but I need a way to differentiate noPenalty from noReward either in `SingleConstraintAssertion` or in `AbstractSingleConstraintAssertion`. Or I can go by the initial implementation with 2 different methods.

triceo:
I'm not sure I follow.

My interpretation of the issue is that you want to check that there is no impact from the constraint. Are you saying that you're looking for a way to write a test such that there is _some_ impact, just not a penalty? So `noPenalty()` would actually mean `hasNoImpact() || rewardsWith(positiveInteger)`?

greyhairredbear:
@PatrickDiallo23 first, thank you for contributing! Maybe to add a bit more context here: 

When building a constraint, we either `penalize` or `reward` the match, there's no way to do both (afaik, but also wouldn't make sense to me). So when testing a constraint using the `ConstraintVerifier`, `penalizesBy` implies that no reward can result from that constraint (and vice versa for `rewardsBy`). 

The description in #1907 is missing the "why" part, that's on me. Let me explain: When testing a constraint, we usually not only check if a certain scenario breaks a constraint as expected, but also if a scenario that doesn't break a constraint does not get a penalty. For the latter case, we usually use `penalizesBy(0)`. 

To me, both `penalizesBy(0)` and `rewardsWith(0)` express the same intent here: There was no match for that constraint[1]. So I like the `hasNoImpact` proposal by @triceo better than having this split up in two different functions. 

Note on the implementation: `assertImpact` in `AbstractSingleConstraintVerification` might be of interest. This is used by `penalizesBy` and also might take `ScoreImpactType.MIXED` as argument. Not sure if this is the best way, maybe @triceo can chime in here.  

[1] side note: that's technically not entirely correct, since there might have been a match with a weight of 0, if I'm not mistaken)

PatrickDiallo23:
> I'm not sure I follow.
> 
> My interpretation of the issue is that you want to check that there is no impact from the constraint. Are you saying that you're looking for a way to write a test such that there is _some_ impact, just not a penalty? So `noPenalty()` would actually mean `hasNoImpact() || rewardsWith(positiveInteger)`?

I will try to provide a clearer context.

I implemented the hasNoImpact by using penalizeBy(0) because I wanted to use what I already have in the interface. Then, instead of writing a new test case, I wanted to modify the existing text that involved no penalties or no rewards.

In some tests there are the following types of scenarios:

a) We have a set of given facts and a constraint that should penalize everything that falls in a certain condition
b) We have a set of given facts and a constraint that should reward everything that falls in a certain condition

For a), if you go by `noPenalty()`, everything is clear since there are chances that there will be no penalization at all. If you go `hasNoImpact()` it is not 100% clear if you mean that there are no penalties only or if you are taking into account that there are also no rewards as well.

For b), if you go by `noReward()`, everything is clear since there are chances that there will be no rewarding at all. If you go `hasNoImpact()` it is not 100% clear if you mean that there are no rewards only, or if you are taking into account that there are also no penalties as well.

In this case, I think `hasNoImpact()` should mean `penalizeBy(0)` && `rewardsWith(0)` or `noPenalty()` && `noReward()`.

I saw your response, @greyhairredbear. I was thinking the same that, if we go by `hasNoImpact()`, I need to provide an implementation to reflect that.

triceo:
@PatrickDiallo23 @greyhairredbear Thanks to the both of you for the input!

I think that `hasNoImpact()` does what we all want.
The question is if we also want to add `doesNotPenalize()` and `doesNotReward()`

These three methods would give the full picture. `hasNoImpact()` checks that there is zero impact, positive or negative, on the score. `doesNotPenalize()` checks that there are no penalties, but there may be rewards. `doesNotReward()` checks that there are no rewards, but there may be penalties.

Personally, I don't think we need the latter two - but I won't stand in the way. If you think we need them, let's introduce them.

greyhairredbear:
+1 for not introducing `doesNotPenalize`/`doesNotReward`. The only value added of these two functions is that you may assert the `ScoreImpactType` of a constraint with a scenario that leads to no impact on the score. But since I can't think of an example where one wouldn't have a test that asserts a constraint matching (which would have to assert the `ScoreImpactType` anyways), I guess that's a **very** weak argument.

PatrickDiallo23:
Alright, I will go with `hasNoImpact() method` then. In this case, it would be better to look for `AbstractSingleConstraintAssertion` for the implementation, right? @greyhairredbear mentioned something about `assertImpact`. Now, I am not sure if I should think about this method as `penalizesBy(0) && rewardsWith(0)` or as something that could give a score impact (I feel penalization/rewarding is a bit different from score impact; correct me if I am wrong). What would be your advice?

P.S: Depending on your answer I think you should be deleting the `Good first issue` since the solution could not be that straightforward.

triceo:
@PatrickDiallo23 If you use `penalizesBy()` or `rewardsWith()`, the assertion message will mention penalty or reward respectively. If we want the message to be neutral, saying "no impact" as opposed to "no penalty", I would not base it on `penalizesBy()` or `rewardsWith`.

PatrickDiallo23:
@triceo @greyhairredbear I tried to implement `hasNoImpact()` as best as I could. Please take a look.

test/src/main/java/ai/timefold/solver/test/impl/score/stream/AbstractSingleConstraintAssertion.java
```diff
@@ -714,6 +738,17 @@ private static String buildAssertionErrorMessage(String type, String constraintI
...
 
+    private String buildNoImpactAssertionErrorMessage(Number actualImpact, String constraintId, String message) {
+        var expectation = message != null ? message : "Broken expectation.";
+        var preformattedMessage = "%s%n%18s: %s%n%18s: no impact%n%18s: %s (%s)%n%n  %s";
+        return String.format(preformattedMessage,
+                expectation,
+                "Constraint", constraintId,
+                "Expected",
+                "Actual impact", actualImpact, actualImpact.getClass(),
+                DefaultScoreExplanation.explainScore(actualScore, constraintMatchTotalCollection, indictmentCollection));
```
greyhairredbear: Moving the hard coded words (`"Constraint"`, `"Expected"`, `"Actual impact"`) into `preformattedMessage` would benefit readability

greyhairredbear:
@PatrickDiallo23 Thanks for putting in the work to update this! 

I'm not sure if `hasNoImpact()` needs an overloaded version with a message, but this has been done for other assertion methods as well, so I'll let Lukáš be the judge of that, he's a maintainer :)

test/src/main/java/ai/timefold/solver/test/impl/score/stream/AbstractSingleConstraintAssertion.java
triceo: Agree.

triceo (approved):
LGTM after the preformatted message is improved. (See comment.)

Wrt. @greyhairredbear's comment on `hasNoImpact(String)`, I think that - for consistency - we should have it.

test/src/main/java/ai/timefold/solver/test/impl/score/stream/AbstractSingleConstraintAssertion.java
```diff
@@ -714,6 +738,17 @@ private static String buildAssertionErrorMessage(String type, String constraintI
         return String.format(preformattedMessage.toString(), params.toArray());
     }
 
+    private String buildNoImpactAssertionErrorMessage(Number actualImpact, String constraintId, String message) {
+        var expectation = message != null ? message : "Broken expectation.";
+        var preformattedMessage = "%s%n%18s: %s%n%18s: no impact%n%18s: %s (%s)%n%n  %s";
```
triceo: In the more recent versions of Java, we have a better way of doing newlines. Multi-line strings to the rescue:

    """
        %s
        %18s: %s
        %18s: no impact
        %18s: %s (%s)

          %s""".formatted(...);

This also makes it simpler to replace `%18s` with just `%s`, because left-padding the string with whitespace will be respected.

PatrickDiallo23:
@triceo @greyhairredbear Done. Now, I am not 100% sure that I did the right formatting of the message. I checked in the unit test and I also ran the build with Maven, but still :)

triceo:
The `%18s` was there to align the text on the right, padded from the left so that the ":" are at the same position.

There was no point for further back and forth, so I just made this small change myself.

triceo (approved):

triceo:
Thank you, @PatrickDiallo23!

PR #1913, opened 2025-11-04
feat: add a producer id to BestSolutionChangedEvent

### LINKED ISSUE #1888: Best solution events should identify which phase they come from
In various strange combinations of phases (e.g. custom+CH+custom+LS), the solver will trigger all kinds of best solution events, and it is not clear which comes from which phase exactly. We should extend the events with some identification of the phase that triggered them.

Maybe phase ID?

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/api/solver/event/BestSolutionChangedEvent.java
```diff
@@ -58,6 +73,16 @@ public long getTimeMillisSpent() {
         return timeMillisSpent;
     }
 
+    /**
+     * @return A string identifying what generated the event, either of the form
+     *         "Event" where "Event" is a String describing the event that cause the update (like "Solving started")
+     *         or "Phase (index)", where "Phase" is a String identifying the type of phase (like "Construction Heuristics")
+     *         and index is the index of the phase in the {@link SolverConfig#getPhaseConfigList()}.
```
triceo: This is an interesting idea. I was originally only thinking about the phase ID, but this is probably more useful.

That said, let us not design stringly typed APIs - how about a record instead, which has two components; event type (enum) and an optional index? People will be parsing this information, so let's make it easy for them.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -45,7 +46,7 @@ public TerminationStatus getTerminationStatus() {
 
     @Override
     public String getPhaseTypeString() {
-        return "Construction Heuristics";
+        return CONSTRUCTION_HEURISTICS_STRING;
```
triceo: If we agree on the record idea above, I recommend centralizing all of this information in that record. Effectively, the record would have different static constructor methods, such as `ofConstructionHeuristics(int)`.

core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultCustomPhase.java
```diff
@@ -23,6 +23,7 @@
 public final class DefaultCustomPhase<Solution_>
         extends AbstractPossiblyInitializingPhase<Solution_>
         implements CustomPhase<Solution_> {
+    public static final String CUSTOM_STRING = "Custom";
```
triceo: I'd recommend adding a suffix "Phase" here, because "Custom" on its own doesn't say much.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -167,7 +170,8 @@ void solve() {
         solution.setValueList(Arrays.asList(new TestdataValue("v1"), new TestdataValue("v2")));
         solution.setEntityList(Arrays.asList(new TestdataEntity("e1"), new TestdataEntity("e2")));
 
-        solution = PlannerTestUtils.solve(solverConfig, solution);
+        solution = PlannerTestUtils.solveAssertingEvents(solverConfig, solution,
+                BestScoreChangedEvent.constructionHeuristic(SimpleScore.ZERO, 0));
```
triceo, line 172: Similar idea, just in a different place. IMO this is the right pattern even for the solver.

core/src/main/java/ai/timefold/solver/core/api/solver/event/BestSolutionChangedEvent.java
```diff
@@ -17,8 +18,10 @@
  */
 // TODO In Solver 2.0, maybe convert this to an interface.
 public class BestSolutionChangedEvent<Solution_> extends EventObject {
```
triceo: I wonder if we should also add the phase id to the first initialized solution event. In that case, it is easier to identify which phase triggered it (it will be the phase before the first LS), but maybe we can make it trivial using this mechanism.

triceo (commented):
Good idea on the event type! I think we can still improve on it.

core/src/main/java/ai/timefold/solver/core/api/solver/event/BestSolutionChangedEvent.java
```diff
@@ -58,6 +70,15 @@ public long getTimeMillisSpent() {
         return timeMillisSpent;
     }
 
+    /**
+     * @return A {@link EventProducerId} identifying what generated the event, either a
+     *         {@link SolveEventProducerId} if the cause is not associated with a Phase,
+     *         or {@link PhaseEventProducerId} if it is.
```
triceo: Please switch the order - start with the case that doesn't require negation; the sentence becomes easier to read.

core/src/main/java/ai/timefold/solver/core/api/solver/event/EventProducerId.java
```diff
@@ -0,0 +1,86 @@
...
+    String simpleProducerName();
+
+    /**
+     * An optional integer, that if present, identify what index in {@link SolverConfig#getPhaseConfigList()}
+     * corresponds to the Phase that produced the event.
+     *
+     * @return The index of the Phase that produced the event, or {@link OptionalInt#empty()} if the event producer
+     *         is not associated with a phase.
+     */
+    OptionalInt eventPhaseIndex();
```
triceo: If we have the distinction between phase and solver producers, why do _both_ have this method? Only the phase producer should have it.

This would also allow us to get rid of the `OptionalInt` - the current agreement is that we don't use `Optional` in the solver.

I am thinking about a different thing as well though - serialization. How likely is it that these events will be serialized by users? IMO it is almost inevitable. And for that, the split into two implementations is not ideal; they will have to jump through a lot of hoops to get that serialized (and more importantly deserialized) into the correct types.

With that in mind, I'd reconsider having it just a single type.

core/src/main/java/ai/timefold/solver/core/impl/phase/PhaseType.java
```diff
@@ -0,0 +1,52 @@
...
+import ai.timefold.solver.core.config.localsearch.LocalSearchPhaseConfig;
+import ai.timefold.solver.core.config.partitionedsearch.PartitionedSearchPhaseConfig;
+import ai.timefold.solver.core.config.phase.NoChangePhaseConfig;
+import ai.timefold.solver.core.config.phase.custom.CustomPhaseConfig;
+import ai.timefold.solver.core.impl.phase.NoChangePhase;
+
+/**
+ * The type of phase (for example, a Construction Heuristic).
+ */
+public enum PhaseType {
```
triceo, line 13: Regardless of the consideration above, I think it's overkill to have both the phase name exposed, and this type. Let's make it one or the other - expose either the phase name, or this type.

In order to have only as small API surface as possible, I'd get rid of this type (right now it is being exposed in public API) and only expose the phase name in the `EventProducerId` type. (I don't mind that it's a String in that case, because the string is just an identifier, and it doesn't carry additional information that'd have to be parsed out, such as phase ID.)

`PhaseType` can remain to facilitate the conversion, if you like, but don't put it in public API then.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -45,7 +48,12 @@ public TerminationStatus getTerminationStatus() {
 
     @Override
     public String getPhaseTypeString() {
```
triceo: Since `Phase` is not public API, we have some freedom here.
If you decide to keep `PhaseType` (see above), maybe we return that here, instead of the String? If we ever need to exhaustively switch over phase types, this will help.

triceo (commented):
I'll have to ask the platform guys about serialization of the events. That will give us some context here.

core/src/main/java/ai/timefold/solver/core/api/solver/event/EventProducerId.java
triceo: Having asked the platform folks, we now know that they do their own serialization; therefore we needn't worry about this aspect. Still, I have a feeling that it'd work better as a single type. (Even though the platform does their own serialization, other users of our API may not; this API has been around since forever, there will be other users.)

core/src/main/java/ai/timefold/solver/core/impl/phase/PhaseType.java
```diff
@@ -0,0 +1,61 @@
...
+     * The type of phase associated with {@link NoChangePhase}.
+     * 
+     * @deprecated Deprecated on account of {@link NoChangePhase} having no use.
+     */
+    @Deprecated(forRemoval = true, since = "1.28.0")
+    NO_CHANGE("No Change"),
+    /**
+     * The type of phase associated with {@link ConstructionHeuristicPhase}.
+     */
+    CONSTRUCTION_HEURISTIC("Construction Heuristics"),
```
triceo: Let's use singular here. Each phase runs exactly one.

core/src/main/java/ai/timefold/solver/core/impl/phase/PhaseType.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * The type of phase associated with {@link RuinRecreateConstructionHeuristicPhase}
+     */
+    RUIN_AND_RECREATE_CONSTRUCTION_HEURISTIC("Ruin & Recreate Construction Heuristics"),
```
triceo: Dtto.

triceo (commented):
We're missing the identification on the first initialized solution consumer. After that, we're good.

core/src/main/java/ai/timefold/solver/core/api/solver/SolverJobBuilder.java
```diff
@@ -100,20 +146,30 @@ public interface SolverJobBuilder<Solution_, ProblemId_> {
...
-            @NonNull FirstInitializedSolutionConsumer<? super Solution_> firstInitializedSolutionConsumer);
+    SolverJobBuilder<Solution_, ProblemId_> withFirstInitializedSolutionEventConsumer(
+            @NonNull Consumer<FirstInitializedSolutionEvent<Solution_>> firstInitializedSolutionEventConsumer);
+
+    /**
+     * As defined by {@link #withSolverJobStartedEventConsumer(Consumer)}.
+     *
+     * @deprecated Use {@link #withSolverJobStartedEventConsumer(Consumer)} instead.
+     */
+    default SolverJobBuilder<Solution_, ProblemId_>
```
triceo, line 160: Missing the `@Deprecated` annotation.

triceo (approved):
Only one comment. As agreed, I'll fix it myself.

PR #1916, opened 2025-11-06
chore: refactor CS index to support random access

triceo:
Refactors indexing in Constraint Streams so that the index supports random access to elements. This will be necessary for Neighborhoods.

While here, I also noticed that the constructors of streams and nodes were getting out of hand. The new store size trackers bring the number of arguments back under control.

First experiments did not show any significant performance changes in either direction, but I need to run more tests to confirm or deny that.


triceo:
Will not be merged, as I couldn't resolve the perf regressions.
This is happening instead:

https://github.com/TimefoldAI/timefold-solver/pull/1942

PR #1917, opened 2025-11-06
chore: improve memory consumption of reachable values structure

zepfred:
The PR updates the method used to create the reachable values structure to reduce memory usage.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
@@ -1,7 +1,6 @@
 package ai.timefold.solver.core.impl.heuristic.selector.common;
 
 import java.util.AbstractList;
-import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collections;
 import java.util.List;
@@ -183,26 +182,26 @@ public final class ReachableValues {
     }
 
     @NullMarked
-    private static final class ArrayIndexedList<T> extends AbstractList<Object> {
+    private static final class ArrayIndexedList<T, V> extends AbstractList<V> {
 
         private final int[] valueIndex;
         private final List<T> allValues;
-        private final @Nullable Function<T, Object> valueExtractor;
+        private final @Nullable Function<T, V> valueExtractor;
 
-        private ArrayIndexedList(int[] valueIndex, List<T> allValues, @Nullable Function<T, Object> valueExtractor) {
+        private ArrayIndexedList(int[] valueIndex, List<T> allValues, @Nullable Function<T, V> valueExtractor) {
             this.valueIndex = valueIndex;
             this.allValues = allValues;
             this.valueExtractor = valueExtractor;
         }
 
         @Override
-        public Object get(int index) {
+        public V get(int index) {
             if (index < 0 || index >= valueIndex.length) {
                 throw new ArrayIndexOutOfBoundsException(index);
             }
             var value = allValues.get(valueIndex[index]);
             if (valueExtractor == null) {
-                return value;
+                return (V) value;
             } else {
                 return valueExtractor.apply(value);
             }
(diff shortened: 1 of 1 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -110,30 +129,102 @@ public boolean matchesValueClass(Object value) {
...
-                randomAccessEntityList.add(entity);
+        boolean containsEntity(int entityIndex) {
+            return entityBitSet.get(entityIndex);
+        }
+
+        boolean containsValue(int valueIndex) {
+            return valueBitSet.get(valueIndex);
+        }
+
+        private static int[] extractAllIndexes(BitSet bitSet) {
```
triceo, line 160: There is a point at which the bitset no longer changes. (When the values/entities are no longer added.) At this point, you can cache the `int[]` and you never need to compute it again. Arguably, at that point, you can also completely clear and remove the `BitSet`s from memory, because they will not be necessary anymore.

This actually tells me that this object has two states:
- "Larval" state when it's being built, when it's bitset-based.
- "Static" state that lasts forever once it's built, when it's no longer bitset-based.

If you refactor the code to behave like this, perhaps introducing a builder (larval state) for the object (static state), you will further improve memory use at rest, and possibly also performance. Not to mention that the code will be much easier to understand.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -110,30 +129,102 @@ public boolean matchesValueClass(Object value) {
...
+            }
+        }
+
+        @Override
+        public int size() {
+            return valueIndex.length;
+        }
+
+        @Override
+        public boolean equals(Object o) {
```
triceo: Do you actually need to compare the lists for equality?
That is very expensive and I'm interested where we actually use it.
If we don't, let's remove equals/hashcode, in order not to suggest that equality comparison is supported.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -110,30 +129,102 @@ public boolean matchesValueClass(Object value) {
...
+        List<Object> getRandomAccessValueList(List<ReachableItemValue> allValues) {
+            if (onDemandRandomAccessValueList == null) {
+                onDemandRandomAccessValueList = new ArrayIndexedList<>(extractAllIndexes(valueBitSet), allValues, v -> v.value);
+            }
+            return onDemandRandomAccessValueList;
+        }
+    }
+
+    @NullMarked
+    private static final class ArrayIndexedList<T> extends AbstractList<Object> {
```
triceo: `T` on the outside, but `Object` on the inside?
The two should match.

triceo (commented):
This is more complex than I originally expected, but I talked to Matej and he confirmed that it works for him. I'm offering further suggestions to clean up this code and possibly improve its performance a bit.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
zepfred: Sonar suggestion

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
zepfred: > Arguably, at that point, you can also completely clear and remove the BitSets from memory, because they will not be necessary anymore.

We still need the `BitSet` to determine whether a value or entity is reachable. The logic from `ReachableValues::isEntityReachable` and `ReachableValues::isValueReachable` uses these data structures to verify that.

The decision to compute the list of values and entities only when needed was made to reduce memory consumption.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
triceo: In this case, I'd ignore it. Introducing code we don't need is not a good idea.
(Unless we actually need it, in which case I'd like to see where.)

triceo (approved):

PR #1930, opened 2025-11-21
chore: improve performance of joins

triceo:
Even though my attempt to unify CS and Neighborhood joining was ultimately unsuccessful, the investigation still bore some fruit.

By avoiding some iteration in the nodes, I managed to improve performance of joining by up to 5 % (statistically significant) in my microbenchmarks.
I also unify the handling of store sizing, which leads to a significant simplification in node constructors.

I recommend doing this review commit-by-commit, as the things will make more sense that way.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIndexedJoinNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIndexedJoinNode.java
@@ -94,7 +94,7 @@ public abstract class AbstractIndexedJoinNode<LeftTuple_ extends AbstractTuple, 
             // Assume the following scenario:
             // - The join is of two entities of the same type, both filtering out unassigned.
             // - One entity became unassigned, so the outTuple is getting retracted.
-            // - The other entity became assigned, as is therefore getting inserted.
+            // - The other entity became assigned, and is therefore getting inserted.
             //
             // This means the filter would be called with (unassignedEntity, assignedEntity),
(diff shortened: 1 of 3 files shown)


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/bi/JoinBiEnumeratingStream.java
```diff
@@ -45,26 +45,13 @@ public void collectActiveEnumeratingStreams(Set<AbstractEnumeratingStream<Soluti
...
+        var positionTracker =
+                buildHelper.getTupleStorePositionTracker(this, leftParent.getTupleSource(), rightParent.getTupleSource());
         var node = indexerFactory.hasJoiners()
-                ? new IndexedJoinBiNode<>(indexerFactory,
-                        buildHelper.reserveTupleStoreIndex(leftParent.getTupleSource()),
-                        buildHelper.reserveTupleStoreIndex(leftParent.getTupleSource()),
-                        buildHelper.reserveTupleStoreIndex(leftParent.getTupleSource()),
-                        buildHelper.reserveTupleStoreIndex(rightParent.getTupleSource()),
-                        buildHelper.reserveTupleStoreIndex(rightParent.getTupleSource()),
-                        buildHelper.reserveTupleStoreIndex(rightParent.getTupleSource()),
```
triceo, line 58: Through the refactoring, I was aiming at this craziness, which was getting propagated through 3 layers of inheritance. Now each node is responsible for reserving their own indexes.

PR #1943, opened 2025-11-24
ci: use Aikido safe-chain with pnpm

triceo:


.github/workflows/pull_request_secure.yml
```diff
@@ -222,11 +222,14 @@ jobs:
           fetch-depth: 0 # Otherwise merge will fail on account of not having history.
       - name: Install pnpm
         uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061
+      - name: Use Aikido Sec Safe Chain
+        run: |
+          pnpm install -g @aikidosec/safe-chain@1.1.7
+          safe-chain setup
```
jennedebleser: For CI they recommend this:
```suggestion
          safe-chain setup-ci
```

triceo:
Since this is using the secure workflow, we will only see these changes once merged. Therefore merging.

PR #1947, opened 2025-11-25
chore: migrate MemberAccessor and SolutionCloner to Gizmo2

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
```diff
@@ -48,21 +48,21 @@ public boolean writeMemberValue(MethodDescriptor setter, BytecodeCreator bytecod
...
 
     @Override
     public Type getType() {
         try {
-            return declaringClass.getDeclaredField(fieldDescriptor.getName()).getGenericType();
+            return declaringClass.getDeclaredField(fieldDescriptor.name()).getGenericType();
         } catch (NoSuchFieldException e) {
             throw new IllegalStateException(
-                    "Cannot find field (" + fieldDescriptor.getName() + ") on class (" + declaringClass + ").",
+                    "Cannot find field (" + fieldDescriptor.name() + ") on class (" + declaringClass + ").",
```
triceo: When we touch old code, we typically refactor the exceptions there to the new `formatted(...)` pattern, which is much easier to read. Please apply consistently throughout this PR.

triceo (approved):
LGTM after comments are resolved; Sonar also makes some good points.

Side note: I appreciate that they removed ASM (maybe we could do the same from the rest of our codebase), but some of these bwd compatibility breaking changes are there seemingly just to annoy people. Renaming `*Descriptor` to `*Desc` is a perfect example of something they did maybe for some internal consistency somewhere, but to the users, it's a completely unjustified pain. Let that be a lesson to us.

PR #1949, opened 2025-11-26
fix: always update successors of single directional graph when non-declarative variables of parent are referenced

### LINKED ISSUE #1948: SingleDirectionalParentVariableReferenceGraph incorrectly assumes successors cannot change if parent did not change
### Discussed in https://github.com/TimefoldAI/timefold-solver/discussions/1946

Given a mixed model:

```java
@PlanningEntity
public class TestdataMixedEntity extends TestdataObject {
    @PlanningListVariable
    List<TestdataMixedValue> valueList;

    public TestdataMixedEntity() {
        valueList = new ArrayList<>();
    }

    public TestdataMixedEntity(String code) {
        super(code);
        valueList = new ArrayList<>();
    }

    public List<TestdataMixedValue> getValueList() {
        return valueList;
    }

    public void setValueList(List<TestdataMixedValue> valueList) {
        this.valueList = valueList;
    }
}
```

```java
@PlanningEntity
public class TestdataMixedValue {
    @PlanningVariable
    Integer delay;

    @InverseRelationShadowVariable(sourceVariableName = "valueList")
    private TestdataMixedEntity vehicle;

    @PreviousElementShadowVariable(sourceVariableName = "valueList")
    private TestdataMixedValue previous;

    @ShadowVariable(supplierName = "previousDelaySupplier")
    private Integer previousDelay;

    @ShadowSources({
            "previous",
            "previous.delay"
    })
    private Integer previousDelaySupplier() {
        if (previous == null) {
            return null;
        } else {
            return previous.delay;
        }
    }

    public Integer getDelay() {
        return delay;
    }

    public void setDelay(Integer delay) {
        this.delay = delay;
    }

    public TestdataMixedEntity getVehicle() {
        return vehicle;
    }

    public void setVehicle(TestdataMixedEntity vehicle) {
        this.vehicle = vehicle;
    }

    public TestdataMixedValue getPrevious() {
        return previous;
    }

    public void setPrevious(TestdataMixedValue previous) {
        this.previous = previous;
    }

    public Integer getPreviousDelay() {
        return previousDelay;
    }

    public void setPreviousDelay(Integer previousDelay) {
        this.previousDelay = previousDelay;
    }
}
```

SingleDirectionalParentVariableReferenceGraph will be chosen (since all shadows only depends on variables on previous).
SingleDirectionalParentVariableReferenceGraph assumes that if no shadows of a changed entity is changed, then no shadows of its successors can change (unless the successor had a variable that changed). This is incorrect; consider `[TestdataMixedValue("a"), TestdataMixedValue("b")]`; if `a.delay` changes from `1` to `2`, then its `previousDelay` is unchanged (as it has no previous), but `b.previousDelay` should change from `1` to `2`.

As such, this optimization should be disabled when any non-declarative variables (genuine or shadow) are accessed from parent.

Christopher-Chianelli:
SingleDirectionalParentVariableReferenceGraph assumes that if no shadows of a changed entity is changed, then no shadows of its successors can change (unless the successor had a variable that changed). This is incorrect; consider [TestdataMixedValue("a"), TestdataMixedValue("b")]; if a.delay changes from 1 to 2, then its previousDelay is unchanged (as it has no previous), but b.previousDelay should change from 1 to 2.

SingleDirectionalParentVariableReferenceGraph now always update successors in the case that a declarative variable depends on a non-declarative variable in its parent.

Fixes https://github.com/TimefoldAI/timefold-solver/issues/1948.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
@@ -193,6 +193,17 @@ public class DefaultShadowVariableSessionFactory<Solution_> {
         var declarativeShadowVariables = graphDescriptor.solutionDescriptor().getDeclarativeShadowVariableDescriptors();
         var sortedDeclarativeVariables = topologicallySortedDeclarativeShadowVariables(declarativeShadowVariables);
 
+        var canTerminateEarly = true;
+        for (var declarativeShadowVariable : declarativeShadowVariables) {
+            for (var source : declarativeShadowVariable.getSources()) {
+                for (var sourceReference : source.variableSourceReferences()) {
+                    if (!sourceReference.isTopLevel() && !sourceReference.isDeclarative()) {
+                        canTerminateEarly = false;
+                    }
+                }
+            }
+        }
+
         var topologicalSorter =
                 getTopologicalSorter(graphDescriptor.solutionDescriptor(),
                         Objects.requireNonNull(graphDescriptor.changedVariableNotifier().innerScoreDirector()),
@@ -200,7 +211,9 @@ public class DefaultShadowVariableSessionFactory<Solution_> {
 
         return new SingleDirectionalParentVariableReferenceGraph<>(graphDescriptor.consistencyTracker(),
                 sortedDeclarativeVariables,
-                topologicalSorter, graphDescriptor.changedVariableNotifier(), graphDescriptor.entities());
+                topologicalSorter, graphDescriptor.changedVariableNotifier(),
+                canTerminateEarly,
+                graphDescriptor.entities());
     }
 
     private static <Solution_> List<DeclarativeShadowVariableDescriptor<Solution_>>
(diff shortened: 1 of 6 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -193,14 +193,27 @@ static <Solution_> VariableReferenceGraph buildSingleDirectionalParentGraph(
         var declarativeShadowVariables = graphDescriptor.solutionDescriptor().getDeclarativeShadowVariableDescriptors();
         var sortedDeclarativeVariables = topologicallySortedDeclarativeShadowVariables(declarativeShadowVariables);
 
+        var canTerminateEarly = true;
+        for (var declarativeShadowVariable : declarativeShadowVariables) {
+            for (var source : declarativeShadowVariable.getSources()) {
+                for (var sourceReference : source.variableSourceReferences()) {
+                    if (!sourceReference.isTopLevel() && !sourceReference.isDeclarative()) {
+                        canTerminateEarly = false;
```
triceo: Aren't we missing a `break;` here?
Once set to false, there is no point in further iterating, is there?

triceo (approved):
LGTM when comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
triceo: Considering the multiple nesting, probably best to create a new method and just `return`.

PR #1952, opened 2025-11-26
fix: add PlanningEntity's comparatorClass and comparatorFactoryClass as bean defining annotation values in Quarkus

### LINKED ISSUE #1951: Native build fails when using comparatorClass in @PlanningEntity
**Describe the bug**

When using `@PlanningEntity(comparatorClass = MyComparator.class)` in a Quarkus application compiled to GraalVM native image, the application fails at runtime with: `IllegalArgumentException: Impossible state: could not find the EntityDescriptor's comparatorClass generated Gizmo supplier`.

The issue occurs because Timefold uses Gizmo to generate supplier classes for comparators at build-time, but these generated classes are not properly registered for GraalVM native-image compilation.

**Expected behavior**

The comparator should work in native mode, just like it does in JVM mode (`mvn quarkus:dev` or `java -jar`). The solver should initialize successfully and use the custom entity difficulty comparator.

**Actual behavior**

The application crashes at runtime when initializing the solver:

```
java.lang.IllegalArgumentException: Impossible state: could not find the EntityDescriptor(ai.timefold.exam.domain.CandidateNotice)'s comparatorClass (ai.timefold.exam.comparator.CandidateNoticeComparator) generated Gizmo supplier.
	at ai.timefold.solver.core.config.util.ConfigUtils.newInstance(ConfigUtils.java:23)
	at ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor.processSorting(EntityDescriptor.java:321)
	at ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor.processEntityAnnotations(EntityDescriptor.java:246)
	at ai.timefold.solver.core.impl.domain.entity.descriptor.EntityDescriptor.processAnnotations(EntityDescriptor.java:203)
	at ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor.buildSolutionDescriptor(SolutionDescriptor.java:157)
	at ai.timefold.solver.core.impl.solver.DefaultSolverFactory.buildSolutionDescriptor(DefaultSolverFactory.java:181)
```

**To Reproduce**

1. Create a planning entity with a comparator:
```java
@PlanningEntity(comparatorClass = CandidateNoticeComparator.class)
public class CandidateNotice {
    @PlanningId
    private Integer id;

    @PlanningVariable(valueRangeProviderRefs = "timeSlotRange")
    private TimeSlot timeSlot;

    // ... other fields
}
```

2. Create the comparator class:
```java
public class CandidateNoticeComparator implements Comparator<CandidateNotice> {
    @Override
    public int compare(CandidateNotice a, CandidateNotice b) {
        return Integer.compare(
            a.getTest().getCandidatesPerSession(),
            b.getTest().getCandidatesPerSession()
        );
    }
}
```

3. Add native profile to `pom.xml`:
```xml
<profiles>
    <profile>
        <id>native</id>
        <properties>
            <quarkus.package.type>native</quarkus.package.type>
        </properties>
    </profile>
</profiles>
```

4. Build native image:
```bash
mvn package -Pnative
```

5. Run the native executable - it crashes when initializing the solver

**Attempted fixes that did NOT work:**
- Adding `@RegisterForReflection` (Quarkus annotation) to the comparator class
- Creating `META-INF/native-image/reflect-config.json` with full reflection configuration:
```json
[
  {
    "name": "com.example.CandidateNoticeComparator",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "queryAllDeclaredConstructors": true,
    "queryAllDeclaredMethods": true,
    "unsafeAllocated": true
  }
]
```

**Workaround:**
Remove `comparatorClass` from the `@PlanningEntity` annotation. The solver will use default entity ordering instead.

## Environment

**Timefold Solver Version or Git ref**: 1.28.0

```
openjdk version "21.0.9" 2025-01-21
OpenJDK Runtime Environment Mandrel-21.0.9+1 (build 21.0.9+1)
OpenJDK 64-Bit Server VM Mandrel-21.0.9+1 (build 21.0.9+1, mixed mode, sharing)
```

```
Linux 6.8.0-87-generic x86_64
```

**Additional dependencies:**
- Quarkus: 3.28.3
- timefold-solver-quarkus: 1.28.0
- timefold-solver-quarkus-jackson: 1.28.0
- Build image: `quay.io/quarkus/ubi9-quarkus-mandrel-builder-image:jdk-21`

## Additional information

- This works perfectly in JVM mode (`mvn quarkus:dev` or running the JAR)
- The issue only occurs with GraalVM native compilation
- The Timefold Quarkus extension should either:
  1. Detect `comparatorClass` attributes during build-time and properly register the Gizmo-generated suppliers for native-image
  2. Or document this limitation in the native-image documentation

The root cause appears to be that the Gizmo-generated supplier code is not being properly included in the native image, despite reflection configuration.

Christopher-Chianelli:
Fixes #1951

### DIFF
--- a/quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/DotNames.java
+++ b/quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/DotNames.java
@@ -152,7 +152,7 @@ public final class DotNames {
         PLANNING_SCORE(DotNames.PLANNING_SCORE, "scoreDefinitionClass"),
         PLANNING_SOLUTION(DotNames.PLANNING_SOLUTION, "solutionCloner"),
         PLANNING_ENTITY(DotNames.PLANNING_ENTITY, "pinningFilter", "difficultyComparatorClass",
-                "difficultyWeightFactoryClass"),
+                "difficultyWeightFactoryClass", "comparatorClass", "comparatorFactoryClass"),
         PLANNING_VARIABLE(DotNames.PLANNING_VARIABLE, "strengthComparatorClass",
                 "strengthWeightFactoryClass"),
         CUSTOM_SHADOW_VARIABLE(DotNames.CUSTOM_SHADOW_VARIABLE, "variableListenerClass"),
--- a/quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/DotNames.java
+++ b/quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/DotNames.java
@@ -1,6 +1,7 @@
 package ai.timefold.solver.quarkus.deployment;
 
-import java.util.Arrays;
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


triceo:
This is very brittle; it's not reasonable to expect that people will remember to add every new field here as well. Case in point: I believe we also introduced the comparator on list variable.

Surely we can reflect over the annotations and detect those fields? These annotations are public API; any field we find is user-facing.

Christopher-Chianelli:
> This is very brittle; it's not reasonable to expect that people will remember to add every new field here as well. Case in point: I believe we also introduced the comparator on list variable.
> 
> Surely we can reflect over the annotations and detect those fields? These annotations are public API; any field we find is user-facing.

I have mixed feelings on this: in theory, we could add an annotation method that return a class that is not a bean, although we haven't yet. Additionally, when a new Annotation class is added, it still need to be added to DotNames (in particular, some annotations define member accessors, and others define planning entities; we cannot deduce that info from an annotation class).

triceo (approved):
I agree it is not perfect. Especially the fact that new annotations still need to be manually added. But IMO it is an improvement over status quo.

PR #1959, opened 2025-12-02
perf: improve runtime performance of value ranges

triceo:
Makes two independent changes:

- Deduplication doesn't actually require creating a bitset. We can compare the value ranges directly. With the hash trick that was already there, it will be equally cheap and faster on account of dropping the `BitSet`.
- The array list was very inefficient. It required creating a list of `Integer` (not `int`, therefore boxing all the time), and even that list was unnecessary as we can perform the same operation on the fly.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
@@ -43,8 +43,7 @@ public final class ReachableValues<Entity_, Value_> {
         this.expectedSupertypeOfValue = expectedSupertypeOfValue;
         this.valueRangeSorter = valueRangeSorter;
         this.acceptsNullValue = acceptsNullValue;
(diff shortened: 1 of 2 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/ReachableValues.java
```diff
@@ -212,50 +204,49 @@
...
-        private final List<Integer> valueIndex;
-        private final List<Type_> allValues;
-        private final @Nullable Function<Type_, Value_> valueExtractor;
-
-        private ArrayIndexedList(List<Integer> valueIndex, List<Type_> allValues,
-                @Nullable Function<Type_, Value_> valueExtractor) {
-            this.valueIndex = valueIndex;
-            this.allValues = allValues;
-            this.valueExtractor = valueExtractor;
+    private static final class BitSetIndexedList<Type_, Value_> extends AbstractList<Value_> {
```
zepfred, line 207: I liked the simplification of the code

zepfred (approved):
LGTM!

PR #1960, opened 2025-12-02
fix: FilteringValueSelector.ofAssigned checks if selection is null

Christopher-Chianelli:
Value selectors can return nulls as a valid selection. In particular, when entity value ranges are used with allows unassigned, `null` may be returned.


zepfred (approved):

zepfred:
@triceo the change makes sense to me. I remember that, when working on the entity-value range nearby support, a solution for when there is no reachable value for a selection was to return `null`. So, that change can explain why we are only now seeing this error, because previously, the Nearby selector couldn't return a null value.

https://github.com/TimefoldAI/timefold-solver-enterprise/blob/bcdea34f2c12f8a34418494e70480bfbf3d09075/core/src/main/java/ai/timefold/solver/enterprise/core/nearby/common/ListVariableNearbyDistanceMatrix.java#L147

triceo:
Replaced by other PRs.

PR #1962, opened 2025-12-03
feat: supporting parameters for the value Range annotated method

zepfred:
The PR extends the solver model configuration by allowing value-range annotated methods to accept the solution as a parameter.

```
    @ValueRangeProvider(id = "valueRange")
    public List<TestdataValue> getValueRange(TestdataEntityProvidingWithParameterSolution solution) {
        return valueRange;
    }
```

### DIFF
--- a/docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
+++ b/docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
@@ -1270,8 +1270,8 @@ Java::
 A `ValueRangeProvider` that uses the solution as a parameter for the planning entity consumes less memory
 compared to the previous approach,
 as it does not require storing information at the entity level.
-Additionally,
-improvements in memory consumption can be observed during the serialization process when using Quarkus or Spring Boot.
+This also helps reduce the overhead of the xref:using-timefold-solver/modeling-planning-problems.adoc#cloningASolution[solution cloning],
+which can become significant for huge data sets.
 ====
 
 [#referencingValueRangeProviders]


core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/AbstractFromPropertyValueRangeDescriptor.java
```diff
@@ -108,9 +109,17 @@ public boolean isCountable() {
         return countable;
     }
 
+    private Object executeGetter(Object bean, Object parameter) {
```
zepfred: @triceo I'm unsure whether we should define a separate extended accessor field and use that instead of `instanceof` operation.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
```diff
@@ -184,28 +187,45 @@ public String generateMethodAccessor(@Nullable AnnotationInstance annotationInst
...
                     Thread.currentThread().getContextClassLoader());
         }
-        var memberInfo = new GizmoMemberInfo(member, requiredReturnType, annotationClass);
+        var memberInfo = new GizmoMemberInfo(descriptor, accessorInfo.returnTypeRequired(),
+                descriptor.getMethodParameterType() != null, annotationClass);
         GizmoMemberAccessorImplementor.defineAccessorFor(generatedClassName, classOutput, memberInfo);
         return generatedClassName;
     }
 
+    private static Method getDeclaredMethod(Class<?> declaringClass, String methodName) throws NoSuchMethodException {
```
zepfred: @triceo, this method may not be backward-compatible because if someone is currently using a value-range method annotated, and another method with a single parameter with the same name exists, it will fail.

zepfred: However, we need that now because we introduced this new feature.

triceo:
I'm asking @Christopher-Chianelli to review first, on account of the Gizmo thing.
I'll take a look when we know the thing is functionally OK.

zepfred:
I'm changing it to a draft because I want to remove the new interface `ExtendedMemberAccessor`.

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/GizmoMemberAccessorEntityEnhancer.java
Christopher-Chianelli: You know the method name, return type, and parameters from `methodInfo`; you can use that look up the exact matching method.

Christopher-Chianelli (changes requested):
GIZMO code looks good, but you should change how `getDeclaredMethod` works in GizmoMemberAccessorEntityEnhancer. The rest LGTM.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/AbstractReadOnlyExtendedGizmoMemberAccessor.java
```diff
@@ -0,0 +1,18 @@
...
+
+    @Override
+    public Object executeGetter(Object bean) {
+        throw new UnsupportedOperationException(
+                "The method executeGetter(Object) without parameter is not supported. Maybe call executeGetter(Object, Object) from ExtendedMemberAccessor instead.");
+    }
+
+    @Override
+    public <Fact_, Result_> Function<Fact_, Result_> getGetterFunction() {
+        throw new UnsupportedOperationException("The method getGetterFunction() is not supported.");
```
triceo: These are impossible states. Therefore the exception doesn't need an explanation, and it doesn't need any maybes. It just needs to start with "Impossible state: "

Please apply consistently throughout the code, wherever the exception is a defensive check for situations which the code is designed not to reach.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -330,11 +331,21 @@ private void processSorting(PlanningEntity entityAnnotation) {
...
             descriptorPolicy.addFromEntityValueRangeProvider(memberAccessor);
         }
     }
 
+    private void assertGetterParameterType(MemberAccessor memberAccessor) {
+        if (memberAccessor.acceptsParameter()
+                && !memberAccessor.getGetterMethodParameterType().equals(getSolutionDescriptor().getSolutionClass())) {
+            throw new IllegalStateException("The parameter type (%s) of the method (%s) must match the solution (%s)."
+                    .formatted(memberAccessor.getGetterMethodParameterType().getTypeName(), memberAccessor.getName(),
+                            getSolutionDescriptor().getSolutionClass().getSimpleName()));
```
triceo: The solution class is a user type, therefore we do not give a simple name, but a fully qualified name. Applies to both the method parameter type and to the solution class itself.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -330,11 +331,21 @@ private void processSorting(PlanningEntity entityAnnotation) {
     private void processValueRangeProviderAnnotation(DescriptorPolicy descriptorPolicy, Member member) {
         if (((AnnotatedElement) member).isAnnotationPresent(ValueRangeProvider.class)) {
             var memberAccessor = descriptorPolicy.getMemberAccessorFactory().buildAndCacheMemberAccessor(member,
-                    FIELD_OR_READ_METHOD, ValueRangeProvider.class, descriptorPolicy.getDomainAccessType());
+                    FIELD_OR_READ_METHOD_WITH_PARAMETER, ValueRangeProvider.class, descriptorPolicy.getDomainAccessType());
```
triceo: Shouldn't this be `FIELD_OR_READ_METHOD_WITH_OPTIONAL_PARAMETER`?
It doesn't feel right to mix this with situations where the parameter is required.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/FromSolutionPropertyValueRangeDescriptor.java
```diff
@@ -21,12 +21,12 @@ public FromSolutionPropertyValueRangeDescriptor(int ordinal, GenuineVariableDesc
...
     @Override
     public <T> ValueRange<T> extractAllValues(Solution_ solution) {
-        return readValueRange(solution);
+        return readValueRange(solution, solution);
     }
 
     @Override
     public <T> ValueRange<T> extractValuesFromEntity(Solution_ solution, Object entity) {
-        return readValueRange(solution); // Needed for composite ranges on solution and on entity.
+        return readValueRange(solution, solution); // Needed for composite ranges on solution and on entity.
```
triceo: Having the solution as an argument twice is strange.
Let's figure out a way to only provide the argument once. It is clearly a special case.

core/src/test/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactoryTest.java
```diff
@@ -36,8 +36,8 @@ void setup() {
     void testReturnedMemberAccessor() throws NoSuchMethodException {
         Method member = TestdataEntity.class.getMethod("getValue");
         MemberAccessor memberAccessor =
-                GizmoMemberAccessorFactory.buildGizmoMemberAccessor(member, PlanningVariable.class, true,
-                        new GizmoClassLoader());
+                GizmoMemberAccessorFactory.buildGizmoMemberAccessor(member, PlanningVariable.class,
+                        AccessorInfo.of(true, false), new GizmoClassLoader());
```
triceo: `AccessorInfo.of(true, false)` says nothing.
`AccessorInfo.withReturnValueAndNoArguments()` says everything; the code explains itself and doesn't require me to see the implementation.

triceo (commented):
I want to see additional test cases:

- Solution value range fails fast when given an argument.
- Entity value range with an argument that is not a solution fails fast.
- Entity value range with an argument that extends a solution works. (Assuming the extension honors the rules of extending solutions.)

Also, please add documentation for this new mechanism by extending this section:

https://docs.timefold.ai/timefold-solver/latest/using-timefold-solver/modeling-planning-problems#valueRangeProviderOnPlanningEntity

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/FromSolutionPropertyValueRangeDescriptor.java
zepfred: Currently, passing a parameter to the solution range is permitted, but I will block that. Additionally, I will create a separate method when only providing the solution.

core/src/test/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactoryTest.java
triceo: Arguably, `AccessorInfo` can be an enum.

Christopher-Chianelli (approved):
I approve the code affecting GIZMO; please address any comments left by @triceo

docs/src/modules/ROOT/pages/using-timefold-solver/modeling-planning-problems.adoc
```diff
@@ -1243,6 +1243,37 @@ A `ValueRangeProvider` on the planning entity consumes more memory than `ValueRa
...
+----
+====
+
+[NOTE]
+====
+A `ValueRangeProvider` that uses the solution as a parameter for the planning entity consumes less memory
+compared to the previous approach,
+as it does not require storing information at the entity level.
+Additionally,
+improvements in memory consumption can be observed during the serialization process when using Quarkus or Spring Boot.
```
triceo: Maybe replace with "This also helps reduce the overhead of solution cloning (insert link), which can become significant for very large data sets."

PR #1963, opened 2025-12-03
ci: bump setup-graalvm action to 1.4.4

triceo:


triceo:
Done through Dependabot.

PR #1964, opened 2025-12-03
fix: correctly handle when HashSet and other set implementations are used as fields in Gizmo

Christopher-Chianelli:
In Gizmo 1, everything was casted to the expected type automatically (even for invalid casts such as `HashSet s = (HashSet) new TreeSet();`). Gizmo 2 doesn't do this, so we need to remove an invalid branches that would never be entered.

### DIFF
--- a/core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
@@ -14,6 +14,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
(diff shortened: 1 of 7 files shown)


core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
```diff
@@ -633,12 +634,29 @@ void cloneSetBasedSolution() {
         originalEntitySet.addAll(Arrays.asList(a, b, c, d));
         original.setEntitySet(originalEntitySet);
 
+        original.setEntityHashSet(new HashSet<>(originalEntitySet));
+        original.setEntityLinkedHashSet(new LinkedHashSet<>(originalEntitySet));
+        original.setEntityTreeSet(new TreeSet<>(entityComparator));
+        original.getEntityTreeSet().addAll(originalEntitySet);
```
triceo, line 640: Do we need to do anything for `List` v. `ArrayList` and `LinkedList`?
Or `Collection` in general?

I see we have some code handling that, but I may have missed a test for it.

triceo (approved):
LGTM when question resolved.

core/src/test/java/ai/timefold/solver/core/impl/domain/solution/cloner/AbstractSolutionClonerTest.java
Christopher-Chianelli: I think I discovered a much simpler approach where we simply reuse `constructCloneCollection` and `constructCloneMap` from `FieldAccessingSolutionCloner`.  Although I don't think we actually test all the cases of `FieldAccessingSolutionCloner` either.

triceo:
The native build is failing now.

Christopher-Chianelli:
Native issue should be fixed, although it was complaining `String.formatted(String[] args)` does not exist. So I look up the method on String.class to get the `String.formatted(String... args)` version which it seems to accept (at least on my machine).

PR #1966, opened 2025-12-04
chore: allow Neighborhoods and Move Selectors to coexist

triceo:


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/placer/PooledEntityPlacer.java
```diff
@@ -42,7 +42,7 @@ protected Placement<Solution_> createUpcomingSelection() {
             if (!moveIterator.hasNext()) {
                 return noUpcomingSelection();
             }
-            return new Placement<>(MoveIteratorFactory.adaptIterator(moveIterator));
+            return new Placement<>(MoveAdapters.toNewMove(moveIterator));
```
zepfred: Let's clarify that this method returns an iterator. It should be named `toNewMoveIterator`, and the same applies to `toLegacyMove`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/MoveAdapters.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * While the Neighborhoods API and Move Selectors API need to coexist,
+ * there are places where moves need to be converted between the old and new types.
+ * This class provides static methods to perform these conversions.
+ */
+@NullMarked
+public final class MoveAdapters {
```
zepfred, line 17: Why is the name written in the plural form?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/NewMoveAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+
+import ai.timefold.solver.core.api.score.director.ScoreDirector;
+import ai.timefold.solver.core.impl.move.director.MoveDirector;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.preview.api.move.Move;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Adapts {@link Move a new move}
```
zepfred, line 14: I understand that "Move" is the best name we have, but using it consistently across different APIs can be confusing.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/NewMoveAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+            ai.timefold.solver.core.impl.heuristic.move.Move<Solution_> {
+
+    @Override
+    public boolean isMoveDoable(ScoreDirector<Solution_> scoreDirector) {
+        return true; // New moves are always doable.
+    }
+
+    @Override
+    public void doMoveOnly(ScoreDirector<Solution_> scoreDirector) {
+        var innerScoreDirector = (InnerScoreDirector<Solution_, ?>) scoreDirector;
```
zepfred: We don't need a cast here because the method `getMoveDirector` handles that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/NewMoveAdapter.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private MoveDirector<Solution_, ?> getMoveDirector(ScoreDirector<Solution_> scoreDirector) {
+        return ((InnerScoreDirector<Solution_, ?>) scoreDirector).getMoveDirector();
+    }
+
+    @Override
+    public ai.timefold.solver.core.impl.heuristic.move.Move<Solution_>
+            rebase(ScoreDirector<Solution_> destinationScoreDirector) {
+        return new NewMoveAdapter<>(newMove.rebase(getMoveDirector(destinationScoreDirector)));
```
zepfred: Should we call `MoveAdapters.toLegacyMove`?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/factory/MoveIteratorFactory.java
```diff
@@ -18,11 +18,6 @@
  */
 public interface MoveIteratorFactory<Solution_, Move_ extends Move<Solution_>> {
 
-    static <Solution_> Iterator<ai.timefold.solver.core.preview.api.move.Move<Solution_>>
```
zepfred, line 21: Much cleaner now!

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -63,26 +59,37 @@ public DefaultLocalSearchPhaseFactory(LocalSearchPhaseConfig phaseConfig) {
...
+        if (neighborhoodsEnabled) {
+            phaseConfigPolicy.ensurePreviewFeature(PreviewFeature.NEIGHBORHOODS);
+        }
+        var moveSelectorConfig = phaseConfig.getMoveSelectorConfig();
+        var moveSelectorsEnabled = moveSelectorConfig != null;
+        if (moveSelectorsEnabled) {
+            if (!neighborhoodsEnabled) {
+                return buildMoveSelectorBasedDecider(phaseConfigPolicy, phaseTermination);
+            } else {
+                return buildMixedDecider(phaseConfigPolicy, phaseTermination, neighborhoodProviderClass);
```
zepfred, line 84: What are your thoughts on supporting this mixed approach? The only reasons I can think of is to gradually support neighborhoods or the fact that we are not removing the move selector feature.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -133,20 +127,39 @@ Convert your entities (%s) to use @%s instead."""
...
-        var moveRepository = new NeighborhoodsBasedMoveRepository<>(moveStreamFactory, moveStream,
+        return new NeighborhoodsBasedMoveRepository<>(moveStreamFactory, moveDefinitionList,
                 pickSelectionOrder() == SelectionOrder.RANDOM);
+    }
 
-        return buildDecider(moveRepository, configPolicy, termination);
+    private LocalSearchDecider<Solution_> buildMixedDecider(HeuristicConfigPolicy<Solution_> configPolicy,
+            PhaseTermination<Solution_> termination,
+            Class<? extends NeighborhoodProvider<Solution_>> neighborhoodProviderClass) {
+        var neighborhoodsMoveSelector = new NeighborhoodsMoveSelector<>(
```
zepfred, line 140: We can now have multiple move definitions in the Neighborhood approach. Would creating a single adapted neighborhood move selector act like adding a union move selector? Could this skew the probability of selection in both approaches?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/NeighborhoodsBasedMoveRepository.java
```diff
@@ -83,14 +88,13 @@ public void stepStarted(AbstractStepScope<Solution_> stepScope) {
...
-        Objects.requireNonNull(neighborhoodSession).settle();
+        neighborhoodSession.settle(); // The step will have made changes to the working memory; settle it again.
     }
 
     @Override
     public void phaseEnded(AbstractPhaseScope<Solution_> phaseScope) {
-        if (neighborhoodSession != null) {
-            neighborhoodSession = null;
-        }
+        neighborhoodSession = null;
```
zepfred, line 96: Let's also release `moveIterableList = null;`

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/RandomOrderNeighborhoodIterator.java
```diff
@@ -0,0 +1,51 @@
...
+            } else {
+                unexhaustedMoveIteratorList.remove(randomIndex);
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public Move<Solution_> next() {
+        if (!hasNext()) {
```
zepfred, line 52: It is possible that we might call `hasNext` twice, since `hasNext` is called first to verify the iterator and then to retrieve the move. We could add an if-block similar to that in `OriginalOrderNeighborhoodIterator`, or use a flag to prevent fetching the move again.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
triceo: This is going to be the key to the success of Neighborhoods. In the beginning, Neighborhoods won't be able to do much - they will improve over time. But we want people to start using them. If we make them exclusive (either move selector, or Neighborhoods, but never both) then that will make people _only_ use selectors until Neighborhoods are perfect - and that is not what we want.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
triceo: Selection probability in the mixed case is something that I'm not concerned with right now. The current behavior is:

- If you're only using selectors, nothing changes.
- If you're only using Neighborhoods, then each move definition has an equal chance of being picked.
- If you're mixing both, you have all move definitions on one side, all move selector on another side, and the first random choice is between Neighborhoods and Selectors.

This is IMO a good enough behavior for the first preview, but we will probably have to invent something better. That said, I intentionally did not spend much time here, because the situation where move selectors and neighborhoods have to live together is not the goal, and will not stay forever. I hope we'll get to the point of separating the two relatively quickly.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/MoveAdapters.java
triceo: It's not an adapter. It's a bunch of them. Also follows the same pattern as `Joiners`, `ConstraintCollectors`...

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/NewMoveAdapter.java
triceo: It is. :-( Not much we can do about it, we've been over this before.

triceo:
@zepfred Resolved your comments.

PR #1968, opened 2025-12-05
perf: dynamically calculate topological order to prevent cache invalidation

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/BaseTopologicalOrderGraph.java
```diff
@@ -39,23 +39,23 @@ public interface BaseTopologicalOrderGraph {
...
      */
-    record NodeTopologicalOrder(int nodeId, int order)
+    record NodeTopologicalOrder(int nodeId, BaseTopologicalOrderGraph graph)
             implements
                 Comparable<NodeTopologicalOrder> {
 
         @Override
         public int compareTo(NodeTopologicalOrder other) {
-            return order - other.order;
+            return graph.getTopologicalOrder(nodeId) - graph.getTopologicalOrder(other.nodeId);
```
triceo, line 73: This could, in theory, be unsafe - if the order of a node changes while the node is already in the collection, this would make the order of that collection invalid.

I think this cannot happen, because the collection is regularly cleared. Still, I think it's a good idea to mention that possible failure condition somewhere, and how we prevent it from happening.

triceo (approved):
LGTM once comment resolved and once Fred confirms the regression is gone.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraphBuilder.java
```diff
@@ -116,6 +116,15 @@ public void addAfterProcessor(GraphChangeType graphChangeType, VariableMetaModel
                 .add(consumer);
     }
 
+    public BaseTopologicalOrderGraph.NodeTopologicalOrder[] buildNodeTopologicalOrderArray(BaseTopologicalOrderGraph graph) {
```
triceo: IMO this method lives in the wrong place. It should live in the graph, as it creates instances of a type from there, and is exclusively called from there.

The only thing it needs from this type is `nodeList.size()`, which the builder can easily return in a getter.

PR #1990, opened 2025-12-21
perf: assorted improvements to help JVM to inline methods

triceo:
These changes are focused mostly on splitting large methods into smaller ones, as the JVM identified them as both hot and too big to inline.

This brings a minor statistically significant performance improvement in select use cases.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/CartesianProductMoveSelector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/CartesianProductMoveSelector.java
@@ -28,8 +28,8 @@ public class CartesianProductMoveSelector<Solution_> extends CompositeMoveSelect
 
     private final boolean ignoreEmptyChildIterators;
 
-    public CartesianProductMoveSelector(List<MoveSelector<Solution_>> childMoveSelectorList,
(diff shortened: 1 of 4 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/CartesianProductMoveSelector.java
```diff
@@ -82,98 +82,104 @@ public Iterator<Move<Solution_>> iterator() {
...
                 if (startingIndex < 0) {
                     return noUpcomingSelection();
                 }
                 // Clone to avoid CompositeMove corruption
                 System.arraycopy(subSelections, 0, moveList, 0, startingIndex);
                 moveList[startingIndex] = moveIteratorList.get(startingIndex).next(); // Increment the 4 in 004999
             }
-            for (int i = startingIndex + 1; i < childSize; i++) { // Increment the 9s in 004999
-                Iterator<Move<Solution_>> moveIterator = childMoveSelectorList.get(i).iterator();
+            for (var i = startingIndex + 1; i < childSize; i++) { // Increment the 9s in 004999
```
zepfred: I'm not sure if this comment, along with the previous one, is helpful in any way.

PR #1991, opened 2025-12-21
perf: further improve performance of precompute

triceo:
The update queue does deduplicate the updates. 
But we can skip iterating the entire collection, if we've already been over it once.

Another small bump in CS performance.

@Christopher-Chianelli Approach this with caution. I have seen this skip the collection more often than it iterates over it, but for that, the performance improvement is very small. There may be something I do not understand about this code.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
@@ -28,8 +28,7 @@ import org.jspecify.annotations.NullMarked;
  * @param <Tuple_>
  */
 @NullMarked
-public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
-        implements Propagator {
+public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple> implements Propagator {
 
     private final Set<Object> retractQueue;
     private final Set<Object> insertQueue;
@@ -42,14 +41,13 @@ public final class RecordAndReplayPropagator<Tuple_ extends AbstractTuple>
     private final Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier;
     private final UnaryOperator<Tuple_> internalTupleToOutputTupleMapper;
     private final Map<Object, List<Tuple_>> objectToOutputTuplesMap;
-    private final Map<Class<?>, Boolean> objectClassToIsEntitySourceClass;
+    private final Set<Object> alreadyUpdatingSet = Collections.newSetFromMap(new IdentityHashMap<>());
+    private final Map<Class<?>, Boolean> objectClassToIsEntitySourceClassMap;
 
     private final StaticPropagationQueue<Tuple_> propagationQueue;
 
-    public RecordAndReplayPropagator(
-            Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier,
-            UnaryOperator<Tuple_> internalTupleToOutputTupleMapper,
-            TupleLifecycle<Tuple_> nextNodesTupleLifecycle, int size) {
+    public RecordAndReplayPropagator(Supplier<BavetPrecomputeBuildHelper<Tuple_>> precomputeBuildHelperSupplier,
+            UnaryOperator<Tuple_> internalTupleToOutputTupleMapper, TupleLifecycle<Tuple_> nextNodesTupleLifecycle, int size) {
         this.precomputeBuildHelperSupplier = precomputeBuildHelperSupplier;
         this.internalTupleToOutputTupleMapper = internalTupleToOutputTupleMapper;
         this.objectToOutputTuplesMap = CollectionUtils.newIdentityHashMap(size);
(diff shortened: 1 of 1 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/bigdecimal/BigDecimalValueRange.java
```diff
@@ -158,11 +158,9 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        var hash = 7;
-        hash = 31 * hash + Objects.hashCode(from);
+        var hash = Objects.hashCode(from);
```
zepfred: Do we want to align the implementation with `Objects#hash`? If that's the case, we should use `var hash = 31 * Objects.hashCode(from);`.

zepfred: The comment may be relevant to all other `hashCode` functions.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
```diff
@@ -88,10 +88,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        var hash = 7;
-        hash = 31 * hash + Boolean.hashCode(isValueImmutable);
-        hash = 31 * hash + Objects.hashCode(list);
-        return hash;
+        return 31 * Boolean.hashCode(isValueImmutable) + Objects.hashCode(list);
```
zepfred:
```suggestion
        return Boolean.hashCode(isValueImmutable) + 31 * Objects.hashCode(list);
```

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/SetValueRange.java
```diff
@@ -89,10 +89,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        var hash = 7;
-        hash = 31 * hash + Boolean.hashCode(isValueImmutable);
-        hash = 31 * hash + Objects.hashCode(set);
-        return hash;
+        return 31 * Boolean.hashCode(isValueImmutable) + Objects.hashCode(set);
```
zepfred:
```suggestion
        return Boolean.hashCode(isValueImmutable) + 31 * Objects.hashCode(set);
```

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/composite/CompositeCountableValueRange.java
```diff
@@ -102,9 +102,6 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        var hash = 7;
-        hash = 31 * hash + Boolean.hashCode(isValueImmutable);
-        hash = 31 * hash + Objects.hashCode(valueRangeList);
-        return hash;
+        return 31 * Boolean.hashCode(isValueImmutable) + Objects.hashCode(valueRangeList);
```
zepfred:
```suggestion
        return Boolean.hashCode(isValueImmutable) + 31 * Objects.hashCode(valueRangeList);
```

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/primdouble/DoubleValueRange.java
```diff
@@ -97,10 +97,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        var hash = 7;
-        hash = 31 * hash + Double.hashCode(from);
-        hash = 31 * hash + Double.hashCode(to);
-        return hash;
+        return 31 * Double.hashCode(from) + Double.hashCode(to);
```
zepfred:
```suggestion
        return Double.hashCode(from) + 31 * Double.hashCode(to);
```

core/src/main/java/ai/timefold/solver/core/impl/util/Pair.java
```diff
@@ -14,17 +14,17 @@ public record Pair<Key_, Value_>(Key_ key, Value_ value) {
...
                 && Objects.equals(value, other.value);
     }
 
     @Override
     public int hashCode() { // Often used in hash-based collections; the JDK-generated default is too slow.
-        var hash = 7;
-        hash = 31 * hash + Objects.hashCode(key);
-        hash = 31 * hash + Objects.hashCode(value);
-        return hash;
+        return 31 * Objects.hashCode(key) + Objects.hashCode(value);
```
zepfred:
```suggestion
        return Objects.hashCode(key) + 31 * Objects.hashCode(value);
```

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/buildin/collection/ListValueRange.java
zepfred: This suggestion and all subsequent comments may need updates to align with `Objects#hash`.

triceo:
I removed all unrelated hashing changes.

zepfred:
> I removed all unrelated hashing changes.

Okay, I'll skip the review then.

Christopher-Chianelli (approved):

PR #1992, opened 2025-12-21
chore: refactor tuples and scoring

triceo:
- Changes the tuples to be all backed by a single type.
- Refactors scoring to use that single type.
- Applies the same partial scoring patterns from bendable to all other score types as well.

This is in preparation for the future, when we will drop the `int`-based scores, making the scoring bimorphic (`long` and `BigDecimal` only). This will improve performance in the future, but did not cause regressions now.

The PR is large, but that's because of all the copy-paste. This PR is, in fact, 3 conceptual changes, copied all over the place 20 times.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/WeightedScoreImpacter.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/WeightedScoreImpacter.java
@@ -4,6 +4,9 @@ import java.math.BigDecimal;
 
 import ai.timefold.solver.core.api.score.Score;
 
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
 /**
  * There are several valid ways how an impacter could be called from a constraint stream:
  *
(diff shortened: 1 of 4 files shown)


ge0ffrey:
Note there is probably a merge conflict on IndexerFactory with https://github.com/TimefoldAI/timefold-solver/pull/1999
As usual, the later one merged will have to figure it out?

triceo:
> Note there is probably a merge conflict on IndexerFactory with #1999 As usual, the later one merged will have to figure it out?

At the moment, I do not know which one that will be. That said, this PR has 2000 lines of code changes - rebasing this would be a nightmare.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/WeightedScoreImpacter.java
```diff
@@ -43,42 +43,42 @@ public interface WeightedScoreImpacter<Score_ extends Score<Score_>, Context_ ex
      * @param constraintMatchSupplier ignored unless constraint match enableds
      * @return never null
      */
-    UndoScoreImpacter impactScore(int matchWeight, ConstraintMatchSupplier<Score_> constraintMatchSupplier);
+    ScoreImpact<Score_> impactScore(int matchWeight, ConstraintMatchSupplier<Score_> constraintMatchSupplier);
```
Christopher-Chianelli: Add `@Nullable` to `constraintMatchSupplier`; it is being passed `null` in every `ImpactHandler` implementation.

PR #2002, opened 2026-01-06
feat: enable partitioned search for list variable

zepfred:
Counterpart for TimefoldAI/timefold-solver-enterprise#411

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
@@ -100,6 +100,7 @@ public sealed class MoveDirector<Solution_, Score_ extends Score<Score_>>
         externalScoreDirector.beforeVariableChanged(variableDescriptor, entity);
         variableDescriptor.setValue(entity, newValue);
         externalScoreDirector.afterVariableChanged(variableDescriptor, entity);
+        externalScoreDirector.triggerVariableListeners();
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -605,6 +605,10 @@ Partitioning only works for use cases where the planning entities and value rang
 without any of the constraints crossing boundaries between partitions.
 ====
 
+[WARNING]
+====
+Models with xref:using-timefold-solver/modeling-planning-problems.adoc#chainedPlanningVariable[chained variables] or xref:using-timefold-solver/modeling-planning-problems.adoc#mixedModels[mixed models] cannot apply the partitioned search method.
```
triceo: I wonder if we should make the mixed model work.
After all, it's an official feature now, and we want to maintain it going forward.

Not doing chained is completely fine.

docs/src/modules/ROOT/pages/enterprise-edition/enterprise-edition.adoc
```diff
@@ -717,6 +721,14 @@ add the `solutionPartitionerCustomProperties` element and use xref:using-timefol
   </partitionedSearch>
 ----
 
+[WARNING]
+====
+When working with a model that includes a xref:using-timefold-solver/modeling-planning-problems.adoc#planningListVariable[list variable],
+it's essential to remember that a xref:using-timefold-solver/modeling-planning-problems.adoc#planningValue[planning value] can be assigned to only one xref:using-timefold-solver/modeling-planning-problems.adoc#planningEntity[entity].
+Therefore,
+ensure that your partitioner returns distinct value lists for each entity
+or the solver may fail to optimize the problem.
```
triceo: IMO this is something we should fail fast on, as opposed to making it the user's problem to watch for it.

triceo (approved):

PR #2010, opened 2026-01-11
Neighborhoods: finishing touches

triceo:
The PR has 3 distinct parts:

- First commit introduces the default neighborhood provider, allowing for easier co-existence between move selectors and neighborhoods.
- Second commit significantly increases performance of neighborhoods by not materializing the lists in indexers if not absolutely necessary.
- Third commit fixes a critical bug when neighborhoods would not be notified of all changes to the solution. This potentially also improves performance of neighborhoods, since it scopes which updates neighborhoods gets to see, but performance is secondary in this respect.

The 3 commits require each other to function, therefore they are provided as a single PR. But they will be merged without squashing.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
@@ -2,6 +2,7 @@ package ai.timefold.solver.core.impl.domain.variable;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 
 import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.index.IndexShadowVariableDescriptor;
@@ -16,6 +17,7 @@ import ai.timefold.solver.core.preview.api.domain.metamodel.PositionInList;
 final class ListVariableState<Solution_> {
 
     private final ListVariableDescriptor<Solution_> sourceVariableDescriptor;
+    private final Consumer<Object> notifier;
 
     private ExternalizedIndexVariableProcessor<Solution_> externalizedIndexProcessor = null;
     private ExternalizedListInverseVariableProcessor<Solution_> externalizedInverseProcessor = null;
@@ -27,8 +29,10 @@ final class ListVariableState<Solution_> {
     private int unassignedCount = 0;
     private Map<Object, MutablePosition> elementPositionMap;
 
-    public ListVariableState(ListVariableDescriptor<Solution_> sourceVariableDescriptor) {
+    public ListVariableState(ListVariableDescriptor<Solution_> sourceVariableDescriptor,
+            Consumer<Object> notifier) {
         this.sourceVariableDescriptor = sourceVariableDescriptor;
+        this.notifier = notifier;
     }
 
     public void linkShadowVariable(IndexShadowVariableDescriptor<Solution_> shadowVariableDescriptor) {
@@ -121,49 +125,26 @@ final class ListVariableState<Solution_> {
                                 .formatted(sourceVariableDescriptor, element, index, oldPosition));
             }
         }
+        var elementUpdateSent = false;
         if (externalizedIndexProcessor != null) {
-            externalizedIndexProcessor.addElement(scoreDirector, element, index);
+            elementUpdateSent = externalizedIndexProcessor.addElement(scoreDirector, element, index);
         }
         if (externalizedInverseProcessor != null) {
-            externalizedInverseProcessor.addElement(scoreDirector, entity, element);
+            elementUpdateSent = externalizedInverseProcessor.addElement(scoreDirector, entity, element) || elementUpdateSent;
         }
         if (externalizedPreviousElementProcessor != null) {
-            externalizedPreviousElementProcessor.setElement(scoreDirector, elements, element, index);
+            elementUpdateSent = externalizedPreviousElementProcessor.setElement(scoreDirector, elements, element, index)
+                    || elementUpdateSent;
         }
         if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index);
+            elementUpdateSent =
+                    externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index) || elementUpdateSent;
         }
         unassignedCount--;
-    }
-
-    public void removeElement(Object entity, Object element, int index) {
-        if (requiresPositionMap) {
-            var oldPosition = elementPositionMap.remove(element);
-            if (oldPosition == null) {
-                throw new IllegalStateException(
-                        "The supply for list variable (%s) is corrupted, because the element (%s) at index (%d) was already unassigned (%s)."
-                                .formatted(sourceVariableDescriptor, element, index, oldPosition));
-            }
-            var oldIndex = oldPosition.getIndex();
-            if (oldIndex != index) {
-                throw new IllegalStateException(
-                        "The supply for list variable (%s) is corrupted, because the element (%s) at index (%d) had an old index (%d) which is not the current index (%d)."
-                                .formatted(sourceVariableDescriptor, element, index, oldIndex, index));
-            }
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
+        if (!elementUpdateSent) {
+            notifier.accept(element);
         }
-        if (externalizedIndexProcessor != null) {
-            externalizedIndexProcessor.removeElement(scoreDirector, element);
-        }
-        if (externalizedInverseProcessor != null) {
-            externalizedInverseProcessor.removeElement(scoreDirector, entity, element);
-        }
-        if (externalizedPreviousElementProcessor != null) {
-            externalizedPreviousElementProcessor.unsetElement(scoreDirector, element);
-        }
-        if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.unsetElement(scoreDirector, element);
-        }
-        unassignedCount++;
     }
 
     public void unassignElement(Object element) {
@@ -175,37 +156,49 @@ final class ListVariableState<Solution_> {
                                 .formatted(sourceVariableDescriptor, element));
             }
         }
+        var elementUpdateSent = false;
         if (externalizedIndexProcessor != null) {
-            externalizedIndexProcessor.unassignElement(scoreDirector, element);
+            elementUpdateSent = externalizedIndexProcessor.unassignElement(scoreDirector, element);
         }
         if (externalizedInverseProcessor != null) {
-            externalizedInverseProcessor.unassignElement(scoreDirector, element);
+            elementUpdateSent = externalizedInverseProcessor.unassignElement(scoreDirector, element) || elementUpdateSent;
         }
         if (externalizedPreviousElementProcessor != null) {
-            externalizedPreviousElementProcessor.unsetElement(scoreDirector, element);
+            elementUpdateSent = externalizedPreviousElementProcessor.unsetElement(scoreDirector, element) || elementUpdateSent;
         }
         if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.unsetElement(scoreDirector, element);
+            elementUpdateSent = externalizedNextElementProcessor.unsetElement(scoreDirector, element) || elementUpdateSent;
         }
         unassignedCount++;
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
+        if (!elementUpdateSent) {
+            notifier.accept(element);
+        }
     }
 
     public boolean changeElement(Object entity, List<Object> elements, int index) {
         var element = elements.get(index);
         var difference = processElementPosition(entity, element, index);
+        var elementUpdateSent = false;
         if (difference.indexChanged && externalizedIndexProcessor != null) {
-            externalizedIndexProcessor.changeElement(scoreDirector, element, index);
+            elementUpdateSent = externalizedIndexProcessor.changeElement(scoreDirector, element, index);
         }
         if (difference.entityChanged && externalizedInverseProcessor != null) {
-            externalizedInverseProcessor.changeElement(scoreDirector, entity, element);
+            elementUpdateSent = externalizedInverseProcessor.changeElement(scoreDirector, entity, element) || elementUpdateSent;
         }
         // Next and previous still might have changed, even if the index and entity did not.
         // Those are based on what happened elsewhere in the list.
         if (externalizedPreviousElementProcessor != null) {
-            externalizedPreviousElementProcessor.setElement(scoreDirector, elements, element, index);
+            elementUpdateSent = externalizedPreviousElementProcessor.setElement(scoreDirector, elements, element, index)
+                    || elementUpdateSent;
         }
         if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index);
+            elementUpdateSent =
+                    externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index) || elementUpdateSent;
+        }
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
+        if (!elementUpdateSent) {
+            notifier.accept(element);
         }
         return difference.anythingChanged;
     }
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/anchor/ExternalizedAnchorVariableSupply.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/anchor/ExternalizedAnchorVariableSupply.java
@@ -2,6 +2,7 @@ package ai.timefold.solver.core.impl.domain.variable.anchor;
 
 import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.function.Consumer;
 
 import ai.timefold.solver.core.impl.domain.variable.BasicVariableChangeEvent;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.VariableDescriptor;
@@ -23,12 +24,15 @@ public class ExternalizedAnchorVariableSupply<Solution_> implements
     protected final VariableDescriptor<Solution_> previousVariableDescriptor;
     protected final SingletonInverseVariableSupply nextVariableSupply;
     protected final Map<Object, @Nullable Object> anchorMap;
+    private final Consumer<Object> notifier;
 
     public ExternalizedAnchorVariableSupply(VariableDescriptor<Solution_> previousVariableDescriptor,
-            SingletonInverseVariableSupply nextVariableSupply) {
+            SingletonInverseVariableSupply nextVariableSupply,
+            Consumer<Object> notifier) {
         this.previousVariableDescriptor = previousVariableDescriptor;
         this.nextVariableSupply = nextVariableSupply;
         this.anchorMap = new IdentityHashMap<>();
+        this.notifier = notifier;
     }
 
     @Override
@@ -72,6 +76,7 @@ public class ExternalizedAnchorVariableSupply<Solution_> implements
         Object nextEntity = entity;
         while (nextEntity != null && anchorMap.get(nextEntity) != anchor) {
             anchorMap.put(nextEntity, anchor);
+            notifier.accept(nextEntity);
             nextEntity = nextVariableSupply.getInverseSingleton(nextEntity);
         }
     }
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
@@ -111,19 +111,17 @@ public sealed class MoveDirector<Solution_, Score_ extends Score<Score_>>
             return moveValueInList(variableMetaModel, sourceEntity, sourceIndex, destinationIndex);
         }
         var variableDescriptor = extractVariableDescriptor(variableMetaModel);
-
         externalScoreDirector.beforeListVariableChanged(variableDescriptor, sourceEntity, sourceIndex, sourceIndex + 1);
         var element = (Value_) variableDescriptor.removeElement(sourceEntity, sourceIndex);
         externalScoreDirector.afterListVariableChanged(variableDescriptor, sourceEntity, sourceIndex, sourceIndex);
-        externalScoreDirector.triggerVariableListeners();
 
-        externalScoreDirector.beforeListVariableChanged(variableDescriptor, destinationEntity, destinationIndex,
-                destinationIndex);
+        externalScoreDirector.beforeListVariableChanged(variableDescriptor,
+                destinationEntity, destinationIndex, destinationIndex);
         variableDescriptor.addElement(destinationEntity, destinationIndex, element);
-        externalScoreDirector.afterListVariableChanged(variableDescriptor, destinationEntity, destinationIndex,
-                destinationIndex + 1);
-        externalScoreDirector.triggerVariableListeners();
+        externalScoreDirector.afterListVariableChanged(variableDescriptor,
+                destinationEntity, destinationIndex, destinationIndex + 1);
 
+        externalScoreDirector.triggerVariableListeners();
         return element;
     }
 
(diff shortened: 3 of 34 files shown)


core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -66,20 +67,39 @@ public LocalSearchPhase<Solution_> buildPhase(int phaseIndex, boolean lastInitia
...
 
+    @SuppressWarnings({ "unchecked", "rawtypes" })
     private LocalSearchDecider<Solution_> buildDecider(HeuristicConfigPolicy<Solution_> phaseConfigPolicy,
             PhaseTermination<Solution_> phaseTermination) {
+        var neighborhoodsEnabled = phaseConfigPolicy.isPreviewFeatureEnabled(PreviewFeature.NEIGHBORHOODS);
         var neighborhoodProviderClass = phaseConfig.<Solution_> getNeighborhoodProviderClass();
-        var neighborhoodsEnabled = neighborhoodProviderClass != null;
-        if (neighborhoodsEnabled) {
-            phaseConfigPolicy.ensurePreviewFeature(PreviewFeature.NEIGHBORHOODS);
+        if (phaseConfigPolicy.isPreviewFeatureEnabled(PreviewFeature.NEIGHBORHOODS)) {
```
zepfred:
```suggestion
        if (neighborhoodsEnabled) {
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -66,20 +67,39 @@ public LocalSearchPhase<Solution_> buildPhase(int phaseIndex, boolean lastInitia
...
+            }
+            if (phaseConfigPolicy.getNearbyDistanceMeterClass() != null) {
+                throw new IllegalArgumentException(
+                        """
+                                The neighborhoodProviderClass (%s) is not compatible with using the top-level property nearbyDistanceMeterClass (%s).
+                                The Neighborhoods API does not yet support nearby selection.
+                                You may still configure nearby selection in move selectors individually."""
+                                .formatted(neighborhoodProviderClass, phaseConfigPolicy.getNearbyDistanceMeterClass()));
+            }
+        } else {
```
zepfred:
```suggestion
        } else if (neighborhoodProviderClass != null) {
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -66,20 +67,39 @@ public LocalSearchPhase<Solution_> buildPhase(int phaseIndex, boolean lastInitia
...
+            if (neighborhoodProviderClass != null) {
+                throw new UnsupportedOperationException("""
+                        The neighborhoodProviderClass (%s) can only be used if the %s preview feature is enabled.
+                        Maybe add <enablePreviewFeature>%s</enablePreviewFeature> to your solver configuration file?"""
+                        .formatted(neighborhoodProviderClass.getCanonicalName(), PreviewFeature.NEIGHBORHOODS,
+                                PreviewFeature.NEIGHBORHOODS));
+            }
```
zepfred:
```suggestion
            
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -133,9 +153,12 @@ private NeighborhoodsBasedMoveRepository<Solution_> buildNeighborhoodsBasedMoveR
...
             PhaseTermination<Solution_> termination,
             Class<? extends NeighborhoodProvider<Solution_>> neighborhoodProviderClass) {
+        var legacyMoveSelector = buildMoveSelector(configPolicy, neighborhoodProviderClass != null);
+        if (legacyMoveSelector == null) { // There were no move selectors configured.
+            return buildNeighborhoodsBasedDecider(configPolicy, termination, neighborhoodProviderClass);
+        }
         var neighborhoodsMoveSelector =
                 new NeighborhoodsMoveSelector<>(buildNeighborhoodsBasedMoveRepository(configPolicy, neighborhoodProviderClass));
-        var legacyMoveSelector = buildMoveSelector(configPolicy);
         if (legacyMoveSelector instanceof UnionMoveSelector<Solution_> unionMoveSelector) {
```
zepfred: We can move `finalMoveSelector` out of the `if` block and create it according to the configuration. Then, we can build the repository and the decider at the end of the method without duplicating the code.

zepfred: Arguably, the `else` block on line #166 may is not necessary. We can either throw the exception or continue executing the method.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -257,30 +280,32 @@ protected LocalSearchForager<Solution_> buildForager(HeuristicConfigPolicy<Solut
...
-                    && NearbyAutoConfigurationEnabled.class.isAssignableFrom(moveSelectorConfig.getClass())
-                    && !UnionMoveSelectorConfig.class.isAssignableFrom(moveSelectorConfig.getClass())) {
-                // The move selector config is not a composite selector, but it accepts Nearby autoconfiguration.
-                // We create a new UnionMoveSelectorConfig with the existing selector to enable Nearby autoconfiguration.
-                var moveSelectorCopy = (MoveSelectorConfig) moveSelectorConfig.copyConfig();
-                var updatedConfig = new UnionMoveSelectorConfig().withMoveSelectors(moveSelectorCopy);
-                moveSelectorFactory = MoveSelectorFactory.create(updatedConfig);
+            if (neighborhoodsEnabled) {
+                // Default moves are already provided by Neighborhoods.
+                return null;
```
zepfred, line 277: This is not completely true. The list variable has `KOptListMoveSelectorConfig`, which is not currently supported by `Neighborhoods`. Should we return a single configuration with the KOpt move for list variables?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
```diff
@@ -257,30 +280,32 @@ protected LocalSearchForager<Solution_> buildForager(HeuristicConfigPolicy<Solut
...
+            } else {
+                return new UnionMoveSelectorFactory<Solution_>(determineDefaultMoveSelectorConfig(configPolicy))
+                        .buildMoveSelector(configPolicy, defaultCacheType, defaultSelectionOrder, true);
             }
-            moveSelector = moveSelectorFactory.buildMoveSelector(configPolicy, defaultCacheType, defaultSelectionOrder, true);
         }
-        return moveSelector;
+
+        AbstractMoveSelectorFactory<Solution_, ?> moveSelectorFactory = MoveSelectorFactory.create(moveSelectorConfig);
```
zepfred: Let's call the factory only after the config is updated. This way, we will call it only once.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodProvider.java
```diff
@@ -0,0 +1,46 @@
...
+                if (variableMetaModel instanceof PlanningListVariableMetaModel<Solution_, ?, ?> listVariableMetaModel) {
+                    hasListVariable = true;
+                    // TODO Implement 2-opt and 3-opt moves for list variables.
+                    builder.add(new ListChangeMoveProvider<>(listVariableMetaModel));
+                    builder.add(new ListSwapMoveProvider<>(listVariableMetaModel));
+                } else if (variableMetaModel instanceof PlanningVariableMetaModel<Solution_, ?, ?> basicVariableMetaModel) {
+                    builder.add(new ChangeMoveProvider<>(basicVariableMetaModel));
+                }
+            }
+            if (!hasListVariable) {
```
zepfred: What if it is a mixed model? We should add it if there is at least one basic variable.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -164,47 +168,115 @@ private void forEachManyIndexers(Object compositeKey, Consumer<T> tupleConsumer)
...
+            default -> getManyIndexers(compositeKey, index);
         };
     }
 
-    private List<? extends ListEntry<T>> asListSingleIndexer(Object compositeKey) {
+    private @Nullable ListEntry<T> getSingleIndexer(Object compositeKey, int index) {
         var indexKey = keyRetriever.apply(compositeKey);
         var entry = comparisonMap.firstEntry();
-        return boundaryReached(entry.getKey(), indexKey) ? Collections.emptyList() : entry.getValue().asList(compositeKey);
+        return boundaryReached(entry.getKey(), indexKey) ? null : entry.getValue().get(compositeKey, index);
```
zepfred: `getManyIndexers` return null if the index exceeds the size. Should we also check the size and return null if the index is greater than the size?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -164,47 +168,115 @@ private void forEachManyIndexers(Object compositeKey, Consumer<T> tupleConsumer)
...
-        return result;
+        return null;
     }
 
     @Override
     public String toString() {
         return "size = " + comparisonMap.size();
     }
 
+    private final class DefaultIterator implements Iterator<T> {
```
zepfred, line 236: Nice!

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/EqualsIndexer.java
```diff
@@ -108,18 +109,29 @@ public void forEach(Object compositeKey, Consumer<T> tupleConsumer) {
...
+    @Nullable
+    public ListEntry<T> get(Object compositeKey, int index) {
         Key_ indexKey = keyRetriever.apply(compositeKey);
         Indexer<T> downstreamIndexer = downstreamIndexerMap.get(indexKey);
         if (downstreamIndexer == null) {
-            return Collections.emptyList();
+            return null;
         }
-        return downstreamIndexer.asList(compositeKey);
+        return downstreamIndexer.get(compositeKey, index);
```
zepfred, line 127: Do we need to check the size?

triceo: No. This will bubble down to the backend, and the backend will throw an exception if we're out of bounds. This is really the only thing we could do here, only it would cost us so many CPU cycles to do that.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/RandomAccessIndexerBackend.java
```diff
@@ -41,13 +41,18 @@ public void forEach(Object compositeKey, Consumer<T> tupleConsumer) {
...
-        return tupleList.isEmpty();
+    public Iterator<T> iterator(Object compositeKey) {
+        return tupleList.iterator();
     }
 
     @Override
-    public List<ElementAwareArrayList.Entry<T>> asList(Object compositeKey) {
-        return tupleList.asList();
+    public ListEntry<T> get(Object compositeKey, int index) {
+        return tupleList.get(index);
```
zepfred, line 50: We can also throw an `IndexOutOfBoundsException` in this context whithout checking the size. IMO, we should clarify the desired behavior for the `Indexer#get` method. Please add documentation describing what occurs if the indexer exceeds the size at `Indexer` contract.

It appears that the `ComparisonIndexer` is the only class that can return `null` if the index exceeds the size. Should we make its behavior consistent with the other classes, or should we modify the others to return `null` as well?

core/src/main/java/ai/timefold/solver/core/impl/localsearch/DefaultLocalSearchPhaseFactory.java
triceo: No. The solution is to add these moves in a future version, not to make the configuration even more complex. This is a preview feature, it's not expected to be complete.

Also, side note: we likely won't do a K-Opt with Neighborhoods. K-Opt is super-complex, and we only ever use 2- and 3-opt. Therefore we'll do much simpler implementations of those.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/FilteredUniqueRandomSequence.java
```diff
@@ -54,17 +45,19 @@ private int pickIndex(Random workingRandom) {
             if (delegate.isEmpty()) {
                 throw new NoSuchElementException("No more elements to pick from.");
             }
-            delegate.remove(actualValueIndex);
+            delegate.removeWithoutReturn(actualValueIndex);
             // We try the same random index again; the underlying sequence will find the next best non-removed element.
             actualValueIndex = delegate.pickIndex(workingRandom, originalRandomIndex);
-            value = originalList.get(actualValueIndex).getElement();
+            value = accessor.get(actualValueIndex);
```
zepfred, line 51: I wonder if we should also move `originalRandomIndex` to `actualValueIndex`. This way, the next bit search would start from the previously filtered out value.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/IndexerBasedElementAccessor.java
```diff
@@ -0,0 +1,23 @@
...
+import ai.timefold.solver.core.impl.bavet.common.index.Indexer;
+
+public record IndexerBasedElementAccessor<T>(Indexer<T> indexer, Object compositeKey)
+        implements
+            ElementAccessor<T> {
+
+    @Override
+    public T get(int index) {
+        var entry = indexer.get(compositeKey, index);
+        if (entry == null) {
```
zepfred: We expect the value to be `null` if the index exceeds the size. My previoues comment connects to these accessor classes.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
triceo: I'll remove the nullability. This is an exceptional state, it can be handled with exceptions.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/FilteredUniqueRandomSequence.java
triceo: This is sensitive code - I have verified that this algorithm guarantees fair random selection. If we want to make optimizations, this experiment needs to be redone, and therefore I'd leave that for some future point in time; the current code is correct.

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/anchor/ExternalizedAnchorVariableSupply.java
```diff
@@ -23,12 +24,15 @@ public class ExternalizedAnchorVariableSupply<Solution_> implements
     protected final VariableDescriptor<Solution_> previousVariableDescriptor;
     protected final SingletonInverseVariableSupply nextVariableSupply;
     protected final Map<Object, @Nullable Object> anchorMap;
+    private final Consumer<Object> notifier;
```
zepfred, line 27: In analyzing the implementation of the supply classes, we could create a common abstract notification class that provides the notification method and stores the consumer. This approach allows us to update the related classes and reuse code while providing a separate common contract for notifications.

core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
```diff
@@ -111,19 +111,17 @@ public final <Entity_, Value_> void changeVariable(PlanningVariableMetaModel<Sol
             return moveValueInList(variableMetaModel, sourceEntity, sourceIndex, destinationIndex);
         }
         var variableDescriptor = extractVariableDescriptor(variableMetaModel);
-
         externalScoreDirector.beforeListVariableChanged(variableDescriptor, sourceEntity, sourceIndex, sourceIndex + 1);
-        var element = (Value_) variableDescriptor.removeElement(sourceEntity, sourceIndex);
+        Value_ element = (Value_) variableDescriptor.removeElement(sourceEntity, sourceIndex);
```
zepfred: let's use `var`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/RandomOrderNeighborhoodIterator.java
```diff
@@ -30,20 +29,19 @@ public RandomOrderNeighborhoodIterator(List<MoveIterable<Solution_>> moveIterabl
...
         }
         while (!unexhaustedMoveIteratorList.isEmpty()) {
             var randomIndex = workingRandom.nextInt(unexhaustedMoveIteratorList.size());
             currentMoveIterator = unexhaustedMoveIteratorList.get(randomIndex);
             if (currentMoveIterator.hasNext()) {
-                hasNext = true;
                 return true;
             } else {
                 unexhaustedMoveIteratorList.remove(randomIndex);
+                currentMoveIterator = null;
```
zepfred: We can set it to `null` only once, and that should be done at the end of the method.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/anchor/ExternalizedAnchorVariableSupply.java
triceo: I'd rather not work on any standardization here. This code will see changes soon, when we remove chained var; that may be a good time to look at what's common, and how this can be improved.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -316,16 +326,14 @@ private void assertInitScoreZeroOrLess() {
 
     @Override
     public void executeMove(Move<Solution_> move) {
+        setAllChangesWillBeUndoneBeforeStepEnds(false);
         moveDirector.execute(move);
+        setAllChangesWillBeUndoneBeforeStepEnds(true);
```
zepfred: I'm not sure we need to set it to `true`. It makes more sense to restore it to its previous state.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -337,7 +345,7 @@ public InnerScore<Score_> executeTemporaryMove(Move<Solution_> move, boolean ass
                     assertWorkingScoreFromScratch(score, move);
                     return score;
                 }) : moveDirector.executeTemporary(move);
-        allChangesWillBeUndoneBeforeStepEnds = false;
+        setAllChangesWillBeUndoneBeforeStepEnds(false);
```
zepfred: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -549,9 +565,13 @@ Attempting to change list variable (%s) on an entity (%s) in range [%d, %d), whi
     }
 
     @Override
-    public void afterListVariableChanged(ListVariableDescriptor<Solution_> variableDescriptor,
-            Object entity, int fromIndex, int toIndex) {
+    public void afterListVariableChanged(ListVariableDescriptor<Solution_> variableDescriptor, Object entity, int fromIndex,
+            int toIndex) {
         variableListenerSupport.afterListVariableChanged(variableDescriptor, entity, fromIndex, toIndex);
+        if (allChangesWillBeUndoneBeforeStepEnds) {
```
zepfred: We can simplify the update operations by calling the neighborhood notifier's accept method, which will ignore the update if it is not being tracked. I wonder if we could add a new method that accepts the operation type, allowing us to insert or retract as needed. This change simplifies the logic concerning neighborhood repository notifications.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -23,7 +20,7 @@
  * meaning that once a particular (A,B) pair has been returned,
  * it will never be returned again by this iterator.
  * This means that this random move iterator will eventually end.
- * 
+ *
```
zepfred:
```suggestion
 * <p>
```

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -37,7 +34,7 @@
  * Instead, it will terminate the iteration after selecting just a handful,
  * as the chance of accepting a move grows more and more with each new move.</li>
  * </ol>
- * 
+ *
```
zepfred:
```suggestion
 * <p>
```

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -56,7 +53,7 @@
  * <li>If the filter rejects the pair, (A,B) is discarded and a new B is selected.
  * This guarantees that A keeps its selection probability of (1/A).</li>
  * </ul>
- * 
+ * <p>
```
zepfred, line 54:
```suggestion
 * <p>
```

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
zepfred:
```suggestion
 * <p>
```

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -121,49 +125,26 @@ public void addElement(Object entity, List<Object> elements, Object element, int
...
-                        "The supply for list variable (%s) is corrupted, because the element (%s) at index (%d) was already unassigned (%s)."
-                                .formatted(sourceVariableDescriptor, element, index, oldPosition));
-            }
-            var oldIndex = oldPosition.getIndex();
-            if (oldIndex != index) {
-                throw new IllegalStateException(
-                        "The supply for list variable (%s) is corrupted, because the element (%s) at index (%d) had an old index (%d) which is not the current index (%d)."
-                                .formatted(sourceVariableDescriptor, element, index, oldIndex, index));
-            }
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
```
zepfred, line 144: Should we send the notification only if there has been a change?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -175,37 +156,49 @@ public void unassignElement(Object element) {
...
         if (externalizedPreviousElementProcessor != null) {
-            externalizedPreviousElementProcessor.unsetElement(scoreDirector, element);
+            elementUpdateSent = externalizedPreviousElementProcessor.unsetElement(scoreDirector, element) || elementUpdateSent;
         }
         if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.unsetElement(scoreDirector, element);
+            elementUpdateSent = externalizedNextElementProcessor.unsetElement(scoreDirector, element) || elementUpdateSent;
         }
         unassignedCount++;
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
```
zepfred, line 173: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
```diff
@@ -175,37 +156,49 @@ public void unassignElement(Object element) {
...
+            elementUpdateSent = externalizedPreviousElementProcessor.setElement(scoreDirector, elements, element, index)
+                    || elementUpdateSent;
         }
         if (externalizedNextElementProcessor != null) {
-            externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index);
+            elementUpdateSent =
+                    externalizedNextElementProcessor.setElement(scoreDirector, elements, element, index) || elementUpdateSent;
+        }
+        // Trigger notifier if none of the previous methods triggered a shadow var update for this element.
+        if (!elementUpdateSent) {
```
zepfred, line 200: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableState.java
triceo: At this point, a change is pretty much guaranteed.
It is better to change it than to not; the updates will be deduplicated later, so it's a much less of an issue as compared to not sending the update.

PR #1999, opened 2025-12-26
feat: Joiners.contain() and Joiners.containedIn()

ge0ffrey:
Before

```
    Constraint speakerUnavailableTimeslot(ConstraintFactory factory) {
        return factory.forEach(Talk.class)
                .join(Speaker.class,
                        filtering((talk, speaker) -> talk.hasSpeaker(speaker)
                                && speaker.getUnavailableTimeslots().contains(talk.getTimeslot())))
                .penalize(...)
                .asConstraint(...);
    }
```

After (**bigO order of magnitude more scalable**)

```
    Constraint speakerUnavailableTimeslot(ConstraintFactory factory) {
        return factory.forEach(Talk.class)
                .join(Speaker.class,
                        contain(Talk::getSpeakers, speaker -> speaker),
                        containedIn(Talk::getTimeslot, Speaker::getUnavailableTimeslots))
                .penalize(...)
                .asConstraint(...);
    }
```

Refactorings

- Unify IndexerFactory.buildIndexer() into one switch across the joiner types
-- Do not create single/composite KeyRetriever in EqualsIndexer and ComparisonIndexer to unify to one constructor.
-- Extract KeyRetriever creation to IndexerFactory (that class is responsibility for those key tricks)
-- Implement equal joiner flip() as no-op, so all right bridges indexer can just get a flip.
- Renamed EqualsIndexer to EqualIndexer because its Joiners.equal() and not Joiners.equals()

Open questions (added as todo's)
- For Neighberhood, why track a left index?
- For neighberhoods, why not flip the right bridge?
    I 'd image this code should feel exactly the same as scoring, but not left side
    so always flip basically?
    
    
   
Tasks

- [X] API
- [ ] javadocs (API, ...)
- [X] Refactorings to plug it in
- [X] Unit tests
- [x] Indexer implementations
- [x] Use it in Quickstarts (conference scheduling, ...)
- [ ] Docs

PR #2022, opened 2026-01-16
feat: containing/containedIn/intersecting joiners

triceo:
Before

```
    Constraint speakerUnavailableTimeslot(ConstraintFactory factory) {
        return factory.forEach(Talk.class)
                .join(Speaker.class,
                        filtering((talk, speaker) -> talk.hasSpeaker(speaker)
                                && speaker.getUnavailableTimeslots().contains(talk.getTimeslot())))
                .penalize(...)
                .asConstraint(...);
    }
```

After (**bigO order of magnitude more scalable**)

```
    Constraint speakerUnavailableTimeslot(ConstraintFactory factory) {
        return factory.forEach(Talk.class)
                .join(Speaker.class,
                        containing(Talk::getSpeakers, speaker -> speaker),
                        containedIn(Talk::getTimeslot, Speaker::getUnavailableTimeslots))
                .penalize(...)
                .asConstraint(...);
    }
```

This is mostly a work of @ge0ffrey, I just rebased it and smoothened some rough edges.
Replaces https://github.com/TimefoldAI/timefold-solver/pull/1999.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/IndexerFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/IndexerFactory.java
@@ -35,7 +35,7 @@ import ai.timefold.solver.core.impl.util.Triple;
  * Parent indexers delegate to their children,
  * until they reach the ultimate {@link IndexerBackend}.
  * <p>
- * Example 1: EQUAL+LESS_THAN joiner will become EqualsIndexer -> ComparisonIndexer -> NoneIndexer.
+ * Example 1: EQUAL+LESS_THAN joiner will become EqualIndexer -> ComparisonIndexer -> NoneIndexer.
  * <p>
  * Indexers have an id, which is the position of the indexer in the chain.
  * Top-most indexer has id 0, and the id increases as we go down the hierarchy.
@@ -497,7 +497,7 @@ public final class IndexerFactory<Right_> {
                 requiresRandomAccess ? RandomAccessIndexerBackend::new : LinkedListIndexerBackend::new;
         if (!hasJoiners()) { // NoneJoiner results in NoneIndexer.
             return backendSupplier.get();
-        } else if (joiner.getJoinerCount() == 1) { // Single joiner maps directly to EqualsIndexer or ComparisonIndexer.
+        } else if (joiner.getJoinerCount() == 1) { // Single joiner maps directly to EqualIndexer or ComparisonIndexer.
             var joinerType = joiner.getJoinerType(0);
             KeyUnpacker<?> keyUnpacker = new SingleKeyUnpacker<>();
             return buildIndexerPart(isLeftBridge, joinerType, keyUnpacker, backendSupplier);
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/KeyUnpacker.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/KeyUnpacker.java
@@ -5,7 +5,7 @@ import java.util.function.Function;
 /**
  * A function that retrieves keys of a composite key for an {@link Indexer}.
  * For example, {@code join(..., equals(), lessThan(), greaterThan())} has 3 keys.
- * Given {@code ("a", 7, 9)} the key retriever for {@code lessThan()} retrieves {@code 7}.
+ * Given {@code ("a", 7, 9)} the key unpacker for {@code lessThan()} retrieves {@code 7}.
  * 
  * @param <Key_>
  */
--- a/docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
+++ b/docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
@@ -563,9 +563,9 @@ pairing a match from each side:
 the paired matches have a `Comparable` property following the prescribed ordering.
 - `overlapping()`: the paired matches have two properties (a _start_ and an _end_ property) of the same `Comparable` type
 that both represent an interval which overlap.
-- `containing()`: the match on the left is a collection containing the match on the right.
-- `containedIn()`: the match on the right is a collection containing the match on the left.
-- `intersecting()`: the collections on both sides have at least one element in common.
+//- `containing()`: the match on the left is a collection containing the match on the right.
+//- `containedIn()`: the match on the right is a collection containing the match on the left.
+//- `containingAnyOf()`: the collections on both sides have at least one element in common.
 
 All `Joiners` methods have an overloaded method to use the same property of the same class on both stream sides.
 For example, calling `equal(Shift::getEmployee)` is the same as calling `equal(Shift::getEmployee, Shift::getEmployee)`.
@@ -649,8 +649,7 @@ and if the joiner does not significantly reduce the number of matches,
 this overhead will outweigh the benefits.
 
 For small datasets, filters are often faster than joiners.
-Some of the more advanced joiners, such as `containing()`,  `containedIn()` or `intersecting()`,
-may require very large datasets to outperform filters.
+Some of the more advanced joiners may require very large datasets to outperform filters.
 
 When it comes to performance, we recommend you follow these guidelines:
 


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/IntersectingIndexer.java
```diff
@@ -0,0 +1,285 @@
...
+                    if (index < seenCount + size) {
+                        return downstreamIndexer.get(queryCompositeKey, index - seenCount);
+                    } else {
+                        seenCount += size;
+                    }
+                }
+            }
+        }
+        throw new IndexOutOfBoundsException("Index: %d".formatted(index));
+    }
```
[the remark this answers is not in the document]
triceo: I think this comment is wrong. `distinctingSet` exists exactly to prevent this behavior.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainedInIndexer.java
```diff
@@ -0,0 +1,194 @@
...
+@NullMarked
+final class ContainedInIndexer<T, Key_, KeyCollection_ extends Collection<Key_>> implements Indexer<T> {
+
+    private final KeyUnpacker<Key_> modifyKeyUnpacker;
+    private final KeyUnpacker<KeyCollection_> queryKeyUnpacker;
+    private final Supplier<Indexer<T>> downstreamIndexerSupplier;
+    /**
+     * See {@link EqualIndexer} for explanation of the parameters.
+     */
+    private final Map<Key_, Indexer<T>> downstreamIndexerMap = new HashMap<>(16, 0.5f);
```
Christopher-Chianelli, line 32: These `HashMap` parameters were found after testing IIRC. Might be better to have them in `CollectionUtils` so they don't need to be copied and pasted everywhere.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingIndexer.java
```diff
@@ -0,0 +1,152 @@
...
+    private long unremovedSize = 0;
+
+    /**
+     * @param keyUnpacker determines if it immediately goes to a {@link IndexerBackend} or if it uses a {@link CompositeKey}.
+     * @param downstreamIndexerSupplier the supplier of the downstream indexer
+     */
+    @SuppressWarnings("unchecked")
+    public ContainingIndexer(KeyUnpacker<Key_> keyUnpacker, Supplier<Indexer<T>> downstreamIndexerSupplier) {
+        this.modifyKeyUnpacker = Objects.requireNonNull((KeyUnpacker<KeyCollection_>) keyUnpacker);
+        this.queryKeyUnpacker = Objects.requireNonNull(keyUnpacker);
```
Christopher-Chianelli, line 43: These two fields always point to the same thing (but with different generics)?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingIndexer.java
```diff
@@ -0,0 +1,152 @@
...
+    @Override
+    public void remove(Object modifyCompositeKey, ListEntry<T> entry) {
+        unremovedSize--;
+        var indexKeyCollection = modifyKeyUnpacker.apply(modifyCompositeKey);
+        var children = ((CompositeListEntry<Key_, T>) entry).children();
+        if (indexKeyCollection.size() != children.size()) {
+            throw new IllegalStateException("""
+                    Impossible state: the tuple (%s) with composite key (%s) has a different number of children (%d) \
+                    than the index key collection size (%d)."""
+                    .formatted(entry, modifyCompositeKey, children.size(), indexKeyCollection.size()));
```
Christopher-Chianelli, line 76: You might want to add to the Javadoc using variable collection properties such as `PlanningListVariable` or `InverseShadowVariable` is not allowed (since those will trigger this exception as they change size).

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/IndexerFactory.java
```diff
@@ -497,48 +499,63 @@ public <T> Indexer<T> buildIndexer(boolean isLeftBridge) {
...
+            }
+            case CONTAINED_IN -> {
+                return new ContainedInIndexer<>(keyUnpacker, downstreamIndexerSupplier);
+            }
+            case INTERSECTING -> {
+                return new IntersectingIndexer<>(keyUnpacker, downstreamIndexerSupplier);
+            }
+            default -> throw new IllegalStateException(
+                    "Impossible state: The joiner type (" + joinerType + ") is not implemented.");
+        }
```
Christopher-Chianelli: You can make this a compiler error when new `Joiner` types are added:

```suggestion
        return switch (joinerType) {
            case EQUAL -> new EqualIndexer<>(keyUnpacker, downstreamIndexerSupplier);
            case LESS_THAN, LESS_THAN_OR_EQUAL, GREATER_THAN, GREATER_THAN_OR_EQUAL -> new ComparisonIndexer<>(joinerType, keyUnpacker, downstreamIndexerSupplier);
            case CONTAINING -> new ContainingIndexer<>(keyUnpacker, downstreamIndexerSupplier);
            case CONTAINED_IN -> new ContainedInIndexer<>(keyUnpacker, downstreamIndexerSupplier);
            case INTERSECTING -> new IntersectingIndexer<>(keyUnpacker, downstreamIndexerSupplier);
        };
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -0,0 +1,285 @@
...
+    private long unremovedSize = 0;
+
+    /**
+     * @param keyUnpacker determines if it immediately goes to a {@link IndexerBackend} or if it uses a {@link CompositeKey}.
+     * @param downstreamIndexerSupplier the supplier of the downstream indexer
+     */
+    @SuppressWarnings("unchecked")
+    public IntersectingIndexer(KeyUnpacker<Key_> keyUnpacker, Supplier<Indexer<T>> downstreamIndexerSupplier) {
+        this.modifyKeyUnpacker = Objects.requireNonNull((KeyUnpacker<KeyCollection_>) keyUnpacker);
+        this.queryKeyUnpacker = Objects.requireNonNull((KeyUnpacker<KeyCollection_>) keyUnpacker);
```
Christopher-Chianelli, line 47: These two fields will always have the exact same value (with the exact same generics)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/IntersectingIndexer.java
```diff
@@ -0,0 +1,285 @@
...
+            while (downstreamIndexerIterator.hasNext()) { // To avoid capturing a lambda on the hot path.
+                var tuple = downstreamIndexerIterator.next();
+                if (distinctingSet.add(tuple)) {
+                    var size = downstreamIndexer.size(queryCompositeKey);
+                    if (index < seenCount + size) {
+                        return downstreamIndexer.get(queryCompositeKey, index - seenCount);
+                    } else {
+                        seenCount += size;
+                    }
+                }
```
Christopher-Chianelli: Since the distinct set carries over, does it mess up the index calculation, since if a duplicate was encountered inside a different `downstreamIndexer`, `seenCount` will not be incremented?

Christopher-Chianelli (changes requested):
I would have at least 1 test that go through the `ScoreDirector` that uses these `Joiners`. Additionally, I do not think these Joiners will work if the backing collection changes (such as a `InverseShadowVariable`) given the same size assertions in the code (IIRC, everything is updated only when `calculateScore` is called, and not at `before`/`after` variable change, so the indexer will encounter a different sized collection if a shadow variable changes).

triceo:
Moving back to draft, as there are score corruptions.

PR #2028, opened 2026-01-20
feat: add ability to run a move during testing

triceo:


core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunner.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Dummy constraint provider that creates a single dummy constraint.
+     * This is needed to create a valid score director factory without actual constraint evaluation.
+     * <p>
+     * This class is an internal detail and must not be exposed to the user.
+     */
+    @NullMarked
+    static final class DummyConstraintProvider implements ConstraintProvider {
```
triceo: This forces some changes to Quarkus/Spring.
I made these manually, as I didn't trust the agent to figure it out without all the speccing and planning prerequisites.

.specify/specs/001-move-running-api/contracts/api-contract.md
```diff
@@ -0,0 +1,389 @@
...
+
+### Example 4: MoveRunner Reuse
+
+```java
+@Test
+void testMoveRunnerReuse() {
+    try (var runner = MoveRunner.build(Solution.class, Entity.class)) {
+        // Execute multiple moves in sequence
+        var context1 = runner.using(solution1);
+        context1.execute(move1);
```
Christopher-Chianelli: Where are `move1`/`move2` defined? (I know this is for a docs and doesn't need to compile per se, but the other examples define their moves)

.specify/specs/001-move-running-api/quickstart.md
```diff
@@ -0,0 +1,691 @@
...
+```java
+var solution = createTestSolution();
+try {
+    try (var runner = MoveRunner.build(...)) {
+        runner.using(solution).executeTemporarily(move, assertions);
+    }
+} catch (Exception e) {
+    // Solution is now in undefined state - discard it
+    solution = createTestSolution(); // Create fresh instance
+}
```
Christopher-Chianelli: If a test looked like this, I would call it a bad unit test.

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -84,7 +84,9 @@ public class ConfigUtils {
     public static <T> @NonNull T newInstance(@NonNull Supplier<String> ownerDescriptor, @NonNull String propertyName,
             @NonNull Class<T> clazz) {
         try {
-            return clazz.getDeclaredConstructor().newInstance();
+            var ctor = clazz.getDeclaredConstructor();
+            ctor.setAccessible(true); // For inner static non-public classes; public noarg ctors are effectively not.
```
Christopher-Chianelli: This will fail when Java enforces encapsulation; when is `setAccessible` needed?

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/Moves.java
```diff
@@ -33,9 +32,11 @@ public static <Solution_, Entity_, Value_> Move<Solution_> change(
...
     }
 
-    public static <Solution_, Entity_> Move<Solution_> swap(
-            PlanningVariableMetaModel<Solution_, Entity_, Object> variableMetaModel, Entity_ leftEntity, Entity_ rightEntity) {
-        return swap(Collections.singletonList(variableMetaModel), leftEntity, rightEntity);
+    @SuppressWarnings("unchecked")
+    public static <Solution_, Entity_, Value_> Move<Solution_> swap(
+            PlanningVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel, Entity_ leftEntity, Entity_ rightEntity) {
+        return swap(List.of((PlanningVariableMetaModel<Solution_, Entity_, Object>) variableMetaModel), leftEntity,
+                rightEntity);
```
Christopher-Chianelli, line 39: The previous implementation was better

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunContext.java
```diff
@@ -0,0 +1,102 @@
...
+     *        invoked only for {@link Exception} subclasses, not {@link Error}s
+     * @throws NullPointerException if move or exceptionHandler is null
+     * @throws IllegalStateException if the parent MoveRunner has been closed
+     * @throws Error if the move throws an Error (Errors are never suppressed)
+     */
+    public void execute(Move<Solution_> move, Consumer<Exception> exceptionHandler) {
+        Objects.requireNonNull(exceptionHandler, "exceptionHandler");
+
+        try {
+            scoreDirector.executeMove(Objects.requireNonNull(move, "move"));
```
Christopher-Chianelli: Should we throw an `AssertionError` when an exception is not thrown? I would only used this method if I am expecting an exception in a test.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunner.java
Christopher-Chianelli: Why do you need a `DummyConstraintProvider`? `MoveAsserter` works without one.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        // Act
+        try (var runner = MoveRunner.build(TestdataSolution.class, TestdataEntity.class)) {
+            var context = runner.using(solution);
+            context.execute(changeMove);
+        }
+
+        // Assert
+        assertThat(entity.getValue()).isEqualTo(newValue);
+        assertThat(entity.getValue()).isNotEqualTo(originalValue);
```
Christopher-Chianelli: Redundant assertion; same as asserting `originalValue != newValue`.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        // Act
+        try (var runner = MoveRunner.build(TestdataSolution.class, TestdataEntity.class)) {
+            var context = runner.using(solution);
+            context.execute(changeMove);
+        }
+
+        // Assert
+        assertThat(entity.getValue()).isNull();
+        assertThat(entity.getValue()).isNotEqualTo(originalValue);
```
Christopher-Chianelli: Ditto

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveTest.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @Test
+    void changeMoveMultipleEntities() {
+        // Arrange
+        var solution = TestdataSolution.generateSolution(3, 3);
+        var entity1 = solution.getEntityList().get(0);
+        var entity2 = solution.getEntityList().get(1);
+        var entity3 = solution.getEntityList().get(2);
```
Christopher-Chianelli, line 50: It is unclear what the values assigned to each entity are initially. I would set them explictly in the test.

core/src/test/java/ai/timefold/solver/core/preview/api/move/MoveRunnerTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class MoveRunnerTest {
+
+    @Test
+    void buildWithNullSolutionClass() {
+        assertThatNullPointerException()
+                .isThrownBy(() -> MoveRunner.build(null, TestdataEntity.class))
+                .withMessageContaining("solutionClass");
```
Christopher-Chianelli, line 18: TBH the exception message is a bit minimal; probably should be "The solutionClass (null) must not be null."

core/src/test/java/ai/timefold/solver/core/preview/api/move/MoveRunnerTest.java
```diff
@@ -0,0 +1,295 @@
...
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> MoveRunner.build(TestdataSolution.class))
+                .withMessageContaining("entityClasses must not be empty");
+    }
+
+    @Test
+    void buildWithNullEntityClasses() {
+        assertThatNullPointerException()
+                .isThrownBy(() -> MoveRunner.build(TestdataSolution.class, (Class<?>[]) null))
+                .withMessageContaining("entityClasses");
```
Christopher-Chianelli, line 30: Ditto

core/src/test/java/ai/timefold/solver/core/preview/api/move/MoveRunnerTest.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    @Test
+    void executeWithNullMove() {
+        var solution = TestdataSolution.generateSolution(2, 2);
+        try (var runner = MoveRunner.build(TestdataSolution.class, TestdataEntity.class)) {
+            var context = runner.using(solution);
+            assertThatNullPointerException()
+                    .isThrownBy(() -> context.execute(null))
+                    .withMessageContaining("move");
```
Christopher-Chianelli: Ditto

core/src/test/java/ai/timefold/solver/core/preview/api/move/MoveRunnerTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+    @Test
+    void usingAfterClose() {
+        var runner = MoveRunner.build(TestdataSolution.class, TestdataEntity.class);
+        runner.close();
+
+        var solution = TestdataSolution.generateSolution(2, 2);
+        assertThatThrownBy(() -> runner.using(solution))
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessageContaining("closed");
```
Christopher-Chianelli: Probably should be "The MoveRunner cannot be used after it is closed."

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,140 @@ Over time, we expect to add all commonly used move types to this collection,
...
+It is not thread-safe and not fine-tuned for performance.
+
+[#neighborhoodsMoveRunnerBasics]
+==== Basic usage
+
+The `MoveRunner` API follows a fluent builder pattern:
+
+[source,java,options="nowrap"]
+----
+try (var runner = MoveRunner.build(SolutionClass.class, EntityClass.class)) {
```
Christopher-Chianelli:
```suggestion
try (var runner = MoveRunner.build(Solution.class, Entity.class)) {
```

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,140 @@ Over time, we expect to add all commonly used move types to this collection,
...
+==== Permanent execution
+
+The `execute()` method applies a move permanently to the solution:
+
+[source,java,options="nowrap"]
+----
+@Test
+void testChangeMove() {
+    // Arrange
+    var solution = TestdataSolution.generateSolution(2, 2);
```
Christopher-Chianelli: I would not use `Testdata` in the docs and would use an example such as Employee Scheduling

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,140 @@ Over time, we expect to add all commonly used move types to this collection,
...
+==== Temporary execution with automatic undo
+
+The `executeTemporarily()` method allows you to test a move's effects without permanently modifying the solution.
+This is useful when you want to verify intermediate state or test that the solver's undo mechanism works correctly:
+
+[source,java,options="nowrap"]
+----
+@Test
+void testMoveThenUndo() {
+    var solution = TestdataSolution.generateSolution(2, 2);
```
Christopher-Chianelli: Ditto

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,140 @@ Over time, we expect to add all commonly used move types to this collection,
...
+[#neighborhoodsMoveRunnerExceptions]
+==== Exception handling
+
+If a move may throw exceptions during execution, you can provide an exception handler:
+
+[source,java,options="nowrap"]
+----
+@Test
+void testMoveWithExceptionHandler() {
+    var solution = createInvalidSolution(); // May cause exceptions
```
Christopher-Chianelli: That comment is pointless/not useful; I would create a move that uses an out of bound index

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -350,6 +351,15 @@ public InnerScore<Score_> executeTemporaryMove(Move<Solution_> move, boolean ass
                 }) : moveDirector.executeTemporary(move);
     }
 
+    @Override
+    public void executeTemporarily(Move<Solution_> move, Consumer<SolutionView<Solution_>> consumer) {
```
zepfred: First, let's rename the method to `executeTemporaryMove` for consistency. Additionally, I suggest merging the logic of the new method with the existing one. This way, we can reuse the logic and pass a dummy consumer when it is not necessary.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -190,6 +192,23 @@ private static <Score_ extends Score<Score_>> List<MatchAnalysis<Score_>> sumMat
      */
     InnerScore<Score_> executeTemporaryMove(Move<Solution_> move, boolean assertMoveScoreFromScratch);
 
+    /**
```
zepfred, line 197: See my previous related comment. We can provide a default implementation for the method without a consumer.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunContext.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @param <Solution_> the planning solution type
+ */
+@NullMarked
+public final class MoveRunContext<Solution_> {
+
+    private final InnerScoreDirector<Solution_, ?> scoreDirector;
+
+    MoveRunContext(InnerScoreDirector<Solution_, ?> scoreDirector) {
+        this.scoreDirector = Objects.requireNonNull(scoreDirector, "scoreDirector");
```
zepfred: I wonder how useful it is to check for nullity when using `@NullMarked`.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunContext.java
```diff
@@ -0,0 +1,102 @@
...
+     * discard the solution instance.</li>
+     * </ul>
+     *
+     * @param move the move to execute temporarily; must not be null
+     * @param assertions callback to verify the modified solution state; receives a {@link SolutionView}
+     *        for read-only access; must not be null
+     * @throws NullPointerException if move or assertions is null
+     * @throws IllegalStateException if the parent MoveRunner has been closed
+     */
+    public void executeTemporarily(Move<Solution_> move, Consumer<SolutionView<Solution_>> assertions) {
```
zepfred: Let's rename it to `executeTemporaryMove`.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunner.java
```diff
@@ -0,0 +1,167 @@
...
+        if (Objects.requireNonNull(entityClasses, "entityClasses").length == 0) {
+            throw new IllegalArgumentException("entityClasses must not be empty");
+        }
+
+        // Create solution descriptor
+        var solutionDescriptor = SolutionDescriptor.buildSolutionDescriptor(
+                Objects.requireNonNull(solutionClass, "solutionClass"), entityClasses);
+
+        // Create a Constraint Streams configuration with a dummy constraint provider
+        var scoreDirectorFactoryConfig = new ScoreDirectorFactoryConfig()
```
zepfred: I don't like the idea of using a dummy score provider by default. I would prefer to include the score provider as a parameter and move `DummyConstraintProvider` to a test package. This way, we can reuse the provider across all move test classes without losing cohesion.

zepfred: We could also provide a facility method that uses the dummy constraint provider by default outside of the `MoveRunner` class and maintains the current logic of the test move classes.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+            // After undo, value should be restored
+            assertThat(entity.getValue()).isEqualTo(originalValue);
+        }
+    }
+
+    @Test
+    void changeMoveTemporaryMultipleEntities() {
+        // Arrange
+        var solution = TestdataSolution.generateSolution(3, 3);
```
zepfred, line 96: Ditto.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunner.java
zepfred: After reviewing all classes, I see how the dummy score provider simplifies testing. Maybe we can rename the provider class to something like `InternalMoveRunnerConstraintProvider` and keep the current implementation.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/Moves.java
triceo: It may have looked like it, but the `Object` was causing trouble for callers. This works for callers, and hides the implementation complexity from them.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunContext.java
triceo: IMO no. I can see it as a way of capturing something in there, to be used later.

core/src/test/java/ai/timefold/solver/core/preview/api/move/MoveRunnerTest.java
triceo: I considered that, but with the understanding that this is already a `NullPointerException`, I decided it's self-explanatory. There is no point in introducing all that extra text just to tell you something which the `@NullMarked` interface told you already anyway.

core/src/main/java/ai/timefold/solver/core/preview/api/move/MoveRunContext.java
triceo: Much like the traffic sign "no parking" doesn't physically prevent people from parking there, neither does `@NullMarked` not do anything to actually stop `null` from propagating. :-)

triceo:
I have made significant changes to the PR, @Christopher-Chianelli @zepfred.

- The spec is gone. Keeping the constitution, as that has value on its own and we also use it to drive Copilot reviews.
- MoveRunner no longer requires try-with-resources. There is nothing to close, so no need for this boilerplate.
- MoveRunner is now an interface, implementation moved to impl. This allowed me to hide some stuff.
- I kept the changes to Quarkus and Spring Boot, since they are overall beneficial; we should not treat our own internal constraint providers (if any) as user-facing code.
- I kept `MoveAsserter`, as that fundamentally does more than `MoveRunner`. (Hence the "assert" in the name.) I added Javadoc to discourage people from using it, and I took some lessons from it to improve the `MoveRunner` implementation.
- I moved some tests around; many were in `testdomain`, where tests just should not be. This was not caused by the AI, but it was also not discovered during earlier reviews.

core/src/main/java/ai/timefold/solver/core/impl/move/DefaultMoveRunContext.java
```diff
@@ -0,0 +1,45 @@
...
+        Objects.requireNonNull(exceptionHandler, "exceptionHandler");
+        try {
+            scoreDirector.executeMove(Objects.requireNonNull(move, "move"));
+        } catch (Exception e) {
+            exceptionHandler.accept(e);
+        }
+    }
+
+    @Override
+    public void executeTemporarily(Move<Solution_> move, Consumer<SolutionView<Solution_>> assertions) {
```
zepfred: Let's rename it to `executeTemporaryMove`.

core/src/main/java/ai/timefold/solver/core/impl/move/DefaultMoveRunContext.java
```diff
@@ -0,0 +1,45 @@
...
+public final class DefaultMoveRunContext<Solution_> implements MoveRunContext<Solution_> {
+
+    private final AbstractScoreDirector<Solution_, ?, ?> scoreDirector;
+
+    DefaultMoveRunContext(AbstractScoreDirector<Solution_, ?, ?> scoreDirector) {
+        this.scoreDirector = Objects.requireNonNull(scoreDirector, "scoreDirector");
+    }
+
+    @Override
+    public void execute(Move<Solution_> move) {
```
zepfred, line 23: Let's rename it to `executeMove`.

core/src/main/java/ai/timefold/solver/core/impl/move/DefaultMoveRunContext.java
```diff
@@ -0,0 +1,45 @@
...
+        scoreDirector.executeMove(Objects.requireNonNull(move, "move"));
+    }
+
+    @Override
+    public void execute(Move<Solution_> move, Consumer<Exception> exceptionHandler) {
```
zepfred: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/move/MoveRunnerScoreDirector.java
```diff
@@ -0,0 +1,77 @@
...
+import ai.timefold.solver.core.api.score.constraint.ConstraintMatchTotal;
+import ai.timefold.solver.core.api.score.constraint.Indictment;
+import ai.timefold.solver.core.impl.neighborhood.MoveRepository;
+import ai.timefold.solver.core.impl.score.director.AbstractScoreDirector;
+import ai.timefold.solver.core.impl.score.director.InnerScore;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+final class MoveRunnerScoreDirector<Solution_, Score_ extends Score<Score_>>
```
zepfred, line 16: Much better!

core/src/main/java/ai/timefold/solver/core/impl/move/DefaultMoveRunContext.java
triceo: I do not see the point. We already have something called `MoveRunner`. It already gets a `Move`. Saying it a third time seems just pointless.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/Moves.java
```diff
@@ -33,9 +32,11 @@ public static <Solution_, Entity_, Value_> Move<Solution_> change(
         return new ChangeMove<>(variableMetaModel, entity, value);
     }
 
-    public static <Solution_, Entity_> Move<Solution_> swap(
-            PlanningVariableMetaModel<Solution_, Entity_, Object> variableMetaModel, Entity_ leftEntity, Entity_ rightEntity) {
-        return swap(Collections.singletonList(variableMetaModel), leftEntity, rightEntity);
+    @SuppressWarnings("unchecked")
+    public static <Solution_, Entity_, Value_> Move<Solution_> swap(
```
zepfred, line 36: As Chris pointed out, the old implementation looks better.

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,117 @@ Over time, we expect to add all commonly used move types to this collection,
...
+==== Temporary execution with automatic undo
+
+The `executeTemporarily()` method allows you to test a move's effects without permanently modifying the solution.
+This is useful when you want to verify intermediate state or test that the solver's undo mechanism works correctly:
+
+[source,java,options="nowrap"]
+----
+@Test
+void testMoveThenUndo() {
+    ... // Setup code as in previous examples.
```
zepfred: Perhaps we could include the setup code to make the test snippet clear.

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,117 @@ Over time, we expect to add all commonly used move types to this collection,
...
+        assertThat(lesson.getRoom()).isEqualTo(newRoom);
+    });
+
+    // Verify automatic undo restored original state
+    assertThat(lesson.getRoom()).isEqualTo(originalRoom);
+}
+----
+
+The callback function receives a `MutableSolutionView` parameter,
+but you typically won't need to use it—you can directly inspect the solution's state
```
zepfred:
```suggestion
but you typically won't need to use it as you can directly inspect the solution's state
```

zepfred (approved):
LGTM!

core/src/main/java/ai/timefold/solver/core/impl/move/DefaultMoveRunContext.java
zepfred: I prefer to keep the names consistent. Since we are already using `executeTemporaryMove`, my suggestion is to maintain it, as the method's goal remains the same: "execute a move and undo the changes".

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -290,6 +290,139 @@ Over time, we expect to add all commonly used move types to this collection,
...
+var runner = MoveRunner.build(solutionMetaModel);
+var context = runner.using(solution);
+var move = ...; // Move you wish to test.
+context.execute(move);
+----
+
+The API requires:
+
+1. **Solution class**: The class annotated with `@PlanningSolution`
+2. **Entity classes**: One or more classes annotated with `@PlanningEntity`
```
TomCools, line 323: What happens if i forget to add one of these? Do we throw a clear error or is it something more vague?

triceo: There will be a fail fast at some point in the code, but really this is hard to catch.
I need them to provide the entity classes, because I have no way of knowing which of them the solution uses. (I don't scan the classpath.)

Therefore I cannot throw good exceptions when some of the entity classes are missing. If I knew what they were, I wouldn't be asking for them.

core/src/test/java/ai/timefold/solver/core/testutil/MoveAsserter.java
```diff
@@ -7,12 +7,18 @@
 import ai.timefold.solver.core.impl.neighborhood.MoveRepository;
 import ai.timefold.solver.core.impl.score.director.InnerScore;
 import ai.timefold.solver.core.preview.api.move.Move;
+import ai.timefold.solver.core.preview.api.move.MoveRunner;
 
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+/**
+ * Prefer {@link MoveRunner}, unless you absolutely need score corruption checks.
```
Christopher-Chianelli: It is solution corruption checks;
In particular:

- Does the solution prior state match the same state after move + undo
- Did any variable change without notification

PR #2035, opened 2026-01-22
fix: remove a faulty fail-fast

triceo:
The fail-fast is not necessary since the latest improvements to Neighborhoods.
I added a lot of test coverage to `DefaultSolverTest` to prove that.
I also removed much of shadow var updating code, replacing it with a custom-made tool which we also recently introduced.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
@@ -98,7 +98,11 @@ final class BiRandomMoveIterator<Solution_, A, B> implements Iterator<Move<Solut
 
         while (!leftTupleSequence.isEmpty()) {
             var leftElement = leftTupleSequence.pick(workingRandom);
-            pickNextMove(leftElement);
+            var rightEmpty = pickNextMove(leftElement);
+            if (rightEmpty) {
+                leftTupleSequence.remove(leftElement.index());
+                leftElement.value().setStore(rightSequenceStoreIndex, null);
+            }
             if (nextMove != null) {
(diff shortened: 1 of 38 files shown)


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -98,7 +98,11 @@ public boolean hasNext() {
 
         while (!leftTupleSequence.isEmpty()) {
             var leftElement = leftTupleSequence.pick(workingRandom);
-            pickNextMove(leftElement);
+            var rightEmpty = pickNextMove(leftElement);
+            if (rightEmpty) {
+                leftTupleSequence.remove(leftElement.index());
+                leftElement.value().setStore(rightSequenceStoreIndex, null);
+            }
```
triceo, line 105: Not a functional change.
I just decided it's best if the sequence is modified in the same method where it's iterated, for locality.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/list/RandomSubListSelectorTest.java
```diff
@@ -82,9 +84,9 @@ void randomWithPinning() {
         var v2 = new TestdataPinnedWithIndexListValue("2");
         var v3 = new TestdataPinnedWithIndexListValue("3");
         var v4 = new TestdataPinnedWithIndexListValue("4");
-        var a = TestdataPinnedWithIndexListEntity.createWithValues("A", v1, v2, v3, v4);
+        var a = new TestdataPinnedWithIndexListEntity("A", v1, v2, v3, v4);
         a.setPlanningPinToIndex(1); // Ignore v1.
-        var b = TestdataPinnedWithIndexListEntity.createWithValues("B");
+        var b = new TestdataPinnedWithIndexListEntity("B");
```
zepfred, line 89: Should we also update the shadow variables?

PR #2041, opened 2026-01-23
chore: convert indexers to internal iteration

triceo:
This will allow complex filtering that will be necessary to make `containingAnyOf()` work.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
@@ -185,11 +185,11 @@ final class ComparisonIndexer<T, Key_ extends Comparable<Key_>>
                 var entry = comparisonMap.firstEntry();
                 if (boundaryReached(entry.getKey(), indexKey)) {
                     yield Collections.emptyIterator();
-                } else {// Boundary condition not yet reached; include the indexer in the range.
+                } else { // Boundary condition not yet reached; include the indexer in the range.
                     if (filter == null) {
                         yield entry.getValue().randomIterator(queryCompositeKey, workingRandom);
                     } else {
-                        yield entry.getValue().randomIterator(queryCompositeKey, workingRandom);
+                        yield entry.getValue().randomIterator(queryCompositeKey, workingRandom, filter);
                     }
                 }
             }
--- a/core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/EqualIndexerTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/EqualIndexerTest.java
@@ -1,11 +1,20 @@
 package ai.timefold.solver.core.impl.bavet.common.index;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Random;
 
 import ai.timefold.solver.core.api.score.stream.Joiners;
 import ai.timefold.solver.core.impl.bavet.bi.joiner.DefaultBiJoiner;
 import ai.timefold.solver.core.impl.bavet.common.tuple.UniTuple;
+import ai.timefold.solver.core.impl.neighborhood.stream.joiner.DefaultBiNeighborhoodsJoiner;
+import ai.timefold.solver.core.preview.api.neighborhood.stream.joiner.NeighborhoodsJoiners;
 
 import org.junit.jupiter.api.Test;
 
@@ -14,6 +23,10 @@ class EqualIndexerTest extends AbstractIndexerTest {
     private final DefaultBiJoiner<TestPerson, TestPerson> joiner =
             (DefaultBiJoiner<TestPerson, TestPerson>) Joiners.equal(TestPerson::gender)
                     .and(Joiners.equal(TestPerson::age));
+
+    private final DefaultBiNeighborhoodsJoiner<TestPerson, TestPerson> neighborhoodsJoiner =
+            (DefaultBiNeighborhoodsJoiner<TestPerson, TestPerson>) NeighborhoodsJoiners.equal(TestPerson::gender)
+                    .and(NeighborhoodsJoiners.equal(TestPerson::age));
 
     @Test
     void isRemovable() {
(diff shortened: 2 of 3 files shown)


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainedInIndexer.java
```diff
@@ -108,12 +109,28 @@ public Iterator<T> iterator(Object queryCompositeKey) {
         if (indexKeyCollection.isEmpty()) {
             return Collections.emptyIterator();
         }
-        return new DefaultIterator(queryCompositeKey, indexKeyCollection);
+        return new DefaultIterator(indexKeyCollection,
+                downstreamIndexer -> downstreamIndexer.iterator(queryCompositeKey));
+    }
+
+    @Override
+    public Iterator<T> randomIterator(Object queryCompositeKey, Random workingRandom) {
```
zepfred, line 107: The `randomIterator` methods, both with and without filters, contain duplicated code in this class and others. To address this, we could modify the `Indexer` contract to accept null filters. This would allow us to create a default implementation for `randomIterator` without a filter, and handle the filtering in the other method:

```
@Override
public Iterator<T> randomIterator(Object queryCompositeKey, Random workingRandom, Predicate<T> filter) {
    var indexKeyCollection = queryKeyUnpacker.apply(queryCompositeKey);
    if (indexKeyCollection.isEmpty()) {
        return Collections.emptyIterator();
    }
    if (filter == null) {
        return new DefaultIterator(indexKeyCollection,
                downstreamIndexer ->     downstreamIndexer.randomIterator(queryCompositeKey,   workingRandom));
    } else {
        return new DefaultIterator(indexKeyCollection,
                downstreamIndexer ->     downstreamIndexer.randomIterator(queryCompositeKey,   workingRandom, filter));
    }
}
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultUniqueRandomIterator.java
```diff
@@ -89,42 +91,32 @@ private int determineActiveIndex(Random workingRandom, int randomIndex) {
...
-     * @throws NoSuchElementException if the index has already been removed
-     */
-    public T remove(int index) {
-        removeWithoutReturn(index);
-        return accessor.get(index);
+    public T next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        indexToOptionallyRemove = nextIndex;
```
zepfred, line 98: The previous contract clearly defined the process for picking or removing values. It seems that the new version requires calling `next()` before removing it, is that right? Is it possible to call `next()` twice and lose one index deletion?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FilteredUniqueRandomIterator.java
```diff
@@ -0,0 +1,71 @@
...
+        this.delegate = new DefaultUniqueRandomIterator<>(source, workingRandom);
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (hasNext) {
+            return true;
+        }
+        while (delegate.hasNext()) {
+            var retrieved = delegate.next();
```
zepfred, line 41: Let's assume the following sequence:

1. Call `next()`, set `indexToOptionallyRemove` to zero and accepts the value;
2. At this point, the `remove()` was not called.
3. Call `next()` again, set `indexToOptionallyRemove` to zero and accepts the value.

The value was returned twice, which should be unique. Does it make sense?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultUniqueRandomIterator.java
```diff
@@ -7,67 +7,69 @@
...
         // If the index has already been removed, find the next closest active one.
         // If no such index is found, pick the previous closest active one.
         // This algorithm ensures that we do not pick the same index twice.
         var randomIndex = workingRandom.nextInt(leftmostIndex, rightmostIndex + 1);
-        return pickIndex(workingRandom, randomIndex);
+        nextIndex = pickIndex(workingRandom, randomIndex);
+        if (nextIndex == -1) {
+            return false;
+        }
+        next = source.get(nextIndex).element();
```
zepfred, line 62: Perhaps we could remove the selected element at this point and address my concern about potentially missing some deletions.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FilteredUniqueRandomIterator.java
triceo: Technically, you're correct. But it was designed intentionally to work like this - sometimes you want to have the option of choosing the same thing twice.

Consider three pairs:

A, B
A, C
B, C

You pick A randomly, which gives you choices B and C for the second pick.
You pick B, getting A+B.
If you removed A, there would be no way anymore to get to A+C - because A could not be randomly picked anymore.

Therefore the iterator is only "unique" if you actually call "remove" on it for everything that you pick. Maybe the name of the class should change to not say "Unique", but the behavior needs to stay.

triceo: Only after you've picked both A+B and A+C, you can remove `A` from the first iterator.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/Indexer.java
```diff
@@ -31,18 +33,6 @@
 public sealed interface Indexer<T>
         permits EqualIndexer, ComparisonIndexer, ContainingIndexer, ContainedInIndexer, ContainingAnyOfIndexer, IndexerBackend {
 
-    /**
-     * Gets the entry at the given position for the given composite key.
```
zepfred, line 35: I'm curious why this method is no longer considered necessary. Does this mean that we will always reach a specific index using an iterator? It seems inefficient if we truly need to retrieve values from specific positions, don’t you think?

triceo: We only needed `get(index)` to be able to do `get(randomIndex)`.
The random iterator completely removes that need, because it iterates in random order. We don't need `get(generateRandomNumber())` - we just call `next()` on the random iterator.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultUniqueRandomIterator.java
triceo: Answered above.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FilteredUniqueRandomIterator.java
zepfred: Does this mean we won't call next on the left iterator until the right iterator is exhausted?

triceo: Yes.

triceo: Or rather this is what happens with original selection.
For random selection, we call next() on both iterators - but we only remove from the left when the one on the right is empty. (We cache the one on the right.)

PR #2056, opened 2026-01-26
fix: make ComparableValue fallback to comparing identity hashcode when compareTo returns 0 (instead of when index are identity equal)

### LINKED ISSUE #2040: NPE in consecutive collectors
10:58:00.888 ERROR [pool-9-thread-1] Solving with id (ad39fb26-8a01-4e36-98a6-e8a43cf6e1d1) failed: message (Cannot read field "index" because "other" is null): java.lang.NullPointerException: Cannot read field "index" because "other" is null
	at ai.timefold.solver.core.impl.score.stream.collector.consecutive.ComparableValue.compareTo(ComparableValue.java:34)
	at ai.timefold.solver.core.impl.score.stream.collector.consecutive.ConsecutiveSetTree.remove(ConsecutiveSetTree.java:270)
	at ai.timefold.solver.core.impl.score.stream.collector.SequenceCalculator.retract(SequenceCalculator.java:32)
	at ai.timefold.solver.core.impl.score.stream.collector.tri.ObjectCalculatorTriCollector.lambda$accumulator$0(ObjectCalculatorTriCollector.java:28)
	at ai.timefold.solver.core.impl.bavet.common.AbstractGroupNode.retract(AbstractGroupNode.java:269)
	at ai.timefold.solver.core.impl.bavet.common.tuple.ConditionalTupleLifecycle.retract(ConditionalTupleLifecycle.java:34)
	at ai.timefold.solver.core.impl.bavet.common.StaticPropagationQueue.propagateRetracts(StaticPropagationQueue.java:83)
	at ai.timefold.solver.core.impl.bavet.NodeNetwork.settleLayer(NodeNetwork.java:62)
	at ai.timefold.solver.core.impl.bavet.NodeNetwork.settle(NodeNetwork.java:52)
	at ai.timefold.solver.core.impl.bavet.AbstractSession.settle(AbstractSession.java:63)
	at ai.timefold.solver.core.impl.score.stream.bavet.BavetConstraintSession.calculateScore(BavetConstraintSession.java:39)
	at ai.timefold.solver.core.impl.score.director.stream.BavetConstraintStreamScoreDirector.calculateScore(BavetConstraintStreamScoreDirector.java:91)
	at ai.timefold.solver.core.impl.score.director.AbstractScoreDirector.assertExpectedUndoMoveScore(AbstractScoreDirector.java:747)
	at ai.timefold.solver.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.doMove(ConstructionHeuristicDecider.java:156)
	at ai.timefold.solver.core.impl.constructionheuristic.decider.ConstructionHeuristicDecider.decideNextStep(ConstructionHeuristicDecider.java:112)
	at ai.timefold.solver.core.impl.constructionheuristic.DefaultConstructionHeuristicPhase.solve(DefaultConstructionHeuristicPhase.java:88)
	at ai.timefold.solver.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:82)
	at ai.timefold.solver.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:196)
	at ai.timefold.solver.core.impl.solver.DefaultSolverJob.call(DefaultSolverJob.java:134)

Since this is happening with the 1.30.0 solver upgrade, this is likely a regression. I strongly suspect the following commit, but I have not verified my suspicion:

https://github.com/TimefoldAI/timefold-solver/commit/0bd527f4d42192d19eda20356a122463179f3b1f

Christopher-Chianelli:
Previous, the code assumed  a.compareTo(b) == 0 if and only if
a == b; this is false; in particular, when LocalDate or other
reference type is used as the point. This cause the fallback identity hashcode
compare to not happen, and thus erases an entry in the `TreeMap`.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
@@ -1,5 +1,6 @@
 package ai.timefold.solver.core.impl.score.stream.collector.consecutive;
 
+import java.util.Objects;
 import java.util.TreeMap;
 import java.util.TreeSet;
 
@@ -30,12 +31,11 @@ record ComparableValue<Value_, Point_ extends Comparable<Point_>>(Value_ value, 
         if (this == other) {
             return 0;
         }
-        var point1 = index;
-        var point2 = other.index;
-        if (point1 == point2) {
+        var out = index.compareTo(other.index);
+        if (out == 0) {
             return compareWithIdentityHashCode(value, other.value);
         }
-        return point1.compareTo(point2);
+        return out;
     }
 
     private int compareWithIdentityHashCode(Value_ o1, Value_ o2) {
@@ -49,4 +49,16 @@ record ComparableValue<Value_, Point_ extends Comparable<Point_>>(Value_ value, 
         return Integer.compare(identityHashCode1, identityHashCode2);
     }
 
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof ComparableValue<?, ?> that)) {
+            return false;
+        }
+        return value == that.value && Objects.equals(index, that.index);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(System.identityHashCode(value), index);
+    }
 }
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
@@ -1,6 +1,5 @@
 package ai.timefold.solver.core.impl.score.stream.collector.consecutive;
 
-import java.util.Objects;
 import java.util.TreeMap;
 import java.util.TreeSet;
 
@@ -54,11 +53,11 @@ record ComparableValue<Value_, Point_ extends Comparable<Point_>>(Value_ value, 
         if (!(object instanceof ComparableValue<?, ?> that)) {
             return false;
         }
-        return value == that.value && Objects.equals(index, that.index);
+        return value == that.value && index.equals(that.index);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(System.identityHashCode(value), index);
+        return 31 * System.identityHashCode(value) + index.hashCode();
     }
 }
--- a/core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ConsecutiveSetTreeTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ConsecutiveSetTreeTest.java
@@ -185,7 +185,7 @@ class ConsecutiveSetTreeTest {
         assertThat(breakList).isEmpty();
 
         tree.remove(c);
-        assertThat(sequenceList).hasSize(0);
+        assertThat(sequenceList).isEmpty();
         assertThat(tree.getBreaks()).isEmpty();
     }
 
--- a/core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ConsecutiveSetTreeTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ConsecutiveSetTreeTest.java
@@ -8,16 +8,15 @@ import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.Comparator;
 import java.util.Iterator;
-import java.util.Map;
 import java.util.Random;
 import java.util.TreeMap;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import ai.timefold.solver.core.api.score.stream.common.Break;
 import ai.timefold.solver.core.api.score.stream.common.Sequence;
 import ai.timefold.solver.core.api.score.stream.common.SequenceChain;
+import ai.timefold.solver.core.impl.util.MutableInt;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.parallel.Execution;
@@ -29,9 +28,9 @@ class ConsecutiveSetTreeTest {
     @Test
     void testNonconsecutiveNumbers() {
         var tree = getIntegerConsecutiveSetTree();
-        var start1 = atomic(3);
-        var middle3 = atomic(5);
-        var end7 = atomic(5);
+        var start1 = mutable(3);
+        var middle3 = mutable(5);
+        var end7 = mutable(5);
 
         tree.add(start1, 1);
         tree.add(middle3, 3);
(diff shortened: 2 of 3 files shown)


triceo:
Some questions:

- You say if you changed back to HashMap, you'd see the tests fail. But the tests were not failing before. This makes me confused.
- Why has the bug not shown before? What's new now? I assumed this was a regression from my recent changes - are you saying that it's not? Matej confirmed that the only thing needed to avoid the regression is to roll back the solver to 1.29.0.
- I wonder if identity is a good approach. If those maps are supposed to hold values like `Integer` or `LocalDate`, they need to be equality-based.
- Whether we standardize on identity or on equality, aren't we breaking backwards compatibility? (For bugfixes, we will if we have to, but I just want to be sure I fully understand what's going on.)

Christopher-Chianelli:
> Some questions:
> 
>     * You say if you changed back to HashMap, you'd see the tests fail. But the tests were not failing before. This makes me confused.
> 
>     * Why has the bug not shown before? What's new now? I assumed this was a regression from my recent changes - are you saying that it's not? Matej confirmed that the only thing needed to avoid the regression is to roll back the solver to 1.29.0.

The tests before used `AtomicInteger`, which does not override `equals`. The tests now use `MutableInt`, which does override equals. The tests assumed `AtomicInteger` override `equals` before, since we explictly changed the `AtomicInteger` values.

>     * I wonder if identity is a good approach. If those maps are supposed to hold values like `Integer` or `LocalDate`, they need to be equality-based.

It stores objects like `Employee` or `Shift`, which are mapped to `int` via `toIndexFunction`.
 
>     * Whether we standardize on identity or on equality, aren't we breaking backwards compatibility? (For bugfixes, we will if we have to, but I just want to be sure I fully understand what's going on.)

We did half with identity, and half with value, which did not make any sense; `ComparableValue` uses identity of value in its `compareTo`.

```java
    @Override
    public int compareTo(ComparableValue<Value_, Point_> other) {
        if (this == other) {
            return 0;
        }
        var point1 = index;
        var point2 = other.index;
        if (point1 == point2) {
            return compareWithIdentityHashCode(value, other.value);
        }
        return point1.compareTo(point2);
    }

    private int compareWithIdentityHashCode(Value_ o1, Value_ o2) {
        if (o1 == o2) {
            return 0;
        }
        // Identity Hashcode for duplicate protection; we must always include duplicates.
        // Ex: two different games on the same time slot
        var identityHashCode1 = System.identityHashCode(o1);
        var identityHashCode2 = System.identityHashCode(o2);
        return Integer.compare(identityHashCode1, identityHashCode2);
    }
```

This does not fix the regression, but does fix a different bug when a user uses an entity inside the constraint collector and overrides equals to depend on a variable.

triceo:
I'm thinking we should just always use a `HashMap`.
The problem with entities being equal is effectively a user error that can be solved:

- By docs, which is where it IMO already is.
- At solution ingestion time, where we can actually check entities for duplicates.

The split between `Hash` and `IdentityHash` is not workable, because we very often do not know what we'll get. Just as in this case - it can be entity, it can be anything else. `HashMap` is the correct way to go here.

Christopher-Chianelli:
It seems the regression was related to https://github.com/TimefoldAI/timefold-solver/commit/0bd527f4d42192d19eda20356a122463179f3b1f given what the fix is (do identity hashcode if compareTo returns 0)

Christopher-Chianelli:
I am of the opinion the `HashMap` should be `IdentityHashMap`; if we are comparing by identity in one place, we should do it in all places.

triceo:
> I am of the opinion the `HashMap` should be `IdentityHashMap`; if we are comparing by identity in one place, we should do it in all places.

I agree that all the maps should be the same map implementation. There is no question there.

But here's why the `HashMap` is IMO the correct solution:

- We do not control what type of object we get from the user. It could be an entity, it may be `Integer` or `LocalDate`.
- Technically, entities force us to use identity; but I already explain above that it's time to move away from this. It only creates problems like this one, and serves no practical benefit. In 2.0, I'll just make it impossible, failing fast when people give two entities which are equal. (And we already say it in Javadocs and docs.)
- For everything else that's not an entity, we _must_ use `HashMap`. Why? Because `new String("a")` and `new String("a")` simply _has to_ be treated as the same thing. That is what the user expects of us; nobody caches string instances. (Or integers, or localdates, or ...)

From these three points, it follows that it must be a `HashMap`, not `IdentityHashMap`. If we're specifically talking about why we're comparing objects by identity - that is to get a unique order of things. We _only_ use identity hash code if the comparison says the two objects are equal. And we only do that so that we can have a fully deterministic algorithm to order items. The identity checks here have _nothing in common_ with whether or not the type belongs in an identity hash map; they are completely different concerns.

Technically, we can remove that check from the comparator - because we can just agree that if the two are equal, they are equal and no further checks are necessary. We simply do not operate on identity at all; which should IMO have been the case all along. Identity hash maps are supposed to be rare, and not everywhere like we do it. IMO, time to move on.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
```diff
@@ -49,4 +49,16 @@ private int compareWithIdentityHashCode(Value_ o1, Value_ o2) {
...
+    public boolean equals(Object object) {
+        if (!(object instanceof ComparableValue<?, ?> that)) {
+            return false;
+        }
+        return value == that.value && Objects.equals(index, that.index);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(System.identityHashCode(value), index);
```
triceo: Let's hand-roll the hashCode, as we do for other carrier classes. (Such as `Pair`.)

`Objects.hash` is a performance nightmare on account of creating a vararg array every time it is used; and here, it is not unlikely that the type would one day end up as a key in some hash map. (If it isn't already.)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ComparableValue.java
```diff
@@ -49,4 +49,16 @@ private int compareWithIdentityHashCode(Value_ o1, Value_ o2) {
         return Integer.compare(identityHashCode1, identityHashCode2);
     }
 
+    @Override
+    public boolean equals(Object object) {
+        if (!(object instanceof ComparableValue<?, ?> that)) {
+            return false;
+        }
+        return value == that.value && Objects.equals(index, that.index);
```
triceo: In other types, I have recently gotten rid of `Objects.equals` (by inlining it). We discussed it together too.

triceo (approved):
LGTM when comments resolved.
Also please update description of the PR so that, when I merge it, I merge something that matches what the PR does.

triceo (changes requested):
After having deeply reviewed the code and its use in examples, quickstarts and models, I came to the following conclusion:

- We need `HashMap` because we need to support value types.
- We must not switch to `IdentityHashMap` since `HashMap` is status quo, all code in the world assumes this to behave like a `HashMap`. Unless we can show that `HashMap` is a bug, it has to stay.
- The `ConsecutiveSetTreeTest` is wrong, because the way it mimics planning var change is not actually how var changes happen. They happen via remove+add, not just by changing the value directly. When that test adjustment is made, almost all tests pass with `MutableInt`. (The rest should be evaluated; are the assertions still correct under these circumstances?)
- The `ConsecutiveSetTreeTest#testNonconsecutiveNumbers` is additionally wrong, because it decided to enforce identity. If it didn't deliberately trigger a failure state, it would not fail.

Therefore I propose:
- Keeping `HashMap` in the codebase.
- Keeping `MutableInt` in the test and fixing the test coverage.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/collector/consecutive/ConsecutiveSetTreeTest.java
```diff
@@ -185,25 +237,25 @@ void testJoinOfTwoChains() {
...
         tree.add(removed4, 4);
         tree.add(breakEnd5, 5);
-        tree.add(atomic(6), 6);
-        tree.add(atomic(7), 7);
+        tree.add(mutable(6), 6);
+        tree.add(mutable(7), 7);
 
-        removed4.set(8); // mimic changing a planning variable
+        removed4.setValue(8); // mimic changing a planning variable
         tree.remove(removed4);
```
triceo: This is not correct. The outside code would trigger such a change like this:

    tree.remove(removed4);
    removed4.setValue(8);
    tree.add(removed4, 8);

Two main reasons why the original is wrong:

- It does not actually exercise the code as the solver would, making it a bad proxy for the real world and therefore not a very useful test.
- It changes the value's `hashCode` while the value is still an active key in the map. This would break any `HashMap` anywhere.

triceo (approved):

PR #2057, opened 2026-01-28
chore: improve MutableSolutionView API around list variables

triceo:
Updates Javadocs of several `MutableSolutionView` methods to better reflect the actual behavior.
Also introduces new `assignValue` options to support some corner cases.
Adds plenty of test coverage for each.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
@@ -18,6 +18,7 @@ import ai.timefold.solver.core.preview.api.domain.metamodel.ElementPosition;
 import ai.timefold.solver.core.preview.api.domain.metamodel.GenuineVariableMetaModel;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningListVariableMetaModel;
 import ai.timefold.solver.core.preview.api.domain.metamodel.PlanningVariableMetaModel;
+import ai.timefold.solver.core.preview.api.domain.metamodel.UnassignedElement;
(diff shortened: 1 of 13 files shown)


triceo:
Tagging @Christopher-Chianelli for review because I had to touch the multi-stage move.

triceo:
Taking it back to draft again, because it still doesn't work right.
List variable listeners are a mess.

core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
```diff
@@ -62,14 +66,46 @@ public final <Entity_, Value_> void assignValue(PlanningListVariableMetaModel<So
...
+        externalScoreDirector.beforeListVariableChanged(variableDescriptor, destinationEntity, destinationIndex,
+                destinationIndex + 1);
+        var actualOldValue = variableDescriptor.setElement(destinationEntity, destinationIndex, planningValue);
+        if (oldValue != actualOldValue) {
+            throw new IllegalStateException(
+                    "Impossible state: The value (%s) at index (%d) of entity (%s) is not as expected (%s)."
+                            .formatted(actualOldValue, destinationIndex, destinationEntity, oldValue));
+        }
+        externalScoreDirector.afterListVariableChanged(variableDescriptor, destinationEntity, destinationIndex,
+                destinationIndex + 1);
```
triceo, line 97: These notifications negate the performance benefit of this approach.
We only replace an element at a given place in the list, meaning nothing in the list needs to move back or forth; yet, these variable change notifications will still move things in the list. 

I wonder if it could somehow be improved...

PR #2058, opened 2026-01-29
chore: improve name for recently introduced MoveDirector API

triceo:
This way, it is consistent with what java.util.List does.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
+++ b/core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
@@ -59,7 +59,7 @@ public interface MutableSolutionView<Solution_> extends SolutionView<Solution_> 
             Value_ value, Entity_ destinationEntity, int destinationIndex);
 
     /**
-     * Puts a given value at a particular index in a given entity's {@link PlanningListVariable planning list variable}.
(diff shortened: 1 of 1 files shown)


core/src/main/java/ai/timefold/solver/core/preview/api/move/MutableSolutionView.java
```diff
@@ -53,11 +55,12 @@ default <Entity_, Value_> void assignValue(PlanningListVariableMetaModel<Solutio
      *        moving the pre-existing value at that index and all subsequent values to the right.
      * @throws IllegalStateException if the value is already assigned to a list variable
      */
-    <Entity_, Value_> void assignValueAndInsert(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
+    <Entity_, Value_> void assignValueAndAdd(PlanningListVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel,
             Value_ value, Entity_ destinationEntity, int destinationIndex);
 
     /**
      * Puts a given value at a particular index in a given entity's {@link PlanningListVariable planning list variable}.
```
triceo:
```suggestion
     * Puts a given value at a particular index in a given entity's {@link PlanningListVariable planning list variable},
```

PR #2067, opened 2026-01-30
chore: upgrade to spring 4.x

zepfred:
This pull request upgrades Spring to version 4. The new spring version now defaults to Jackson 3, while Quarkus 3.x continues to use Jackson 2.

The proposed updates upgrade the current Jackson module to version 3 and move the Jackson 2 logic to the module `quarkus-jackson`. As a result, the solver codebase and the Spring module now operate with Jackson 3, while the Quarkus integration remains on Jackson 2.

Once Quarkus starts using Jackson 3, we will remove the Jackson 2 logic and use the Jackson 3 module instead.


triceo:
On reflection:

- The "jackson3" module should just be called "jackson". When we remove "jackson2", it will look significantly better.
- From the outset, the "jackson2" module should be marked as deprecated in the module description. It should be made clear that this module would be removed the first chance we get, and that users should prefer to use the "jackson" module unless they need to use Quarkus 3.

persistence/jackson2/pom.xml
```diff
@@ -11,13 +11,13 @@
 
   <artifactId>timefold-solver-jackson</artifactId>
```
triceo:
```suggestion
  <artifactId>timefold-solver-jackson2</artifactId>
```

persistence/jackson2/pom.xml
```diff
@@ -11,13 +11,13 @@
...
 
-  <name>Timefold Solver Jackson</name>
+  <name>Timefold Solver Jackson 2</name>
   <description>
     Timefold solves planning problems.
     This lightweight, embeddable planning engine implements powerful and scalable algorithms
     to optimize business resource scheduling and planning.
 
-    This module contains the Jackson integration.
+    This module contains the Jackson 2 integration.
```
triceo: Needs to explicitly state that this module is deprecated and will be removed during the lifetime of Timefold Solver 2.x. People should be using `timefold-solver-jackson` instead.

persistence/jackson3/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>ai.timefold.solver</groupId>
+    <artifactId>timefold-solver-persistence-parent</artifactId>
+    <version>999-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>timefold-solver-jackson3</artifactId>
```
triceo:
```suggestion
  <artifactId>timefold-solver-jackson</artifactId>
```

build/build-parent/pom.xml
```diff
@@ -102,6 +104,20 @@
...
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>tools.jackson</groupId>
+        <artifactId>jackson-bom</artifactId>
+        <version>${version.tools.jackson}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
```
triceo: I question if these imports should be here.
Maybe we only import them in the `jackson2` and `jackson` modules respectively? IMO we're asking for dependency conflicts like this.

spring-integration/spring-boot-integration-test/pom.xml
```diff
@@ -151,6 +139,12 @@
                 <!-- Ob == fast build -->
                 <buildArg>-Ob</buildArg>
                 <buildArg>--no-fallback -H:+ReportExceptionStackTraces</buildArg>
+                <!--
+                    JUnit 6.0.x introduced an issue for native compilation
+                    affecting GraalVM 21. Newer versions of GraalVM are not impacted by this issue,
+                    and parameters may be removed once a fix is merged and published.
+                 -->
+                <buildArg>--initialize-at-build-time=org.junit.jupiter.engine.discovery.MethodSegmentResolver</buildArg>
```
triceo: Do not do this.
By the time this is merged, the branch will no longer contain GraalVM 21.

spring-integration/pom.xml
```diff
@@ -42,6 +42,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <!-- Solution to make Maven work when the BOM has not been published -->
+      <dependency>
+        <groupId>ai.timefold.solver</groupId>
+        <artifactId>timefold-solver-jackson3</artifactId>
+        <version>999-SNAPSHOT</version>
+        <scope>import</scope>
+      </dependency>
```
triceo: This is highly unusual. What exactly is broken here?

build/build-parent/pom.xml
zepfred: They have changed the `groupId`, so I don't see how this import could cause conflicts. All tests are passing without major changes to the pom definitions. It could be done, but I'm unsure whether it is really necessary.

spring-integration/spring-boot-integration-test/pom.xml
zepfred: Well, I've been blocked by a native compilation issue I was trying to address. Good to know we are upgrading the GrallVM.

spring-integration/pom.xml
zepfred: I found the issue, and this workaround is no longer necessary.

build/build-parent/pom.xml
triceo: We do not know what they'll be importing exactly. Sure, they will hopefully only keep it to Jackson, but there is no proof. I'd rather not expose us to that.

triceo: Actually, do we need those imports at all?
How many Jackson JARs do we actually depend on? And if it's the 3 that I expect, why not just list them manually?

zepfred: After the refactoring, the POM import was moved into the quarkus module.

zepfred:
@triceo I've confirmed that native compilation works with GraalVM 25 for both Quarkus and Spring.

quarkus-integration/quarkus-jackson/runtime/src/main/java/ai/timefold/solver/quarkus/jackson/api/domain/solution/AbstractConstraintWeightOverridesDeserializer.java
```diff
@@ -0,0 +1,43 @@
+package ai.timefold.solver.quarkus.jackson.api.domain.solution;
```
triceo: In Quarkus Jackson, the API package must not exist.
This is not API, this will eventually be removed.

quarkus-integration/quarkus-jackson/runtime/src/main/java/ai/timefold/solver/quarkus/jackson/score/stream/common/DeserializableSequenceChain.java
```diff
@@ -0,0 +1,57 @@
...
+
+record DeserializableSequenceChain<Value_, Difference_ extends Comparable<Difference_>>(
+        List<DeserializableSequence<Value_, Difference_>> sequences)
+        implements
+            SequenceChain<Value_, Difference_> {
+
+    @Override
+    public @NonNull Collection<Sequence<Value_, Difference_>> getConsecutiveSequences() {
+        return Collections.unmodifiableCollection(sequences());
+    }
```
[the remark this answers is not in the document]
zepfred: `DeserializableSequence` implements `Sequence` and no cast is necessary.

spring-integration/spring-boot-integration-test/pom.xml
```diff
@@ -2,163 +2,139 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
```
triceo, line 2: Why are we reformatting this file (and only this file)?

spring-integration/spring-boot-integration-test/src/main/resources/application.properties
```diff
@@ -1 +1 @@
-timefold.solver.termination.best-score-limit=0
\ No newline at end of file
+timefold.solver.termination.best-score-limit=0
```
triceo: Unnecessary change.

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverBeanFactory.java
```diff
@@ -211,11 +210,11 @@ ConstraintVerifier<ConstraintProvider_, SolutionClass_> constraintVerifier() {
 
     // @Bean wrapped by static class to avoid classloading issues if dependencies are absent
     @Configuration(proxyBeanMethods = false)
-    @ConditionalOnClass({ Jackson2ObjectMapperBuilder.class, Score.class })
+    @ConditionalOnClass({ JsonMapper.class, Score.class })
     static class TimefoldJacksonConfiguration {
 
         @Bean
-        Module jacksonModule() {
+        tools.jackson.databind.JacksonModule jacksonModule() {
```
triceo: Why the FQN?

quarkus-integration/quarkus-jackson/pom.xml
```diff
@@ -15,6 +15,18 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.fasterxml.jackson</groupId>
+        <artifactId>jackson-bom</artifactId>
+        <version>${version.com.fasterxml.jackson}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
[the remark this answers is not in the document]
triceo: This makes sense. Maybe we don't need this import, since the Jackson version used here comes straight from Quarkus, which we already do import?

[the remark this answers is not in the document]
triceo: Which probably also means we do not need the jackson deps in dep management in the build parent. Let's try to do the minimal thing, whatever it is.

triceo (approved):
LGTM when comments resolved.

quarkus-integration/quarkus-jackson/runtime/src/main/java/ai/timefold/solver/quarkus/jackson/solution/JacksonSolutionFileIO.java
```diff
@@ -0,0 +1,68 @@
...
+    public JacksonSolutionFileIO(Class<Solution_> clazz) {
+        this(clazz, new ObjectMapper());
+    }
+
+    public JacksonSolutionFileIO(Class<Solution_> clazz, ObjectMapper mapper) {
+        this.clazz = clazz;
+        this.mapper = mapper;
+        // Loads TimefoldJacksonModule via ServiceLoader, as well as any other Jackson modules on the classpath.
+        mapper.findAndRegisterModules();
+    }
```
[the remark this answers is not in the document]
zepfred: Jackson 3 addresses this issue, and we won't change the current behavior of Jackson 2 logic.

quarkus-integration/quarkus-jackson/pom.xml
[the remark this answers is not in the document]
zepfred: Yes, the Quarkus import comes from the `build-parent`.

PR #2068, opened 2026-01-30
chore: standardize on Spring Boot 3.5.x

zepfred:


triceo (approved):
Before we merge this though, @zepfred please check the docs if we say anywhere which versions of Spring we support. With this, it should only be the latest 3.x LTS.

(We also need to change start.spring.io when we do the next 1.x release.)

PR #2069, opened 2026-02-02
chore: fail-fast validation for invalid value ranges

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -816,6 +833,78 @@ that could cause the scoreDifference (%s).""".formatted(scoreDifference, beforeM
...
+        var basicVariableDescriptorList = entityDescriptor.getGenuineBasicVariableDescriptorList();
+        var listVariableDescriptor = entityDescriptor.getGenuineListVariableDescriptor();
+        assertValueRangeForBasicVariables(this, lookUpManager, basicVariableDescriptorList, entity);
+        assertValueRangeForListVariable(this, lookUpManager, listVariableDescriptor, entity);
+    }
+
+    private static <Solution_> void assertValueRangeForBasicVariables(InnerScoreDirector<Solution_, ?> scoreDirector,
+            LookUpManager lookUpManager, List<GenuineVariableDescriptor<Solution_>> basicVariableDescriptorList,
+            Object entity) {
+        if (basicVariableDescriptorList == null || basicVariableDescriptorList.isEmpty() || lookUpManager == null) {
```
triceo: Why is it important that lookup is enabled?
Isn't lookup for multi-threaded only? (Not 100 % sure here; but this is suspicious.)

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -816,6 +833,78 @@ that could cause the scoreDifference (%s).""".formatted(scoreDifference, beforeM
...
+        }
+        for (var variableDescriptor : basicVariableDescriptorList) {
+            var value = variableDescriptor.getValue(entity);
+            if (value == null || !lookUpManager.isLookUpEnabled(value)) {
+                continue;
+            }
+            var valueRange = scoreDirector.getValueRangeManager()
+                    .getFromEntity(variableDescriptor.getValueRangeDescriptor(), entity);
+            // We use the lookup search instead of rebasing in order to control the expected error message
+            var rebasedValue = scoreDirector.lookUpWorkingObjectOrReturnNull(value);
```
triceo: So we really do this because of multi-threaded?

Why not just skip this entire thing in multi-threaded? The user does not control `setWorkingSolution()` in multi-threaded, therefore we can assume stuff is correct. IMO there is no need to bring lookups into this.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -816,6 +833,78 @@ that could cause the scoreDifference (%s).""".formatted(scoreDifference, beforeM
...
+            if (rebasedValue == null) {
+                rebasedValue = value;
+            }
+            if (!valueRange.contains(rebasedValue)) {
+                throw new IllegalStateException(
+                        "The value (%s) from the planning variable (%s) has been assigned to the entity (%s), but it is outside of the related value range %s."
+                                .formatted(value, variableDescriptor.getVariableName(), entity,
+                                        StreamSupport.stream(Spliterators.spliterator(valueRange.createOriginalIterator(),
+                                                valueRange.getSize(), Spliterator.SIZED), false)
+                                                .toList()));
```
triceo: You want to print the entire value range? That's not wise. They will be big.
How about just using the value range ref?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -1030,6 +1120,12 @@ public Builder_ withExpectShadowVariablesInCorrectState(boolean expectShadowVari
             return (Builder_) this;
         }
 
+        @SuppressWarnings("unchecked")
+        public Builder_ withEnvironmentMode(EnvironmentMode environmentMode) {
+            this.environmentMode = environmentMode;
+            return (Builder_) this;
+        }
```
triceo: We didn't have this information before? Surely we did, score director does assertions regularly. Why does this need to be introduced now?

core/src/test/java/ai/timefold/solver/core/testdomain/clone/customcloner/TestdataCorrectlyClonedSolution.java
```diff
@@ -24,11 +24,13 @@ public class TestdataCorrectlyClonedSolution implements SolutionCloner<TestdataC
...
     private TestdataEntity entity = new TestdataEntity("A");
 
+    private static final List<TestdataValue> staticList = List.of(new TestdataValue("1"), new TestdataValue("2"));
+
     @ValueRangeProvider(id = "valueRange")
     @ProblemFactCollectionProperty
     public List<TestdataValue> valueRange() {
         // two values needed to allow for at least one doable move, otherwise the second step ends in an infinite loop
-        return Arrays.asList(new TestdataValue("1"), new TestdataValue("2"));
+        return new ArrayList<>(staticList);
```
triceo, line 33: Why the change?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -554,6 +568,9 @@ public void afterListVariableChanged(ListVariableDescriptor<Solution_> variableD
             int toIndex) {
         variableListenerSupport.afterListVariableChanged(variableDescriptor, entity, fromIndex, toIndex);
         neighborhoodsElementUpdateNotifier.accept(entity);
+        if (isStepAssertOrMore) {
+            assertValueRangeForEntity(entity);
```
triceo: Wouldn't it make sense to only check the changed indexes?
No need to iterate over the entire list.

triceo (commented):
Much better! IMO still room for improvement.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: ~~For some reason, I had to execute the lookup in my previous approach, which seems unnecessary now that the tests are passing.~~

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: ~~No need for a lookup anymore.~~

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: Yep. The value range has its own `toString()` implementation.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: The tests are failing, and I need to understand why.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: I couldn't find any place in score director where we could extract that information about the environment.

core/src/test/java/ai/timefold/solver/core/testdomain/clone/customcloner/TestdataCorrectlyClonedSolution.java
zepfred: Test failing because every time it calls value range, different values are returned.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: Yes, that makes sense. Let me improve it.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: The tests failed due to this new fail-fast check, and a different check is needed when using chained variables.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: The score director does not store any assertion flag. That's why I chose not to implement it inside the score director in my first approach. So, if we need the environment, we need to pass it to the score director.

triceo (approved):

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -119,6 +119,7 @@ public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> ge
                 .withLookUpEnabled(true)
                 .withConstraintMatchPolicy(
                         constraintMatchEnabled ? ConstraintMatchPolicy.ENABLED : ConstraintMatchPolicy.DISABLED)
+                .withEnvironmentMode(environmentMode)
```
triceo: I just realized something.

There are _many_ places which build score directors. _None_ of them call this on their builders. Which means the environment mode is effectively almost never set correctly.

I can see a way of fixing this:

- The score director factory is typically created from the solver config. Therefore it can store the environment mode there.
- The score director builder is always given the factory as an argument, therefore it can read the environment mode and store it as the default value.
- The builder method then only serves as override.

zepfred: The environment mode is available during factory creation, except in two places: `DefaultMoveRunner` and `ShadowVariableUpdateHelper`. For these special cases, I'm setting the default environment mode to `PHASE_ASSERT`.

Also, I don't see a need for an environment override feature, as we are not using it anywhere.

triceo:
Having just merged my other PR, this one now has conflicts. :-(
Also, I see some more Copilot comments.

PR #2070, opened 2026-02-03
ci: reduce permissions of GitHub tokens

triceo:


triceo:
Github is complaining that some workflows may have more permissions that they actually need.
@jennedebleser Would you mind reviewing? I'm not exactly sure this is what will actually work; we will only find out once we merge this.

.github/workflows/codeql.yml
triceo, line 1: Removed since we already have enough checks in Sonar and Aikido.

jennedebleser (approved):

Issue #1710, opened 2025-07-26
In Quarkus, setters are not used if they are present


triceo:
When using DSV in Quarkus, the updater returns a value and that value is directly stored into the field. It does not go through the setter, even though the setter exists. This is a bug, because the setter may have some additional logic which the updater will effectively ignore.

(This can be reproduced by putting a breakpoint in the updater, and then a breakpoint in the setter. The updater breakpoint will trigger, the setter breakpoint will not.)

In the past few months we've seen several problems with member accessors in Quarkus, so let's agree on and implement safe behavior, which will also fix this bug:

- If a getter exists, a setter must exist.
- If a setter exists, a getter must exist.
- If a getter/setter pair exists, any access to the field will go through the pair. (Not just in DSV.)

Christopher-Chianelli:
Fixed by https://github.com/TimefoldAI/timefold-solver/pull/2072

PR #2072, opened 2026-02-04
fix: Use field getters/setters when available, enforce getters/setters are public

### LINKED ISSUE #1862: No longer rely on deep reflection
In the near future, Java will start restricting deep reflection - for example, `setAccessible(true)` will become a problem.

To prepare for this future, the solver needs to do two things.

## Require field access via getters/setters

If a class field being accessed (variable, shadow variable, fact/entity collection etc.) is not public, the solver has to access it via a getter and setter.

The field can still be annotated; read-only reflection should still be OK. But regardless of what is annotated, the access needs to happen through the getter/setter.

As this is a backwards incompatible behavior, we can only make this change in the next major version of the solver.

## Redesign cloners

The current approach to solution cloning relies heavily on deep reflection, and it can not be fixed. A new approach to cloning will be necessary. Possibilities include:

- `@JsonCreator`-like approach.
- `PlanningClonable` interface.
- Wait until Java finishes serialization 2.0, which will likely provide mechanisms for this. Unfortunately, as this would require an upgrade of the minimum version of the JDK, this is unlikely to be applicable any time soon; a solution to this issue will likely be required earlier.

Christopher-Chianelli:
Before, this is how domain accessors were created:

- Annotation on the field? Make that field accessible and read/write directly to it.
- Annotation on a getter method? Make that method accessible, and pair it with the corresponding setter (if available) and read/write using the method pair.

Now, domain accessors are created like this:

- Annotation on the field? If the field has a getter, enforce that getter is public. If there is a setter but no getter, fail-fast. Use the field getter/setter for the MemberAccessor; only directly read/write the field if the field is public and there are no getters or setters for the field. Fail-fast otherwise.
- Annotation on a getter method? Enforce the getter is public, and pair it with the corresponding setter (if available, fail if there a setter and not public), and read/write using the method pair.

Quarkus generated MemberAccessors will also follow these same rules and use getters/setters if available.

This is a breaking change. Do not merge until 2.0 is branched.

Fixes #1862 and #1710.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -32,7 +32,7 @@
     <dependency>
       <groupId>io.quarkus.gizmo</groupId>
       <artifactId>gizmo2</artifactId>
-      <optional>true</optional>
+      <!--      <optional>true</optional>-->
     </dependency>
 
     <!-- External dependencies -->
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/common/ReflectionHelper.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/common/ReflectionHelper.java
@@ -22,6 +22,7 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -276,6 +277,10 @@ public final class ReflectionHelper {
     }
 
     public static void assertGetterMethod(Method getterMethod) {
+        if (!Modifier.isPublic(getterMethod.getModifiers())) {
+            throw new IllegalStateException("The getterMethod (%s) must be public."
+                    .formatted(getterMethod));
+        }
         if (getterMethod.getParameterTypes().length != 0) {
             throw new IllegalStateException("The getterMethod (%s) must not have any parameters (%s)."
                     .formatted(getterMethod, Arrays.toString(getterMethod.getParameterTypes())));
@@ -304,8 +309,12 @@ public final class ReflectionHelper {
     }
 
     public static void assertGetterMethod(Method getterMethod, Class<? extends Annotation> annotationClass) {
+        if (!Modifier.isPublic(getterMethod.getModifiers())) {
+            throw new IllegalStateException("The getterMethod (%s) with a @%s annotation must be public."
+                    .formatted(getterMethod, annotationClass.getSimpleName()));
+        }
         if (getterMethod.getParameterTypes().length != 0) {
-            throw new IllegalStateException("The getterMethod (%s) with a %s annotation must not have any parameters (%s)."
+            throw new IllegalStateException("The getterMethod (%s) with a @%s annotation must not have any parameters (%s)."
                     .formatted(getterMethod, annotationClass.getSimpleName(),
                             Arrays.toString(getterMethod.getParameterTypes())));
         }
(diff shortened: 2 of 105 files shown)


triceo (commented):
I like this!

We need to figure out what to do with Gizmo. Personally, I'd just remove Gizmo from `core`, I don't see much point. It's not being used, and it complicates the codebase.

That said, I'm open to keeping Gizmo in core, if we can make it the default without changing any of the rules we just defined. In that case, it would only have benefits and therefore reflection-based access could be removed entirely.

Side note: there seem to be failing tests.

Christopher-Chianelli:
Gizmo in core was modified to use getters/setters for fields if available.

triceo:
Wrt. Gizmo - as I said above, unless we're willing to make Gizmo the default and the only approach, IMO we should not have it in the core at all. Either it is safe and performant, and therefore a good choice for everyone, or it's not and therefore it is an unwelcome presence in the codebase. What are the downsides of Gizmo which make it impractical to be the default?

Also, I question if Gizmo is still as fast as it used to be. The current "reflection" isn't really reflection, it's unreflected method handles. It no longer performs access checks, which was much of the slow part of reflection. What exactly does Gizmo bring, which the current approach does not have? IMO this should be re-benchmarked; the assumption of Gizmo giving significant perf improvements may no longer be true, too many circumstances have changed.

Wrt. Quarkus - in the past, I've been bitten by Quarkus and Core's handling of accessors being subtly different. This is inevitable, since both are defined in different places, and implemented by different code. I wonder if we can do something to bring them closer together, so that the key decisions are only made in one place.

Christopher-Chianelli:
> Wrt. Gizmo - as I said above, unless we're willing to make Gizmo the default and the only approach, IMO we should not have it in the core at all. Either it is safe and performant, and therefore a good choice for everyone, or it's not and therefore it is an unwelcome presence in the codebase. What are the downsides of Gizmo which make it impractical to be the default?

In the past, the primary downside was that Gizmo uses ASM, which was a dependency that cause conflicts if another library uses it but with a different version. Gizmo2 no longers uses ASM (check all dependencies in https://production.portal.central.sonatype.com/artifact/io.quarkus.gizmo/gizmo2/2.1.0/dependencies), so that downside is eliminated.

A secondary downside that we enforced on ourself was that cloning cannot use public getters/setters, since those getters/setters might not be simple. But if we were to eliminate `setAccessible` calls, reflection too must use public getters/setters to do a solution clone.

Since domain accessors are all (both member accessors and cloning) or nothing (niether member accessors or cloning), the cloning limitations prevents using member accessors. The Gizmo member accessors does not really have downsides. 

> Also, I question if Gizmo is still as fast as it used to be. The current "reflection" isn't really reflection, it's unreflected method handles. It no longer performs access checks, which was much of the slow part of reflection. What exactly does Gizmo bring, which the current approach does not have? IMO this should be re-benchmarked; the assumption of Gizmo giving significant perf improvements may no longer be true, too many circumstances have changed.

When tested, only static final MethodHandle get converted to a direct call: https://timefold.ai/blog/java-reflection-but-much-faster.

Doing a quick test using the vehicle routing quickstart, GIZMO is still 10% faster than REFLECTION.
 
> Wrt. Quarkus - in the past, I've been bitten by Quarkus and Core's handling of accessors being subtly different. This is inevitable, since both are defined in different places, and implemented by different code. I wonder if we can do something to bring them closer together, so that the key decisions are only made in one place.

FYI both REFLECTION and GIZMO did not use getters/setters if the annotation was on the field prior.

triceo (approved):

Issue #2042, opened 2026-01-23
feat: Use RandomGenerator instead of Random


Christopher-Chianelli:
**Is your feature request related to a problem? Please describe.**
Currently, the solver directly references the `Random` class in many places such as Move Selectors, Acceptors, and SolverScope. This make it difficult to use the new `RandomGenerator` implementations.

**Describe the solution you'd like**
Refactor the solver to use `RandomGenerator` wherever `Random` was used.

PR #2074, opened 2026-02-04
feat: Change usage of Random to RandomGenerator

Christopher-Chianelli:
Note: there is a tempoarily workaround since
Collections.shuffle(List, RandomGenerator) is not available until JDK 21.

This is a backward-incompatible change since it also affects custom move iterators.
Do not merge until Timefold Solver 2.0 is branched.


Christopher-Chianelli:
Fixes https://github.com/TimefoldAI/timefold-solver/issues/2042

triceo (approved):

PR #2076, opened 2026-02-05
chore: Neighborhoods improvements for ListChangeMove

triceo:


core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/Moves.java
```diff
@@ -27,18 +79,70 @@ public static <Solution_> Move<Solution_> compose(Move<Solution_>... moves) {
...
     public static <Solution_, Entity_, Value_> Move<Solution_> swap(
             PlanningVariableMetaModel<Solution_, Entity_, Value_> variableMetaModel, Entity_ leftEntity, Entity_ rightEntity) {
         return swap(List.of((PlanningVariableMetaModel<Solution_, Entity_, Object>) variableMetaModel), leftEntity,
                 rightEntity);
     }
 
+    /**
+     * Creates a move that swaps the values of multiple planning variables between two entities.
+     * <p>
+     * All variables in the list must belong to the same entity class.
```
Christopher-Chianelli: So we cannot use interfaces (with multiple implementations) for values?

i.e. `interface Visit`, `class CatVisit implements Visit`, `class DogVisit implements Visit`?

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListChangeMoveTest.java
```diff
@@ -2,142 +2,514 @@
...
+                    (ListChangeMove<TestdataListSolution, TestdataListEntity, TestdataListValue>) Moves.change(
+                            variableMetaModel,
+                            entity, 0, entity, 1);
+
+            // getPlanningValues requires the move to be executed first to know the value
+            MoveRunner.build(solutionMetaModel)
+                    .using(solution)
+                    .execute(changeMove);
+
+            assertThat(changeMove.getPlanningValues()).containsExactly(value1);
```
Christopher-Chianelli, line 350: Should `value2` be included, since it was shifted from position 1 to position 0?
(Note: the moves `Moves.change(metamodel, entity, 0, entity, 1)` and `Moves.change(metamodel, entity, 1, entity, 0)` are equivalent but would return different planning values?

triceo: I think that's a question of semantics. This is a change move, its goal is to change a single value's position.

We can make a decision that this method will return any planning value which had any of its shadow variables affected by the move. But is that more useful? (Think of tabu search, which is what this method is for. I think that the current behavior is better, and also status quo - this has been the way solver's been doing this move since forever.)

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/Moves.java
triceo: I don't see the relation - maybe I phrased it wrong?

The idea here is that all variables being swapped (variables, not their values) are declared on the same entity. Now that I think of it, it's pretty much guaranteed and impossible to do any other way, so maybe this text just shouldn't be there at all.

Christopher-Chianelli: "All variables in the list must belong to the same entity class" is quite similar to "All variables in the list must belong to the same class", and thus make it seem like the values must all be instances of the same `@PlanningEntity` class (no subclasses allowed). I would remove it, causes more confusion since we only allow one `@PlanningListVariable` currently anyway.

core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
```diff
@@ -176,20 +176,45 @@ public final <Entity_, Value_> Value_ moveValueInList(
...
 
         var variableDescriptor = extractVariableDescriptor(variableMetaModel);
+        var list = variableDescriptor.getValue(sourceEntity);
+        var listSize = list.size();
+        if (sourceIndex >= listSize) {
+            throw new IllegalArgumentException(
+                    "The sourceIndex (%d) must be less than the list size (%d).".formatted(sourceIndex, listSize));
+        } else if (destinationIndex > listSize) { // destinationIndex == listSize is allowed (append to the end of the list)
+            throw new IllegalArgumentException(
+                    "The destinationIndex (%d) must be less than or equal to the list size (%d)."
```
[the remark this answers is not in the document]
triceo: This method handles both shifting before destination and after destination, therefore this is actually a valid case and cannot be guarded against.

PR #2077, opened 2026-02-05
chore: prevent the generation of invalid moves

zepfred:


core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
```diff
@@ -449,11 +449,11 @@ void randomWithEntityValueRangeAndFiltering() {
         // IgnoreBValueSelectionFilter is applied to the value selector used by the destination selector,
         // and that causes the B destination to become an invalid destination
         assertCodesOfNeverEndingMoveSelector(moveSelector,
-                "1 {A[1]->A[1]}",
+                "1 {A[1]->A[2]}",
+                "3 {B[0]->A[1]}",
+                "3 {B[0]->A[1]}",
```
triceo: Same move twice?

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
```diff
@@ -613,11 +613,11 @@ void randomAllowsUnassignedValuesWithEntityValueRange() {
         phaseStarted(solverScope, moveSelector);
 
         assertCodesOfNeverEndingMoveSelector(moveSelector,
-                "1 {A[1]->A[1]}",
-                "3 {B[0]->B[1]}",
-                "3 {B[0]->B[1]}",
+                "1 {A[1]->A[0]}",
                 "3 {B[0]->B[1]}",
-                "3 {B[0]->B[1]}");
+                "1 {A[1]->A[1]}",
```
triceo, line 616: A no-op? We're trying to avoid those.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java
```diff
@@ -39,12 +41,23 @@ public ElementPositionRandomIterator(ListVariableStateSupply<Solution_, Object,
...
+        // but the destination entity iterator still points to the previous outdated value.
+        if (maybeOutdated && isEntityRange && entityIterator.hasNext()) {
+            // Discard the current stale value
+            entityIterator.next();
+        }
+
         // The valueSelector's hasNext() is insignificant.
         // The next random destination exists if and only if there is a next entity.
+        maybeOutdated = true;
         return entityIterator.hasNext();
```
triceo: Since we already asked for `next()`, shouldn't `maybeOutdated` be `false` now?

This logic has gradually become impossible to understand.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
zepfred: It is not an error. Because we are now calling `entityIterator.next()`, the order of operations has changed, causing the move to be applied twice in this specific test.

triceo:
We also no longer have confidence (at least I don't) that these selector produce _every possible move_. (= No valid move is skipped.) And with equal probabilities.

core/src/test/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ListChangeMoveSelectorTest.java
zepfred: `ElementPositionRandomIterator` can generate these no-op moves because it increments the index of the generated position.

```
// +1 to include the destination after the final element in the list.
return ElementPosition.of(positionInList.entity(), positionInList.index() + 1);
```

Let's consider an entity `e[b, a]`. The value iterator picks `a` and the element iterator picks `b`. The result is the position of `b + 1`.

PR #2079, opened 2026-02-05
chore: enable the list variable to use an anonymous entity provider

zepfred:


triceo (commented):
If I provide IDs to the ranges (make them not anonymous) does this still work? Tom also reported that as a problem.

PR #2089, opened 2026-02-09
chore: start work on Timefold Solver 2.0.0

triceo:


docs/src/modules/ROOT/nav.adoc
```diff
@@ -33,7 +33,6 @@
 * Upgrading Timefold Solver
 ** xref:upgrading-timefold-solver/overview.adoc[leveloffset=+1]
 ** xref:upgrading-timefold-solver/upgrade-to-latest-version.adoc[leveloffset=+1]
-** xref:upgrading-timefold-solver/upgrade-from-optaplanner.adoc[leveloffset=+1]
```
Christopher-Chianelli, line 36: Do we want to remove this section? Although upgrading from OptaPlanner to Timefold 2.0 is not as straightforward as upgrading to Timefold 1.x or 0.x, people might still search for it.

Christopher-Chianelli (approved):
I am unsure about removing references to OptaPlanner; the rest LGTM.

docs/src/modules/ROOT/nav.adoc
triceo: The 1.x docs will still be around (we still have the 0.8 docs even), which IMO should be enough. We'll adjust if we see it's not.

PR #2091, opened 2026-02-10
refactor: remove deprecated chained variable support and clean up related files

triceo:
Chained variable is removed.
I went through all code looking for instances of the word "chain"; only unrelated mentions are left.

I removed the bare minimum of tests.

I accidentally pushed directly into `enterprise`, so these changes are already there on `main`.
(Technically this should not be allowed, but since I am repo owner, Github let me do that. This was totally not intentional, and I apologize.)


core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/pillar/SubPillarConfigPolicy.java
```diff
@@ -72,8 +72,8 @@ private void validateSizes() {
                     + ") must be at least 1.");
         }
         if (minimumSubPillarSize > maximumSubPillarSize) {
-            throw new IllegalStateException("The minimumPillarSize (" + minimumSubPillarSize
-                    + ") must be at least maximumSubChainSize (" + maximumSubPillarSize + ").");
+            throw new IllegalStateException("The minimumPillarSize (%d) must be <= maximumPillarSize (%d)."
```
zepfred: Let's use `format` for the previous message as well.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/RootVariableSource.java
```diff
@@ -156,7 +154,7 @@ public static <Entity_, Value_> RootVariableSource<Entity_, Value_> from(
                                 "The source path (%s) starting from root entity (%s) referencing multiple facts (%s, %s) in a row."
                                         .formatted(variablePath, rootEntityClass.getSimpleName(),
                                                 chainToVariable.get(chainToVariable.size() - 2).getName(),
-                                                chainToVariable.get(chainToVariable.size() - 1).getName()));
+                                                chainToVariable.getLast().getName()));
```
zepfred, line 157: I understand the name `chainToVariable` is unrelated to chained variables, but I wonder whether we should use a different name.

triceo: We no longer need to. :-)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/inverserelation/InverseRelationShadowVariableDescriptor.java
```diff
@@ -95,35 +91,26 @@ private void linkShadowSources(DescriptorPolicy descriptorPolicy) {
...
-                throw new IllegalArgumentException("The entityClass (" + entityDescriptor.getEntityClass()
-                        + ") has an @" + InverseRelationShadowVariable.class.getSimpleName()
-                        + " annotated property (" + variableMemberAccessor.getName()
-                        + ") which does not return a " + Collection.class.getSimpleName()
-                        + " with sourceVariableName (" + sourceVariableName
-                        + ") which is neither a list variable @" + PlanningListVariable.class.getSimpleName()
-                        + " nor a chained variable @" + PlanningVariable.class.getSimpleName()
-                        + "(graphType=" + PlanningVariableGraphType.CHAINED + ")."
-                        + " Only list and chained variables support a singleton inverse.");
+            if (!list) {
```
zepfred, line 95: Let's use `format` for the previous messages.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -318,40 +315,19 @@ private ValueSelector<Solution_> buildBaseValueSelector(GenuineVariableDescripto
...
     protected ValueSelector<Solution_> applyFiltering(ValueSelector<Solution_> valueSelector,
             ClassInstanceCache instanceCache) {
-        var variableDescriptor = valueSelector.getVariableDescriptor();
-        if (hasFiltering(variableDescriptor)) {
+        if (config.getFilterClass() != null) {
             List<SelectionFilter<Solution_, Object>> filterList = new ArrayList<>(config.getFilterClass() == null ? 1 : 2);
             if (config.getFilterClass() != null) {
                 filterList.add(instanceCache.newInstance(config, "filterClass", config.getFilterClass()));
             }
             // Filter out pinned entities
```
zepfred:
```suggestion

```

core/src/main/resources/solver.xsd
```diff
@@ -111,7 +111,7 @@
                     
           <xs:element minOccurs="0" name="constraintStreamAutomaticNodeSharing" type="xs:boolean"/>
                     
-          <xs:element minOccurs="0" name="constraintStreamProfilingMode" type="tns:constraintProfilingMode"/>
+          <xs:element minOccurs="0" name="constraintStreamProfilingEnabled" type="xs:boolean"/>
```
zepfred, line 114: Why are we changing this element?

core/src/main/resources/solver.xsd
```diff
@@ -1759,20 +1621,6 @@
       
   </xs:simpleType>
     
-  <xs:simpleType name="constraintProfilingMode">
```
zepfred, line 1762: Ditto.

core/src/main/resources/solver.xsd
triceo: The XSD was wrong. For some reason, it was never updated. A bug in the XSD plugin?
Anyway... same fix needed for 1.31.0.

triceo: It looks like the plugin has a regression. It does not update the XSD file when it is already there. Which means our CI needs to delete it before running the plugin.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/dummy/MultipleConstraintProviderSpringTestConfiguration.java
```diff
@@ -1,11 +0,0 @@
-package ai.timefold.solver.spring.boot.autoconfigure.dummy;
-
-import org.springframework.boot.autoconfigure.AutoConfigurationPackage;
-import org.springframework.context.annotation.Configuration;
-
-@Configuration
-@AutoConfigurationPackage(basePackages = { "ai.timefold.solver.spring.boot.autoconfigure.normal.domain",
```
zepfred, line 7: It seems we are also using this class to test basic variables. Why are we deleting it instead of just removing the chained package?

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/dummy/MultipleEasyScoreConstraintSpringTestConfiguration.java
```diff
@@ -1,11 +0,0 @@
-package ai.timefold.solver.spring.boot.autoconfigure.dummy;
-
-import org.springframework.boot.autoconfigure.AutoConfigurationPackage;
-import org.springframework.context.annotation.Configuration;
-
-@Configuration
-@AutoConfigurationPackage(basePackages = { "ai.timefold.solver.spring.boot.autoconfigure.normal.domain",
```
zepfred, line 7: Ditto.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/dummy/MultipleIncrementalScoreConstraintSpringTestConfiguration.java
```diff
@@ -1,11 +0,0 @@
-package ai.timefold.solver.spring.boot.autoconfigure.dummy;
-
-import org.springframework.boot.autoconfigure.AutoConfigurationPackage;
-import org.springframework.context.annotation.Configuration;
-
-@Configuration
-@AutoConfigurationPackage(basePackages = { "ai.timefold.solver.spring.boot.autoconfigure.normal.domain",
```
zepfred, line 7: ditto.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/dummy/MultipleSolutionsSpringTestConfiguration.java
```diff
@@ -1,10 +0,0 @@
-package ai.timefold.solver.spring.boot.autoconfigure.dummy;
-
-import org.springframework.boot.autoconfigure.domain.EntityScan;
-import org.springframework.context.annotation.Configuration;
-
-@Configuration
-@EntityScan(basePackages = { "ai.timefold.solver.spring.boot.autoconfigure.normal.domain",
```
zepfred, line 7: ditto.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverMultipleSolverAutoConfigurationTest.java
```diff
@@ -478,34 +445,6 @@ void noSolutionClass() {
                 .cause().message().contains("No classes were found with a @PlanningSolution annotation.");
     }
 
-    @Test
```
zepfred, line 481: I believe we should keep this test and the related Spring configuration classes. Perhaps we can modify them to utilize list variables instead of the chained classes.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverMultipleSolverAutoConfigurationTest.java
```diff
@@ -527,144 +466,6 @@ void noConstraintClass() {
                         "No classes found that implement EasyScoreCalculator, ConstraintProvider, or IncrementalScoreCalculator.");
     }
 
-    @Test
```
zepfred, line 530: I'm concerned about removing all these tests. If they use chained variables, let's have them use list variables instead.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverUnspecifiedAutoConfigurationTest.java
```diff
@@ -45,17 +35,6 @@ void noSolutionClass() {
                 .cause().message().contains("No classes were found with a @PlanningSolution annotation.");
     }
 
-    @Test
```
zepfred, line 48: ditto.

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverUnspecifiedAutoConfigurationTest.java
```diff
@@ -75,82 +54,6 @@ void noConstraintClass() {
                         "No classes found that implement EasyScoreCalculator, ConstraintProvider, or IncrementalScoreCalculator.");
     }
 
-    @Test
```
zepfred, line 78: ditto.

zepfred (changes requested):
I'm concerned about removing certain tests from the Spring integration. I believe these tests should be retained. Please double-check if we can replace any use of chained variables with list variables instead of removing the related tests.

docs/src/modules/ROOT/nav.adoc
```diff
@@ -34,7 +34,4 @@
 ** xref:upgrading-timefold-solver/overview.adoc[leveloffset=+1]
 ** xref:upgrading-timefold-solver/upgrade-to-latest-version.adoc[leveloffset=+1]
 ** xref:upgrading-timefold-solver/backwards-compatibility.adoc[leveloffset=+1]
-** Migration Guides
-*** xref:upgrading-timefold-solver/migration-guides/variable-listeners-to-custom-shadow-variables.adoc[leveloffset=+1]
-*** xref:upgrading-timefold-solver/migration-guides/chained-variables-to-planning-list-variable.adoc[leveloffset=+1]
-* xref:enterprise-edition/enterprise-edition.adoc[leveloffset=+1]
\ No newline at end of file
+* xref:enterprise-edition/enterprise-edition.adoc[leveloffset=+1]
```
zepfred, line 37:
```suggestion
** xref:enterprise-edition/enterprise-edition.adoc[leveloffset=+1]
```

spring-integration/spring-boot-autoconfigure/src/test/java/ai/timefold/solver/spring/boot/autoconfigure/dummy/MultipleConstraintProviderSpringTestConfiguration.java
```diff
@@ -6,6 +6,6 @@
 @Configuration
 @AutoConfigurationPackage(basePackages = { "ai.timefold.solver.spring.boot.autoconfigure.normal.domain",
         "ai.timefold.solver.spring.boot.autoconfigure.normal.constraints",
```
zepfred: I wonder if we should use basic instead of normal in the package names. What do you think?

docs/src/modules/ROOT/nav.adoc
triceo: No, this is intentionally one level up.

benchmark/src/test/resources/logback-test.xml
```diff
@@ -3,7 +3,7 @@
 
   <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <!-- %L lowers performance, %C and %c break indentation and therefore reduce readability, normal %t is verbose -->
+      <!-- %L lowers performance, %C and %c break indentation and therefore reduce readability, basic %t is verbose -->
```
triceo:
```suggestion
      <!-- %L lowers performance, %C and %c break indentation and therefore reduce readability, normal %t is verbose -->
```

core/src/main/java/ai/timefold/solver/core/api/score/buildin/bendablebigdecimal/BendableBigDecimalScore.java
```diff
@@ -310,7 +310,7 @@ public BendableBigDecimalScore power(double exponent) {
         var actualExponent = BigDecimal.valueOf(exponent);
         // The (unspecified) scale/precision of the exponent should have no impact on the returned scale/precision
         // TODO FIXME remove .intValue() so non-integer exponents produce correct results
-        // None of the normal Java libraries support BigDecimal.pow(BigDecimal)
+        // None of the basic Java libraries support BigDecimal.pow(BigDecimal)
```
triceo:
```suggestion
        // None of the normal Java libraries support BigDecimal.pow(BigDecimal)
```

core/src/main/java/ai/timefold/solver/core/api/score/buildin/hardmediumsoft/HardMediumSoftScore.java
```diff
@@ -147,7 +147,7 @@ public int getHardScore() {
      * Their weight is included in the total.
      * The medium score is usually a negative number because most use cases only have negative constraints.
      * <p>
-     * In a normal score comparison, the medium score is irrelevant if the 2 scores don't have the same hard score.
+     * In a basic score comparison, the medium score is irrelevant if the 2 scores don't have the same hard score.
```
triceo:
```suggestion
     * In a normal score comparison, the medium score is irrelevant if the 2 scores don't have the same hard score.
```

core/src/main/java/ai/timefold/solver/core/impl/heuristic/move/Move.java
```diff
@@ -71,7 +71,7 @@ public interface Move<Solution_> extends ai.timefold.solver.core.preview.api.mov
      * <li>Either it's simply not possible to do (for example due to built-in hard constraints).</li>
      * </ul>
      * <p>
-     * It is recommended to keep this method implementation simple: do not use it in an attempt to satisfy normal
+     * It is recommended to keep this method implementation simple: do not use it in an attempt to satisfy basic
```
triceo:
```suggestion
     * It is recommended to keep this method implementation simple: do not use it in an attempt to satisfy normal
```

PR #2095, opened 2026-02-11
chore: move modules around

triceo:
This only moves modules around.
It does not make any meaningful changes to implementation, or to tests.


docs/TODO.md
```diff
@@ -0,0 +1,11 @@
+# Changes to mention in release notes for 2.0.0
+
+- [ ] Chained var removal.
+- [ ] Spring 4 upgrade + changes to Jackson.
+- [ ] Random -> RandomGenerator
+- [ ] New rules for getters/setters.
+- [ ] `webui` module removal.
+- [ ] `benchmark-aggregator` split into separate module.
+- [ ] `benchmark` no longer has the method to open in a browser.
+
```
Christopher-Chianelli, line 10: Shouldn't JSONB modules being removed also be mentioned?

tools/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/statistic/subsingle/pickedmovetypestepscore/PickedMoveTypeStepScoreDiffStatisticPoint.java
```diff
@@ -11,7 +10,7 @@ public class PickedMoveTypeStepScoreDiffStatisticPoint extends StatisticPoint {
     private final long timeMillisSpent;
     /**
      * Not a {@link Class}{@code <}{@link Move}{@code >} because {@link CompositeMove}s need to be atomized
-     * and because that {@link Class} might no longer exist when {@link BenchmarkAggregator} aggregates.
+     * and because that {@link Class} might no longer exist when benchmark aggregator runs.
```
Christopher-Chianelli: I would use `the` or `a` here:

```suggestion
     * and because that {@link Class} might no longer exist when the benchmark aggregator runs.
```

tools/benchmark-aggregator/pom.xml
```diff
@@ -5,25 +5,27 @@
...
-  <name>(Deprecated) Timefold Solver JSON-B</name>
+  <name>Timefold Solver Benchmark Aggregator</name>
   <description>
     Timefold solves planning problems.
     This lightweight, embeddable planning engine implements powerful and scalable algorithms
     to optimize business resource scheduling and planning.
 
-    This module contains the JSON-B integration.
-    This module is deprecated and will be removed in a future major version of Timefold Solver.
+    This module contains the benchmarker aggregator,
```
Christopher-Chianelli: Inconsistent naming; in `PickedMoveTypeStepScoreDiffStatisticPoint`, you call this the "benchmark aggregator"; here it is "benchmarker aggregator"; pick one.

PR #2102, opened 2026-02-11
fix: make XXE protection test locale-resilient

Dolverin:
## Summary
- The `GenericJaxbIOTest.readOverridingNamespaceIsProtectedFromXXE` test asserts on the English error message `"DOCTYPE is disallowed"`, which fails on non-English JVM locales (e.g. `de_DE` produces `"DOCTYPE ist nicht zulässig"`)
- Changed the assertion to match `"disallow-doctype-decl"` instead — the XML parser feature name, which is locale-independent and always present in the `SAXParseException` message

## Test plan
- [x] Verified test passes on `de_DE` locale (previously failed)
- [x] Verified test passes on `en_US` locale (still works)
- [x] Tested with JDK 21 (Temurin 21.0.10)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


triceo (approved):
Thank you for the contribution, @Dolverin!
LGTM once the `CLAUDE.md` file is removed.

PR #2104, opened 2026-02-12
fix: copy solver XSD file

zepfred:


zepfred:
> LGTM assuming that either Copilot is wrong, or you fix it.

The comment was correct, and I applied the suggestion.

PR #2106, opened 2026-02-12
refactor: remove deprecations around score and constraints

triceo:


core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintStream.java
```diff
@@ -1617,728 +1417,4 @@ <C> TriConstraintStream<A, B, C> concat(@NonNull TriConstraintStream<A, B, C> ot
...
-     * @return never null
-     * @deprecated Prefer {@link #impactBigDecimal(Score, BiFunction)} and {@link ConstraintWeightOverrides}.
-     */
-    @Deprecated(forRemoval = true)
-    default Constraint impactConfigurableBigDecimal(String constraintPackage, String constraintName,
-            BiFunction<A, B, BigDecimal> matchWeigher) {
-        return impactConfigurableBigDecimal(matchWeigher)
-                .asConstraint(constraintPackage, constraintName);
-    }
-
```
zepfred, line 2343: Cool!

core/src/main/java/ai/timefold/solver/core/api/score/ScoreExplanation.java
```diff
@@ -64,19 +62,16 @@ public interface ScoreExplanation<Solution_, Score_ extends Score<Score_>> {
...
     String getSummary();
 
     /**
      * Explains the {@link Score} of {@link #getScore()} by splitting it up per {@link Constraint}.
      * <p>
      * The sum of {@link ConstraintMatchTotal#getScore()} equals {@link #getScore()}.
      *
-     * @return the key is the constraintId
-     *         (to create one, use {@link ConstraintRef#composeConstraintId(String, String)}).
+     * @return the key is the constraint name
```
zepfred, line 72: The comment seems incomplete.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/BendableScoreContext.java
```diff
@@ -79,7 +80,8 @@ public BendableScore toScore() {
     }
 
     @NullMarked
-    private record SingleHardImpact(BendableScoreContext ctx, int impact) implements ScoreImpact<BendableScore> {
+    private record SingleHardImpact(BendableScoreContext ctx,
```
zepfred: Let's format it.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/BendableScoreContext.java
```diff
@@ -93,8 +95,8 @@ public BendableScore toScore() {
     }
 
     @NullMarked
-    private record ComplexImpact(BendableScoreContext ctx, int[] hardImpacts,
-            int[] softImpacts) implements ScoreImpact<BendableScore> {
+    private record ComplexImpact(BendableScoreContext ctx, long[] hardImpacts,
```
zepfred: Ditto

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/HardMediumSoftScoreContext.java
```diff
@@ -50,7 +51,7 @@ public ScoreImpact<HardMediumSoftScore> changeScoreBy(int matchWeight,
 
     @NullMarked
     private record SoftImpact(HardMediumSoftScoreInliner inliner,
-            int softImpact) implements ScoreImpact<HardMediumSoftScore> {
+            long softImpact) implements ScoreImpact<HardMediumSoftScore> {
```
zepfred: I noticed some strange formatting, but it was like that before.

triceo:
> I've noticed that we are removing many migration recipes without including any new ones. This means that users will need to update everything manually. Is that the intended plan?

The recipes need the final migrated state to compile; which is why they cannot exist for any of the things that I removed. (Which is pretty much all the things we had migration recipes for.)

I still need to write new recipes for many of the things I just did. It's on my list. But overall, the process of migration from 1.x will have to start by running all the old recipes while still on 1.x.

core/src/main/java/ai/timefold/solver/core/api/score/ScoreExplanation.java
triceo: It is complete. The value is obvious from the name of the type, and from the name of the method.

zepfred (approved):

PR #2109, opened 2026-02-12
ci: configure Dependabot for the branch `1.x`

zepfred:


.github/dependabot.yml
```diff
@@ -33,7 +33,39 @@ updates:
   - package-ecosystem: "github-actions"
```
triceo, line 33: Why the changes here?
Github Actions on `main` should not be reconfigured.

.github/dependabot.yml
```diff
@@ -33,7 +33,39 @@ updates:
...
+      regular:
+        update-types:
+          - "patch"
+        patterns:
+          - "*"
+  - package-ecosystem: "github-actions"
+    target-branch: "1.x"
+    directory: "/"
+    schedule:
+      interval: monthly
```
triceo: Doesn't this also need the cron setting?

.github/dependabot.yml
zepfred: I've updated the wrong ecosystem.

PR #2111, opened 2026-02-13
refactor: value range cleanup

triceo:
All value ranges are countable.
The solver now implicitly trusts entity equals().


core/src/main/java/ai/timefold/solver/core/api/domain/valuerange/ValueRange.java
```diff
@@ -1,43 +1,27 @@
...
+ * use {@link ValueRangeFactory} to get an instance of a {@link ValueRange}.
+ * 
  * @see ValueRangeFactory
  */
 @NullMarked
 public interface ValueRange<T> {
 
     /**
-     * In a {@link CountableValueRange}, this must be consistent with {@link CountableValueRange#getSize()}.
+     * This must be consistent with {@link #getSize()}.
```
zepfred, line 24: We can implement a default method here to reduce implementation details for child classes.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/AbstractValueRange.java
```diff
@@ -0,0 +1,30 @@
...
+ * Abstract superclass for {@link ValueRange}.
+ *
+ * @see ValueRange
+ * @see ValueRangeFactory
+ */
+@NullMarked
+public abstract class AbstractValueRange<T> implements ValueRange<T>, SortableValueRange<T> {
+
+    @Override
+    public boolean isEmpty() {
```
zepfred, line 22: I just noticed this now. My earlier comment related to this issue can be resolved.

core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/AbstractValueRange.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Abstract superclass for {@link ValueRange}.
+ *
+ * @see ValueRange
+ * @see ValueRangeFactory
+ */
+@NullMarked
+public abstract class AbstractValueRange<T> implements ValueRange<T>, SortableValueRange<T> {
```
zepfred: All ranges appear to be grouped under the same package. Let's seal this class.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/composite/CompositeMoveSelector.java
```diff
@@ -34,14 +30,7 @@ protected CompositeMoveSelector(List<MoveSelector<Solution_>> childMoveSelectorL
...
                                 + ") has neverEnding (" + childMoveSelector.isNeverEnding()
-                                + ") with randomSelection (" + randomSelection + ")."
-                                + (childMoveSelector.isCountable() ? ""
-                                        : "\nThe selector is not countable, check the "
-                                                + ValueRange.class.getSimpleName() + "s involved.\n"
-                                                + "Verify that a @" + ValueRangeProvider.class.getSimpleName()
-                                                + " does not return " + ValueRange.class.getSimpleName()
-                                                + " when it can return " + CountableValueRange.class.getSimpleName()
-                                                + " or " + Collection.class.getSimpleName() + "."));
+                                + ") with randomSelection (" + randomSelection + ").");
```
zepfred: Let's use `format`.

PR #2121, opened 2026-02-13
refactor: clean up around problem changes

triceo:


zepfred (approved):
LGTM!

Christopher-Chianelli (approved):
LGTM after Copilot suggestions are applied (in particular, if there are multiple `ProblemChanges`, a later `ProblemChange` may have to deal with a working solution with stale shadow variables since the adapter (that was removed) used to call `updateShadowVariables`). We probably should add a test for this case.

PR #2122, opened 2026-02-13
refactor: clean up around solver config

triceo:


core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/entity/EntitySelectorConfig.java
```diff
@@ -432,7 +358,7 @@ public static <Solution_, T> SelectionSorter<Solution_, T> determineSorter(
         return switch (entitySorterManner) {
             case NONE:
                 throw new IllegalStateException("Impossible state: hasSorter() should have returned null.");
-            case DECREASING_DIFFICULTY, DECREASING_DIFFICULTY_IF_AVAILABLE, DESCENDING, DESCENDING_IF_AVAILABLE:
+            case DESCENDING, DESCENDING_IF_AVAILABLE:
```
Christopher-Chianelli: I would convert this old switch (`case VALUE:`) to the new switch (`case VALUE -> {}`) for consistency.

core/src/main/java/ai/timefold/solver/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
```diff
@@ -288,21 +208,11 @@ public void setStepCountingHillClimbingType(@Nullable StepCountingHillClimbingTy
         return this;
     }
 
-    public @NonNull LocalSearchAcceptorConfig withValueTabuRatio(@NonNull Double valueTabuRatio) {
```
Christopher-Chianelli, line 291: The field + get/set methods were deprecated, but the `with` method never was?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/SubListChangeMoveSelectorFactory.java
```diff
@@ -155,19 +155,19 @@ private SubListChangeMoveSelectorConfig buildChildMoveSelectorConfig(ListVariabl
...
 
-        subListSelectorConfig = Objects.requireNonNull(subListChangeMoveSelectorConfig.getSubListSelectorConfig());
+        var effectiveSubListSelectorConfig = Objects.requireNonNull(subListChangeMoveSelectorConfig.getSubListSelectorConfig());
         SubListConfigUtil.transferDeprecatedMinimumSubListSize(
                 subListChangeMoveSelectorConfig,
-                SubListChangeMoveSelectorConfig::getMinimumSubListSize,
+                s -> effectiveSubListSelectorConfig.getMinimumSubListSize(),
                 "subListSelector",
                 subListSelectorConfig);
         SubListConfigUtil.transferDeprecatedMaximumSubListSize(
```
Christopher-Chianelli: Should this method be renamed, since the deprecated method was removed?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -214,69 +213,27 @@ protected boolean determineBaseRandomSelection(SelectionCacheType resolvedCacheT
...
+        var comparatorFactoryClass = config.getComparatorFactoryClass();
         if (sorterManner != null) {
             if (!ValueSelectorConfig.hasSorter(sorterManner, variableDescriptor)) {
                 return null;
             }
             sorter = ValueSelectorConfig.determineSorter(sorterManner, variableDescriptor);
         } else if (comparatorClass != null) {
             Comparator<Object> sorterComparator =
-                    instanceCache.newInstance(config, determineComparatorPropertyName(config), comparatorClass);
+                    instanceCache.newInstance(config, "comparatorClass", comparatorClass);
```
Christopher-Chianelli, line 232: I wonder if we should introduce a constant `COMPARATOR_CLASS_NAME = "comparatorClass"` somewhere, since this name is used as a String constant in a few places.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
```diff
@@ -214,69 +213,27 @@ protected boolean determineBaseRandomSelection(SelectionCacheType resolvedCacheT
...
             sorter = new ComparatorSelectionSorter<>(sorterComparator,
                     SelectionSorterOrder.resolve(config.getSorterOrder()));
         } else if (comparatorFactoryClass != null) {
-            var comparatorFactory = instanceCache.newInstance(config, determineComparatorFactoryPropertyName(config),
+            var comparatorFactory = instanceCache.newInstance(config, "comparatorFactoryClass",
```
Christopher-Chianelli, line 236: Ditto but for `COMPARATOR_FACTORY_CLASS_NAME = "comparatorFactoryClass"`

core/src/main/java/ai/timefold/solver/core/config/localsearch/decider/acceptor/LocalSearchAcceptorConfig.java
triceo: I guess so. These things are difficult to keep consistent; I'm surprised this happened only once.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/value/ValueSelectorFactory.java
triceo: Since it ultimately is the name of a method, it doesn't help much.
But I've done some other things to reduce the duplication around this.

Issue #2110, opened 2026-02-13
Make Gizmo the default member accessor


triceo:
- Detect when we can enable Gizmo by trying a super-quick bytecode manipulation.
- If possible, enable Gizmo; otherwise keep Reflection.
- Log to `TRACE` what the effective method is.
- Remove any means of the user choosing the method for themselves.

The code around member accessors and how they are built is also quite complex, on account of Quarkus choosing a completely different code path. The factories are a bit of a mess. Bonus points if it could be simplified.

Christopher-Chianelli:
Done by #2123

PR #2123, opened 2026-02-13
chore: use Gizmo MemberAccessors when possible, remove domainAccessType from SolverConfig

Christopher-Chianelli:
Note: SolutionCloner will still use REFLECTION by default, since GIZMO cannot copy final fields and requires all fields to be accessible (i.e. either public or have public getters and setters).

Partially completes https://github.com/TimefoldAI/timefold-solver/issues/2110 (Did not remove `AutoDiscoverMemberType` as that is unrelated to GIZMO).


core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
```diff
@@ -36,18 +36,20 @@ final class GizmoFieldHandler implements GizmoMemberHandler {
                         .formatted(fieldDescriptor.name(), declaringClass.getName()));
             }
             if (!ignoreChecks && !isFieldPublic) {
-                throw new IllegalStateException("""
-                        Member (%s) of class (%s) is not public."""
-                        .formatted(fieldDescriptor.name(), declaringClass.getName()));
+                throw new IllegalArgumentException(
+                        """
```
triceo: Let's add some vertical density an put `"""` on the previous line.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoFieldHandler.java
```diff
@@ -36,18 +36,20 @@ final class GizmoFieldHandler implements GizmoMemberHandler {
...
             getterDescriptor = null;
             setterDescriptor = null;
             this.canBeWritten = canBeWritten;
         } else {
             if (!ignoreChecks && !Modifier.isPublic(getterMethod.getModifiers())) {
-                throw new IllegalStateException("""
-                        Member (%s) of class (%s) is not public."""
-                        .formatted(getterMethod.getName(), getterMethod.getDeclaringClass().getName()));
+                throw new IllegalArgumentException(
+                        """
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactory.java
```diff
@@ -35,18 +41,56 @@ public static String getGeneratedClassName(Member member) {
...
-                    Thread.currentThread().getContextClassLoader());
-        } catch (ClassNotFoundException e) {
-            throw new IllegalStateException("""
-                    When using the domainAccessType (%s) the classpath or modulepath must contain io.quarkus.gizmo:gizmo2.
-                    Maybe add a dependency to io.quarkus.gizmo:gizmo2.""".formatted(DomainAccessType.GIZMO));
-        }
         return GizmoMemberAccessorImplementor.createAccessorFor(member, annotationClass, accessorInfo, gizmoClassLoader);
     }
 
+    public static boolean isGizmoSupported(GizmoClassLoader gizmoClassLoader) {
```
triceo: IMO this check should happen within `GizmoClassloader` the first time the support status is asked. How that happens is an internal detail, and IMO outside classes shouldn't be making that decision and then _telling_ the classloader about it. Simpler that way.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactory.java
```diff
@@ -35,18 +41,56 @@ public static String getGeneratedClassName(Member member) {
...
+            case SUPPORTED -> true;
+            case UNSUPPORTED -> false;
+            case UNKNOWN -> {
+                var classPackage = GizmoMemberAccessorFactory.class.getPackage().getName();
+                var bytecodeHolder = new AtomicReference<byte[]>();
+                var gizmo = Gizmo.create((className, bytecode) -> {
+                    bytecodeHolder.set(bytecode);
+                });
+
+                var classDesc = ClassDesc.of("%s.Test".formatted(classPackage));
```
triceo: I'd rather avoid the word "test" - that already means something. Maybe "proof"?

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorFactory.java
```diff
@@ -35,18 +41,56 @@ public static String getGeneratedClassName(Member member) {
...
+                        // Should be impossible, but a native image might decide to optimize out
+                        // instance if it is unused
+                        gizmoClassLoader.setGizmoSupportStatus(GizmoSupportStatus.UNSUPPORTED);
+                        yield false;
+                    } else {
+                        gizmoClassLoader.setGizmoSupportStatus(GizmoSupportStatus.SUPPORTED);
+                        yield true;
+                    }
+                } catch (IllegalAccessException | NoSuchMethodException | InstantiationException
+                        | InvocationTargetException | Error e) {
```
triceo: Since we cannot guarantee what will be thrown, and we already catch an `Error`, I think this is one of the few cases when catching a `Throwable` is warranted. Let's be defensive a bit - _whatever_ happens here, we cannot enable Gizmo.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorImplementor.java
```diff
@@ -135,7 +136,7 @@ static MemberAccessor createAccessorFor(Member member, Class<? extends Annotatio
         }
         final MutableReference<byte @NonNull []> classBytecodeHolder = new MutableReference<>(null);
```
triceo: We generally don't use `final` for fields.
Also, inconsistent `var` use throughout the class.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoMemberAccessorImplementor.java
```diff
@@ -235,77 +236,95 @@ private static void createGetDeclaringClass(GeneratedClassInfo generatedClassInf
...
-                        Maybe rename the method (get%s)?"""
-                        .formatted(methodName, method.returnType(), methodName.substring(2)));
+
+        var memberAccessorType = accessorInfo.memberAccessorType();
+
+        var methodType = (methodName.startsWith("get") || methodName.startsWith("is")) ? "getterMethod" : "readMethod";
+        if (memberAccessorType == MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD ||
+                memberAccessorType == MemberAccessorFactory.MemberAccessorType.FIELD_OR_GETTER_METHOD_WITH_SETTER) {
+            // Only enforce getters/setters naming rules for getters/setters
+            // we don't want to enforce them on @ShadowSources suppliers
```
triceo: But do we enforce that they are public?
I am not too happy that this validation logic is clearly still in two different places. That can only lead to inconsistencies between reflection and Gizmo.

Can we do something to unify this? If we cannot unify this in the codebase, then we should follow the same thing we do for cloners - have the same test run on both configurations, guaranteeing they behave identically.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoSupportStatus.java
```diff
@@ -0,0 +1,7 @@
+package ai.timefold.solver.core.impl.domain.common.accessor.gizmo;
+
+public enum GizmoSupportStatus {
```
triceo, line 3: When the comment above is applied, this can possibly be hidden or removed entirely. At the very least, the `UNKNOWN` status can be.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/MemberAccessorFactory.java
```diff
@@ -164,10 +174,16 @@ public MemberAccessorFactory() {
...
-    public MemberAccessorFactory(Map<String, MemberAccessor> memberAccessorMap) {
+    public MemberAccessorFactory(@Nullable Map<String, MemberAccessor> memberAccessorMap) {
         // The MemberAccessorFactory may be accessed, and this cache both read and updated, by multiple threads.
         this.memberAccessorCache =
                 memberAccessorMap == null ? new ConcurrentHashMap<>() : new ConcurrentHashMap<>(memberAccessorMap);
+        // If the memberAccessorMap is not empty, we are in Quarkus using pregenerated member accessors
+        this.isGizmoSupported =
+                (memberAccessorMap != null && !memberAccessorMap.isEmpty()) || GizmoMemberAccessorFactory.isGizmoSupported(
+                        gizmoClassLoader);
+        LOGGER.trace("Using domain access type {} for member accessors",
```
triceo:
```suggestion
        LOGGER.trace("Using domain access type {} for member accessors.",
```

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/MemberAccessorFactory.java
```diff
@@ -164,10 +174,16 @@ public MemberAccessorFactory() {
      *
      * @param memberAccessorMap key is the fully qualified member name
      */
-    public MemberAccessorFactory(Map<String, MemberAccessor> memberAccessorMap) {
+    public MemberAccessorFactory(@Nullable Map<String, MemberAccessor> memberAccessorMap) {
         // The MemberAccessorFactory may be accessed, and this cache both read and updated, by multiple threads.
         this.memberAccessorCache =
                 memberAccessorMap == null ? new ConcurrentHashMap<>() : new ConcurrentHashMap<>(memberAccessorMap);
+        // If the memberAccessorMap is not empty, we are in Quarkus using pregenerated member accessors
```
triceo, line 184: Arguably this comment should be in the `@param` of the Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/ReflectionBeanPropertyMemberAccessor.java
```diff
@@ -73,7 +73,7 @@ public ReflectionBeanPropertyMemberAccessor(Method getterMethod, AnnotatedElemen
                                         declaringClass.getCanonicalName()));
             }
             try {
-                setterMethod.setAccessible(true);
+                this.setterMethod.setAccessible(true);
```
triceo, line 53: There are two cases here:

1. REFLECTION for cloning.
2. REFLECTION for anything else.

IMO 2 should not attempt `setAccessible`.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -725,17 +725,11 @@ private void initSolutionCloner(DescriptorPolicy descriptorPolicy) {
...
-                    solutionCloner = new FieldAccessingSolutionCloner<>(this);
-                    break;
-                default:
-                    throw new IllegalStateException("The domainAccessType (" + domainAccessType
-                            + ") is not implemented.");
-            }
+            solutionCloner = switch (descriptorPolicy.getDomainAccessType()) {
+                case GIZMO -> GizmoSolutionClonerFactory.build(this, memberAccessorFactory.getGizmoClassLoader());
+                // AUTO means we are probably in plain Java, so we need to use reflection so we can clone final fields
+                case AUTO, REFLECTION -> new FieldAccessingSolutionCloner<>(this);
```
triceo: See, `AUTO` is confusing... _probably_ in plain Java. Why not use the effective domain access type here, the one we already confirmed as opposed to the assumed one?

quarkus-integration/quarkus/deployment/src/main/java/ai/timefold/solver/quarkus/deployment/TimefoldProcessor.java
```diff
@@ -981,155 +969,157 @@ private GeneratedGizmoClasses generateDomainAccessors(Map<String, SolverConfig>
...
-                                        Maybe you included a parameter which is not a planning solution (%s)?
-                                        Maybe you misspelled the supplierName name?"""
-                                        .formatted(ShadowVariable.class.getSimpleName(), memberName, targetMethodName,
-                                                classInfo.name().toString(), solutionClassInfo.name().toString()));
-                    }
-                    buildMethodAccessor(annotatedMember, generatedMemberAccessorsClassNameSet, entityEnhancer, classOutput,
-                            classInfo, methodInfo, AccessorInfo.of(true, !methodInfo.parameterTypes().isEmpty()), transformers);
+                if (methodInfo == null) {
+                    throw new IllegalArgumentException(
+                            """
```
triceo: Same comment on vertical density.

triceo (commented):
IMO we have more work to do. Other than the notes inline, I think we should also:

- Add a note on the configuration disappearing to the Markdown TODO file in `docs/`, so that I have a full list of things to mention in upgrade recipe later.
- Check the documentation for the domain access type; I find it strange that this setting would never be mentioned, yet we're not updating the docs.
- Consider that maybe we only enable GIZMO for Enterprise? (Not sure if good idea, but it crossed my mind as I was reviewing the PR.)

Also, Sonar.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
Christopher-Chianelli: Cloner requirements are too strict to enable GIZMO outside of Quarkus (since it pretty much bans having any final fields). Do note DomainAccessType is now internal. `GIZMO` = Use Gizmo for everything, `REFLECTION` = Use reflection for everything, `AUTO` = Use the best thing possible (so GIZMO except for native image for MemberAccessors, and always REFLECTION for Cloner). The only think that actually use `GIZMO`/`REFLECTION` is Quarkus (Quarkus uses `REFLECTION` temporarily due to build time shenanigans/restrictions).

triceo: Useful context. In general, this sort of information would look great in the code itself - either as Javadoc of the enum, or in some other way. Essentially - what will these do, under what circumstances.

triceo (approved):
I resolved the Sonar comments myself, in order to not block other PRs.

LGTM.

PR #2124, opened 2026-02-14
refactor: remove remaining deprecation from public API

triceo:


core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningId.java
```diff
@@ -1,4 +1,4 @@
-package ai.timefold.solver.core.api.domain.lookup;
+package ai.timefold.solver.core.api.domain.entity;
```
Christopher-Chianelli: This will be a PITA for existing users, and problem facts can also have ID's.

core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -232,76 +230,28 @@ private void processEntityAnnotations() {
...
-            throw new IllegalStateException(
-                    "The entityClass (%s) cannot have a %s (%s) and a %s (%s) at the same time.".formatted(getEntityClass(),
-                            "difficultyWeightFactoryClass", difficultyWeightFactoryClass.getName(), "comparatorFactoryClass",
-                            comparatorFactoryClass.getName()));
-        }
         // Selected settings
         var selectedComparatorPropertyName = "comparatorClass";
         var selectedComparatorClass = comparatorClass;
         var selectedComparatorFactoryPropertyName = "comparatorFactoryClass";
         var selectedComparatorFactoryClass = comparatorFactoryClass;
```
Christopher-Chianelli: Couldn't these be removed, since there is now only one `comparator`/`comparatorFactory` property?

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
```diff
@@ -69,80 +68,4 @@ interface NullComparator<T> extends Comparator<T> {
...
-     * This applies to all the planning variables of this planning entity.
-     * <p>
-     * The method {@link PinningFilter#accept(Object, Object)} returns false if the selection entity is pinned
-     * and it returns true if the selection entity is movable
-     *
-     * @return {@link NullPinningFilter} when it is null (workaround for annotation limitation)
-     * @deprecated Prefer using {@link PlanningPin}.
-     */
-    @Deprecated(forRemoval = true, since = "1.23.0")
-    Class<? extends PinningFilter> pinningFilter() default NullPinningFilter.class;
```
Christopher-Chianelli, line 84: Can a `@PlanningPin` method accept a `@PlanningSolution`? If not, I suggest we do allow it, since it does make things easier to express (example: in scheduling, a shift is pinned if it is before the draft start date).

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/descriptor/BasicVariableDescriptor.java
```diff
@@ -40,78 +40,36 @@ protected void processPropertyAnnotations(DescriptorPolicy descriptorPolicy) {
...
             comparatorFactoryClass = null;
         }
-        if (strengthWeightFactoryClass != null && comparatorFactoryClass != null) {
-            throw new IllegalStateException(
-                    "The entityClass (%s) property (%s) cannot have a %s (%s) and a %s (%s) at the same time.".formatted(
-                            entityDescriptor.getEntityClass(), variableMemberAccessor.getName(), "strengthWeightFactoryClass",
-                            strengthWeightFactoryClass.getName(), "comparatorFactoryClass", comparatorFactoryClass.getName()));
-        }
         // Selected settings
         var selectedComparatorPropertyName = "comparatorClass";
```
Christopher-Chianelli: These can be removed/inlined since there is only one comparator/comparatorFactory class.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
```diff
@@ -140,9 +143,6 @@ public void close() {
 
     @Override
     public void afterEntityAdded(EntityDescriptor<Solution_> entityDescriptor, Object entity) {
-        if (entity == null) {
```
Christopher-Chianelli, line 143: There a null check for `fact` and no null check for `entity`?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -923,36 +915,6 @@ private static <Solution_> void assertValueRangeForListVariable(InnerScoreDirect
         }
     }
 
-    public SolutionTracker.SolutionCorruptionResult getSolutionCorruptionAfterUndo(Move<Solution_> move,
```
Christopher-Chianelli, line 926: Was this unused? Do note that although we don't have undo moves in the new API, this also verifies that our generated Undo works properly.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningId.java
triceo: I can move it to just `domain` - but `lookup` is no more, and I will not keep the package around of 1 member.

triceo: Also, this will be easy to provide automatic upgrade for.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningEntity.java
triceo: I'll file an issue for this, but the point here is not to introduce new features; we're removing old cruft.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
triceo: This was indeed unused. I was wondering why that is, but couldn't remember removing any calls to this.

core/src/main/java/ai/timefold/solver/core/api/domain/entity/PlanningId.java
Christopher-Chianelli: I am okay with moving to `domain`

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: I didn't see any removed calls in the code; I will investigate; not a blocker.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirector.java
triceo: I removed he null check, as the very next line will produce an NPE as well:

    if (!getSolutionDescriptor().hasEntityDescriptor(entity.getClass())) {

PR #2137, opened 2026-02-22
feat: maximum function over connected ranges

plentydone:
Let me know what you think about the concept of this feature and it's implementation. I brought it up in Discord some weeks back. Naming things is hard.

The (very minimal) doc example is more or less my use case, calculating equipment usage at all times based on a job attribute, compared to an equipment capacity value that is lessened based on the quantity of jobs. Sort of an enhanced version of maximum overlap.

In implementing it I was starting to feel like a grouping joiner that yields tuples for distinct overlapping ranges of values (possibly with an optional filter function?) would be a more flexible and maintainable interface, but I wanted to send something along to see if you have any early opinions. It feels like a lot of use cases would be answered by a tuple of `(Collection of Range_ simultaneously overlapping), (start Point_ of overlapping range), (end Point_ of overlapping range)`

Happy to refactor any of the interface, implementation, tests, docs, etc according to your input.

### DIFF
--- a/docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
+++ b/docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
@@ -1066,6 +1066,37 @@ Java::
 ----
 ====
 
+In the event that data about distinct ranges of overlapping values is needed beyond the total count, `ConnectedRange` provides methods to calculate a max sum
+of an integer function over each distinct arrangement of overlapping entities:
+
+[tabs]
+====
+Java::
++
+[source,java,options="nowrap"]
+----
+.groupBy(Job::getEquipment,
+    ConstraintCollectors.toConnectedRanges(
+                          Job::getStart,
+                          Job::getEnd
+    )
+).expand(
+    connectedRangeChain -> connectedRangeChain.getMaximumValueForDistinctRanges(
+        (activeJobList, duration) -> {
+            var required = activeJobList.stream().mapToInt(Job::getCapacityRequired).sum()
+            var capacity = equipment.getCapacity() - (activeJobList.size() - 1);  // Lose a little capacity per additional job
+            return Math.max(0, capacity - required);
+        }
+    )
+)
+.filter((connectedRangeChain, amountOverCapacity) -> amountOverCapacity > 0)
+.penalize((connectedRangeChain, amountOverCapacity) -> amountOverCapacity)
+)
+)
+----
+====
+
+
 [#collectorsLoadBalance]
 ==== Load balancing collectors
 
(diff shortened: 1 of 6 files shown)


docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
```diff
@@ -1066,6 +1066,37 @@ Java::
...
+    ConstraintCollectors.toConnectedRanges(
+                          Job::getStart,
+                          Job::getEnd
+    )
+).expand(
+    connectedRangeChain -> connectedRangeChain.getMaximumValueForDistinctRanges(
+        (activeJobList, duration) -> {
+            var required = activeJobList.stream().mapToInt(Job::getCapacityRequired).sum()
+            var capacity = equipment.getCapacity() - (activeJobList.size() - 1);  // Lose a little capacity per additional job
+            return Math.max(0, capacity - required);
```
plentydone, line 1088: could do something like multiplying by the duration here to get a scaled penalty

triceo:
Hello @plentydone, and thanks for the PR!

After a first quick read, I have 2 comments:

- Why should there be a function to do `max` specifically? `max` is just one type of a reduction function, others could be `min`, `avg`, ... IMO this feature should be generified to allow for any reduction function, similarly to map+reduce in Java streams. The `ConstraintCollector` already has all you need for that; I'm just not sure we should use that interface here directly.
- I am a bit worried that this code is designed to not be incremental. In the world of optimization, iterating over collections kills performance. You could argue that these collections will be short, and therefore not a problem, and you may be right. But it gives people ideas that processing collections in CS is good or acceptable, and I'm not sure how I feel about that. At the same time, making this incremental probably takes the complexity of this issue to another level entirely.

plentydone:
@triceo love the feedback, agreed with all points. This spun off a discussion where  @Christopher-Chianelli mentioned that he saw value in the max feature, so I figured I would prompt some ideas with an MR.

Do you think the feature should be in line with my comment `a (new) grouping joiner that yields tuples for distinct overlapping ranges of values (possibly with an optional filter function?)` ? I agree that makes the feature relatively large, but I'm happy to go back to the drawing board and take a swing at it at least as a thought experiment. (That would also allow me to pull it into my own project more easily without maintaining a build.)

triceo:
> Do you think the feature should be in line with my comment a (new) grouping joiner that yields tuples for distinct overlapping ranges of values (possibly with an optional filter function?)

TBH I cannot really imagine right now how that would look. Do you have an API strawman you wanna paste here in chat?

FYI we are already working on joiners which allow for overlap of different collections.

triceo:
Closing due to a long period of inactivity.
We are still interested in this contribution, provided that the behavior can be made incremental. 

Feel free to open another PR if you intend to continue working on this.

PR #2144, opened 2026-02-24
feat: add fast-fail for invalid value ranges

zepfred:
This PR adds a new fail-fast that evaluates whether the value range includes distinct instances of values that have the same identity according to the hashCode/equals semantics.
1. The best place to add the check is during the cache initialization, since we are already touching the structure. This way, the check will not incur any performance penalties.
2. The entity value ranges create a `CompositeValueRange`, and it checks the value range when the value range manager requests the complete value range.
3. The solution range requires the cache to be initialized in order to trigger the check.
4. The complete value range is always requested for both the solution and entity ranges when updating the current working solution.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeCache.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.random.RandomGenerator;
 
+import ai.timefold.solver.core.impl.domain.solution.cloner.DeepCloningUtils;
 import ai.timefold.solver.core.impl.domain.valuerange.sort.ValueRangeSorter;
 import ai.timefold.solver.core.impl.heuristic.selector.common.iterator.CachedListRandomIterator;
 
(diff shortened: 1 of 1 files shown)


triceo:
I've been thinking about this some more, and I no longer think we need this one.

The value ranges will automatically deduplicate the values, because they use `Set`. This means that, even if they do have duplicates in there, there is no way how they could cause problems, and therefore no reason to fail fast. But I think there is another place where we may want to fail fast.

We've seen it with the customer on Monday. When you have the value range, and then independently of that value range the value is already assigned to an entity, _that_ is when we _maybe_ want to fail fast on duplicates. Because that value will just move around through swap moves, won't it? There will be no way to deduplicate this one. (A change move may just entirely eliminate it.)

Although I have to ask... is this a problem? Does this actually cause problems somewhere? IMO we should only fail fast for things that will cause actual issues.

How do we handle duplicate entities? Do we allow that? Maybe there should be special handling for values _if_ they have `@ProblemId`? A lot of questions. But overall, I think this PR as currently implemented is not necessary.

zepfred:
After reconsidering the fail-fast PR, I realized it's not entirely clear that we need it. My major concern was about the constraint stream, but it won't be a problem. To iterate over values, they must be either a problem fact collection or a planning entity collection. In both cases, we are covered by the current entity visitor, which already checks for duplicate values. So, if we have distinct instances of values across the entity value ranges, the merged fix will resolve them and won't affect the score calculation in case of propagated changes.

PR #2148, opened 2026-02-27
chore: final check of codebase post refactors

triceo:


core/src/main/java/ai/timefold/solver/core/api/score/Score.java
```diff
@@ -15,8 +15,10 @@
  * @see HardSoftScore
  */
 @NullMarked
-public interface Score<Score_ extends Score<Score_>>
-        extends Comparable<Score_> {
+public sealed interface Score<Score_ extends Score<Score_>>
```
zepfred, line 18: I’m wondering whether we should restrict users from creating their own score classes.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/DelegatingListVariableStateSupply.java
```diff
@@ -0,0 +1,101 @@
...
+import ai.timefold.solver.core.impl.domain.variable.nextprev.NextElementShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.nextprev.PreviousElementShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.preview.api.domain.metamodel.ElementPosition;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+record DelegatingListVariableStateSupply<Solution_>(ListVariableStateSupply<Solution_, Object, Object> delegate,
```
zepfred, line 19: I understand that only KOpt is using this class, but in my opinion, it should be located in the supply package.

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
triceo: Yes, we should and we are. Creating a score hasn't been possible for a long time - you also need support in ScoreDefinition, in constraint streams... This is simply not an option. We need to extinguish all hope.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/DelegatingListVariableStateSupply.java
triceo: If the need for this shows up, we will share it. I don't think it will - this is a replacement for the index, and no other place was using it. Every other place was using the supply directly.

PR #2149, opened 2026-03-01
feat: convert solver to JPMS

triceo:


core/src/main/java/module-info.java
```diff
@@ -0,0 +1,251 @@
...
+            ai.timefold.solver.benchmark,
+            ai.timefold.solver.benchmark.aggregator,
+            ai.timefold.solver.spring.boot.autoconfigure, ai.timefold.solver.quarkus.deployment,
+            ai.timefold.solver.enterprise.core;
+    exports ai.timefold.solver.core.impl.heuristic.selector.entity to ai.timefold.solver.jackson, ai.timefold.solver.benchmark,
+            ai.timefold.solver.enterprise.core;
+    exports ai.timefold.solver.core.impl.heuristic.selector.entity.pillar
+            to ai.timefold.solver.jackson, ai.timefold.solver.benchmark,
+            ai.timefold.solver.enterprise.core;
+    exports ai.timefold.solver.core.impl.heuristic.selector.list;
```
Christopher-Chianelli: What classes from `ai.timefold.solver.core.impl.heuristic.selector.list` that users are expected to use (there no restrictions on this export)?

core/src/main/java/module-info.java
```diff
@@ -0,0 +1,251 @@
...
+    exports ai.timefold.solver.core.impl.heuristic.selector.move to ai.timefold.solver.jackson, ai.timefold.solver.benchmark,
+            ai.timefold.solver.enterprise.core;
+    exports ai.timefold.solver.core.impl.heuristic.selector.move.factory;
+    exports ai.timefold.solver.core.impl.heuristic.selector.move.generic;
+    exports ai.timefold.solver.core.impl.heuristic.selector.move.generic.list;
```
Christopher-Chianelli: Ditto

core/src/main/java/module-info.java
Christopher-Chianelli: There a section for broad impl export, why aren't these exports there?

quarkus-integration/quarkus/pom.xml
```diff
@@ -22,4 +22,22 @@
...
     <module>devui-integration-test</module>
   </modules>
+
+  <!-- Fixes quarkus https://github.com/microprofile/microprofile-config/issues/811 -->
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.eclipse.microprofile.config</groupId>
+        <artifactId>microprofile-config-api</artifactId>
+        <version>3.1.1-RC1</version>
```
Christopher-Chianelli: We typically store versions either in build-parent or a property.

tools/benchmark/src/main/java/ai/timefold/solver/benchmark/config/PlannerBenchmarkConfig.java
```diff
@@ -107,10 +106,10 @@ public class PlannerBenchmarkConfig {
      */
     public static @NonNull PlannerBenchmarkConfig createFromXmlResource(@NonNull String benchmarkConfigResource,
             @Nullable ClassLoader classLoader) {
-        ClassLoader actualClassLoader = classLoader != null ? classLoader : Thread.currentThread().getContextClassLoader();
-        try (InputStream in = actualClassLoader.getResourceAsStream(benchmarkConfigResource)) {
+        var actualClassLoader = classLoader != null ? classLoader : Thread.currentThread().getContextClassLoader();
+        try (var in = actualClassLoader.getResourceAsStream(benchmarkConfigResource)) {
             if (in == null) {
-                String errorMessage = "The benchmarkConfigResource (" + benchmarkConfigResource
+                var errorMessage = "The benchmarkConfigResource (" + benchmarkConfigResource
```
Christopher-Chianelli: We use `"The x (%s) does not y (%s).".formatted(x, y)` typically for creating strings with variable references.

tools/benchmark/src/main/java/ai/timefold/solver/benchmark/config/PlannerBenchmarkConfig.java
```diff
@@ -255,10 +244,10 @@ public class PlannerBenchmarkConfig {
      */
     public static @NonNull PlannerBenchmarkConfig createFromFreemarkerXmlResource(@NonNull String templateResource,
             @Nullable Object model, @Nullable ClassLoader classLoader) {
-        ClassLoader actualClassLoader = classLoader != null ? classLoader : Thread.currentThread().getContextClassLoader();
-        try (InputStream templateIn = actualClassLoader.getResourceAsStream(templateResource)) {
+        var actualClassLoader = classLoader != null ? classLoader : Thread.currentThread().getContextClassLoader();
+        try (var templateIn = actualClassLoader.getResourceAsStream(templateResource)) {
             if (templateIn == null) {
-                String errorMessage = "The templateResource (" + templateResource
+                var errorMessage = "The templateResource (" + templateResource
```
Christopher-Chianelli: Ditto

tools/migration/src/main/resources/META-INF/rewrite/ToLatest.yml
```diff
@@ -9,8 +9,6 @@ recipeList:
   - org.openrewrite.java.DeleteMethodArgument:
       methodPattern: ai.timefold.solver.core.api.score.constraint.ConstraintRef of(String, String)
       argumentIndex: 0
+  - ai.timefold.solver.migration.one.PersistenceCommonMigrationRecipe
   - ai.timefold.solver.migration.one.TestingAPIsMigrationRecipe
-  - org.openrewrite.java.RemoveUnusedImports
-  - org.openrewrite.maven.RemoveDependency:
-      groupId: ai.timefold.solver
-      artifactId: timefold-solver-test
```
Christopher-Chianelli, line 16: Why was this removed? As far as I can tell, the module `timefold-solver-test` no longer exists on main.

core/src/main/java/module-info.java
triceo: I have changed this section just now, created a new one specifically for this.
The `heuristic` subpackage is a mess of tangled classes, some of which will be necessary for users to play with custom moves. We will find some dependency problems here, when people start using this, I'm sure.

tools/migration/src/main/resources/META-INF/rewrite/ToLatest.yml
triceo: It's still there, just moved to Java code.

Christopher-Chianelli: I don't see a `RemoveDependency` rewrite in https://github.com/TimefoldAI/timefold-solver/blob/main/tools/migration/src/main/java/ai/timefold/solver/migration/one/TestingAPIsMigrationRecipe.java (which only move packages and does not modify maven packages), nor do any rewrites in this PR actually remove the `timefold-solver-test` dependency.
(Edit: I see it now)

PR #2156, opened 2026-03-02
chore: add escape hatch for SolverConfig in Quarkus build time

Christopher-Chianelli:
Due to what I assume to be classloader shenanigans, the generated Gizmo classes cannot access annotations on fields/methods at Quarkus build time.

An escape hatch was added to SolverConfig for those users that need to build a Solver or related components at Quarkus build time.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
@@ -212,32 +212,32 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
     // Warning: all fields are null (and not defaulted) because they can be inherited
     // and also because the input config file should match the output config file
     @XmlElement(name = "enablePreviewFeature")
-    protected Set<PreviewFeature> enablePreviewFeatureSet = null;
-    protected EnvironmentMode environmentMode = null;
-    protected Boolean daemon = null;
-    protected RandomType randomType = null;
-    protected Long randomSeed = null;
-    protected Class<? extends RandomFactory> randomFactoryClass = null;
-    protected String moveThreadCount = null;
-    protected Integer moveThreadBufferSize = null;
-    protected Class<? extends ThreadFactory> threadFactoryClass = null;
+    private Set<PreviewFeature> enablePreviewFeatureSet = null;
+    private EnvironmentMode environmentMode = null;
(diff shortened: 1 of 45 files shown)


core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -643,10 +646,22 @@ public boolean canTerminate() {
...
+            return DomainAccessType.AUTO;
+        }
+        return resolvedDomainAccessType;
+    }
+
     // ************************************************************************
     // Builder methods
     // ************************************************************************
 
+    public SolverConfig adaptForQuarkusBuildTime() {
```
triceo: I don't like this. Not because it's a config setting, but because it leaks Quarkus into code which has nothing to do with it. `core` shouldn't do anything about Quarkus.

Can we not improve the Gizmo detection? We now know that, for whatever reason, it cannot read annotations. So why don't we make _that_ the check? Or combine that with the check we already have.

Christopher-Chianelli: It already exists in SolverConfig, via `GizmoMemberAccessorMap` and `GizmoSolutionClonerMap`. Any check would be way more complicated and require "rewinding" the SolutionDescriptor when an error occurs. `GizmoMemberAccessorMap` would be non-null at both build time and run time, so it cannot be used to determine we are at Quarkus build time.

triceo: I don't understand... when we run the check whether Gizmo is enabled, that has to happen - logically - _before_ any members are accessed. Because it is used to determine which member accessors are to be generated.

Therefore it should be possible, in the checking logic, to pick some annotated class and read its annotations. There are plenty - override, nullmarked, ... Why is this not possible?

triceo (changes requested):
As discussed on Slack, we should make sure the SDK can access the constraint metamodel through the build item which already exists.

The build item probably needs to do the same hack that the PR already uses. The benefit is though that the build item need not do this publicly and therefore the method on solver config need not exist.

(Arguably, we should do something similar with the gizmo stuff on solver config as well. We have access to internals, we should just push this information through wherever we need this, without bothering the user with this nasty impl detail.)

triceo (approved):
I like this!

PR #2157, opened 2026-03-03
perf: significantly improve assertion speed

triceo:
The basic variable descriptor list was computed every time the assertion was called. Which was on every basic variable change. 

As a result of this change, a test which used to take 2 minutes now takes under 30 seconds. It also allocates less than 10 % of the memory it originally did.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
@@ -14,13 +14,13 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.SequencedCollection;
+import java.util.SequencedMap;
 import java.util.function.BiPredicate;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
 import ai.timefold.solver.core.api.domain.entity.PlanningPin;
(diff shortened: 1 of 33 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/entity/descriptor/EntityDescriptor.java
```diff
@@ -667,21 +668,21 @@ public boolean hasAnyDeclaredGenuineVariableDescriptor() {
...
-    public Collection<CascadingUpdateShadowVariableDescriptor<Solution_>>
+    public SequencedCollection<CascadingUpdateShadowVariableDescriptor<Solution_>>
             getDeclaredCascadingUpdateShadowVariableDescriptors() {
-        return declaredCascadingUpdateShadowVariableDecriptorMap.values();
+        return declaredCascadingUpdateShadowVariableDecriptorMap.sequencedValues();
     }
 
-    public Collection<VariableDescriptor<Solution_>> getDeclaredVariableDescriptors() {
-        Collection<VariableDescriptor<Solution_>> variableDescriptors = new ArrayList<>(
+    public SequencedCollection<VariableDescriptor<Solution_>> getDeclaredVariableDescriptors() {
```
zepfred, line 679: This method aligns with the PR goal and should be cached.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -1046,14 +1051,6 @@ public long getGenuineVariableCount(Solution_ solution) {
         return result.longValue();
     }
 
-    public List<ShadowVariableDescriptor<Solution_>> getAllShadowVariableDescriptors() {
```
zepfred, line 1049: The method `getDeclarativeShadowVariableDescriptors()` aligns with the PR goal and should also be cached.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/nextprev/AbstractNextPrevElementShadowVariableDescriptor.java
```diff
@@ -61,11 +60,11 @@ private void linkShadowSources(DescriptorPolicy descriptorPolicy) {
                     + entitiesWithSourceVariable + ").");
         }
         VariableDescriptor<Solution_> variableDescriptor =
-                entitiesWithSourceVariable.get(0).getVariableDescriptor(sourceVariableName);
+                entitiesWithSourceVariable.getFirst().getVariableDescriptor(sourceVariableName);
         if (variableDescriptor == null) {
             throw new IllegalStateException(
                     "Impossible state: variableDescriptor (" + variableDescriptor + ") is null"
-                            + " but previous checks indicate that the entityClass (" + entitiesWithSourceVariable.get(0)
+                            + " but previous checks indicate that the entityClass (" + entitiesWithSourceVariable.getFirst()
```
zepfred: Format pattern?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -762,10 +767,13 @@ public PlanningSolutionMetaModel<Solution_> getMetaModel() {
...
-        return getGenuineEntityDescriptors().stream()
-                .flatMap(entityDescriptor -> entityDescriptor.getGenuineBasicVariableDescriptorList().stream())
-                .map(descriptor -> (BasicVariableDescriptor<Solution_>) descriptor)
+        if (basicVariableDescriptorList != null) {
+            return basicVariableDescriptorList;
+        }
+        basicVariableDescriptorList = getGenuineEntityDescriptors().stream()
+                .flatMap(entityDescriptor -> entityDescriptor.getBasicVariableDescriptorList().stream())
                 .toList();
+        return basicVariableDescriptorList;
```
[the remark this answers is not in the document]
triceo: This actually is safe. Worst that can happen is that this information will be computed multiple times before finally being applied to all threads.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
zepfred, line 1033: I also think this one should be cached.

PR #2161, opened 2026-03-04
fix: score corruption caused by duplicate, implement randomIterator

### LINKED ISSUE #8: CS: new joiners: intersecting(), disjoint(), containsAll()
Reqs:

- Hashtable index performance/scalability
- Intersecting: Person Ann and Beth have a hobby in common, so they can sit at the same table.
- Disjoint: Talk A and B don't have a topic in common, so they can happen in parallel.
- ContainsAll: Job A has a set of required skills. Person Ann's skillSet contains all of those required skills.

```
                .forEachUniquePair(Job.class,
                        overlapping(Job::getStartDate, Job::getEndDate),
                        // TODO Make this hash powered
                        filtering((job1, job2) -> !Collections.disjoint(
                                job1.getTagSet(), job2.getTagSet())))
```

To reproduce: see MaintenanceScheduleConstraintProvider#tagConflict

Christopher-Chianelli:
- Move UnfinishedJoiners to Joiners
- Score corruption was caused when a duplicate was encounter, causing the forEach iterator to skip the remaining elements in its current downstream iterator and proceed to the next downstream iterator instead. It now exhausts the current downstream iterator beforce proceeding to the next one.
- Due to the existance of duplicates, the implemented randomIterator is not completely fair; to create a fair random iterator, all the elements must be put into a set first.

Draft, since it probably need more tests, but the modified examples pass `FULL_ASSERT`.


core/src/main/java/ai/timefold/solver/core/api/score/stream/Joiners.java
```diff
@@ -256,6 +257,92 @@ public final class Joiners {
...
+     *
+     * @param leftMapping mapping function to apply to A
+     * @param rightMapping mapping function to apply to B
+     * @param <A> the type of object on the left
+     * @param <B> the type of object on the right
+     * @param <Property_> the type of the property to compare
+     */
+    public static <A, B, Property_> @NonNull BiJoiner<A, B> containingAnyOf(
+            @NonNull Function<A, Collection<Property_>> leftMapping,
+            @NonNull Function<B, Collection<Property_>> rightMapping) {
```
triceo: Didn't we say this was going to require `SequencedCollection`?
Is that no longer the case?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -187,12 +188,24 @@ private Iterator<T> iteratorSingleKey(Object queryCompositeKey, KeyCollection_ i
...
+            return Collections.emptyIterator();
+        }
+        if (filter == null) {
+            return new RandomIterator(indexKeyCollection, workingRandom,
+                    downstreamIndexer -> downstreamIndexer.randomIterator(queryCompositeKey, workingRandom),
+                    size(queryCompositeKey));
+        } else {
+            return new RandomIterator(indexKeyCollection, workingRandom,
+                    downstreamIndexer -> downstreamIndexer.randomIterator(queryCompositeKey, workingRandom, filter),
+                    size(queryCompositeKey));
```
triceo: Computing the size in this indexer is quite expensive. Since the random iterators are going to be created pretty frequently and their performance will dictate the performance of Neighborhoods, I've been wondering - can we replace `size(...)` internally by keeping a counter?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +275,72 @@ public T next() {
...
+
+        public RandomIterator(KeyCollection_ indexKeyCollection, RandomGenerator workingRandom,
+                Function<Indexer<T>, Iterator<T>> downstreamIndexerIteratorFunction,
+                int uniqueElementCount) {
+            this.downstreamIteratorList = new ArrayList<>(indexKeyCollection.size());
+            this.removedSet = new HashSet<>();
+            this.workingRandom = workingRandom;
+            this.uniqueElementCount = uniqueElementCount;
+            for (var indexKey : indexKeyCollection) {
+                this.downstreamIteratorList.add(downstreamIndexerIteratorFunction.apply(downstreamIndexerMap.get(indexKey)));
```
triceo: I'd rather we built this on-demand.
In a typical use case, there will only be one pick, which will immediately return something; therefore creating all of this would be a waste. We should IMO only create the iterator first time we need it.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +275,72 @@ public T next() {
...
+            while (!downstreamIteratorList.isEmpty()) {
+                var remainingIteratorCount = downstreamIteratorList.size();
+                var selectedIndex = workingRandom.nextInt(remainingIteratorCount);
+                var downstreamIterator = downstreamIteratorList.get(selectedIndex);
+                if (!downstreamIterator.hasNext()) {
+                    downstreamIteratorList.remove(selectedIndex);
+                    continue;
+                }
+
+                next = downstreamIterator.next();
```
triceo: IMO we should remove the iterator here already, if we find out it no longer `hasNext()`. That way, we can never pick an empty iterator again.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +275,72 @@ public T next() {
...
+                if (!removedSet.contains(next)) {
+                    return true;
+                }
```
triceo, line 384: I wonder if instead of picking another random iterator again, we should just pick from the same iterator. It was randomly picked already, IMO that means we should select from it.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +275,72 @@ public T next() {
...
+            current = next;
+            next = null;
+            return current;
+        }
+
+        @Override
+        public void remove() {
+            // Note: since we have multiple downstream iterators, and they may have
+            // duplicates, we need to keep track of removed elements ourselves
+            if (current == null) {
```
triceo, line 403: Don't we want to support `null` as a valid value? I'm not exactly sure if the other indexers do or don't, but we should be consistent either way. (I think `equal` at least does support `null` as a value.)

core/src/main/java/module-info.java
```diff
@@ -154,8 +154,6 @@
             ai.timefold.solver.enterprise.core;
     exports ai.timefold.solver.core.impl.score.director.incremental to ai.timefold.solver.benchmark,
             ai.timefold.solver.enterprise.core;
-    exports ai.timefold.solver.core.impl.score.stream to ai.timefold.solver.jackson,
-            ai.timefold.solver.enterprise.core;
```
triceo, line 157: Why are we removing this? Especially the export to Jackson suggests that something here was important.

triceo (commented):
Adding some comments after first reading.

core/src/main/java/module-info.java
Christopher-Chianelli: No classes in `ai.timefold.solver.core.impl.score.stream` exist now; the sole class that was exported was `UnfinishedJoiners`.

triceo: Then I wonder why it was exported in the first place, but fair enough.

core/src/main/java/ai/timefold/solver/core/api/score/stream/Joiners.java
Christopher-Chianelli: Not needed;  the score corruption was caused by this:
"""
Score corruption was caused when a duplicate was encounter, causing the forEach iterator to skip the remaining elements in its current downstream iterator and proceed to the next downstream iterator instead. It now exhausts the current downstream iterator beforce proceeding to the next one.
"""

I suspect when `SequencedCollection`  was tested, it changed the iteration order and thus avoided the issue, but it still have the same bug (when I tested with `TreeSet`).

In particular, this is the fix for the bug:

```java
- if (downstreamIterator != null && downstreamIterator.hasNext()) {
-     var tuple = downstreamIterator.next();
-     if (distinctingSet.add(tuple)) {
-         next = tuple;
-         return true;
+ if (downstreamIterator != null) {
+     while (downstreamIterator.hasNext()) {
+         var tuple = downstreamIterator.next();
+         if (distinctingSet.add(tuple)) {
+             next = tuple;
+             return true;
+         }
      }
  }
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
triceo: No, because we never insert `null`, as seen by the contract of `put()`. 
We insert `Tuple(null)`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
Christopher-Chianelli: That would affect fairness and we still need to change iterators once the current one exhausted.

triceo: I don't think it would. In fact, it would mean one less random number generated.

Picking an iterator randomly and finding that its next value was already selected, that's effectively a no-op. I could argue that if you skip iterators because some of their values have already been selected, then you start skewing in favor of iterators which - irrelevantly - only contain values which are in no other iterator.

triceo:
One more thing that crossed my mind: the indexer for intersections is expensive, because it has to aggregate things from multiple collections and keep track of that. I'm wondering - would it make sense to create a specialized implementation for cases of only 1 collection? It's a question of whether this situation is likely to happen.

core/src/main/java/ai/timefold/solver/core/api/score/stream/Joiners.java
```diff
@@ -256,6 +257,92 @@ public final class Joiners {
...
+     * <li>{@code []} containing {@code null} is {@code false}</li>
+     * </ul>
+     *
+     * @param leftMapping mapping function to apply to A
+     * @param rightMapping mapping function to apply to B
+     * @param <A> the type of object on the left
+     * @param <B> the type of object on the right
+     * @param <Property_> the type of the property to compare
+     */
+    public static <A, B, Property_> @NonNull BiJoiner<A, B> containing(@NonNull Function<A, Collection<Property_>> leftMapping,
```
triceo: I know we said we don't need `SequencedCollection` here, but let's consider the impact on reproducibility.

If these collections are iterated in different orders between different runs of the JVM, that means different moves will be picked in different runs of the JVM. Meaning the solver will not be reproducible.

IMO this needs to be addressed. I see two ways of addressing this:

- These joiners only accept sequenced collections.
- The indexers' internal representation is a sequenced collection.

From my point of view, it seems simpler to just force the user to use a sequenced collection.
In which case, an explanatory note should be added to the javadocs of these methods.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +273,111 @@ public T next() {
...
+        public RandomIterator(KeyCollection_ indexKeyCollection, RandomGenerator workingRandom,
+                Function<Indexer<T>, Iterator<T>> downstreamIndexerIteratorFunction) {
+            this.downstreamIteratorSupplierList = new ArrayList<>(indexKeyCollection.size());
+            this.removedSet = new HashSet<>();
+            this.workingRandom = workingRandom;
+            this.downstreamIndexerIteratorFunction = downstreamIndexerIteratorFunction;
+            for (var indexKey : indexKeyCollection) {
+                this.downstreamIteratorSupplierList.add(new DownstreamIteratorSupplier(indexKey));
+            }
+        }
```
triceo: Consider the main (and currently the only) use case of the random iterator, Neighborhoods.

Late Acceptance is a fast-stepping algorithm, therefore typically it will only select 1 move in a step. This means this iterator will be created once, its `next()` will be called once, and then it will be thrown away.

In this light, this constructor is still far too heavy. It creates one possibly large collection, and fills it entirely. It also instantiates another hash set. This is quite the overhead for something which will be happening all the time, and only be used once.

IMO this constructor needs to be as lazy as it can possibly be. This class needs to be optimized for the 1-time use path. And then possibly delegate to a different implementation which is optimized for the other path, where there are multiple selections.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +273,111 @@ public T next() {
...
+                        downstreamIterator.remove();
+                    }
+                }
+                downstreamIteratorSupplierList.remove(currentIteratorSupplier);
+            }
+            while (!downstreamIteratorSupplierList.isEmpty()) {
+                var remainingIteratorCount = downstreamIteratorSupplierList.size();
+                var selectedIndex = workingRandom.nextInt(remainingIteratorCount);
+                currentIteratorSupplier = downstreamIteratorSupplierList.get(selectedIndex);
+                var downstreamIterator = currentIteratorSupplier.iterator();
```
triceo: I would declare this variable outside of this loop; that way, you don't need to call the supplier over and over again, making it clear that the iterator is still the same instance.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +273,111 @@ public T next() {
...
+        public void remove() {
+            // Note: since we have multiple downstream iterators, and they may have
+            // duplicates, we need to keep track of removed elements ourselves
+            if (current == null) {
+                throw new IllegalStateException("next() must be called before remove().");
+            }
+            removedSet.add(current);
+            currentIteratorSupplier.iterator().remove();
+            if (!currentIteratorSupplier.iterator().hasNext()) {
+                downstreamIteratorSupplierList.remove(currentIteratorSupplier);
```
triceo: See previous comment; the iterator is the same in both cases, yet the user cannot be sure without looking at the implementation of the supplier.

triceo (approved):
LGTM after comments are addressed.
(There are some valid ones in Sonar as well.)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +273,111 @@ public T next() {
...
+                    } else {
+                        downstreamIterator.remove();
+                    }
+                }
+                downstreamIteratorSupplierList.remove(currentIteratorSupplier);
+            }
+            while (!downstreamIteratorSupplierList.isEmpty()) {
+                var remainingIteratorCount = downstreamIteratorSupplierList.size();
+                var selectedIndex = workingRandom.nextInt(remainingIteratorCount);
+                currentIteratorSupplier = downstreamIteratorSupplierList.get(selectedIndex);
```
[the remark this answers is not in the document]
triceo: Random selection is hard. :-( I think there is some logic to this comment.

We cannot do this absolutely perfectly, the possibility of duplicates screws us here. But selecting the iterators proportionally makes sense; if something has 2x the size, it should be selected 2x as likely.

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/ContainedInIndexerTest.java
```diff
@@ -8,17 +8,21 @@
...
 
 class ContainedInIndexerTest extends AbstractIndexerTest {
 
     private final DefaultBiJoiner<TestJob, TestWorker> joiner =
-            (DefaultBiJoiner<TestJob, TestWorker>) UnfinishedJoiners.containedIn(TestJob::skill, TestWorker::skills)
+            (DefaultBiJoiner<TestJob, TestWorker>) Joiners.containedIn(TestJob::skill, TestWorker::skills)
                     .and(Joiners.equal(TestJob::department, TestWorker::department));
 
+    private final DefaultBiNeighborhoodsJoiner<TestWorker, TestJob> randomAccessSingleJoiner =
+            new DefaultBiNeighborhoodsJoiner<>(TestWorker::skills, JoinerType.CONTAINED_IN, TestJob::skill);
```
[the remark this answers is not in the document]
triceo: I only now noticed... why does this use a neighborhoods joiner? It should use a standard joiner, no?

Does this PR even have support for neighborhoods? I don't see us introducing neighborhood joiners. This is suspicious!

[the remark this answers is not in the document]
Christopher-Chianelli: To test random iterator, since the regular joiners fail fast when `randomIterator` is called.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
Christopher-Chianelli: You do realize `SequencedCollection` cannot be accessed by `index` nor any way of randomly accessing its elements? The only way to randomly select an element from a `SequencedCollection` is to store it inside a list or another `RandomAccess` structure. If we always picked the first element, it is no longer a `randomIterator`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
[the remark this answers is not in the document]
Christopher-Chianelli: But isn't `size()` expensive (or expensive enough to justify removing it from the constructor)?

Christopher-Chianelli:
I have a suggestion: since Neighborhoods apparently require Joiners from `NeighborhoodJoiners` instead of `Joiners`, make `NeighborhoodJoiners` use `SequencedCollection`, and regular `Joiners` use `Collection`. Forcing `SequencedCollection` means users won't be able to use `Set` (which may be better for their domain model) even if they don't use neighborhoods.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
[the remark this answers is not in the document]
triceo: It is expensive. But:

- Maybe it needn't be. Perhaps the indexer can maintain a counter, instead of always calling `size()` on the downstreams. It knows when it inserts and when it removes. Maybe it has no way of knowing when it removes a duplicate?
- Correctness over performance; Copilot is right here, we skew the selection probability towards the smaller indexers.

It sucks though, because if we can't make `size()` perform reasonably, this really will be slow.

triceo:
> Forcing `SequencedCollection` means users won't be able to use `Set` (which may be better for their domain model) even if they don't use neighborhoods. 

This is the bit I don't understand. `SequencedSet extends SequencedCollection` - so they can always use `LinkedHashSet`. Or `TreeSet`. Yes, they can no longer use `HashSet`; but most uses of `HashSet` around the solver are wrong anyway, because they introduce non-reproducibility. IMO nothing bad about forcing (teaching) this when we can now.

(For example: we recently found non-reproducibility in one of the examples, because Jackson was deserializing JSON into a `Set` and decided to use `HashSet`. The collection then ended up as a value range, and that caused non-determinism. Had we forced `SequencedSet`, Jackson would have no choice but to use `LinkedHashSet` and this tricky well-hidden bug would never have shown up.)

Christopher-Chianelli:
> > Forcing `SequencedCollection` means users won't be able to use `Set` (which may be better for their domain model) even if they don't use neighborhoods.
> 
> This is the bit I don't understand. `SequencedSet extends SequencedCollection` - so they can always use `LinkedHashSet`. Or `TreeSet`. Yes, they can no longer use `HashSet`; but most uses of `HashSet` around the solver are wrong anyway, because they introduce non-reproducibility. IMO nothing bad about forcing (teaching) this when we can now.
> 
> (For example: we recently found non-reproducibility in one of the examples, because Jackson was deserializing JSON into a `Set` and decided to use `HashSet`. The collection then ended up as a value range, and that caused non-determinism. Had we forced `SequencedSet`, Jackson would have no choice but to use `LinkedHashSet` and this tricky well-hidden bug would never have shown up.)

`Set.of(...)` returns a `Set` and not `SequencedSet`, and there no `of` method for `SequencedSet`.

Christopher-Chianelli:
(Java unfortantly makes using `SequencedSet` extremely annoying)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
triceo: From this point of view, it matters not if it's a `Collection` or `SequencedCollection`; random access is not possible either way. `Sequenced` gives us more information without taking any away.


I'm probably missing something in your comment, because I don't see how it's relevant to this code.

Christopher-Chianelli: "In this light, this constructor is still far too heavy. It creates one possibly large collection, and fills it entirely. It also instantiates another hash set. This is quite the overhead for something which will be happening all the time, and only be used once."

i.e. You need to create one possible large collection and fill it regardless so you have random access to keys.

triceo: But you don't need this one large collection. The logic is pretty much what it is now - pick one iterator, pick one random thing from it. If more values are needed, that's when the switch happens to the heavier logic.

That said, this may be a moot point once we address the comment on fair random selection and kill performance doing it.

Christopher-Chianelli: In order to be fair, you must be able to pick any iterator, which always means iterating the entire collection since random access is not available. And if you want to take an expensive `size()` calculation into account for fairness, that has to be stored in some datastructure.

triceo:
> (Java unfortantly makes using `SequencedSet` extremely annoying)

True. I wonder whether - if we provided a PR to create `SequencedSet.of` - it would be accepted or not.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,131 @@ public T next() {
...
     }
 
+    private final class RandomIterator implements Iterator<T> {
+
+        private final List<DownstreamIteratorSupplier> downstreamIteratorSupplierList;
+        private final RandomGenerator workingRandom;
+        private final Function<Indexer<T>, Iterator<T>> downstreamIndexerIteratorFunction;
+        private final int[] distribution;
+        private int distributionSum;
+        private @Nullable Set<T> removedSet;
```
triceo: I would add an explanatory comment as to why these are nullable.
(It's not obvious that this is an optimization aimed at the fast-stepping algorithms.)

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,131 @@ public T next() {
...
+        private @Nullable T next = null;
+        private @Nullable T current = null;
+        private @Nullable DownstreamIteratorSupplier currentIteratorSupplier = null;
+        private @Nullable Iterator<T> currentIterator = null;
+
+        private class DownstreamIteratorSupplier {
```
triceo: I wonder... why not make this be `Iterator`? We already have `remove()` - sounds like we're half-way.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,131 @@ public T next() {
...
+                throw new NoSuchElementException();
+            }
+            current = next;
+            next = null;
+            return current;
+        }
+
+        @Override
+        public void remove() {
+            // Note: since we have multiple downstream iterators, and they may have
```
triceo: The "Note" is implied by this being a comment.

```suggestion
            // Since we have multiple downstream iterators, and they may have
```

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomUtils.java
```diff
@@ -49,18 +47,24 @@ public static double nextDouble(RandomGenerator random, double n) {
...
+     * @return An index between 0 and {@code distribution.length}, biased towards indices in distribution with larger values.
      */
-    public static <Item_> void shuffle(List<Item_> shuffledList, RandomGenerator random) {
-        var listSize = shuffledList.size();
-        for (var rightIndex = listSize - 1; rightIndex > 0; rightIndex--) {
-            var leftIndex = random.nextInt(rightIndex + 1); // leftIndex <= rightIndex
-            var leftValue = shuffledList.get(leftIndex);
-            var rightValue = shuffledList.set(rightIndex, leftValue);
-            shuffledList.set(leftIndex, rightValue);
+    public static int sampleWithDistribution(RandomGenerator random, int distributionSum,
```
triceo, line 59: Wouldn't this be more efficient if the array was replaced with a `TreeSet`?
Spatially, probably not.
Temporally though, the loop would be avoided.

triceo (commented):
Since we have substantial new code, I have added a small number of new comments.

We'll release Beta 1 later today, at which point I'd like to start merging some of these longer-standing PRs.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomUtils.java
Christopher-Chianelli: I doubt indexKeyCollection would have more than 16 elements in the vast majority of cases, in which case, an array is more efficient. Additional, the array is constantly mutated by the remove.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,134 @@ public T next() {
         }
     }
 
+    private final class RandomIterator implements Iterator<T> {
+
+        private final List<DownstreamIterator> downstreamIteratorList;
+        private final RandomGenerator workingRandom;
+        private final int[] distribution;
+        private int distributionSum;
```
triceo, line 295: Let's add comments to these field describing their purpose (fair random selection), and how it works.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,134 @@ public T next() {
...
+        private @Nullable T next = null;
+        private @Nullable T current = null;
+
+        // These are nullable as an optimization for fast-step algorithms which
+        // will only call next() once.
+        private @Nullable Set<T> removedSet;
+        private @Nullable DownstreamIterator currentIterator = null;
+
+        private class DownstreamIterator implements Iterator<T> {
```
triceo, line 304: Now that the class is an `Iterator`, the methods are missing `@Override`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
```diff
@@ -260,4 +274,134 @@ public T next() {
...
+        @Override
+        public boolean hasNext() {
+            if (next != null) {
+                return true;
+            }
+            if (currentIterator != null) {
+                while (currentIterator.hasNext()) {
+                    next = currentIterator.next();
+                    if (removedSet == null || !removedSet.contains(next)) {
+                        return true;
```
triceo, line 356: A thought just crossed my mind.

Why are we not removing the item here?
Doesn't not removing the item mean that it can be picked again?
Or is the remove call coming from above in Neighborhoods?

Either way, this deserves a comment.

Christopher-Chianelli: An item is added to `removeSet` only in `remove`. If `removedSet` does contains the item, it is removed from the current iterator in the `else` block. What would you want a comment to say?

triceo: I would expect that, when something is selected, that means it can no longer be selected again. Therefore it has to be removed. 

Remove here doesn't actually remove from the indexer; it removes from _the iterator_. Meaning that it prevents a repeated random selection of the same thing; it is a way of doing exhaustive random selection. At least that was the case the last time I reviewed the code holistically together with Neighborhoods.

Christopher-Chianelli: `ContainedInIndexer`'s `RandomIterator` only calls `remove` in `remove` (so by default, I assume returning duplicate elements are allowed, and if you don't want duplicate elements, you need to explictly call remove).

Christopher-Chianelli: (And if we `remove` in `next`, it makes no sense to have a `remove` method.)

Christopher-Chianelli: Looking at `DefaultUniqueRandomIterator`, an element is only removed from the iterator in `remove` (i.e. `DefaultUniqueRandomIterator` will return duplicates unless `remove` is called). So I am pretty sure it is the neighborhood code responsibility to call `remove` if it does not want to get duplicate elements.

PR #2167, opened 2026-03-05
fix: in event consumption, always release the lock if we already aquired it

triceo:
The `activeConsumption` semaphore was only released conditionally, which led to an issue with final best solution consumer and therefore failing to terminate on very high CPU loads.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
@@ -18,6 +18,19 @@ import ai.timefold.solver.core.api.solver.event.SolverJobStartedEvent;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+/**
+ * Support class for consuming solver events in a separate thread.
+ * It ensures that the events are consumed in the correct order and handles exceptions properly.
+ * <p>
+ * The public consume* methods in this class are called by the Solver thread,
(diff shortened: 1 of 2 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -49,26 +50,60 @@ public ConsumerSupport(ProblemId_ problemId, @Nullable Consumer<NewBestSolutionE
...
+
+    /**
+     * Called both on the Solver thread and the Consumer thread.
+     * Don't call without locking, otherwise multiple consumptions may be scheduled.
+     * 
+     * @return future which completes when the consumption is done; can be unlocked then
+     */
+    private CompletableFuture<Void> scheduleIntermediateBestSolutionConsumption() {
+        return CompletableFuture.runAsync(() -> {
+            BestSolutionContainingProblemChanges<Solution_> bestSolutionContainingProblemChanges = bestSolutionHolder.take();
```
Christopher-Chianelli:
```suggestion
            var bestSolutionContainingProblemChanges = bestSolutionHolder.take();
```

triceo:
> Why do you make the caller release the semaphore instead of releasing it unconditionally inside the function?

If you look carefully, you will notice that the method that acquires the permit is now the same method that releases the permit; this is much cleaner, and if it had been done like this from the start, the error of only conditionally releasing the permit would probably not have been missed. It would have been painfully obvious, because it would have been in the same method, in the same context.

Issue #2172, opened 2026-03-06
Bug: ValueRangeCache IdentityHashSet breaks value range deduplication with JPA entities


paoloantinori:
## Bug Report: ValueRangeCache IdentityHashSet breaks value range deduplication with JPA entities

**Version**: Timefold Solver Community Edition 1.31.0  
**Severity**: High (blocks solver initialization)  
**Type**: Design contradiction between deduplication and validation

### Summary

`ValueRangeCache` uses `IdentityHashSet` (`==`) for containment checks on user-defined types, but value range deduplication (`HashedValueRange`) uses `.equals()` to match ranges. When entity instances differ by identity but match by `.equals()`, validation fails with:

```
IllegalStateException: The value (Room@456) has been assigned to the entity (Lesson@789), 
but it is outside of the related value range [Room@123, ...].
```

### Reproducible Case

This blocks legitimate scenarios with:
- **Filtering value range providers** (e.g., `@FilteringValueRangeProvider`)
- **Pre-assignment** with JPA entities
- Entities persisted in one transaction, loaded in another (different object instances, same `id`)

```java
@Entity
public class Lesson extends PanacheEntityBase {
    @PlanningId public Long id;
    
    @PlanningVariable(valueRangeProviderRefs = {"rooms"})
    public Room room;
}

@Entity  
public class Room extends PanacheEntityBase {
    public Long id;
    
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (!(o instanceof Room)) return false;
        return id != null && id.equals(((Room) o).id);
    }
    
    @Override
    public int hashCode() {
        return id == null ? getClass().hashCode() : id.hashCode();
    }
}
```

### Root Cause

**ValueRangeState deduplicates ranges using `.equals()`**:
```java
// ValueRangeState.java (v1.31.0)
Entity_ fromEntity = valueRangeDeduplicationCache.get(hashedValueRange);
// HashedValueRange.equals() uses ListValueRange.equals() → Room.equals()
```

**ValueRangeCache uses `IdentityHashSet` for validation**:
```java
// ValueRangeCache.java (v1.31.0) - FOR_USER_VALUES
return new ValueRangeCache<>(size, CollectionUtils.newIdentityHashSet(size), false);
```

**The contradiction**: Deduplication says two ranges are "equal" by `.equals()`, but validation says a value is "in range" only if `==` (same object).

### Suggested Fix

[...]

PR #2173, opened 2026-03-06
test: Add regression test for issue #2172 (ValueRangeCache equals vs identity)

paoloantinori:
## Summary

Add regression test for issue #2172 - ValueRangeCache should use `.equals()` not `==` for JPA entities.

## Background

Issue #2172 reported that `ValueRangeCache.Builder.FOR_USER_VALUES` used `IdentityHashSet` (`==`) instead of `HashSet` (`.equals()`) for containment checks. This caused `IllegalStateException` when:

1. Using filtering value range providers
2. With JPA entities that implement `equals()` by ID
3. Entities are persisted in one transaction and loaded in another (different object instances, same ID)

## Fix Status

This was already fixed in PR #2111 ("refactor: value range cleanup"), but the fix wasn't explicitly tracked as a bug fix for this scenario. The commit message says:

> "The solver now implicitly trusts entity equals()."

## This PR

Adds a regression test that would have caught this bug in 1.31.0. The test:

1. Creates entities with `.equals()` by ID (different instances, same ID)
2. Uses a planning variable with a value range provider
3. Simulates pre-assignment (canonical range + entity with "different" but equal value)
4. Verifies the solver doesn't throw `IllegalStateException`

## Benefits

- **Documents the specific scenario** that was broken
- **Prevents regression** if someone reintroduces identity-based checks
- **Serves as documentation** for users with similar use cases (JPA + filtering VRPs)

## Related

- Issue #2172: https://github.com/TimefoldAI/timefold-solver/issues/2172
- PR #2111: https://github.com/TimefoldAI/timefold-solver/pull/2111
- Commit: https://github.com/TimefoldAI/timefold-solver/commit/2f99a1d72b

### DIFF
--- a/core/src/test/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeEqualsVsIdentityRegressionTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/domain/valuerange/ValueRangeEqualsVsIdentityRegressionTest.java
@@ -1,0 +1,212 @@
+package ai.timefold.solver.core.impl.domain.valuerange;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.List;
+import java.util.ArrayList;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
+import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.api.domain.valuerange.ValueRangeProvider;
+import ai.timefold.solver.core.config.solver.SolverConfig;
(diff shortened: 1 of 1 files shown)


paoloantinori:
Hi there. As you can guess this has been AI-supported.
I just wanted to add some fully human edited context:

I've stumbled on the bug reported, using the released version. It throws a proper exception. We have put together a fix and a corresponding unit test to show the problem in action.

We've noticed that the fix is already in place, despite being tracked more generically as a refactor. All good, we will wait for the next release. But we still see value in adding the original unit test we had prepared contextually to the fix, to avoid eventual future regressions.

HTH
Paolo

triceo:
Hello, and thanks for the PR - as you noticed, the fix is already available.

As to the test - this is too low-level to be tested by running the entire solver. Also, adding 200 lines of code for this simple thing seems like an overkill.

Furthermore, all sufficient test coverage is already present, as you can verify in the commit. Therefore I will be closing this PR as not necessary.

PR #2178, opened 2026-03-09
chore: 2.0 migration docs and final API cleanup

triceo:


core/src/main/java/ai/timefold/solver/core/api/solver/phase/PhaseCommandContext.java
```diff
@@ -65,30 +65,41 @@ public interface PhaseCommandContext<Solution_>
...
+     * @param move the move to execute temporarily
+     * @return the score of the temporarily modified solution after executing the move
+     */
+    <Score_ extends Score<Score_>> Score_ executeTemporarily(Move<Solution_> move);
+
+    /**
+     * As defined by {@link #executeTemporarily(Move)},
+     * with the guarantee of a fresh score at the end of the method's invocation.
+     */
+    <Score_ extends Score<Score_>> Score_ executeTemporarilyAndCalculateScore(Move<Solution_> move);
```
zepfred, line 95: This method will be helpful for the FSR migration.

core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultPhaseCommandContext.java
```diff
@@ -57,8 +57,30 @@ public <Score_ extends Score<Score_>> Score_ executeAndCalculateScore(Move<Solut
...
     @Override
     public @Nullable <Result_> Result_ executeTemporarily(Move<Solution_> move,
-            Function<Solution_, @Nullable Result_> temporarySolutionConsumer, boolean guaranteeFreshScore) {
-        return moveDirector.executeTemporary(move, temporarySolutionConsumer, guaranteeFreshScore);
+            Function<Solution_, @Nullable Result_> temporarySolutionConsumer) {
+        return moveDirector.executeTemporary(move, temporarySolutionConsumer, false);
+    }
+
+    @Override
+    public <Score_ extends Score<Score_>> Score_ executeTemporarily(Move<Solution_> move) {
```
zepfred, line 65: It appears that we won't call the score calculation logic in this call, as it will invoke the previous method, which sets guaranteeFreshScore to `false`.

tools/migration/src/main/java/ai/timefold/solver/migration/v2/GeneralTypeChangeMigrationRecipe.java
```diff
@@ -65,69 +65,6 @@ public List<Recipe> getRecipeList() {
                         "ai.timefold.solver.core.impl.domain.valuerange.CompositeValueRange", true),
                 new ChangeType(
                         "ai.timefold.solver.core.impl.domain.valuerange.buildin.composite.NullAllowingCountableValueRange",
-                        "ai.timefold.solver.core.impl.domain.valuerange.NullAllowingValueRange", true),
-                // Move API
-                new ChangeType("ai.timefold.solver.core.impl.heuristic.move.Move",
```
zepfred, line 70: Perhaps we should maintain the migration of the `Move` class.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/migration-guides/chained-variables-to-planning-list-variable.adoc
```diff
@@ -0,0 +1,196 @@
+[#migrateChainedToPlanningListVariable]
+= Chained planning variable to planning list variable
+:doctype: book
+:icons: font
+
+This section explains how to update your planning model to use
+the declarative xref:using-timefold-solver/modeling-planning-problems.adoc#planningListVariable[planning list variable] approach
```
zepfred: I'm uncertain about using the term "declarative" in this context. I understand it refers to a declarative variable, but we usually use it to shadow variables.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/migration-guides/chained-variables-to-planning-list-variable.adoc
```diff
@@ -0,0 +1,196 @@
...
+instead of the chained planning variable,
+which has been deprecated in Timefold Solver 1.x and removed entirely in Timefold Solver 2.0.
+
+Chained planning variables allowed planning entities to point to each other and form chains,
+which was used for sequence-based problems like TSP and vehicle routing.
+The planning list variable provides a modern,
+more intuitive way to model sequences as lists of values on a parent entity.
+
+If you modeled vehicle routing, TSP, or any ‘ordered route’ problem using chained planning variables,
```
zepfred:
```suggestion
If you modeled vehicle routing, TSP, or any ‘ordered sequece’ problem using chained planning variables,
```

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
```diff
@@ -0,0 +1,893 @@
...
+
+=== Conceptual changes
+
+These changes are most likely to require you to change your code in a non-mechanical way,
+and they can only be automated partially, if at all.
+
+.icon:exclamation-triangle[role=red] Java 21 or later required
+[%collapsible%open]
+====
+Timefold Solver 2.0.0 requires Java 21 or later to run,
```
zepfred:
```suggestion
Timefold Solver 2.x requires Java 21 or later to run,
```

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
```diff
@@ -0,0 +1,893 @@
...
+[%collapsible%open]
+====
+`BendableScore` and `BendableLongScore` have been merged into a single `BendableScore` backed by `long[]`.
+`BendableScore.of(int[], int[])` is now `BendableScore.of(long[], long[])`.
+
+Before in `*.java`:
+
+[source,java]
+----
+BendableScore score = BendableScore.of(new int[]{0, -1}, new int[]{-2});
```
zepfred, line 624: I like this example. Why are we only doing this for bendable scores?

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
```diff
@@ -0,0 +1,893 @@
...
+Use `penalize`, `reward`, and `impact` respectively; they now also accept `long` weights.
+====
+
+'''
+
+.icon:info-circle[role=yellow] Deprecated `SolverManager` `solve` methods removed
+[%collapsible%open]
+====
+Various previously deprecated overloads of `SolverManager.solve()` and `SolverManager.solveAndListen()` were removed.
+Use the `SolverManager.solveBuilder()` method to build `SolverJob`s instead, as it provides more flexibility.
```
zepfred:
```suggestion
Use the `SolverManager.solveBuilder()` method to build `SolverJob` instead, as it provides more flexibility.
```

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
```diff
@@ -0,0 +1,893 @@
...
+* Same renames apply to `ConstraintMatch` and `ConstraintMatchTotal`.
+
+==== Constraint Stream Terminal Methods Refactored
+
+* All `penalize(constraintName, score)`, `reward(constraintName, score)`, `impact(constraintName, score)` forms
+(and their `Configurable`/`Long`/`BigDecimal` variants, across Uni/Bi/Tri/Quad streams)
+are replaced by a two-call chain:
+`penalize(score).asConstraint(constraintName)`, moving the constraint name out of the scoring method and into a dedicated
+`asConstraint()` call.
+* Where the old form passed both a package and a name (`penalize(pkg, name, score)`), the package and name are merged into a single string for `asConstraint()`: `penalize(score).asConstraint(pkg + "." + name)`.
```
zepfred: Should we continue using the `pkg` variable in the `asConstraint()` function?

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
```diff
@@ -0,0 +1,893 @@
...
+* `ConstraintBuilder.asConstraint(packageName, constraintName)` → `asConstraint(packageName + "." + constraintName)`
```
zepfred: Should we add an example?

zepfred (approved):
The documentation is very detailed. LGTM!

core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultPhaseCommandContext.java
triceo: Correct. That is this method's contract, and that's what it does.

tools/migration/src/main/java/ai/timefold/solver/migration/v2/GeneralTypeChangeMigrationRecipe.java
triceo: I considered this. Since the two types are very different, this recipe would be guaranteed to produce code that doesn't compile. This is not acceptable, as it prevents the rest of the migration scripts from going forward.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/upgrade-from-v1.adoc
triceo: Because non-bendable scores do not have this issue. If you pass an int to a long, it just works; not the case with arrays.

core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultPhaseCommandContext.java
zepfred: Does it make sense to return the score?

PR #2183, opened 2026-03-11
fix: provide a more accurate problem scale log when entity dependent value ranges are used for list variables

### LINKED ISSUE #2075: ValueRangeProvider: inconsistent Problem Scale reporting.
**Describe the bug**
I have used valueRangeProvider to reduce the problem scale, but this isn't shown in that metric.

Context:

- 240 vehicles (10 vehicles, planned for 24 days, so i made separate entities for each day) 
- 6874 visits

```
WITH ValueRangeProvider on ENTITY (soft filter): 
Problem scale: entity count (240), variable count (240), approximate value count (1414344), approximate problem scale (1,210169 ? 10^23848).

WITH ValueRangeProvider on ENTITY (more in depth filter): 
Problem scale: entity count (240), variable count (240), approximate value count (341902), approximate problem scale (1,210166 ? 10^23848).

WITH ValueRangeProvider on SOLUTION: 
Problem scale: entity count (240), variable count (240), approximate value count (6874), approximate problem scale (1,210166 ? 10^23848)
```

As far as i grasp it, on Entity (soft/depth) has 1414344/341902 possible values. (the sum of all vehicle value ranges)
But on the solution, it should be 240 vehicles * 6874 visits = 1649520, which is a lot higher.

Shouldn't the approximate problem scale with ValueRangeProvider on Entity be smaller than the one on Solution?

**Expected behavior**

I would expect that the approximate problem scale for ValueRangeProvider on Entity would be _smaller_ , not _larger_ than the one on Solution.

**Actual behavior**

Problem scale is larger when placed on Entity, which doesn't make sense because that one is filtered on which values would be allowed on which vehicles. 

**To Reproduce**
<reproducer on demand>

## Environment

**Timefold Solver Version or Git ref**: 1.30

**Output of `java -version`:** 21

## Additional information

Provide any and all other information which might be relevant to the issue.

Christopher-Chianelli:
Previously, the logic just assumed the values from entity dependent value ranges are unique, which massively inflated the problem scale log for list variables.

Now, we track the number of entities a value is allowed to be in, and use that to scale the problem scale. For instance, if a value is only allowed to be in 25% of entities, it should reduce the problem scale by 25%.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/ProblemScaleTracker.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/ProblemScaleTracker.java
@@ -4,69 +4,29 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.Set;
 
+import ai.timefold.solver.core.api.domain.valuerange.ValueRange;
+import ai.timefold.solver.core.impl.score.director.ListValueRangeStatistics;
 import ai.timefold.solver.core.impl.util.MathUtils;
 
 public class ProblemScaleTracker {
     private final long logBase;
     private final Set<Object> visitedAnchorSet = Collections.newSetFromMap(new IdentityHashMap<>());
-
+    private final ListValueRangeStatistics listValueRangeStatistics = new ListValueRangeStatistics();
     private long basicProblemScaleLog = 0L;
-    private int listPinnedValueCount = 0;
-    private int listTotalEntityCount = 0;
-    private int listMovableEntityCount = 0;
-    private int listTotalValueCount = 0;
+    private boolean listAllowsUnassignedValues = false;
 
     public ProblemScaleTracker(long logBase) {
         this.logBase = logBase;
     }
 
     // Simple getters
-    public long getBasicProblemScaleLog() {
-        return basicProblemScaleLog;
+    public long getProblemScaleLog() {
+        return basicProblemScaleLog + listValueRangeStatistics.computeListProblemScaleLog(listAllowsUnassignedValues, logBase);
     }
 
-    public int getListPinnedValueCount() {
-        return listPinnedValueCount;
-    }
-
-    public int getListTotalEntityCount() {
-        return listTotalEntityCount;
-    }
-
-    public int getListMovableEntityCount() {
-        return listMovableEntityCount;
-    }
-
-    public int getListTotalValueCount() {
-        return listTotalValueCount;
-    }
-
-    public void setListTotalValueCount(int listTotalValueCount) {
-        this.listTotalValueCount = listTotalValueCount;
-    }
-
-    // Complex methods
-    public boolean isAnchorVisited(Object anchor) {
-        if (visitedAnchorSet.contains(anchor)) {
-            return true;
-        }
-        visitedAnchorSet.add(anchor);
-        return false;
-    }
-
-    public void addListValueCount(int count) {
-        listTotalValueCount += count;
-    }
-
-    public void addPinnedListValueCount(int count) {
-        listPinnedValueCount += count;
-    }
-
-    public void incrementListEntityCount(boolean isMovable) {
-        listTotalEntityCount++;
-        if (isMovable) {
-            listMovableEntityCount++;
-        }
+    public void processListValueRange(boolean listAllowsUnassignedValues, ValueRange<?> valueRange) {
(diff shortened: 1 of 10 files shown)


triceo:
You may not be aware, but since not so long ago, the solver actually has a new component - `ValueRangeManager`. It already does a lot of heavy lifting around value ranges, and I think it could easily be extended to providing the information for the scale computation.

The core idea here being that these loops over all entities and values aren't cheap, and it'd be best if they were performed just once. Can you please look into integrating this with the `ValueRangeManager`, so that the amount of iteration over large collections can be minimized?

Christopher-Chianelli:
Are you suggesting combining `computeInitializationStatistics` and `getProblemScale` in `ValueRangeStatistics`? The `ValueRangeManager` already manages problem scale via `ProblemSizeStatistics`.

triceo:
Looking at the code, I see `ValueRangeManager.computeInitializationStatistics` already does a loop over all entities, so it could IMO make sense to use that loop to prepare the data for the problem scale as well. 

More importantly, though - `ValueRangeState` already holds all the values of all value ranges enumerated. Which means there already was a loop which created all of this data. And we should hook into _that_ loop. Maybe we can update `ValueRangeState` to hold the value counts. See this:

    private @Nullable Map<Value_, Integer> fromSolutionValueIndexMap;
    private @Nullable Map<Entity_, ValueRangeItem<Solution_, Entity_, ValueRange<Value_>, Value_>> fromEntityMap;

The value in these maps can be extended to carry the information about value count, or perhaps another map added. And with that minimal update, we have all the information we need to create the problem scale when asked for it.

Here's the thing: problem scale is exclusively used to write information to log output. It has no business being expensive. It would be ridiculous if the solver had to be held back by this. If we now have to enumerate all value ranges to make the computation correct, then let's hook into the place which already does it, as opposed to duplicating those expensive loops.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/ProblemScaleTracker.java
```diff
@@ -1,72 +1,25 @@
...
 
-    public ProblemScaleTracker(long logBase) {
+    public ProblemScaleTracker(ListVariableDescriptor<Solution_> listVariableDescriptor,
+            ValueRangeManager<Solution_> valueRangeManager,
+            long logBase) {
         this.logBase = logBase;
+        this.listValueRangeStatistics = new ListValueRangeStatistics<>(listVariableDescriptor, valueRangeManager);
     }
 
     // Simple getters
```
triceo: Arguably not a simple getter.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/ProblemScaleTracker.java
```diff
@@ -1,72 +1,25 @@
...
-        listPinnedValueCount += count;
-    }
-
-    public void incrementListEntityCount(boolean isMovable) {
-        listTotalEntityCount++;
-        if (isMovable) {
-            listMovableEntityCount++;
-        }
+    public long getProblemScaleLog() {
+        return basicProblemScaleLog + listValueRangeStatistics.computeListProblemScaleLog(logBase);
```
triceo: Are we sure this will only be called once? Otherwise we're spending quite a lot of time recomputing the scale log over and over. In the spirit of defensive programming, I'd just cache it here.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ListValueRangeStatistics.java
```diff
@@ -0,0 +1,51 @@
...
+        var valueSet = reachableValues.extractAllValuesAsSet();
+        var valueCount = valueSet.size();
+        var validPercentageLog = 0L;
+        var additionalCount = allowsUnassignedValues ? 1 : 0;
+
+        for (var value : valueSet) {
+            validPercentageLog += MathUtils.getScaledApproximateLog(MathUtils.LOG_PRECISION, logBase,
+                    ((double) reachableValues.extractEntitiesAsList(value).size() + additionalCount) / entityCount);
+        }
+
```
triceo, line 47: This is still very expensive. The entity list will be created for every value in the valueSet, only so that we can find its size. Surely we can find its size without creating it.

triceo (commented):
Already much better, but I think we can still improve.

triceo (approved):
LGTM when green.

PR #2200, opened 2026-03-21
fix: Handle null planning values in ValueTabuAcceptor

### LINKED ISSUE #2199: NullPointerException in ValueTabuAcceptor with nullable planning variables
### Summary

ValueTabuAcceptor throws `NullPointerException` when using Value Tabu (`<valueTabuSize>` or `<propertyTabuSize>`) with planning problems that have **nullable planning variables**.

### Version

Timefold Solver 1.31.0 (also affects latest main branch)

### Steps to Reproduce

1. Create a planning problem with nullable planning variables
2. Configure solver to use Value Tabu acceptor:
   ```xml
   <acceptor>
     <valueTabuSize>100</valueTabuSize>
   </acceptor>
   ```
3. During local search, when a move's `getPlanningValues()` returns a collection containing `null` (representing unassigned planning variables), the acceptor crashes

### Root Cause

In `AbstractTabuAcceptor.adjustTabuList()` (line 118):
```java
tabuSequenceDeque.add(tabu);  // NPE when tabu is null
```

The `ValueTabuAcceptor.findNewTabu()` returns `getPlanningValues()` which can contain null values for nullable planning variables. `ArrayDeque.add()` throws `NullPointerException` when trying to add null elements.

### Stack Trace

```
java.lang.NullPointerException: Cannot invoke "ArrayDeque.addLast(Object)" because "deque" is null
    at java.base/java.util.ArrayDeque.addLast(ArrayDeque.java:192)
    at java.base/java.util.ArrayDeque.add(ArrayDeque.java:305)
    at ai.timefold.solver.core.impl.localsearch.decider.acceptor.tabu.AbstractTabuAcceptor.adjustTabuList(AbstractTabuAcceptor.java:118)
    at ai.timefold.solver.core.impl.localsearch.decider.acceptor.tabu.AbstractTabuAcceptor.stepEnded(AbstractTabuAcceptor.java:97)
```

### Expected Behavior

The acceptor should skip null planning values (unassigned state) rather than crashing. Tracking "unassigned" as tabu has no meaningful purpose.

### Workaround

Use Entity Tabu instead of Value Tabu for problems with nullable planning variables.

### Proposed Fix

Filter out null values in `adjustTabuList()` before adding to the deque:

```java
// Add the new tabu(s)
for (var tabu : tabus) {
    // Skip null planning values (unassigned state)
    if (tabu == null) {
        continue;
    }
    // ... rest of processing
}
```

**Rationale:**
- Null values in planning variables represent "unassigned" state
- Tabu tracking of "unassigned" has no meaningful purpose
- The same null value would be indistinguishable from other null values anyway

paoloantinori:
Fixes #2199

### Summary

ValueTabuAcceptor throws `NullPointerException` when processing moves with nullable planning variables that contain null values. The null values represent "unassigned" state and cannot be tracked in the tabu deque.

### Root Cause

`AbstractTabuAcceptor.adjustTabuList()` tried to add null elements directly to `ArrayDeque`, which throws `NullPointerException`.

### Fix

Skip null values in `adjustTabuList()` before adding to tabu tracking structures. Null planning values represent "unassigned" state and tracking them as tabu has no meaningful purpose.

### Test

Added `nullablePlanningValue()` test in `ValueTabuAcceptorTest` which verifies:
- Moves with null planning values don't throw NPE
- Null values are skipped (not tracked as tabu)
- Non-null values are correctly tracked as tabu

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/AbstractTabuAcceptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/AbstractTabuAcceptor.java
@@ -109,6 +109,10 @@ public abstract class AbstractTabuAcceptor<Solution_> extends AbstractAcceptor<S
         }
         // Add the new tabu(s)
         for (var tabu : tabus) {
+            // Skip null planning values (unassigned state)
+            if (tabu == null) {
+                continue;
+            }
             // Push tabu to the end of the line
             if (tabuToStepIndexMap.containsKey(tabu)) {
                 tabuToStepIndexMap.remove(tabu);
--- a/core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/ValueTabuAcceptorTest.java
+++ b/core/src/test/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/tabu/ValueTabuAcceptorTest.java
@@ -4,7 +4,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 import ai.timefold.solver.core.api.score.SimpleScore;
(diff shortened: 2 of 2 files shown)


triceo:
Closing. See rationale here:
https://github.com/TimefoldAI/timefold-solver/issues/2199#issuecomment-4127888913

PR #2214, opened 2026-03-27
perf: improve runtime performance of exhaustive search

triceo:


core/src/main/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirector.java
```diff
@@ -66,9 +67,13 @@ private VariableChangeRecordingScoreDirector(@Nullable InnerScoreDirector<Soluti
...
-    @SuppressWarnings("unchecked")
-    public List<ChangeAction<Solution_>> copyChanges() {
-        return List.copyOf(variableChanges);
+    public Move<Solution_> createUndoMove() {
+        // The list would normally be copied here to prevent any outside modification.
+        // However, copying this list on the hot path would be a major performance issue.
+        // Instead, the list is passed as a reference here, and instead of it being cleared by undoChanges(),
+        // the reference is replaced; that way, the move does not actually share the list with anyone,
+        // and copying of its contents can be avoided.
+        return new RecordedUndoMove<>(variableChanges);
```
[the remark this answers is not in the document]
triceo: Incorrect. The original code also doesn't reverse.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
```diff
@@ -102,35 +101,56 @@ public boolean isEntityReinitializable(Object entity) {
...
-            assignNode = assignNode.getParent();
-        }
-        Collections.reverse(assignMoveList);
-        var allMoves = new ArrayList<Move<Solution_>>(unassignMoveList.size() + assignMoveList.size());
-        allMoves.addAll(unassignMoveList);
-        allMoves.addAll(assignMoveList);
-        if (allMoves.isEmpty()) {
-            // No moves to restore, so the working solution is already correct
+        //First, undo all previous changes.
+        var unassignMoves = listAllUndoMoves(phaseScope.getLastCompletedStepScope().getExpandingNode());
```
zepfred, line 105: The while loop logic seems better the current recursive method. Can we determine the size of the undo move array based on the layer of the last completed step? If we can know the size in advance, we can iterate just like the previous while loop.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
```diff
@@ -102,35 +101,56 @@ public boolean isEntityReinitializable(Object entity) {
...
+        // Next, rebuild the solution starting from the current search element.
+        var assignMoves = listAllMovesInReverseOrder(stepScope.getExpandingNode());
```
zepfred, line 107: The same applies here. I prefer using a while loop rather than a recursive method.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/node/comparator/OptimisticBoundFirstNodeComparator.java
```diff
@@ -31,26 +31,18 @@ public int compare(ExhaustiveSearchNode<Solution_> a, ExhaustiveSearchNode<Solut
         Score aScore = a.getScore().raw();
```
zepfred, line 31: Let's make it consistent with the other comparators by keeping only the variable `scoreComparison`.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/node/comparator/ScoreFirstNodeComparator.java
```diff
@@ -24,33 +24,23 @@ public int compare(ExhaustiveSearchNode<Solution_> a, ExhaustiveSearchNode<Solut
         Score aScore = a.getScore().raw();
```
zepfred, line 24: Let's make it consistent with the other comparators by keeping only the variable `scoreComparison`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementDestinationSelector.java
```diff
@@ -136,27 +135,27 @@ public Iterator<ElementPosition> iterator() {
             }
             if (isExhaustiveSearch) {
                 // The exhaustive search method requires elements to be placed only at the end of the entity's list
-                Stream<ElementPosition> stream = StreamSupport.stream(entitySelector.spliterator(), false)
-                        .map(entity -> ElementPosition.of(entity, listVariableDescriptor.getListSize(entity)));
-                return stream.iterator();
+                return new MappingIterator<>(entitySelector.iterator(),
```
zepfred, line 138: Cool!

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/CompositeMove.java
```diff
@@ -26,82 +27,81 @@
...
                     .formatted(CompositeMove.class.getSimpleName()));
-            case 1 -> moves[0];
-            default -> new CompositeMove<>(moves);
+            case 1 -> moveList.getFirst();
+            default -> new CompositeMove<>(moveList);
         };
     }
 
-    private final Move<Solution_>[] moves;
+    private final List<? extends Move<Solution_>> moveList;
```
zepfred, line 42: Why are we changing the type to `List`? The previous logic seems better as we don't need to instantiate the `ArrayList`.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/node/comparator/OptimisticBoundFirstNodeComparator.java
triceo: It is consistent with every other node comparator.
I am not inlining these variables as the chain would be too long to be readable.

    a.getScore().raw().compareTo(b.getScore().raw())

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/node/comparator/ScoreFirstNodeComparator.java
triceo: Explained above.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/CompositeMove.java
triceo: Turning a list to an array requires `list.toArray()`, which copies the entire list to an array. Requiring a list instead does not force any such conversions; `Arrays.asList(...)` just wraps the underlying array. 

Turning a list to an array has a cost. Turning an array to a list doesn't.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
triceo: Sizing the array is not the main problem. Filling the array in reverse order is only possible with recursion; otherwise there `reverse()` call still needs to be there.

I argue recursion is not a problem here. The exhaustive algorithm will run out of heap much sooner than it runs out of stack. (I ran an experiment with TSP. 12 visits took 20 seconds, 14 visits already struggled for memory.)

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
triceo: See above.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/CompositeMove.java
zepfred: I don't understand. Where are we turning the list into an array?

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
zepfred: I understand your point. The ES method isn't suitable for some problems. However, I believe we can use the while loop and the array type to create the move list, which simplifies the logic.

zepfred (approved):
I'm going to approve the PR, but the question about the changes on the composite move remains unclear to me.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/CompositeMove.java
triceo: There are 2 ways how to construct the composite move:

1. From a list of moves.
2. From an array of moves.

There are 2 ways how to store the moves inside of the composite move. If you store them as a list, you may have to convert a move array to a move list. This is simple to do with `Arrays.asList`.

However, if you store them as array and you're given a list, you have to do `list.toArray`, and that is expensive.

Therefore storing as a list is preferred, as that has cheap cases for both 1 and 2.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/decider/ListVariableExhaustiveSearchDecider.java
triceo: Not sure how you want me to do that, other than to manually reverse the array at the end.

PR #2219, opened 2026-03-31
feat: improve unassign move generation

zepfred:
The list change move selector may be prematurely removed from the union move selector iterator during a given step when it picks a random value and all reachable entities from that value are pinned, leaving no valid entities available.

We are not verifying whether any values are assigned, so at least one valid destination could be generated. This PR updates the `hasNext` logic of the iterator to address that. One downside of configuring the list change selector to return true whenever there are any assigned values is that it can generate no-change moves, which waste CPU cycles. This situation can occur if the planning value is unassigned and all of its reachable entities are pinned.

Some experiments found no regressions from this change in the FSR.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
@@ -22,6 +22,7 @@ public abstract class UpcomingSelectionIterator<S> extends SelectionIterator<S> 
 
     protected boolean upcomingCreated = false;
     protected boolean hasUpcomingSelection = true;
+    private boolean recheckUpcomingSelection = false;
     protected S upcomingSelection;
 
     @Override
@@ -29,6 +30,11 @@ public abstract class UpcomingSelectionIterator<S> extends SelectionIterator<S> 
         if (!upcomingCreated) {
             upcomingSelection = createUpcomingSelection();
             upcomingCreated = true;
+            if (recheckUpcomingSelection) {
+                // We ensure that the variable remains consistent if "discardUpcomingSelection" was called previously.
+                hasUpcomingSelection = upcomingSelection != null;
+                recheckUpcomingSelection = false;
+            }
         }
         return hasUpcomingSelection;
     }
@@ -54,8 +60,7 @@ public abstract class UpcomingSelectionIterator<S> extends SelectionIterator<S> 
 
     public void discardUpcomingSelection() {
         upcomingCreated = false;
-        upcomingSelection = null;
-        hasUpcomingSelection = true;
+        recheckUpcomingSelection = true;
     }
 
     @Override
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
@@ -54,6 +54,8 @@ public abstract class UpcomingSelectionIterator<S> extends SelectionIterator<S> 
 
     public void discardUpcomingSelection() {
         upcomingCreated = false;
+        upcomingSelection = null;
+        hasUpcomingSelection = true;
     }
 
     @Override
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
@@ -41,6 +41,12 @@ public abstract class UpcomingSelectionIterator<S> extends SelectionIterator<S> 
 
     @Override
     public S next() {
+        if (recheckUpcomingSelection) {
+            upcomingSelection = createUpcomingSelection();
+            // We ensure that the variable remains consistent if "discardUpcomingSelection" was called previously.
+            hasUpcomingSelection = upcomingSelection != null;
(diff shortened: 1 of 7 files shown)


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
```diff
@@ -54,6 +54,8 @@ protected S noUpcomingSelection() {
 
     public void discardUpcomingSelection() {
         upcomingCreated = false;
+        upcomingSelection = null;
+        hasUpcomingSelection = true;
```
triceo: How can you tell that this should be `true`? What if there isn't any?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java
```diff
@@ -44,6 +45,8 @@ public ElementPositionRandomIterator(ListVariableStateSupply<Solution_, Object,
                     .formatted(totalSize));
         }
         this.allowsUnassignedValues = allowsUnassignedValues;
+        this.maybeAssignedMovableValues =
+                allowsUnassignedValues && (valueSelector.getSize() - listVariableStateSupply.getUnassignedCount()) > 0;
```
triceo: Sizes are notoriously unreliable. They are likely to be less than what `getSize()` says - which means that `maybeAssignedMovableValues` will likely always be `true`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
```diff
@@ -54,6 +54,8 @@ protected S noUpcomingSelection() {
 
     public void discardUpcomingSelection() {
```
triceo, line 63: This method is being called from exactly one place. I wonder if this is the best design; only one other thing needs this, yet the class exposes this for everyone.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
zepfred: The value defaults to `true` and is set to `false` if `noUpcomingSelection()` is called. When discarding the upcoming value, we need to ensure the variable is reset, or we might end up with a valid upcoming value but return `false` from `hasNext()`.

zepfred: I have revised the logic to address the copilot comment.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java
zepfred: I don't think so. We're using the `valueSelector` to get the size, which will use the value range manager to fetch all planning values without duplicates (`IterableFromSolutionPropertyValueSelector`). In that case, the size is reliable; otherwise, it would lead to an inconsistent state in the list variable supply when counting unassigned elements and skew the probabilities in `ElementPositionRandomIterator`.

I moved the variable definition to the selector to simplify the logic.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/common/iterator/UpcomingSelectionIterator.java
zepfred: I see your point, and thought about it. Let's recall why we added it in the first place. We encountered a problem in which the destination iterator selected a new value, but the entity iterator was not refreshed to return only reachable entities for that value. That happened because a filtering iterator created by pinning settings was on top of the entity iterator. The filtering iterator implements `UpcomingSelectionIterator`.

My main argument for keeping it is that the move selector has many combinations, and it can be hard to anticipate all of them. Making logic part of the base contract of `UpcomingSelectionIterator` maximizes the coverage of the combination addressed by this specific fix. IMO, the method does not violate the class contract because it simply adds an additional related feature.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java
```diff
@@ -88,6 +96,9 @@ public boolean hasNext() {
 
     @Override
     public ElementPosition next() {
+        if (!hasNextValue) {
```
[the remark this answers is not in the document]
zepfred: No, `hasNextValue` is always set in `hasNext`, and it must be used instead of calling `hasNext`.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementPositionRandomIterator.java
```diff
@@ -72,9 +75,16 @@ private boolean hasNextValue() {
...
+            // or if there is at least one non-pinned assigned value,
+            // which would result in an unassigning move.
+            return entityIterator.hasNext() || maybeMovableValues;
         }
-        return selectedValue != null && entityIterator.hasNext();
+        // There will be a valid destination if the entity iterator has a next element
+        // or if there is at least one non-pinned assigned value,
+        // which would result in an unassigning move.
+        return selectedValue != null
+                && (entityIterator.hasNext() || maybeMovableValues);
```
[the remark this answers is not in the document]
zepfred: The comment is correct.

triceo (approved):

PR #2220, opened 2026-04-02
chore: remove an outdated TODO

triceo:
We will not be changing the value to 2, because switching sublists of length 1 with another sublists of a greater length is still a valuable move.


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/SubListSelectorFactory.java
```diff
@@ -21,7 +21,7 @@
 
 public final class SubListSelectorFactory<Solution_> extends AbstractFromConfigFactory<Solution_, SubListSelectorConfig> {
 
-    private static final int DEFAULT_MINIMUM_SUB_LIST_SIZE = 1; // TODO Bump this to 2 in Timefold Solver 2.0
+    private static final int DEFAULT_MINIMUM_SUB_LIST_SIZE = 1;
```
zepfred, line 24: I believe maintaining it as `1` is the right choice. The minimum value of `1` still applies in certain cases, such as specific neighborhoods used by HGS.

zepfred (approved):

PR #2234, opened 2026-04-14
chore: make constraint metadata more flexible

triceo:
Enterprise counterpart:
https://github.com/TimefoldAI/timefold-solver-enterprise/pull/517


core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintBuilder.java
```diff
@@ -9,40 +10,23 @@ public interface ConstraintBuilder {
...
-    /**
-     * As defined by {@link #asConstraintDescribed(String, String, String)},
-     * placing the constraint in the {@link Constraint#DEFAULT_CONSTRAINT_GROUP default constraint group}.
-     *
-     * @param constraintName shows up in {@link ScoreAnalysis}
-     * @param constraintDescription can contain any character, but it is recommended to keep it short and concise.
-     */
-    default Constraint asConstraintDescribed(String constraintName, String constraintDescription) {
-        return asConstraintDescribed(constraintName, constraintDescription, Constraint.DEFAULT_CONSTRAINT_GROUP);
+    default Constraint asConstraint(String id) {
```
zepfred, line 17: Let's also add an default method with id and description since the models use the old `asConstraintDescribed`.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintBuilder.java
```diff
@@ -9,40 +10,23 @@ public interface ConstraintBuilder {
...
      *
-     * @param constraintName shows up in {@link ScoreAnalysis}
-     * @param constraintDescription can contain any character, but it is recommended to keep it short and concise.
-     * @param constraintGroup not used by the solver directly, but may be used by external tools to group constraints together,
-     *        such as by their source or by their purpose
+     * @param description identifies and describes the constraint;
+     *        {@link ConstraintDescription#id()} shows up in {@link ScoreAnalysis}
      */
-    Constraint asConstraintDescribed(String constraintName, String constraintDescription, String constraintGroup);
+    Constraint asConstraint(ConstraintDescription description);
```
zepfred: The `ConstraintDescription` class does not include a description method. Is that correct?

zepfred: The default class `DefaultConstraintDescription` does not make sense to me. It is intended to store the constraint description, but lacks the description attribute.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintBuilder.java
triceo: No. The description/group fields are never used by the solver, and as such should have never been there. The purpose of the new mechanism is for people to be able to add their own fields, whatever they need them to be, on their own implementation of `ConstraintDescription`. By default, constraint description only includes the `id`, because that is the only thing that the solver actually needs.

core/src/main/java/ai/timefold/solver/core/api/score/stream/ConstraintBuilder.java
triceo: See the previous comment above.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/DefaultConstraintMetaModel.java
```diff
@@ -16,44 +14,22 @@
...
-            // Constraint groups are an optional feature, therefore most people won't use them,
-            // therefore sizing the list assuming all constraints end up in the default group.
-            perGroupMap.computeIfAbsent(constraint.getConstraintGroup(), k -> new ArrayList<>(constraintCount))
-                    .add(constraint);
-        }
-        return new DefaultConstraintMetaModel(
-                Collections.unmodifiableMap(perRefMap),
-                Collections.unmodifiableMap(perGroupMap));
+        var perRefMap = constraints.stream()
+                .collect(Collectors.toMap(Constraint::getConstraintRef, v -> v, (a, b) -> a, LinkedHashMap::new));
```
zepfred, line 24: We can also use the collector the `toUnmodifiableMap`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -96,7 +96,7 @@ public interface InnerScoreDirector<Solution_, Score_ extends Score<Score_>>
      * Call {@link #calculateScore()} before calling this method,
      * unless that method has already been called since the last {@link PlanningVariable} changes.
      *
-     * @return never null, the key is the constraint name.
+     * @return never null, the key is the constraint ID.
```
zepfred: The key is not the id, but the constraint ref.

tools/migration/src/main/java/ai/timefold/solver/migration/v2/ConstraintMetadataMigrationRecipe.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public List<Recipe> getRecipeList() {
+        return List.of(
+                // asConstraintDescribed(String name, String description) -> asConstraint(String name), drop arg 1
+                new ChangeMethodName(
+                        "ai.timefold.solver.core.api.score.stream.ConstraintBuilder asConstraintDescribed(String, String)",
+                        "asConstraint", true, false),
+                new DeleteMethodArgument(
```
zepfred, line 32: We should keep the description in the constraint and not delete it.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/DefaultConstraintMetaModel.java
triceo: Unfortunately we can't. We need `Linked`, and `toUnmodifiableMap` doesn't let me do that.

tools/migration/src/main/java/ai/timefold/solver/migration/v2/ConstraintMetadataMigrationRecipe.java
triceo: As explained above, keeping the description is against the goal of this PR.

docs/src/modules/ROOT/pages/constraints-and-score/score-calculation.adoc
```diff
@@ -341,16 +325,53 @@ constraint stream.
...
++
+[source,java,options="nowrap"]
+----
+    public record MyConstraintDescription(String id, String displayName) implements ConstraintDescription {
+    }
+
+    private Constraint penalizeEveryShift(ConstraintFactory factory) {
+        return factory.forEach(Shift.class)
+                .penalize(HardSoftScore.ONE_SOFT)
+                .asConstraint(new MyConstraintDescription("penalizeEveryShift", "Penalize every shift"));
```
zepfred: Here is an example that includes a description, and IMO, it should be part of the default contract.

zepfred (changes requested):
The new code improved the cohesion, and the class `ConstraintDescription` is an good solution. However, I don't understand why the class lacks description information. In my opinion, both ID and description should be the key information regarding the constraints.

triceo:
The goal of this PR is for the solver to only expose what the solver needs - and the solver only needs `id`. There are literally zero consumers of the name, the description, or the group. All of these fields were introduced for the platform, and the platform now needs more fields than this; so we agreed that they would get this mechanism instead.

Description will not be on the interface, because there is no need for us to provide any.

zepfred (approved):

PR #2241, opened 2026-04-15
feat: automatically enable automatic node sharing when possible

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/config/score/director/AutomaticNodeSharing.java
```diff
@@ -0,0 +1,7 @@
+package ai.timefold.solver.core.config.score.director;
+
+public enum AutomaticNodeSharing {
```
triceo: If we want to keep the name, then the values should IMO be something like this:

`ENABLED`
`DISABLED`
`ENABLED_IF_POSSIBLE`

`AutomaticNodeSharing.AUTO` seems like it says nothing.

core/src/main/java/ai/timefold/solver/core/config/score/director/ScoreDirectorFactoryConfig.java
```diff
@@ -40,7 +40,7 @@ public final class ScoreDirectorFactoryConfig extends AbstractConfig<ScoreDirect
 
     @XmlJavaTypeAdapter(JaxbCustomPropertiesAdapter.class)
     private Map<String, String> constraintProviderCustomProperties = null;
-    private Boolean constraintStreamAutomaticNodeSharing;
+    private AutomaticNodeSharing constraintStreamAutomaticNodeSharing;
```
triceo: If you don't want this field to be `@Nullable`, make it have a default value.

core/src/main/java/ai/timefold/solver/core/config/score/director/AutomaticNodeSharing.java
triceo: Considering that it's used as an argument in the solver config, then I think changing the name of the type is a better option:

`EnableAutomaticNodeSharing.YES`
`EnableAutomaticNodeSharing.NO`
`EnableAutomaticNodeSharing.AUTO`

spring-integration/spring-boot-autoconfigure/src/main/java/ai/timefold/solver/spring/boot/autoconfigure/TimefoldSolverAutoConfiguration.java
```diff
@@ -261,14 +262,31 @@ private void loadSolverConfig(IncludeAbstractClassesEntityScanner entityScanner,
...
+                        case OFF -> AutomaticNodeSharing.OFF;
+                        case AUTO -> {
+                            if (NativeDetector.inNativeImage()) {
+                                yield AutomaticNodeSharing.OFF;
+                            }
+                            yield AutomaticNodeSharing.AUTO;
+                        }
+                    });
+        } else if (NativeDetector.inNativeImage()) {
+            // Explicitly set it to disabled in a native image if unspecified so
```
triceo, line 280: I wonder if we should `DEBUG` log something at this point.
The expectation is that it is enabled by default. If it's suddenly disabled, there should be a message saying that, and perhaps even why.

triceo (commented):
Overall, I like this, but with comments.
Most importantly, I think the enum needs a naming conversation.

core/src/main/java/ai/timefold/solver/core/config/score/director/AutomaticNodeSharing.java
triceo: Or maybe: `NodeSharing.MANUAL`, `NodeSharing.FORCED`, `NodeSharing.AUTO`.

core/src/main/java/ai/timefold/solver/core/config/score/director/ScoreDirectorFactoryConfig.java
Christopher-Chianelli: The config classes are a mess with `@Nullable/@NullMarked`; basically none of them are `@NullMarked`, and there are many nullable fields without `@Nullable`.

triceo: I know. But do we need to make it worse by continuing that? :-)

Christopher-Chianelli: I made it nullable; I can do the rest of the config classes in a different PR

triceo (approved):
LGTM. Pushed small changes to docs and renamed a test.

PR #2251, opened 2026-04-18
feat(neighborhoods): allow uni pick() and split change moves

triceo:
This PR was largely AI-generated.


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/DefaultMoveStreamFactory.java
```diff
@@ -130,7 +139,7 @@ public <A> UniEnumeratingStream<Solution_, A> forEachUnfiltered(Class<A> sourceC
         if (!variableMetaModel.allowsUnassignedValues()) {
             return (UniEnumeratingStream) forEachDestination(variableMetaModel);
         }
-        var unpinnedEntities = forEach(variableMetaModel.entity().type(), true);
+        var unpinnedEntities = forEach(variableMetaModel.entity().type(), false);
```
zepfred: Is this correct? Setting it to false will make it impossible to perform the unassign move?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/UniOriginalMoveIterator.java
```diff
@@ -0,0 +1,61 @@
...
+    public UniOriginalMoveIterator(UniMoveStreamContext<Solution_, A> context) {
+        this.context = Objects.requireNonNull(context);
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (nextMove != null) {
+            return true;
+        }
+        if (tupleIterator == null) {
```
zepfred: Why not declare `tupleIterator` as final and initialize it in the constructor?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/DefaultMoveStreamFactory.java
zepfred: The `DefaultNeighborhoodProvider` now includes two separate providers: one for assignments and another for unassignments. Perhaps this is why we are not generating this specific move, but the name and documentation of this function are no longer consistent.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodProvider.java
```diff
@@ -31,9 +33,17 @@ public Neighborhood defineNeighborhood(NeighborhoodBuilder<Solution_> builder) {
...
                 } else if (variableMetaModel instanceof PlanningVariableMetaModel<Solution_, ?, ?> basicVariableMetaModel) {
                     hasBasicVariable = true;
                     builder.add(new ChangeMoveProvider<>(basicVariableMetaModel));
+                    if (basicVariableMetaModel.allowsUnassigned()) {
+                        builder.add(new AssignMoveProvider<>(basicVariableMetaModel));
+                        builder.add(new UnassignMoveProvider<>(basicVariableMetaModel));
+                    }
                 }
             }
             if (hasBasicVariable) {
```
zepfred, line 52: Why is the swap move included only at this point? It is unclear why it is not added to the related if block.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/AssignMoveProvider.java
```diff
@@ -0,0 +1,51 @@
...
+                    "The variableMetaModel (%s) must allow unassigned values, but it does not."
+                            .formatted(variableMetaModel));
+        }
+    }
+
+    @Override
+    public MoveStream<Solution_> build(MoveStreamFactory<Solution_> moveStreamFactory) {
+        var nodeSharingSupportFunctions = ((DefaultMoveStreamFactory<Solution_>) moveStreamFactory)
+                .getNodeSharingSupportFunctions(variableMetaModel);
+        var unassignedEntities = moveStreamFactory.forEach(variableMetaModel.entity().type(), false)
```
zepfred, line 43: Why not create a node-sharing function for `view.getValue(variableMetaModel, e) == null`?

zepfred: In reviewing the `MoveStreamFactory` contract, I noticed some building blocks for list variables. I believe we should consider standardizing these building blocks for retrieving unpinned and assigned/unassigned values for both variable types. This could be a common requirement for users who are building custom moves.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/AssignMoveProviderTest.java
```diff
@@ -0,0 +1,82 @@
...
+    void constructorRejectsNonUnassignedVariable() {
+        var solutionMetaModel = TestdataSolution.buildMetaModel();
+        var variableMetaModel = solutionMetaModel.genuineEntity(TestdataEntity.class)
+                .basicVariable();
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> new AssignMoveProvider<>(variableMetaModel));
+    }
+
+    @Test
+    void fromSolution() {
```
zepfred, line 115: We should also include a test case that uses value ranges.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveProviderTest.java
```diff
@@ -32,48 +32,32 @@ void fromSolution() {
 
         var solution = TestdataSolution.generateSolution(2, 2);
         var firstEntity = solution.getEntityList().get(0);
-        firstEntity.setValue(null);
         var secondEntity = solution.getEntityList().get(1);
-        secondEntity.setValue(null);
         var firstValue = solution.getValueList().get(0);
         var secondValue = solution.getValueList().get(1);
+        firstEntity.setValue(firstValue);
```
zepfred: This call seems unnecessary

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ChangeMoveProviderTest.java
```diff
@@ -171,46 +139,25 @@ void fromEntityAllowsUnassigned() {
                 .basicVariable();
 
         var solution = TestdataAllowsUnassignedEntityProvidingSolution.generateSolution(2, 2);
-        var firstEntity = solution.getEntityList().get(0);
         var secondEntity = solution.getEntityList().get(1);
-        var firstValue = firstEntity.getValueRange().get(0);
+        var firstValue = solution.getEntityList().get(0).getValueRange().getFirst();
```
zepfred, line 144:
```suggestion
        var firstValue = solution.getEntityList().getFirst().getValueRange().getFirst();
```

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListAssignMoveProviderTest.java
```diff
@@ -0,0 +1,84 @@
...
+    void constructorRejectsNonUnassignedVariable() {
+        var solutionMetaModel = TestdataListSolution.buildMetaModel();
+        var variableMetaModel = solutionMetaModel.genuineEntity(TestdataListEntity.class)
+                .listVariable();
+        assertThatIllegalArgumentException()
+                .isThrownBy(() -> new ListAssignMoveProvider<>(variableMetaModel));
+    }
+
+    @Test
+    void fromSolution() {
```
zepfred, line 120: We should also include a test case that uses value ranges.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListChangeMoveProviderTest.java
```diff
@@ -96,25 +65,21 @@ void fromEntity() {
                 .listVariable();
 
         var solution = TestdataListEntityProvidingSolution.generateSolution();
-        var e1 = solution.getEntityList().get(0);
-        var v1 = solution.getValueList().get(0);
-        var v2 = solution.getValueList().get(1);
-        var v3 = solution.getValueList().get(2);
+        var e1 = solution.getEntityList().getFirst();
```
zepfred, line 88: We sometimes call `getFirst()`, but other times we do not. Please double-check for consistency regarding this. For instance, the class `ListAssignMoveProviderTest` does not use it.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/UnassignMoveProviderTest.java
```diff
@@ -0,0 +1,81 @@
...
+import ai.timefold.solver.core.testdomain.TestdataSolution;
+import ai.timefold.solver.core.testdomain.TestdataValue;
+import ai.timefold.solver.core.testdomain.unassignedvar.TestdataAllowsUnassignedEntity;
+import ai.timefold.solver.core.testdomain.unassignedvar.TestdataAllowsUnassignedSolution;
+
+import org.jspecify.annotations.NullMarked;
+import org.junit.jupiter.api.Test;
+
+@NullMarked
+class UnassignMoveProviderTest {
```
zepfred, line 20: Let's include a test case for pinned entities as well.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListUnassignMoveProviderTest.java
```diff
@@ -0,0 +1,92 @@
...
+import ai.timefold.solver.core.testdomain.list.TestdataListSolution;
+import ai.timefold.solver.core.testdomain.list.unassignedvar.TestdataAllowsUnassignedValuesListEntity;
+import ai.timefold.solver.core.testdomain.list.unassignedvar.TestdataAllowsUnassignedValuesListSolution;
+import ai.timefold.solver.core.testdomain.list.unassignedvar.TestdataAllowsUnassignedValuesListValue;
+
+import org.jspecify.annotations.NullMarked;
+import org.junit.jupiter.api.Test;
+
+@NullMarked
+class ListUnassignMoveProviderTest {
```
zepfred, line 22: Let's include a test case for pinned entities as well.

zepfred (changes requested):
The overall logic looks good, though I have minor comments and suggest increasing coverage.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListChangeMoveProviderTest.java
triceo: There is some logic to this. You only want to do this if it's consistent with the rest of the code. For example:

            var value1 = entity.getValueList().get(0);
            var value2 = entity.getValueList().get(1);
            var value3 = entity.getValueList().get(2);

Doing it here would just make the code look more confusing.
But there are some places in the tests where this is not the case, and I will fix those.

core/src/main/java/ai/timefold/solver/core/preview/api/move/builtin/AssignMoveProvider.java
triceo: Both valid points. Over time, as the API develops, I can see us adding methods for these more frequent patterns; right now though, we do not have enough data to tell what these patterns are. When in doubt, leave it out.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodProvider.java
triceo: Swap move is the only move which switches all variables of an entity, and not just one variable. Which is why it only needs to be included once per entity.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/UniOriginalMoveIterator.java
triceo: Memory and performance. I only want these expensive things to be created when we're actually guaranteed they'll be used.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/DefaultMoveStreamFactory.java
triceo: I reverted back to `true` and that makes the comment consistent, and the code will actually do that.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/UniOriginalMoveIterator.java
zepfred: Okay. Should we use the same approach for the random iterator?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodProvider.java
zepfred: Please add a comment containing that explanation.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListAssignMoveProviderTest.java
```diff
@@ -0,0 +1,169 @@
...
+        var move1 = getListAssignMove(moveList, 0);
+        assertSoftly(softly -> {
+            softly.assertThat(move1.getDestinationEntity()).isEqualTo(e1);
+            softly.assertThat(move1.getDestinationIndex()).isEqualTo(0);
+            softly.assertThat(move1.getPlanningValues()).containsExactly(v2);
+        });
+
+        var move2 = getListAssignMove(moveList, 1);
+        assertSoftly(softly -> {
+            softly.assertThat(move2.getDestinationEntity()).isEqualTo(e2);
```
zepfred, line 106: Let's check the index

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/ListAssignMoveProviderTest.java
```diff
@@ -0,0 +1,169 @@
...
+            softly.assertThat(move2.getPlanningValues()).containsExactly(v3);
+        });
+
+        var move3 = getListAssignMove(moveList, 2);
+        assertSoftly(softly -> {
+            softly.assertThat(move3.getDestinationEntity()).isEqualTo(e2);
```
zepfred, line 113: Ditto.

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/SwapMoveProviderTest.java
```diff
@@ -125,6 +132,59 @@ void multivariate() {
...
+        var solution = new TestdataEntityProvidingSolution("s1");
+        solution.setEntityList(List.of(e1, e2, e3));
+
+        // e1(v1, range={v1,v2}) ↔ e2(v2, range={v1,v2}): valid swap.
+        // e1 ↔ e3: v3 not in e1's range → excluded.
+        // e2 ↔ e3: v3 not in e2's range → excluded.
+        var moveList = NeighborhoodTester.build(new SwapMoveProvider<>(entityMetaModel), solutionMetaModel)
+                .using(solution)
+                .getMovesAsList(move -> (SwapMove<TestdataEntityProvidingSolution, TestdataEntityProvidingEntity>) move);
+        assertThat(moveList).hasSize(2);
```
zepfred, line 173: Does this mean two moves: e1 <-> e2 and e2 <-> e1?

zepfred (approved):
LGTM!

zepfred (approved):
LGTM!

core/src/test/java/ai/timefold/solver/core/preview/api/move/builtin/SwapMoveProviderTest.java
triceo: Yep.

PR #2259, opened 2026-04-22
chore: split random at the start of step so results are reproducible across higher thread counts

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
```diff
@@ -32,7 +32,8 @@ public DefaultRandomFactory(RandomType randomType, Long randomSeed) {
     public RandomGenerator createRandom() {
         switch (randomType) {
```
triceo: We really should have removed this in 2.0. :-( I completely forgot about this. It's entirely unused.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
```diff
@@ -32,7 +32,8 @@ public DefaultRandomFactory(RandomType randomType, Long randomSeed) {
...
             case JDK:
-                return randomSeed == null ? new Random() : new Random(randomSeed);
+                return randomSeed == null ? new DelegatingSplittableRandomGenerator(new Random().nextLong())
+                        : new DelegatingSplittableRandomGenerator(randomSeed);
```
triceo: Why does the default case generate a random seed? Every other random in this case (and even the code previously) just uses the default random seed. This is related to logging. In this case, we will say the random seed is zero in the logs. Which is no longer the case here. 

IMO if a random seed is not explicitly specified, we should start all randoms with a hard-coded zero.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DelegatingSplittableRandomGenerator.java
```diff
@@ -0,0 +1,99 @@
+package ai.timefold.solver.core.impl.solver.random;
+
+import java.util.SplittableRandom;
+import java.util.random.RandomGenerator;
+
+public final class DelegatingSplittableRandomGenerator implements RandomGenerator {
```
triceo, line 22: Why does this need to exist? Can't we create and use a `SplittableRandom` directly?

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
```diff
@@ -32,7 +32,8 @@ public DefaultRandomFactory(RandomType randomType, Long randomSeed) {
...
             case MERSENNE_TWISTER:
```
triceo: Since I stupidly forgot to delete this code, we now have to give it the same treatment as any other random. Which means the fix needs to apply to these randoms too, not just to the `JDK` random.

I'm toying with an idea... If we do a hotfix for this, we just delete these extra randoms and the config to supply them. We can justify that by the fact that we're fixing a bug. (And also by the fact that it's 2.0.1, and not any later minor release.)

Doing a fail-fast for these other randoms will practically disable this feature anyway, because the solver can no longer work without a splittable random.

Also, removing this would get us one step closer to removing `commons-math` from `core`.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -123,10 +127,12 @@ public void stepStarted(AbstractStepScope<Solution_> stepScope) {
         bestSolutionRecaller.stepStarted(stepScope);
         phaseLifecycleSupport.fireStepStarted(stepScope);
         globalTermination.stepStarted(stepScope);
+        savedRandom = ((DelegatingSplittableRandomGenerator) stepScope.getWorkingRandom()).split();
         // Do not propagate to phases; the active phase does that for itself and they should not propagate further.
     }
 
     public void stepEnded(AbstractStepScope<Solution_> stepScope) {
+        ((DelegatingSplittableRandomGenerator) stepScope.getWorkingRandom()).setDelegate(savedRandom);
```
triceo: This logic confuses me.

- stepStarted splits the random, but doesn't change the delegate.
- then ST and MT do all the work with the delegate, not with the split.
- stepEnded stores the split random as the working random for the next step.

I would have expected instead:

- stepStarted splits the random and uses the split for ST and MT.
- stepEnded restores the random to the one pre-split.

core/src/main/java/module-info.java
```diff
@@ -202,6 +202,7 @@
     exports ai.timefold.solver.core.impl.neighborhood to ai.timefold.solver.enterprise.core;
     exports ai.timefold.solver.core.impl.partitionedsearch to ai.timefold.solver.enterprise.core;
     exports ai.timefold.solver.core.impl.phase to ai.timefold.solver.enterprise.core;
+    exports ai.timefold.solver.core.impl.solver.random to ai.timefold.solver.enterprise.core;
```
triceo, line 204: Why? I don't see any enterprise reference here, and there is no enterprise counterpart.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -1397,8 +1397,8 @@ void solveStaleBuiltinShadows() {
 
         var solution = PlannerTestUtils.solve(solverConfig, problem);
 
-        assertThat(solution.getEntityList().getFirst().getValue().getCode()).isEqualTo("v1");
-        assertThat(solution.getEntityList().get(1).getValue().getCode()).isEqualTo("v2");
+        assertThat(solution.getEntityList().getFirst().getValue().getCode()).isEqualTo("v2");
+        assertThat(solution.getEntityList().get(1).getValue().getCode()).isEqualTo("v1");
```
triceo: Maybe we refactor this to assert the entire list instead? I'm assuming the list only has 2 entities.

triceo (commented):
First read through the code.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
Christopher-Chianelli: `new Random()` generates a random seed. In the default case, randomSeed is set. `randomSeed` is only null in `NON_REPRODUCIBLE` mode IIRC.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DelegatingSplittableRandomGenerator.java
Christopher-Chianelli: Iterators may save the random from `SolverScope`; this make it so we can swap the random without worrying about if an iterator (potentially even a user iterator) saved the random into a field.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
Christopher-Chianelli: With how splittable random works, both approaches should be equivalent. Using the current random and restoring to the split random (i.e. current approach) takes less code than using the split random and restoring to the original random.

core/src/main/java/module-info.java
Christopher-Chianelli: Probably can be removed; there was a prior version where enterprise swapped the random.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DelegatingSplittableRandomGenerator.java
triceo: That's good thinking. But the class should explain this, because it's not at all obvious.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
triceo: Personally, I'd prefer logic that feels more natural, even if it's more code.
Either way, there should be a comment here explaining why it's being done at all.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
triceo: Yeah, let's remove `RandomType`. `RandomFactory` can probably go as well.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomFactory.java
triceo: Fair point with `NON_REPRODUCIBLE`, forgot about that.
Still, I'd rather not generate the seed and let the underlying generator pick its own default.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DelegatingSplittableRandomGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+    private RandomGenerator.SplittableGenerator delegate;
+    private final Thread ownerThread;
+
+    public DelegatingSplittableRandomGenerator(long seed) {
+        this.delegate = new SplittableRandom(seed);
+        this.ownerThread = Thread.currentThread();
+    }
+
+    private void assertIsOwnedByCurrentThread() {
```
triceo, line 39: Also deserves an explanation in comments as to why we enforce this, and what would happen if it were not enforced.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -353,9 +353,8 @@ public SolverScope<Solution_> createChildThreadSolverScope(ChildThreadType child
         childThreadSolverScope.monitoringTags = monitoringTags;
         childThreadSolverScope.solverMetricSet = solverMetricSet;
         childThreadSolverScope.startingSolverCount = startingSolverCount;
-        // TODO FIXME use RandomFactory
         // Experiments show that this trick to attain reproducibility doesn't break uniform distribution
-        childThreadSolverScope.workingRandom = new Random(workingRandom.nextLong());
+        childThreadSolverScope.workingRandom = new DelegatingSplittableRandomGenerator(workingRandom.nextLong());
```
triceo: Why not splitting? we have a splittable random now.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -212,25 +212,15 @@ private <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> b
...
-            var randomType_ = Objects.requireNonNullElse(solverConfig.getRandomType(), RandomType.JDK);
-            var randomSeed_ = solverConfig.getRandomSeed();
-            if (solverConfig.getRandomSeed() == null && environmentMode_ != EnvironmentMode.NON_REPRODUCIBLE) {
-                randomSeed_ = DEFAULT_RANDOM_SEED;
-            }
-            return new DefaultRandomFactory(randomType_, randomSeed_);
+            randomSeed_ = RandomGenerator.getDefault().nextLong();
         }
+        Long finalRandomSeed_ = randomSeed_;
+        return () -> new DelegatingSplittableRandomGenerator(finalRandomSeed_);
```
triceo: Several places can get this and create their own randoms. I wonder if instead there should be one global random, and everyone who needs their own would split that one.

core/src/main/resources/solver.xsd
```diff
@@ -1832,5 +1804,29 @@
     </xs:restriction>
       
   </xs:simpleType>
+    
+  <xs:simpleType name="randomType">
```
triceo: Why is this still here?

tools/benchmark/src/test/java/ai/timefold/solver/benchmark/impl/result/PlannerBenchmarkResultTest.java
```diff
@@ -37,19 +36,19 @@ void createMergedResult() {
         var p1SolverX = new SolverBenchmarkResult(p1);
         p1SolverX.setName("Solver X");
         var p1SolverConfigX = new SolverConfig();
-        p1SolverConfigX.setRandomType(RandomType.JDK);
```
triceo, line 40: The solver used to log something about the random type at startup, next to the seed.
I didn't notice that logging being removed - please check for it.

triceo (changes requested):
All comments need resolving. Also, I am missing some docs changes:

- Effective move speed needs to be explained in multi-threaded documentation. Especially the seeming contradiction of effective speed going down with more move threads being added.
- The format of several logging messages has changed (or will change) as a result of this PR. Let's do a quick grep of the docs, in case they're shown anywhere.
- Check for the Random configuration options; if we mention them in the docs, those mentions need to disappear.

Also, Sonar.

core/src/main/java/module-info.java
Christopher-Chianelli: Used in enterprise now.

tools/benchmark/src/test/java/ai/timefold/solver/benchmark/impl/result/PlannerBenchmarkResultTest.java
Christopher-Chianelli: Logging is from the factory/supplier instance (and in particular, the randomType and seed was included in the factory's toString).

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
Christopher-Chianelli: Required to be in a supplier so that the delegating random gets created on the thread `solve` is called in (which is different for `SolverManager` and when user creates a `Solver` instance in a seperate thread).

Only the creating thread can access the random; all other threads will throw `IllegalStateException` to make it impossible to affect reproducibility.

triceo (approved):
CI is failing because of Javadoc. 
After that, LGTM.

PR #2270, opened 2026-04-29
refactor: introduce incremental constraint collectors for improved performance

triceo:
This benchmark run proves there are no regressions from adding the incremental support to the GroupNode:
https://github.com/TimefoldAI/timefold-solver-benchmarks/actions/runs/25095715799/job/73533079940

This benchmark proves no regressions after the new collector implementations were integrated:
https://github.com/TimefoldAI/timefold-solver-benchmarks/actions/runs/25427319378

The most expensive collectors are not actually part of those benchmarks, so the biggest benefit of this PR does not show up there.


triceo:
@Christopher-Chianelli Ready for review now.

I wasn't able to keep the uni changes separate from the bi/tri/quad, so this PR is large. But if you review the uni parts, you can just skim the rest; they are mechanical copies with updates for the higher arities.

triceo:
It would be great if we had `update()` variants for consecutive sequences and connected intervals; those collectors are expensive - if we can save some work there, it will immediately show.

Christopher-Chianelli:
@triceo GitHub still not allowing me to create any review comments, so I putting my review comments directly on the the PR.

UniConstraintCollector - Does a user need to implement both `accumulator` and `incrementalAccumulator`? I notice some of our builtin ConstraintCollectors (such as `AndThenUniCollector`) implement both?

Christopher-Chianelli:
- Per the Javadoc of `...AccumulatedValue`:
```java
/**
 * <li>{@link #add(Object)} will be called externally exactly once, when the value enters the group.
 * An instance of {@link UniConstraintCollectorAccumulatedValue} will only be created if there is a value to add.</li>
 * /
```

- But in `ConditionalUniCollector`:
```java
        private final UniConstraintCollectorAccumulatedValue<A> innerValue;
        private boolean active = false;

        AccumulatedValue(ResultContainer_ container) {
            this.innerValue = innerIncremental.intoGroup(container);
        }

        @Override
        public void add(A a) {
            if (!predicate.test(a)) {
                return;
            }
            active = true;
            innerValue.add(a);
        }

        @Override
        public void update(A a) {
            boolean nowActive = predicate.test(a);
            if (active && nowActive) {
                innerValue.update(a);
            } else if (active) {
                active = false;
                innerValue.remove();
            } else if (nowActive) {
                active = true;
                innerValue.add(a);
            }
        }
```

i.e. `add` and `remove` can both be called multiple times.

triceo:
> i.e. `add` and `remove` can both be called multiple times.

They can, but they will not be. That is a contract that we guarantee.
(The alternative was to create more garbage, essentially add() would create a new container - and that was negatively affecting benchmarks.)

Christopher-Chianelli:
> > i.e. `add` and `remove` can both be called multiple times.
> 
> They can, but they will not be. That is a contract that we guarantee. (The alternative was to create more garbage, essentially add() would create a new container - and that was negatively affecting benchmarks.)

Something like this would cause `add/remove` to be called multiple times currently:

```java
constraintFactory.forEach(Employee.class)
    .join(Shift.class, equal(id(), Shift::getEmployee)
    // Employee has an InverseCollectionShadowVariable to get shift count
    .groupBy((employee, shift) -> employee, conditionally((employee, shift) -> employee.getShiftCount() > 5, sum((employee, shift) -> shift.getDuration())))
    .filter((employee, hours) -> hours != null && hours > 24)
    .penalize(HardSoftScore.ONE_HARD, (employee, hours) -> hours - 24)
    .asConstraint("Overworked employee")
```

In particular, the `conditionally` can and will call the `sum`'s add multiple times. It seems our actually contract is:

- `add` can only be called when the value is not "added" (and once called, the value is "added")
- `update` can only be called when the value is "added" (and does not affect the state of "added")
- `remove` can only be called when the value is "added" (and once called, the value is not "added")

i.e. this is explictly allowed:
```
var a = new Data();
value.add(a);
a.change();
value.update(a);
value.remove(a);
value.add(a);
```

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractLongSumSlot.java
```diff
@@ -0,0 +1,33 @@
+package ai.timefold.solver.core.impl.score.stream.collector;
+
+public abstract class AbstractLongSumSlot {
+
+    public static final class State {
```
Christopher-Chianelli: MutableLong?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractLongAverageSlot.java
```diff
@@ -0,0 +1,39 @@
+package ai.timefold.solver.core.impl.score.stream.collector;
+
+public abstract class AbstractLongAverageSlot {
+
+    public static final class State {
```
Christopher-Chianelli, line 5: Why seperate state into its own class? It a private field, so it not like subclasses can access the state field directly.

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintCollector.java
```diff
@@ -22,28 +23,36 @@
...
     /**
-     * A lambda that extracts data from the matched facts,
-     * accumulates it in the result container
-     * and returns an undo operation for that accumulation.
-     *
-     * @return the undo operation. This lambda is called when the facts no longer matches.
+     * As defined by {@link UniConstraintCollector#accumulator()}, but for {@link BiConstraintStream}.
      */
-    @NonNull
     TriFunction<ResultContainer_, A, B, Runnable> accumulator();
```
Christopher-Chianelli, line 39: So a user needs to define both `accumulator` and `incrementalAccumlator` if they want to implement an `incrementalAccumlator`?

core/src/main/java/ai/timefold/solver/core/api/score/stream/uni/UniConstraintCollectorValueHandle.java
```diff
@@ -0,0 +1,55 @@
...
+     * In some cases, the default implementation (remove+add) will be enough,
+     * but this method exists for advanced cases where updates can be done more efficiently.
+     *
+     * @param a The component of the tuple.
+     *        It is expected to be the component of the very same tuple that was used to {@link #add(Object)}.
+     *        The component may be an entirely different object instance than was added.
+     *        For other tuples, a fresh {@link UniConstraintCollectorAccumulatedValue} will be obtained;
+     *        an instance of this interface is guaranteed to work on one and only one tuple.
+     */
+    default void update(@Nullable A a) {
```
Christopher-Chianelli: Maybe `set` or `changeTo`? The `update` name make me think `a` must be the same object that was added.

core/src/main/java/ai/timefold/solver/core/impl/bavet/bi/AbstractGroupBiNode.java
```diff
@@ -2,42 +2,59 @@
...
 import ai.timefold.solver.core.impl.bavet.common.AbstractGroupNode;
 import ai.timefold.solver.core.impl.bavet.common.tuple.BiTuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.Tuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
+import ai.timefold.solver.core.impl.score.stream.collector.bi.BiCollectorUtils;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 
 abstract class AbstractGroupBiNode<OldA, OldB, OutTuple_ extends Tuple, GroupKey_, ResultContainer_, Result_>
```
Christopher-Chianelli, line 18: Make this class `@NullMarked`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/bi/AbstractGroupBiNode.java
```diff
@@ -2,42 +2,59 @@
...
-            BiConstraintCollector<OldA, OldB, ResultContainer_, Result_> collector,
+            @NonNull BiConstraintCollector<OldA, OldB, ResultContainer_, Result_> collector,
             TupleLifecycle<OutTuple_> nextNodesTupleLifecycle, EnvironmentMode environmentMode) {
-        super(groupStoreIndex, undoStoreIndex,
-                groupKeyFunction,
-                collector == null ? null : collector.supplier(),
-                collector == null ? null : collector.finisher(),
-                nextNodesTupleLifecycle, environmentMode);
-        accumulator = collector == null ? null : collector.accumulator();
+        super(groupStoreIndex, groupKeyFunction, collector.supplier(), collector.finisher(), nextNodesTupleLifecycle,
```
Christopher-Chianelli, line 28: How about when no collector function is provided? i.e.
```java
constraintFactory.forEach(Shift.class)
     .groupBy(Shift::getEmployee)
    // ...
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractGroupNode.java
Christopher-Chianelli: This is a confusing name for what is really `hasKeyFunction`.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/BiCollectorUtils.java
```diff
@@ -0,0 +1,84 @@
...
+@NullMarked
+public final class BiCollectorUtils {
+
+    public static <ResultContainer_, A, B> TriFunction<ResultContainer_, A, B, Runnable>
+            fromIncremental(BiConstraintCollectorAccumulator<ResultContainer_, A, B> incrementalAccumulator) {
+        if (incrementalAccumulator instanceof BiFromAccumulatorAdapter<ResultContainer_, A, B>(TriFunction<ResultContainer_, A, B, Runnable> accumulator)) {
+            return accumulator;
+        }
+        return new BiFromIncrementalAdapter<>(incrementalAccumulator);
+    }
```
Christopher-Chianelli: Why do we want to turn incremental accumulators back to normal accumulators? (Might need a different name)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractLongAverageSlot.java
triceo: `State` and `Slot` are two distinct concepts.

- `State` is the overall counter for the group. It exists on the collector level, it is the `ResultContainer_`.
- `Slot` represents the individual accumulated element. It is created every time a new value is being added into the group, and removed when the element is removed from the group - it will update the overall state, but it is different from it.

triceo:
> Something like this would cause `add/remove` to be called multiple times currently:

Good point. For collectors which contain other collectors (and only for them), the contract as specified doesn't actually hold.

Two ways out of this:
- These nesting collectors would correctly recreate the inner instances on every remove+add, so that the contract still holds.
- Or we update the contract to what you outlined in your previous message.

Since we want to prevent unnecessary garbage from being collected, the second option is probably the way to go.

triceo:
> UniConstraintCollector - Does a user need to implement both accumulator and incrementalAccumulator? I notice some of our builtin ConstraintCollectors (such as AndThenUniCollector) implement both?

User typically only implements one, and uses the isIncremental() method to say which they implemented. Our own default collectors indeed implement both, and only for backwards compatibility reasons - if for whatever reason the user decided to call the old non-incremental collector manually, they'd get an exception if they were not implemented, breaking backwards compatibility.

triceo:
@Christopher-Chianelli I have resolved your comments so far, but I feel as if the review wasn't yet complete.

In addition to whatever you will to point out, I'm looking for a discussion of the following points specifically:

- Is "incremental collector" a good name for the concept? Aren't all collectors incremental? "updatable collector" ain't much better, all collectors are updatable, only right now it's done through remove+add.
- The switching between incremental and non-incremental collectors; the isIncremental() method. There was no way to map the old collector API to be incremental backwards compatibly, which is why I had to introduce this. But I don't like it - it forces the user to make a choice, and based on that choice, some methods of the API become useless; it's ugly. Can you think of a different way of doing this?
- `UniConstraintCollectorAccumulatedValue` is IMO a horrible name. Doesn't say much. Any suggestions for improvement?

Christopher-Chianelli:
> @Christopher-Chianelli I have resolved your comments so far, but I feel as if the review wasn't yet complete.

I am still going through the review; keeping context and making notes for a 8000+ line changes is not trivial, even if most of the changes are mechanical.

>     * Is "incremental collector" a good name for the concept? Aren't all collectors incremental? "updatable collector" ain't much better, all collectors are updatable, only right now it's done through remove+add.

Slot-based? The old collector API use a global state, whereas this uses a per-tuple state.
 
>     * The switching between incremental and non-incremental collectors; the isIncremental() method. There was no way to map the old collector API to be incremental backwards compatibly, which is why I had to introduce this. But I don't like it - it forces the user to make a choice, and based on that choice, some methods of the API become useless; it's ugly. Can you think of a different way of doing this?

Create a sealed interface `AbstractConstraintCollector`, make the old and new constraint collector interfaces two seperate implementations of it, and change the CS groupBy API to take `AbstractConstraintCollector` instead.

>     * `UniConstraintCollectorAccumulatedValue` is IMO a horrible name. Doesn't say much. Any suggestions for improvement?

`UniConstraintCollectorValueHandle`? https://en.wikipedia.org/wiki/Handle_(computing)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ComposeTwoBiCollector.java
```diff
@@ -7,9 +7,12 @@
...
 import ai.timefold.solver.core.api.function.TriFunction;
 import ai.timefold.solver.core.api.score.stream.bi.BiConstraintCollector;
+import ai.timefold.solver.core.api.score.stream.bi.BiConstraintCollectorAccumulatedValue;
+import ai.timefold.solver.core.api.score.stream.bi.BiConstraintCollectorAccumulator;
 import ai.timefold.solver.core.impl.util.Pair;
 
 import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 
 final class ComposeTwoBiCollector<A, B, ResultHolder1_, ResultHolder2_, Result1_, Result2_, Result_>
```
Christopher-Chianelli, line 16: Since the only code different between the various `ComposeX` implementations is the finisher, you could have an Abstract Base Class that put the accumulators into an array and only leaves `finisher` as the only unimplemented method.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ConditionalBiCollector.java
```diff
@@ -59,4 +65,45 @@ public boolean equals(Object object) {
...
+            if (!predicate.test(a, b)) {
+                return;
+            }
+            active = true;
+            innerValue.add(a, b);
+        }
+
+        @Override
+        public void update(A a, B b) {
+            boolean nowActive = predicate.test(a, b);
```
Christopher-Chianelli: `var` ?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/CountDistinctBiCollector.java
```diff
@@ -1,20 +1,56 @@
...
+        return new Slot(state);
+    }
+
+    private final class Slot extends AbstractLongDistinctSlot<Mapped_>
+            implements BiConstraintCollectorAccumulatedValue<A, B> {
+        Slot(AbstractLongDistinctSlot.State<Mapped_> state) {
+            super(state);
+        }
+
+        @Override
```
Christopher-Chianelli, line 41: Thei implementation of these methods are identical across several classes, suggesting they should be moved into an abstract base class

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/MaxPropertyBiCollector.java
```diff
@@ -5,12 +5,14 @@
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import ai.timefold.solver.core.impl.score.stream.collector.MinMaxUndoableActionable;
+import ai.timefold.solver.core.api.score.stream.bi.BiConstraintCollectorAccumulatedValue;
+import ai.timefold.solver.core.impl.score.stream.collector.AbstractMinMaxSlot;
 
 import org.jspecify.annotations.NonNull;
 
 final class MaxPropertyBiCollector<A, B, Result_, Property_ extends Comparable<? super Property_>>
```
Christopher-Chianelli, line 13: The `MinProperty` and `MaxProperty` collectors are basically identical except in their finisher, suggesting it can be an ABC (potentially with its two implementations within the abstract class itself).

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/MinPropertyBiCollector.java
```diff
@@ -5,12 +5,14 @@
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import ai.timefold.solver.core.impl.score.stream.collector.MinMaxUndoableActionable;
+import ai.timefold.solver.core.api.score.stream.bi.BiConstraintCollectorAccumulatedValue;
+import ai.timefold.solver.core.impl.score.stream.collector.AbstractMinMaxSlot;
 
 import org.jspecify.annotations.NonNull;
 
 final class MinPropertyBiCollector<A, B, Result_, Property_ extends Comparable<? super Property_>>
```
Christopher-Chianelli, line 13: You could in theory replace this with the `Comparator` version, using `Comparator.naturalOrder()`

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractLoadBalanceSlot.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    protected void addMapped(Balanced_ balanced, long load, long initialLoad) {
+        cachedBalanced = balanced;
+        cachedLoad = load;
+        container.registerBalanced(balanced, load, initialLoad);
+    }
+
+    protected void updateMapped(Balanced_ balanced, long load, long initialLoad) {
+        if (Objects.equals(cachedBalanced, balanced) && cachedLoad == load) {
```
Christopher-Chianelli, line 24: I wonder how much caching helps here; load balance should be quick to update.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractMinMaxSlot.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+            cachedItem = item;
+            cachedCount = cachedInnerMap.computeIfAbsent(item, ignored -> new MutableInt());
+            cachedCount.increment();
+            return;
+        }
+        removeMapped();
+        cachedItem = item;
+        cachedKey = newKey;
+        cachedInnerMap = state.propertyToItemCountMap.computeIfAbsent(newKey, ignored -> new LinkedHashMap<>());
```
Christopher-Chianelli: It is unclear that `propertyToItemCountMap` is sorted; maybe `sortedPropertyToItemCountMap`?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractMinMaxSlot.java
```diff
@@ -0,0 +1,116 @@
...
+        cachedCount = cachedInnerMap.computeIfAbsent(item, ignored -> new MutableInt());
```
Christopher-Chianelli, line 96: Does `cachedInnerMap` only have one key at a time? `removedMapped()` is always called before this. It seems you are using a `Map` strictly for its entry, so you can have a (value, count) tuple.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractReferenceSumSlot.java
```diff
@@ -0,0 +1,50 @@
...
+        this.state = state;
+    }
+
+    protected void addMapped(Result_ input) {
+        cachedValue = input;
+        state.current = state.adder.apply(state.current, input);
+    }
+
+    protected void updateMapped(Result_ input) {
+        if (Objects.equals(cachedValue, input)) {
```
Christopher-Chianelli: I wonder if checking for `equals` helps here or not, given how fast sum and subtract should be.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractToCollectionSlot.java
```diff
@@ -0,0 +1,47 @@
...
+            this.collectionFunction = collectionFunction;
+        }
+
+        public Collection_ result() {
+            var out = collectionFunction.apply(list.size());
+            if (list.isEmpty()) {
+                // Avoid exception if out is an immutable collection
+                return out;
+            }
+            out.addAll(list);
```
Christopher-Chianelli, line 23: If the user cannot modify the list, currently you just return a `Collection.unmodifiableList(list)`?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractToListSlot.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+
+import ai.timefold.solver.core.impl.util.ElementAwareArrayList;
+
+public abstract class AbstractToListSlot<Mapped_> {
+    public static final class State<Mapped_> {
+        private final ElementAwareArrayList<Mapped_> resultList = new ElementAwareArrayList<>();
+
+        public List<Mapped_> result() {
+            return resultList;
```
Christopher-Chianelli, line 12: Might want to make `Collections.unmodifiableList(resultList)`

core/src/test/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetAdvancedGroupByConstraintStreamTest.java
```diff
@@ -1,589 +1,12 @@
...
-
-import org.junit.jupiter.api.TestTemplate;
 
 final class BavetAdvancedGroupByConstraintStreamTest extends AbstractAdvancedGroupByConstraintStreamTest {
 
     public BavetAdvancedGroupByConstraintStreamTest(ConstraintMatchPolicy constraintMatchPolicy) {
         super(new BavetConstraintStreamImplSupport(constraintMatchPolicy));
     }
 
-    @TestTemplate
```
Christopher-Chianelli, line 40: Why were these tests removed?

Christopher-Chianelli (changes requested):

core/src/test/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetAdvancedGroupByConstraintStreamTest.java
triceo: Duplicates of the same in another classs. (The parent.)

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractToCollectionSlot.java
triceo: I could, but it's been shown over and over that the GC overhead caused by unmodifiable wrappers can be significant, if on the hot path. Which is why I'm generally leaving it to the Javadoc to let people know now to touch this.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractToListSlot.java
triceo: Explained above.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractReferenceSumSlot.java
triceo: Fair point.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractLoadBalanceSlot.java
triceo: Actually, it updates a map - and that is not free. 
In this case, I think that the caching will help; I do not have a benchmark for this specifically though.

core/src/main/java/ai/timefold/solver/core/api/score/stream/bi/BiConstraintCollector.java
triceo: User typically only implements one, and uses the isIncremental() method to say which they implemented. (The other may throw UnsupportedOperationException.) Our own default collectors indeed implement both, and only for backwards compatibility reasons - if for whatever reason the user decided to call the old non-incremental collector manually, they'd get an exception if they were not implemented, breaking backwards compatibility.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/BiCollectorUtils.java
triceo: Already explained above. We need these impls for bwd compatibility, and I didn't want to reimplement - so I adapted them.

core/src/main/java/ai/timefold/solver/core/impl/bavet/bi/AbstractGroupBiNode.java
triceo: Would require _a lot_ of refactoring which would not be strictly related to this PR - considering its size, I decided against it.

core/src/main/java/ai/timefold/solver/core/impl/bavet/bi/AbstractGroupBiNode.java
triceo: Handled in the other constructor right above. This has not changed.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/ComposeTwoBiCollector.java
triceo: IMO unrelated to the goal of this PR, and the PR is already large enough. I have not changed this from the merged state - it also had all these specializations.

Also not as easy to do as it may seem; the different generic returns and holders of the different collectors make this difficult, if not impossible. I can do the array trick only if I want to lose the generics.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/CountDistinctBiCollector.java
triceo: Not really possible due to the generics. THe parent class of this is already a `Bi` collector, which means I have nothing to put these arity-unrelated methods on.

Also, same comment as above - did not change in this PR, not related to the goal of this PR, and PR already big enough.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/bi/MaxPropertyBiCollector.java
triceo: Same answer as above; not doing any unrelated refactors here.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/AbstractMinMaxSlot.java
triceo: The map lives in the global state, and many items may have entries there. This code is correct (and didn't change much from the committed version).

triceo:
@Christopher-Chianelli This is the finalized code, with the incremental switches completely removed.

Please review if you find some time between Team Days and your PTO, as I'd like to merge this week. **Do not work on this during Team Days, please** - if you don't review before your PTO, I will handle this with Fred instead.

core/src/main/java/ai/timefold/solver/core/api/score/stream/quad/QuadConstraintCollectorAccumulator.java
```diff
@@ -0,0 +1,36 @@
...
+import ai.timefold.solver.core.api.function.PentaFunction;
+import ai.timefold.solver.core.api.score.stream.uni.UniConstraintCollectorAccumulator;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * As defined by {@link UniConstraintCollectorAccumulator},
+ * only for {@link QuadConstraintCollector}.
+ * <p>
+ * Extends {@link PentaFunction} for detection purposes; {@link #apply} always throws.
```
Christopher-Chianelli: Bi...Accumulator and Quad...Accumlator have different Javadoc descriptions?

core/src/main/java/ai/timefold/solver/core/api/score/stream/tri/TriConstraintCollectorAccumulator.java
```diff
@@ -0,0 +1,36 @@
...
+import ai.timefold.solver.core.api.function.QuadFunction;
+import ai.timefold.solver.core.api.score.stream.uni.UniConstraintCollectorAccumulator;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * As defined by {@link UniConstraintCollectorAccumulator},
+ * only for {@link TriConstraintCollector}.
+ * <p>
+ * This interface extends {@link QuadFunction}, but {@link #apply(Object, Object, Object, Object)} always throws
```
Christopher-Chianelli: Also different from both Bi...and Quad...

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractGroupNode.java
```diff
@@ -270,13 +272,20 @@ public final void retract(InTuple_ tuple) {
...
+            groupRetract(tuple);
         }
-        killTuple(group);
+        var newParentCount = --group.parentCount;
+        killOutTuple(group, newParentCount == 0);
+    }
+
+    protected abstract void groupInsert(ResultContainer_ resultContainer, InTuple_ tuple);
+
+    protected void groupUpdate(ResultContainer_ resultContainer, InTuple_ tuple) {
```
Christopher-Chianelli: Why provide a default implementation that should be overwritten so it use `replaceWith` in the subclasses?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/quad/AverageQuadCollector.java
```diff
@@ -1,20 +1,54 @@
...
 import java.util.function.Supplier;
 
 import ai.timefold.solver.core.api.function.ToLongQuadFunction;
-import ai.timefold.solver.core.impl.score.stream.collector.LongAverageCalculator;
+import ai.timefold.solver.core.api.score.stream.quad.QuadConstraintCollectorValueHandle;
+import ai.timefold.solver.core.impl.score.stream.collector.AbstractLongAverageSlot;
 
 import org.jspecify.annotations.NonNull;
 
 final class AverageQuadCollector<A, B, C, D>
```
Christopher-Chianelli, line 12: I feel like we can massively reduce code duplication by only implementing the uni version, and have an adapter class that converts the uni version to a bi, tri or quad version. Many of these classes effectively operate on one value computed by a mapping function given to the collector.

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/quad/AverageQuadCollector.java
triceo: You are probably right, but in the interest of not making this PR even bigger than it already is, I won't do it now. This code had duplication before, it will continue to do so.

PR #2271, opened 2026-04-29
perf: make Random splittable (2.0.x)

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -332,14 +326,6 @@ public void setDaemon(@Nullable Boolean daemon) {
         this.daemon = daemon;
     }
 
-    public @Nullable RandomType getRandomType() {
```
zepfred, line 335: Should we consider deprecating it instead of completely removing it?

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -348,14 +334,6 @@ public void setRandomSeed(@Nullable Long randomSeed) {
         this.randomSeed = randomSeed;
     }
 
-    public @Nullable Class<? extends RandomFactory> getRandomFactoryClass() {
```
zepfred, line 351: Ditto.

Christopher-Chianelli:
Closing since it already merged inside the 2.0.x branch: https://github.com/TimefoldAI/timefold-solver/commits/2.0.x/

PR #2276, opened 2026-05-03
deps: remove commons-math3 from solver-core

triceo:
commons-math3 hasn't seen a release in a long time, and gives JPMS-related warnings; removing it from where we can.
It still remains in benchmark, in enterprise-core and in some tests.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
@@ -1,7 +1,8 @@
+/**
+ * Includes code takes from Apache's commons-math library, v3.6.1 (Apache 2.0-licensed)
+ * These methods are clearly marked in comments, their modifications listed.
+ */
 package ai.timefold.solver.core.impl.util;
-
-import org.apache.commons.math3.util.CombinatoricsUtils;
-import org.apache.commons.math3.util.FastMath;
 
 public class MathUtils {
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
```diff
@@ -65,4 +68,282 @@ public static long getSpeed(long count, long timeMillisSpent) {
         // Avoid divide by zero exception on a fast CPU
         return count * 1000L / (timeMillisSpent == 0L ? 1L : timeMillisSpent);
     }
+
+    public static long factorial(int n) {
+        if (n < 0) {
+            throw new IllegalArgumentException("Factorial of %d is undefined."
+                    .formatted(n));
+        } else if (n > FACTORIAL_MAX_N) {
```
zepfred, line 76: Does this limit match the replaced API?

core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
```diff
@@ -65,4 +68,282 @@ public static long getSpeed(long count, long timeMillisSpent) {
...
+                yield FACTORIAL_CACHE[factorialCacheIndex] = n * factorial(n - 1);
+            }
+        };
+    }
+
+    private static double factorialLog(int n) {
+        if (n <= FACTORIAL_MAX_N) {
+            return Math.log(factorial(n));
+        }
+        var logSum = 0.0d;
```
zepfred, line 97: This approach appears to be incorrect. Where is the factorial being calculated?

zepfred (commented):
Please double-check that the implementation of `factorialLog` is correct.

core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
Christopher-Chianelli: Long max value is `9223372036854775807` (19 digits). `20!` is `2432902008176640000` (19 digits). `21!` is `51090942171709440000` (20 digits). So yes, it match the limit in the replaced API (20).

core/src/main/java/ai/timefold/solver/core/impl/util/MathUtils.java
Christopher-Chianelli: Recall, factorial is multiplication: `fact(n) = 1 * 2 * 3 * ... * n`. This calculates the log of a factorial `log(fact(n))`.

Recall from the rules of logs: `log(a * b) = log(a) + log(b)`.

Thus `log(fact(n)) = log(1) + log(2) + ... + log(n)`

zepfred (approved):

PR #2284, opened 2026-05-04
fix: use config class loader, then context class loader, then parent in GIZMO

### LINKED ISSUE #2264: fix(core): GizmoClassLoader parent-first delegation breaks ConstraintProvider class identity under Spring Boot DevTools
**Describe the bug**

`SolverFactory` creation fails with a misleading error whenever the user's domain classes are reachable through the `timefold-solver-core` JAR's classloader (i.e. above `GizmoClassLoader.getParent()`). Under Spring Boot DevTools every startup hits this; isolated test classloaders and JPMS layers can hit it too.

```
java.lang.IllegalArgumentException: Cannot use class
(com.example.MyProblemFact) in a constraint stream
as it is neither the same as, nor a superclass or superinterface
of one of planning entities or problem facts.
Ensure that all from(), join(), ifExists() and ifNotExists() building
blocks only reference classes assignable from planning entities or
problem facts ([..., com.example.MyProblemFact, ...])
annotated on the planning solution (com.example.MySolution).
```

The class named *is* in the listed problem facts. Two `Class<?>` instances exist for the same canonical name, loaded by different classloaders.

**Expected behavior**

`SolverFactory` is created successfully, regardless of whether user classes are reachable through both the parent and the TCCL of `GizmoClassLoader`.

**Actual behavior**

`SolverFactory.create(solverConfig)` throws `IllegalArgumentException: Cannot use class ... in a constraint stream` for the first problem-fact class joined in a constraint.

**Root cause**

`GizmoClassLoader`'s parent is `GizmoClassLoader.class.getClassLoader()` (line 37). The standard `ClassLoader.loadClass` is parent-first, so when a Gizmo-generated `MemberAccessor` resolves its `declaringClass` constant or `field.getGenericType()` for `List<MyProblemFact>`, the JVM calls `loadClass`, which delegates to the parent before reaching `findClass` (lines 50–51). Whenever user classes are reachable through that parent, the parent wins and the TCCL fallback in `findClass` is dead code. The user's `ConstraintProvider` was loaded by TCCL (e.g. `RestartClassLoader`), so its `MyProblemFact.class` literal resolves via TCCL — different `Class` identity, `assertValidFromType` rejects it.

**To Reproduce**

Spring Boot 4 + `timefold-solver-spring-boot-starter:2.0.0` + `spring-boot-devtools` on the classpath, with any constraint that does `.join(SomeProblemFact.class, ...)`. The same app started via `java -jar` (no DevTools, single classloader for project classes) starts cleanly. Happy to attach a minimal Gist reproducer if helpful.

## Environment

**Timefold Solver Version or Git ref**: `2.0.0` (release tag `v2.0.0`)

**Output of `java -version`**:
```
openjdk version "25" 2025-09-16 LTS
OpenJDK Runtime Environment Temurin-25+36 (build 25+36-LTS)
```

**Output of `uname -a`**: `Darwin 25.3.0 arm64`

**Other relevant context**:
- Spring Boot `4.0.6`
- `spring-boot-devtools` `4.0.6` on the classpath (optional dependency)
- App uses an explicit `solverConfig.xml` with `<solutionClass>`, `<entityClass>`, `<constraintProviderClass>`. Auto-discovery exhibits the same bug.

## Suggested fix

Override `GizmoClassLoader.loadClass` so non-Gizmo classes are looked up through the thread context classloader before delegating to the parent:

```java
@Override
protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
    synchronized (getClassLoadingLock(name)) {
        Class<?> loadedClass = findLoadedClass(name);
        if (loadedClass == null) {
            if (hasBytecodeFor(name)) {
                loadedClass = findClass(name);
            } else {
                ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
                if (contextClassLoader != null && contextClassLoader != this) {
                    try {
                        loadedClass = contextClassLoader.loadClass(name);
                    } catch (ClassNotFoundException ignored) {
                        // Fall through to parent delegation below.
                    }
                }
                if (loadedClass == null) {
                    loadedClass = super.loadClass(name, false);
                }
            }
        }
        if (resolve) {
            resolveClass(loadedClass);
        }
        return loadedClass;
    }
}
```

The constructor's parent setting (line 37) and the comment on lines 32–36 about the Quarkus MemberAccessor double-load invariant are **not** changed — only the delegation order. Under Quarkus, Gizmo-generated MemberAccessors are still resolved at step 2 (`hasBytecodeFor`) and never reach TCCL, so that invariant still holds. Under Spring Boot DevTools, user classes now hit step 3 (TCCL = `RestartClassLoader`) instead of being silently captured by the parent.

Native image: `isGizmoSupported()` already gates the Gizmo path, and TCCL on SubstrateVM is fixed at build time, so AOT behaviour is unchanged.

Verified locally on a Spring Boot 4 + TimeFold `2.0.0` + DevTools project: `Started ... in 5.7 seconds` on the `restartedMain` thread, no exception.

## Additional information

I instrumented `assertValidFromType` and dumped each `factType`'s loader: the planning entity (registered through `solverConfig.entityClassList`) was on `RestartClassLoader`, but every `@ProblemFactCollectionProperty` element type was on `AppClassLoader`, while `fromType` from the constraint provider was `RestartClassLoader`. The patch above made all problem-fact `Class` instances match the constraint provider's loader.

Happy to send a PR with the patch and a regression test (AssertJ) that loads a class through `GizmoClassLoader` while a child TCCL contains a same-named class, asserting `loadClass` returns the TCCL's instance.

Christopher-Chianelli:
Note: the thread local solution is ugly, but I don't see another way of passing this required data to where it needs to be without a massive refactoring of everything.

The test basically copies the existing `TestdataObject`, `TestdataEntity`, `TestdataValue` and `TestdataSolution` class into a new class loader, and can reproduce the exception when the solver config's classloader is not set. 

Fixes #2264

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
@@ -203,6 +203,7 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
     @XmlTransient
     private Clock clock = null;
     @XmlTransient
+    @Deprecated(since = "2.1.0", forRemoval = true)
     private ClassLoader classLoader = null;
 
     // Warning: all fields are null (and not defaulted) because they can be inherited
@@ -263,6 +264,7 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         this.clock = Objects.requireNonNull(clock);
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public SolverConfig(@Nullable ClassLoader classLoader) {
         this.classLoader = classLoader;
     }
@@ -294,10 +296,12 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         this.clock = clock;
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public @Nullable ClassLoader getClassLoader() {
         return classLoader;
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public void setClassLoader(@Nullable ClassLoader classLoader) {
         this.classLoader = classLoader;
     }
@@ -501,6 +505,7 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         return this;
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public @NonNull SolverConfig withClassLoader(@NonNull ClassLoader classLoader) {
         this.setClassLoader(classLoader);
         return this;
--- a/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
@@ -296,11 +296,17 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         this.clock = clock;
     }
 
+    /**
+     * @deprecated does not have any effect.
+     */
     @Deprecated(since = "2.1.0", forRemoval = true)
     public @Nullable ClassLoader getClassLoader() {
         return classLoader;
     }
 
+    /**
+     * @deprecated does not have any effect.
+     */
     @Deprecated(since = "2.1.0", forRemoval = true)
     public void setClassLoader(@Nullable ClassLoader classLoader) {
         this.classLoader = classLoader;
@@ -505,6 +511,9 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         return this;
     }
 
+    /**
+     * @deprecated does not have any effect.
+     */
     @Deprecated(since = "2.1.0", forRemoval = true)
     public @NonNull SolverConfig withClassLoader(@NonNull ClassLoader classLoader) {
         this.setClassLoader(classLoader);
--- a/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
@@ -264,6 +264,9 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         this.clock = Objects.requireNonNull(clock);
     }
 
+    /**
+     * @deprecated does not have any effect.
+     */
     @Deprecated(since = "2.1.0", forRemoval = true)
     public SolverConfig(@Nullable ClassLoader classLoader) {
         this.classLoader = classLoader;
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoClassLoader.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoClassLoader.java
@@ -6,8 +6,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
-import ai.timefold.solver.core.impl.domain.solution.descriptor.SolutionDescriptor;
-
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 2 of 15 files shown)


triceo:
I don't like this approach at all. What would it take to do this in a way that is specific to Spring, which is the place where this bug is being observed?

Or alternatively... would the issue go away if we simply removed Gizmo from Spring?

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -198,6 +198,7 @@ private SolutionDescriptor<Solution_> buildSolutionDescriptor() {
         }
         return SolutionDescriptor.buildSolutionDescriptor(solverConfig.getEnablePreviewFeatureSet(),
                 domainAccessType,
+                solverConfig.getClassLoader(),
```
triceo: I do not see anyone actually _setting_ the classloader into the solver config.
Where does this happen? For a Spring fix, it is surprising it does not touch a single Spring-related class.

Christopher-Chianelli: The fix occurs in the `GizmoClassLoader`; before a class is located in this order `bytecode, classloader of GizmoClassLoader`. Now a class is located in this order `bytecode, solver config classloader, context classloader, classloader of GizmoClassLoader`. Do note that `classloader of GizmoClassLoader` was used to workaround Quarkus (otherwise it would of been context classloader).

Christopher-Chianelli: The test for this fix create copies of our `TestdataEntity`, `TestdataValue` and `TestdataSolution` class inside a different class loader, and set the ClassLoader of the SolverConfig to it in the test.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -294,10 +296,12 @@ public void setClock(@Nullable Clock clock) {
         this.clock = clock;
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public @Nullable ClassLoader getClassLoader() {
         return classLoader;
     }
 
+    @Deprecated(since = "2.1.0", forRemoval = true)
     public void setClassLoader(@Nullable ClassLoader classLoader) {
```
triceo, line 314: Let's add a Javadoc saying that this setting has no actual effect.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoClassLoader.java
```diff
@@ -44,6 +44,43 @@ public String getName() {
         return "Timefold Solver Gizmo ClassLoader";
     }
 
+    @Nullable
+    private Class<?> loadClassFrom(@Nullable ClassLoader otherClassLoader, String name) {
+        if (otherClassLoader == null || otherClassLoader == this) {
+            return null;
```
triceo, line 50: I'd rather throw on `null`. You're saying you want it executed with a given classloader - if it's null, it's clearly a wrong argument.

Also, what is the rationale for returning `null` if the classloader is `this`? Wouldn't it make more sense to just load the class?

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
triceo, line 1: Unnecessary changes; the file should not be included.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
triceo, line 1: Dtto.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -1773,6 +1774,23 @@ private void solveListVarEntityRangeModel(MoveSelectorConfig<?> moveSelectionCon
         assertThat(bestEntity3.getValueList()).doesNotContain(value1, value2, value3);
     }
 
+    @Test
+    void solveCustomClassLoader() {
```
triceo, line 1780: IMO this test should have a comment describing what it is addressing.

It is really unintuitive that a change for something that happens in Spring Boot is fixed and tested entirely outside of Spring Boot. The best possible test case would test what actually happens when Spring Boot Dev Tools are enabled. This doesn't actually guarantee that the original problem is fixed.

triceo (approved):
Some comments. LGTM when resolved.

core/src/main/java/ai/timefold/solver/core/impl/domain/common/accessor/gizmo/GizmoClassLoader.java
Christopher-Chianelli: It returns `null` when `otherClassLoader` is `this` to prevent a recursive loop since `loadClassFrom`  is called within `loadClass`. It handles null since both the SolverConfig classloader and `Thread.getContextClassLoader` can be `null`: https://stackoverflow.com/a/3459235

PR #2285, opened 2026-05-05
perf: improve ConstraintCollectors.toList()

triceo:
The underlying collection is now optimized for removals, as those operations will be frequent inside group nodes. It still retains fast random access to elements, as that is required to reuse it in Neighborhoods.

The collection supports null values, and implements the full list interface; this makes it a suitable replacement to ArrayList.

Compared to ArrayList, remove() is super-fast if done via the entry. This introduces small overhead in addition and iteration, as the arrays may need to be compacted. Some operations compare equally poorly in both collections, such as additions into the middle of the array.

Note: this was co-developed with Claude code. It was reviewed by 2 different LLMs, as well as by me.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
@@ -1,272 +1,465 @@
 package ai.timefold.solver.core.impl.util;
 
+import java.util.AbstractList;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.function.Consumer;
 
 import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
 
 /**
- * {@link ArrayList}-backed list which allows for a cheap {@link #remove(Entry) removal of an element},
+ * {@link AbstractList} implementation which allows for a cheap {@link Entry#remove() removal of an element},
  * while still providing fast iteration and random access.
  * The order of iteration is guaranteed to be the insertion order.
+ * {@code null} is a valid value.
  * <p>
  * It uses internal state of the entry to track insertion position of the element.
- * When an element is removed, the underlying collection isn't actually touched;
+ * When an entry is removed, its slot in the underlying collection is replaced with {@code null} (a gap);
  * therefore, the insertion position of later elements isn't changed.
- * This position is called a gap.
- * Gaps are removed (the list is compacted) when {@link #forEach(Consumer)} or {@link #asList()} is called.
- * This keeps the overhead low while giving us all benefits of {@link ArrayList},
- * such as memory efficiency, random access, and fast iteration.
+ * Gaps are removed (the list is fully compacted) when {@link #forEach(Consumer)} or {@link #add(int, Object)} is called.
+ * {@link #get(int)} and related index-based operations compact only the prefix up to the requested index.
+ * This keeps the overhead low while giving us most benefits of {@link ArrayList}.
  * <p>
- * This class is very thread-unsafe.
+ * Primary fast-path methods are {@link #addEntry(Object)} and {@link Entry#remove()}, both run in O(1).
+ * All standard {@link List} methods are also available and may run in O(n) or worse.
+ * <p>
+ * This class is so very not thread safe.
  *
  * @param <T>
  */
 @NullMarked
-public final class ElementAwareArrayList<T> {
+public final class ElementAwareArrayList<T extends @Nullable Object> extends AbstractList<T> {
 
-    private final List<Entry<T>> elementList = new ArrayList<>();
+    private static final int REMOVED_POSITION = -1;
+
+    private final List<@Nullable Entry> entryList = new ArrayList<>();
     private int lastElementPosition = -1;
-    private int gapCount = 0;
+    private int gapCount = 0; // Always equals the total number of null slots in entryList.
 
     /**
-     * Appends the specified element to the end of this collection.
+     * Appends the specified element to the end of this list.
      *
-     * @param element element to be appended to this collection
+     * @return the entry for later O(1) removal via {@link Entry#remove()}
      */
-    public Entry<T> add(T element) {
-        var newEntry = new Entry<>(element, ++lastElementPosition);
-        elementList.add(newEntry);
+    public Entry addEntry(T element) {
+        modCount++;
+        if (gapCount > 0 && entryList.get(lastElementPosition) == null) { // Reuse a gap if it exists.
+            var newEntry = new Entry(element, lastElementPosition);
+            entryList.set(lastElementPosition, newEntry);
+            gapCount--;
+            return newEntry;
+        }
+        var newEntry = new Entry(element, ++lastElementPosition);
+        entryList.add(newEntry);
         return newEntry;
     }
 
-    public Entry<T> get(int index) {
-        compact();
-        return elementList.get(index);
+    private Entry getEntry(int index) {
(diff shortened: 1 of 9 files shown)


triceo:
Converting back to a draft for another experiment.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -1,274 +1,449 @@
...
-     * Removes the first occurrence of the specified element from this collection, if present.
-     *
-     * @param entry entry to be removed from this collection
-     * @throws IllegalStateException if the element wasn't found in this collection
-     */
-    public void remove(Entry<T> entry) {
-        if (entry.isRemoved()) {
-            throw new IllegalStateException("The entry (%s) was already removed."
-                    .formatted(entry));
+    private Entry partialCompact(int rightBoundaryPosition) {
```
Christopher-Chianelli, line 76: I think if `partialCompact` is ever called on the last element on a list without gaps, an exception will be raised since `rightBoundaryPosition + 1 = lastElementPosition + 1 = list.size()`  cannot be used as the inclusive start index. Might want to just return `entryList.get(rightBoundaryPosition)` if `gapCount` is 0.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -1,274 +1,449 @@
...
+        for (var currentPosition = 0; currentPosition <= lastElementPosition; currentPosition++) {
+            var entry = entryList.get(currentPosition);
+            if (entry == null) {
+                continue;
+            }
+            elementConsumer.accept(entry.element());
+            if (currentPosition != compactPosition) {
+                entry.moveTo(compactPosition);
+                entryList.set(compactPosition, entry);
+                entryList.set(currentPosition, null); // Prevent stale data.
```
Christopher-Chianelli, line 268: Since the tail of the `entryList` is cleared after visiting every present element, doesn't the clear remove stale data?

Christopher-Chianelli (commented):
Changes look good upon initial review. Any benchmarks?

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: Technically it does, but I opted for the list to be consistent at all times - if something is moved to a new place, it should be removed from the old place. The subsequent `clear()` will remove the tail gaps, making the list be the exact size it needs to be.

triceo:
> Changes look good upon initial review. Any benchmarks?

I did benchmark some target use cases. First, `n` additions to the end of the list:

    Benchmark                            (gapPercent)  (size)  Mode  Cnt     Score     Error  Units
    arrayList                                     N/A      10  avgt    4     0.017 ±   0.001  us/op
    arrayList                                     N/A     100  avgt    4     0.188 ±   0.008  us/op
    arrayList                                     N/A    1000  avgt    4     1.758 ±   0.209  us/op
    arrayList                                     N/A   10000  avgt    4    19.007 ±   0.647  us/op
    eaal                                          N/A      10  avgt    4     0.045 ±   0.005  us/op
    eaal                                          N/A     100  avgt    4     0.304 ±   0.018  us/op
    eaal                                          N/A    1000  avgt    4     2.937 ±   0.243  us/op
    eaal                                          N/A   10000  avgt    4    30.550 ±   3.088  us/op

As you can see, the entry allocation and the extra indirection adds some overhead. This is irrelevant though, because in the context of CS/neighborhoods, additions never happen alone. We also do frequent removals (undo) from the middle of the array (benchmark removes all elements, but in random order):

    arrayList                                N/A      10  avgt    4     0.050 ±   0.001  us/op
    arrayList                                N/A     100  avgt    4     0.719 ±   0.014  us/op
    arrayList                                N/A    1000  avgt    4    21.370 ±   2.931  us/op
    arrayList                                N/A   10000  avgt    4  1925.237 ±  12.879  us/op
    eaal                                     N/A      10  avgt    4     0.059 ±   0.050  us/op
    eaal                                     N/A     100  avgt    4     0.406 ±   0.059  us/op
    eaal                                     N/A    1000  avgt    4     3.183 ±   1.212  us/op
    eaal                                     N/A   10000  avgt    4    46.344 ±  10.818  us/op

If we assume that vast majority of additions are followed with a removal, the small overhead of additions is insignificant. But removals create gaps that need to be compacted later, which is why I decided to benchmark randomly removing half the elements from the array + subsequent full iteration:

    arrayList           N/A      10  avgt    4     0.034 ±   0.002  us/op
    arrayList           N/A     100  avgt    4     0.453 ±   0.037  us/op
    arrayList           N/A    1000  avgt    4    19.874 ±   0.237  us/op
    arrayList           N/A   10000  avgt    4  2018.050 ± 512.852  us/op
    eaal                N/A      10  avgt    4     0.025 ±   0.010  us/op
    eaal                N/A     100  avgt    4     0.210 ±   0.016  us/op
    eaal                N/A    1000  avgt    4     3.191 ±   0.120  us/op
    eaal                N/A   10000  avgt    4    28.774 ±   5.234  us/op

Just to be sure, I also benchmarked raw iteration, no gaps:

    arrayList                           N/A      10  avgt    4     0.004 ±   0.001  us/op                                                               
    arrayList                           N/A     100  avgt    4     0.022 ±   0.001  us/op                                                               
    arrayList                           N/A    1000  avgt    4     0.210 ±   0.001  us/op                                                               
    arrayList                           N/A   10000  avgt    4     6.621 ±   3.537  us/op                                                               
    eaal                                N/A      10  avgt    4     0.008 ±   0.001  us/op                                                               
    eaal                                N/A     100  avgt    4     0.064 ±   0.005  us/op                                                               
    eaal                                N/A    1000  avgt    4     0.631 ±   0.087  us/op                                                               
    eaal                                N/A   10000  avgt    4     8.338 ±   0.982  us/op     

Iteration when compact is slower, probably on account of the `Entry` indirection. (Although the error is high here; I only ran 2 forks and 2 iterations each, otherwise this would have taken a day.)

Overall though, there is no beating `ElementAwareArrayList` if you ever need to do removals from the middle. There are other operations which are likely to perform poorly (additions in the middle etc.), but those perform poorly in `ArrayList` as well, and they are not the target here; every kind of data structure has a downside.

(Ignore the `gapPercent` column - that is not used in any of these benchmarks.)

Christopher-Chianelli (approved):
LGTM.

Christopher-Chianelli:
It seems a comment I made got loss, but in `forEachCompacting`, I am pretty sure the second if is impossible since the first if is always taken if the second if condition was true.

Christopher-Chianelli:
(line 261, GitHub not allowing me to comment)

Christopher-Chianelli:
Also, does `remove` and `add` preserving the original list order matter? If we only care about random access (but not the indicies), there a neat trick we can do: when an element is removed, swap it with the tail of the list. This invalidates only the position of the tail and eliminates the need for compacting, but it no longer folllows proper list semantics. Since inserting at a particular index does not really make sense since indices can be invalidated, you can only add to the end.

```
l = [1, 2, 3, 4]
l.remove(2)
// l = [1, 4, 3]
l.remove(1)
// l = [3, 4]
```

triceo:
> It seems a comment I made got loss, but in `forEachCompacting`, I am pretty sure the second if is impossible since the first if is always taken if the second if condition was true.

That method (`clearIfPossible`) is weirdly named. It does clear the list if possible, but it returns true if the list _is compact_ - which means either empty, or non-empty but with no gaps. Which is why the second condition still makes sense. I refactored it slightly now to make the code easier to understand.

> Also, does `remove` and `add` preserving the original list order matter? If we only care about random access (but not the indicies), there a neat trick we can do: when an element is removed, swap it with the tail of the list.

I originally had it implemented like that, but then I decided that insertion order is important. You expect that out of a list.

PR #2300, opened 2026-05-11
chore: fix value-range validation with null values

### LINKED ISSUE #2273: ValueRange should support SortedSet (regression?)
> AbstractFromPropertyValueRangeDescriptor<>.processValueRange() seems to call (set.contains(null)). This appears directly in the source code, the null is not being derived from somewhere else. On the standard Kotlin sorted set, which is a TreeSet, that causes an immediate exception if the set is using natural ordering. That means, it's necessary to write a custom comparator that sorts nulls, even if the ValueRangeProvider that offers the set can never include null. 
> Could this be avoided by checking the comparator() of the sortedset, and if it is null, then null cannot possibly be a member of the set (it cannot be added if there is no custom comparator for the set)

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/domain/valuerange/descriptor/AbstractFromPropertyValueRangeDescriptor.java
```diff
@@ -120,7 +120,11 @@ protected <Value_> ValueRange<Value_> processValueRange(Object valueRangeObject,
                             Use SortedSet or LinkedHashSet to ensure solver reproducibility.
                             """
                             .formatted(ValueRangeProvider.class.getSimpleName(), memberAccessor, bean, set.getClass()));
-                } else if (set.contains(null)) {
+                } else if (!(collection instanceof SortedSet<Value_>) && set.contains(null)) {
```
triceo: I know that exceptions should not be used for control flow, but I wonder if we should still call contains and catch the NPE - because the fact that something is a SortedSet is not a guarantee of anything. Any type can decide whether it accepts null or not, and we have no control over it.

Either we do this, or we remove this fail-fast altogether. Not sure which is best.

triceo (commented):
Not sure about this one.

triceo (approved):

PR #2301, opened 2026-05-12
fix: entities use equality comparison

### LINKED ISSUE #2274: Auto-upgrader should recommend changing HashSet to SequencedSet, not SortedSet
To avoid triggering https://github.com/TimefoldAI/timefold-solver/issues/2273
the OpenRewrite recipe shouldn't recommend SortedSet, but SequencedSet instead.

> I was previously just using HashSets. After the upgrade, the text said I needed to use SortedSet instead to run reproducibly.
> It appeared on the console when I ran the version that used HashSets after upgrading.

triceo:
Removes identity comparisons from Constraint Streams, since that is a bug which we recently fixed for value ranges as well.
Also improves Javadoc around entities, facts and collections.

Downstream bug fix: https://github.com/TimefoldAI/timefold-quickstarts/pull/1077#pullrequestreview-4269937228


zepfred (approved):
LGTM!

PR #2305, opened 2026-05-13
perf: perform shadow variable update exactly once

### LINKED ISSUE #2303: perf: shadow variable update is unneccessary called multiple times
**Describe the bug**
For my problem it is not possible to achieve the same speed for timefold 2.0.0 and timefold 1.3.3.
Reason: The handler method is always called again when the shadow variable field changed.

[TestOld.java](https://github.com/user-attachments/files/27656985/TestOld.java)
[TestNew.java](https://github.com/user-attachments/files/27656986/TestNew.java)

**Expected behavior**
It should be possible to get the same performance as before in timefold 2.0.0.
A handler method should not be called twice in a row although the shadow sources did not change.

**Actual behavior**
The execution time for timefold 2.0.0 is twice as large.
The handler method is always called again when it was found that the shadow variable value has changed.

**To Reproduce**
I made a one file example for the old and new behavior:

Christopher-Chianelli:
There were a couple of inefficiently discovered

- When a declarative shadow variable changes, it goes through the same plumbing as other variables. This is wasteful, since by definition, declarative shadow variables cannot affect the structure of the graph, and the graphs already modify the queues directly.

- When group alignment is used in a fixed graph, the same entity's update might be called multiple times. This is because the change set is a priority queue, which allows duplicate elements.

Now:

- We shortcut before/after variable change to return immediately if the graph is currently updating

- For fixed graphs, we have a seperate bitset to ensure the same node cannot be added to the changed queue multiple times.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
@@ -16,7 +16,7 @@ import ai.timefold.solver.core.preview.api.domain.metamodel.VariableMetaModel;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 
-public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeSet_> implements VariableReferenceGraph
+public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeTracker_> implements VariableReferenceGraph
         permits DefaultVariableReferenceGraph, FixedVariableReferenceGraph {
 
     // These structures are immutable.
@@ -28,9 +28,13 @@ public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeSet
 
     // These structures are mutable.
     protected final DynamicLinearProbeNonNegativeIntCounter[] edgeCount;
-    protected final ChangeSet_ changeSet;
+    protected final ChangeTracker_ changeTracker;
     protected final TopologicalOrderGraph graph;
 
+    /**
+     * True if we are currently doing a declarative shadow variable update,
+     * false otherwise.
+     */
     protected boolean isUpdating;
 
     AbstractVariableReferenceGraph(VariableReferenceGraphBuilder<Solution_> outerGraph,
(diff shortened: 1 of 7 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
```diff
@@ -64,6 +67,17 @@ public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeSet
         }
     }
 
+    abstract void innerUpdateChanged();
+
+    abstract void markChanged(GraphNode<Solution_> changed);
+
+    @Override
+    public final void updateChanged() {
```
triceo, line 93: IMO for the proper understanding of the algorithm, this method (+ isUpdating field + markChanged) should get some comments. Specifically - what is referenced by "changed", when is this called, what does this mean, and why does this allow us to perform the subsequent optimization.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultTopologicalOrderGraph.java
```diff
@@ -84,10 +84,9 @@ public void forEachEdge(EdgeConsumer edgeConsumer) {
 
     @Override
     public PrimitiveIterator.OfInt nodeForwardEdges(int fromNode) {
-        return componentMap.get(fromNode).stream()
-                .flatMap(member -> forwardEdges[member].stream())
+        return forwardEdges[fromNode].stream()
```
triceo, line 87: Suddenly much less nodes. Why?
It looks as if the previous code was not only inefficient, it was incorrect; including more nodes than necessary.

triceo (approved):
I like it! LGTM when comments resolved.

I started the microbenchmark suite; I may have some questions after I see the results.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultTopologicalOrderGraph.java
Christopher-Chianelli: IIRC, before the session updater update by components, but now it updates by node. I notice the mismatch since enterprise already does this.

Christopher-Chianelli: (i.e. enterprise already just return the forward edges for the node not the component)

PR #2306, opened 2026-05-14
feat: implement direct testing of score/impact of constraints

### LINKED ISSUE #2007: Constraint testing: score generated by a single contraint
**Is your feature request related to a problem? Please describe.**
I am trying to unit test that a constraint in situation X generates a (better/worse/different) penalty than in situation Y. I would prefer not to hard code expected penalty values, since these may change, just that two situations are different.
However, a `SingleConstraintVerification` only provides methods to compare with known values: `penalizes(long times)`, `penalizedBy(long matchWeightTotal)`, `penalizesByMoreThan(String s, long l)`, etc.

**Describe the solution you'd like**
There are two possible solutions I can think of, which would both work for me:

**Comparing two solutions through the constraintVerifier**
`constraintVerifier.verifyThat(someConstraint).givenSolution(Solution_ someSolution).penalizesByMoreThan(Solution_ someOtherSolution)`

**Being able to obtain a score through the constraint verifier**
Although I realise the API might be a little confusing in my example below (getting a score from a `ConstraintVerifier` ) something like `constraintVerifier.verifyThat(someConstraint).givenSolution(Solution_ someSolution).score()` would allow me to write the assertion myself:
```
HardSoftScore score1 = constraintVerifier.verifyThat(someConstraint).givenSolution(Solution_ someSolution).score()
HardSoftScore score2 = constraintVerifier.verifyThat(someConstraint).givenSolution(Solution_ otherSolution).score()

assertTrue(score1.compareTo(score2) > 0)
```

**Describe alternatives you've considered**
I am currently working through the `SolutionManager` with `solutionManager.analyse(solution)` and comparing those scores. However, this includes all constraints not just the one I'm trying to test, so this feels more fragile.

PatrickDiallo23:
This PR is introducing the capability to get the score/impact of the constraints given a set of entities or an entire solution addressing the #2007 .

Even though it was proposed to add in `ConstraintVerifier` the *getScore* and *getImpact* methods, I observed that it was counterintuitive for the user to use such methods because it would change the chaining method (also making the implementation more complex than it should be) to `constraintVerifier.getScore().forConstraint(constraint).given(entities)`.

I was thinking of providing a new chaining to obtain the score like `constraintVerifier.evaluate(constraint).given(entities).score()`, where `evaluate()` would be the same as `verifyThat()` resulting in duplicated code which is not a good idea.

So I used the solution provided by the author of the issue, which is the most efficient and straightforward solution that allows writing custom assertions to compare different solutions or subsets of solutions.


core/src/main/java/ai/timefold/solver/core/api/score/stream/test/MultiConstraintAssertion.java
```diff
@@ -27,4 +27,28 @@ default void scores(@NonNull Score<?> score) {
...
+    /**
+     * Returns the {@link Score} produced by all constraints in the {@link ConstraintProvider} for the given set of facts.
+     * <p>
+     * Unlike {@link #scores(Score)}, this method does not perform any assertion.
+     * Instead, it returns the raw score, allowing the caller to compare scores
+     * between different scenarios without hard-coding expected values.
+     * <p>
+     * Usage example:
+     *
+     * <pre>{@code
```
triceo: Since Java 18, we now have the capability of doing `@snippet` - see https://www.baeldung.com/java-doc-code-snippets. This is a much better way of showing code, because it will get its syntax validated etc.

Please convert these. (No need for external snippets.)

core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
```diff
@@ -649,4 +650,56 @@ default void rewardsLessThan(long times) {
...
+    /**
+     * Returns the {@link Score} produced by the {@link Constraint} being tested for the given set of facts.
+     * <p>
+     * Unlike assertion methods such as {@link #penalizesBy(int)} or {@link #rewardsWith(int)},
+     * this method does not perform any assertion. Instead, it returns the raw score, allowing the caller
+     * to compare scores between different scenarios without hard-coding expected values.
+     * <p>
+     * Usage example:
+     * 
+     * <pre>{@code
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
```diff
@@ -649,4 +650,56 @@ default void rewardsLessThan(long times) {
...
+     * Unlike {@link #score()}, which returns the full score (match weight × constraint weight),
+     * this method returns only the <b>match weight</b> — the same number you would pass
+     * to assertion methods like {@link #penalizesBy(int)} or {@link #rewardsWith(int)}.
+     * <p>
+     * This is useful for comparing the <em>severity</em> of a constraint across scenarios
+     * without needing to know the constraint weight.
+     * <p>
+     * Usage example:
+     *
+     * <pre>{@code
```
triceo: Dtto.

triceo (commented):
I agree that the solution presented here is the better of the proposals. 

Let's mention this feature in the docs where we talk about constraint testing - I envision a new subsection, something to the tune to "Employing custom assertions", showing how to use this feature with `AssertJ`. (Much like the Javadoc snippets already do.)

triceo (approved):

triceo:
Thanks, @PatrickDiallo23!

greyhairredbear:
Quite late to the party, but happy to see this merged, thanks @PatrickDiallo23!

PR #2314, opened 2026-05-22
Assorted fixes

### LINKED ISSUE #2287: Improve error message: misconfigured neighborhood feature.
**Describe the bug**
When configuring the feature as in our docs: 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<solver xmlns="https://timefold.ai/xsd/solver" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="https://timefold.ai/xsd/solver https://timefold.ai/xsd/solver/solver.xsd">
    <enablePreviewFeature>NEIGHBORHOODS</enablePreviewFeature>
    <localSearch>
        <neighborhoodProviderClass>com.acme.MyNeighborhoodProvider</neighborhoodProviderClass>
    </localSearch>
</solver>
```

But the class `com.acme.MyNeighborhoodProvider` doesn't exist, I get a vague error. 

**Expected behavior**
Better error message: something like "configured Neighborhood Provider could not be found?" 

**Actual behavior**

On first solve attempt: 

```
2026-05-05 15:13:35,556 ERROR [org.acme.vehiclerouting.rest.VehicleRoutePlanResource] (pool-8-thread-1) Failed solving jobId (6e571640-8e92-41d3-afbf-ac9e59d0435e).: java.lang.IndexOutOfBoundsException: Index 11 out of bounds for length 9
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
        at java.base/java.util.Objects.checkIndex(Objects.java:365)
        at java.base/java.util.ArrayList.get(ArrayList.java:428)
        at ai.timefold.solver.core.impl.util.ElementAwareArrayList.get(ElementAwareArrayList.java:49)
        at ai.timefold.solver.core.impl.bavet.common.index.DefaultUniqueRandomIterator.hasNext(DefaultUniqueRandomIterator.java:62)
        at ai.timefold.solver.core.impl.bavet.common.index.FilteredUniqueRandomIterator.hasNext(FilteredUniqueRandomIterator.java:40)
        at ai.timefold.solver.core.impl.neighborhood.stream.BiRandomMoveIterator.pickNextMove(BiRandomMoveIterator.java:109)
        at ai.timefold.solver.core.impl.neighborhood.stream.BiRandomMoveIterator.hasNext(BiRandomMoveIterator.java:85)
        at ai.timefold.solver.core.impl.neighborhood.RandomOrderNeighborhoodIterator.hasNext(RandomOrderNeighborhoodIterator.java:32)
        at ai.timefold.solver.core.impl.localsearch.decider.LocalSearchDecider.decideNextStep(LocalSearchDecider.java:93)
        at ai.timefold.solver.core.impl.localsearch.DefaultLocalSearchPhase.solve(DefaultLocalSearchPhase.java:89)

```

**To Reproduce**
Add the xml above in a solverconfig.xml to any project. 

## Environment

**Timefold Solver Version or Git ref**:  Using Timefold Solver Community Edition v2.0.0.

**Output of `java -version`:** 21

triceo:
See attached issue for a description of what was wrong.
Independent bugs, independent commits. Will not be squashed.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/score/stream/test/MultiConstraintAssertion.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/stream/test/MultiConstraintAssertion.java
@@ -3,9 +3,10 @@ package ai.timefold.solver.core.api.score.stream.test;
 import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.score.stream.ConstraintProvider;
 
-import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+@NullMarked
 public interface MultiConstraintAssertion {
 
     /**
@@ -14,7 +15,7 @@ public interface MultiConstraintAssertion {
      * @param score total score calculated for the given set of facts
      * @throws AssertionError when the expected score does not match the calculated score
      */
-    default void scores(@NonNull Score<?> score) {
+    default void scores(Score<?> score) {
         scores(score, null);
     }
 
@@ -25,7 +26,7 @@ public interface MultiConstraintAssertion {
      * @param message description of the scenario being asserted
      * @throws AssertionError when the expected score does not match the calculated score
      */
-    void scores(@NonNull Score<?> score, @Nullable String message);
+    void scores(Score<?> score, @Nullable String message);
 
     /**
      * Returns the {@link Score} produced by all constraints in the {@link ConstraintProvider} for the given set of facts.
@@ -39,16 +40,15 @@ public interface MultiConstraintAssertion {
      * {@snippet :
      * HardSoftScore scoreA = constraintVerifier.verifyThat()
      *         .givenSolution(solutionA)
-     *         .score();
+     *         .getScore();
      * HardSoftScore scoreB = constraintVerifier.verifyThat()
      *         .givenSolution(solutionB)
-     *         .score();
+     *         .getScore();
      * assertThat(scoreA).isGreaterThan(scoreB);
      * }
      *
      * @return the score produced by all constraints for the given facts, never null
      */
-    @NonNull
-    <S extends Score<S>> S score();
+    <S extends Score<S>> S getScore();
 
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
@@ -7,9 +7,10 @@ import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.score.stream.Constraint;
 import ai.timefold.solver.core.api.score.stream.ConstraintJustification;
 
-import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+@NullMarked
 public interface SingleConstraintAssertion {
 
     /**
@@ -19,9 +20,8 @@ public interface SingleConstraintAssertion {
      * @param message description of the scenario being asserted
      * @throws AssertionError when the expected penalty is not observed
      */
-    @NonNull
     SingleConstraintAssertion justifiesWith(@Nullable String message,
-            @NonNull ConstraintJustification @NonNull... justifications);
+            ConstraintJustification... justifications);
 
     /**
      * Asserts that the {@link Constraint} being tested, given a set of facts, results in a given
@@ -30,7 +30,7 @@ public interface SingleConstraintAssertion {
      * @param justifications the expected justifications.
      * @throws AssertionError when the expected penalty is not observed
      */
-    default @NonNull SingleConstraintAssertion justifiesWith(@NonNull ConstraintJustification @NonNull... justifications) {
+    default SingleConstraintAssertion justifiesWith(ConstraintJustification... justifications) {
         return justifiesWith(null, justifications);
     }
 
@@ -41,9 +41,8 @@ public interface SingleConstraintAssertion {
      * @param message description of the scenario being asserted
      * @throws AssertionError when the expected penalty is not observed
      */
-    @NonNull
     SingleConstraintAssertion justifiesWithExactly(@Nullable String message,
-            @NonNull ConstraintJustification @NonNull... justifications);
+            ConstraintJustification... justifications);
 
     /**
      * Asserts that the {@link Constraint} being tested, given a set of facts, results in a given
@@ -52,8 +51,8 @@ public interface SingleConstraintAssertion {
      * @param justifications the expected justifications.
      * @throws AssertionError when the expected penalty is not observed
      */
-    default @NonNull SingleConstraintAssertion
-            justifiesWithExactly(@NonNull ConstraintJustification @NonNull... justifications) {
+    default SingleConstraintAssertion
+            justifiesWithExactly(ConstraintJustification... justifications) {
         return justifiesWithExactly(null, justifications);
     }
 
@@ -126,7 +125,7 @@ public interface SingleConstraintAssertion {
      * @param matchWeightTotal at least 0, expected sum of match weights of matches of the constraint.
      * @throws AssertionError when the expected penalty is not observed
      */
-    default void penalizesBy(@NonNull BigDecimal matchWeightTotal) {
+    default void penalizesBy(BigDecimal matchWeightTotal) {
         penalizesBy(null, matchWeightTotal);
     }
 
@@ -137,7 +136,7 @@ public interface SingleConstraintAssertion {
      * @param matchWeightTotal at least 0, expected sum of match weights of matches of the constraint.
      * @throws AssertionError when the expected penalty is not observed
      */
-    void penalizesBy(@Nullable String message, @NonNull BigDecimal matchWeightTotal);
+    void penalizesBy(@Nullable String message, BigDecimal matchWeightTotal);
 
     /**
      * Asserts that the {@link Constraint} being tested, given a set of facts, results in a given number of penalties.
@@ -232,7 +231,7 @@ public interface SingleConstraintAssertion {
      * @param matchWeightTotal at least 0, expected sum of match weights of matches of the constraint.
      * @throws AssertionError when the expected reward is not observed
      */
-    default void rewardsWith(@NonNull BigDecimal matchWeightTotal) {
+    default void rewardsWith(BigDecimal matchWeightTotal) {
         rewardsWith(null, matchWeightTotal);
     }
 
@@ -243,7 +242,7 @@ public interface SingleConstraintAssertion {
      * @param matchWeightTotal at least 0, expected sum of match weights of matches of the constraint.
      * @throws AssertionError when the expected reward is not observed
      */
-    void rewardsWith(@Nullable String message, @NonNull BigDecimal matchWeightTotal);
+    void rewardsWith(@Nullable String message, BigDecimal matchWeightTotal);
 
     /**
      * Asserts that the {@link Constraint} being tested, given a set of facts, results in a given number of rewards.
@@ -341,7 +340,7 @@ public interface SingleConstraintAssertion {
      * @param matchWeightTotal at least 0, expected sum of match weights of matches of the constraint.
      * @throws AssertionError when the expected penalty is not observed
      */
-    default void penalizesByMoreThan(@NonNull BigDecimal matchWeightTotal) {
+    default void penalizesByMoreThan(BigDecimal matchWeightTotal) {
         penalizesByMoreThan(null, matchWeightTotal);
     }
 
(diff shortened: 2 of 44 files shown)


core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/placer/QueuedValuePlacerConfig.java
```diff
@@ -30,7 +30,7 @@ public final class QueuedValuePlacerConfig extends EntityPlacerConfig<QueuedValu
 
     public static final String XML_ELEMENT_NAME = "queuedValuePlacer";
 
-    private Class<?> entityClass = null;
+    private String entityClass = null;
```
zepfred, line 33: Why are we changing the type? This change is inconsistent with other configuration classes that use the `Class` type. Can we solve this issue without changing the type? Have you validated the XML parsing with this modification?

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/common/nearby/NearbySelectionConfig.java
```diff
@@ -44,7 +44,7 @@ public final class NearbySelectionConfig extends SelectorConfig<NearbySelectionC
     private SubListSelectorConfig originSubListSelectorConfig = null;
     @XmlElement(name = "originValueSelector")
     private ValueSelectorConfig originValueSelectorConfig = null;
-    private Class<? extends NearbyDistanceMeter> nearbyDistanceMeterClass = null;
+    private String nearbyDistanceMeterClass = null;
```
zepfred, line 47: I still don't understand why we need to change the variable type, but I've noticed that you've modified several configuration classes. However, there are still some classes that are using the type `Class`. If changing the type resolves the issue, should we consider changing all related variables to type String?

Refer to the `MoveSelectorConfig` for an example of how to use the Class type.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/value/ValueSelectorConfig.java
```diff
@@ -234,7 +235,7 @@ public ValueSelectorConfig withMimicSelectorRef(String mimicSelectorRef) {
     }
 
     public ValueSelectorConfig withDowncastEntityClass(Class<?> entityClass) {
-        this.setDowncastEntityClass(entityClass);
+        this.downcastEntityClass = entityClass.getName();
```
zepfred, line 238: This pattern repeats across all modified configuration classes. I understand that we should be passing a null value here, but if we do, it won't have the same behavior as the set method, and I wonder if we should check for null before getting the class name. I believe we still check for nullity even when the class is marked with @NullMarked in some cases.

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
```diff
@@ -45,6 +45,26 @@ public class ConfigUtils {
 
     private static final AlphabeticMemberComparator alphabeticMemberComparator = new AlphabeticMemberComparator();
 
+    @SuppressWarnings("unchecked")
```
zepfred, line 48: If we want to ensure that a class is present in the classpath during deserialization and provide a helpful message, could we instead add a custom deserializer for the type `Class` to avoid changing all the configuration classes?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -101,10 +108,13 @@ Please refactor your code (%s) to use the new Move API."""
     }
 
     private boolean pickNextMove(UniTuple<A> leftTuple) {
```
zepfred, line 109: Upon reading this method again, I wonder if we should return `true` only if there is a next move. The method's name suggests that it should return `true` if a next move exists.

quarkus-integration/quarkus/reflection-integration-test/pom.xml
```diff
@@ -36,7 +36,7 @@
     <!-- test dependencies -->
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-junit5</artifactId>
+      <artifactId>quarkus-junit</artifactId>
```
zepfred, line 39: Why are we making this change again? Was anything missing from the original related PR?

zepfred (commented):
I have some questions regarding the variable type changes. If we want to proceed with the current approach, we should double-check all configuration classes and ensure that the same approach is used.

core/src/main/java/ai/timefold/solver/core/config/constructionheuristic/placer/QueuedValuePlacerConfig.java
triceo: Hmm, I see I did miss some classes. Will do the same change there as well.

This is a necessary change - JAXB is stupid. If the class doesn't exist, unmarshalling automatically converts it to `null`. No exception, no warning, just a silent `null`. This doesn't happen when the class is represented as a `String`. I tried everything that a marshaller can do - listener, adapter, nothing helped.

XML is entirely unaffected. The XSD always had classes as `string` - which is why you do not see any changes to the XSD.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/common/nearby/NearbySelectionConfig.java
triceo: Explained above.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/value/ValueSelectorConfig.java
triceo: I was considering it. If you do null checks when `@NullMarked`, you get warnings.

If you do not do null checks when `@NullMarked`, you are ignoring a clear instruction and in that case, you get an NPE. If I did catch that NPE in this method and threw a different exception, nothing would meaningfully change; you'd still get an exception. Just additional boilerplate code.

core/src/main/java/ai/timefold/solver/core/config/util/ConfigUtils.java
triceo: Doesn't work. `Class<?>` first needs to be loaded to be checked; `String` does not.

quarkus-integration/quarkus/reflection-integration-test/pom.xml
triceo: There was no original PR in this repo. The other PRs you've seen have been for other repos.

core/src/main/java/ai/timefold/solver/core/api/score/stream/test/MultiConstraintAssertion.java
```diff
@@ -48,7 +49,6 @@ default void scores(@NonNull Score<?> score) {
      *
      * @return the score produced by all constraints for the given facts, never null
      */
-    @NonNull
-    <S extends Score<S>> S score();
+    <S extends Score<S>> S getScore();
```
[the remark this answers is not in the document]
triceo: Not relevant; the original API was not yet released.

core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
```diff
@@ -671,13 +670,12 @@ default void rewardsLessThan(long times) {
      *
      * @return the score produced by this single constraint for the given facts, never null
      */
-    @NonNull
-    <S extends Score<S>> S score();
+    <S extends Score<S>> S getScore();
```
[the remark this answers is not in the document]
triceo: Not relevant; the original API was not yet released.

core/src/main/java/ai/timefold/solver/core/api/score/stream/test/SingleConstraintAssertion.java
```diff
@@ -699,7 +697,6 @@ default void rewardsLessThan(long times) {
      * @return the match weight total produced by this single constraint, never null.
      *         Positive for rewards, negative for penalties, zero when there is no impact.
      */
-    @NonNull
-    Number impact();
+    Number getImpact();
```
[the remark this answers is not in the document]
triceo: Not relevant; the original API was not yet released.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -364,27 +365,31 @@ public void setMoveThreadBufferSize(@Nullable Integer moveThreadBufferSize) {
...
+        if (entityClassList == null) {
+            return null;
+        }
+        return ConfigUtils.resolveClasses(entityClassList, "entityClass", this);
     }
 
     public void setEntityClassList(@Nullable List<Class<?>> entityClassList) {
-        this.entityClassList = entityClassList;
+        this.entityClassList = entityClassList == null ? null
+                : entityClassList.stream().map(Class::getName).collect(Collectors.toList());
```
zepfred, line 392: This list can be made immutable: `toList()`.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -478,22 +483,22 @@ public void setMonitoringConfig(@Nullable MonitoringConfig monitoringConfig) {
...
 
     public @NonNull SolverConfig withSolutionClass(@NonNull Class<?> solutionClass) {
-        this.solutionClass = solutionClass;
+        this.solutionClass = solutionClass.getName();
         return this;
     }
 
     public @NonNull SolverConfig withEntityClassList(@NonNull List<Class<?>> entityClassList) {
-        this.entityClassList = entityClassList;
+        this.entityClassList = entityClassList.stream().map(Class::getName).collect(Collectors.toList());
```
zepfred, line 496: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/test/AbstractMultiConstraintAssertion.java
```diff
@@ -59,7 +59,7 @@ public void scores(@NonNull Score<?> score, String message) {
 
     @Override
     @SuppressWarnings("unchecked")
-    public <S extends Score<S>> S score() {
+    public <S extends Score<S>> S getScore() {
```
zepfred, line 62: Should we add a migration recipe?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/test/AbstractSingleConstraintAssertion.java
```diff
@@ -254,13 +254,13 @@ public void rewardsWithLessThan(@Nullable String message, @NonNull BigDecimal ma
 
     @Override
     @SuppressWarnings("unchecked")
-    public <S extends Score<S>> S score() {
+    public <S extends Score<S>> S getScore() {
```
zepfred, line 257: Should we add a migration recipe?

core/src/main/java/ai/timefold/solver/core/impl/score/stream/test/AbstractSingleConstraintAssertion.java
```diff
@@ -254,13 +254,13 @@ public void rewardsWithLessThan(@Nullable String message, @NonNull BigDecimal ma
...
         ensureInitialized();
         return (S) actualScore.raw();
     }
 
     @Override
-    public Number impact() {
+    public Number getImpact() {
```
zepfred, line 263: Ditto.

zepfred (approved):
Do we know whether the type change affects the Quarkus or Spring modules?

For example, we are still using the type `Class` in `SolverBuildTimeConfig#nearbyDistanceMeterClass`.

Other than that, the change looks good to me.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/test/AbstractMultiConstraintAssertion.java
triceo: There is nothing to migrate. This API was not released yet.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/test/AbstractSingleConstraintAssertion.java
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
triceo: I could, but I won't. This could be backwards incompatible behavior; somebody setting a mutable list and then trying to mutate the result would be an issue.

triceo:
There is no indication that this change is backwards incompatible in any way. Quarkus tests pass, Quickstarts work without any changes.

PR #2321, opened 2026-05-29
feat: Add effective map location to metadata events

diogodanielsoaresferreira:
Fetch resolved map location from enricher and set it in metadata, so that events have access to the resolved map location.


mswiderski (approved):

rsynek (approved):
Thanks! Good to merge after exploring if we can introduce some reasonable test coverage the SonarCloud complains about.

To be fair, there are tests in the other related PR.

rsynek (approved):

PR #2322, opened 2026-05-29
ci: update release workflow

zepfred:
This PR updates the release workflow:

1. Deleted .github/workflows/release.yml — removed the old GitHub Actions release workflow.
2. Deleted jreleaser.yml (renamed to jreleaser-maven.yml, then removed along with jreleaser-jfrog.yml).
3. Added `.github/scripts/change_versions.sh` — new reusable script that updates the Maven project version.
4. Updated `finish_release.yml` — replaced inline `mvn versions:set` call with the new `change_versions.sh` script, and fixed a shell injection risk by passing `${{ github.ref }}` through an env.


triceo:
My recommendation:

- Do not move away from the YAMLs. The XML configuration is both harder to read and harder to write; XML is just not a good format for this, IMO.
- I'd prefer if the configs lived in the enterprise repository. Surely JReleaser can accept the config file location via a system property.

zepfred:
> My recommendation:
> 
> * Do not move away from the YAMLs. The XML configuration is both harder to read and harder to write; XML is just not a good format for this, IMO.
> * I'd prefer if the configs lived in the enterprise repository. Surely JReleaser can accept the config file location via a system property.

There is one issue with the YAML files. I couldn't find a way to switch the `working-dir` when updating multiple projects. For different config files, you can specify a configuration file argument in the JReleaser action indeed.

triceo:
If at all possible, I want _everything related to a release_ located in Enterprise. I have a good reason for it - our releases require secrets, and the community repository allows untrusted forks. This is ultimately a bad situation, and causes a lot of compromises in the CI, and constant worries if the mitigations are enough or if we're still at risk of leaking anything.

Another reason is that our release process is nobody's business - people needn't know anything about JFrog, or the two-step process. By exposing this information here, you are giving adversaries clues for possible attacks.

I think the following should happen:

- All YAMLs are in Enterprise repo.
- All workflows are in Enterprise repo.
- No release-related configuration exists in the Community repo.

This is perfectly achievable. You do not need to run JReleaser as a Maven plugin. You can run JReleaser as a command-line app - just like we did until now. This will allow you to centralize all config.

rsynek:
@triceo me and @zepfred agreed to split the work into two parts:

1. finish the 2-phase release with reusing the same binaries
2. move everything to the enterprise repo

Is it ok to merge it the current way, try the release, and then finish 2, or do you prefer solving 2 before merging anything?

triceo:
Personally, I see doing a release with something that will never be used again as a waste of time. But if you insist, I won't stop you.

.github/workflows/release.yml
```diff
@@ -47,45 +57,113 @@ jobs:
...
+      ######
+      # 3 - Clean up the release branch when deploying in Maven or clean up the tag instead
+      ######
+      - name: Delete release branch from timefold-solver (if exists)
+        if: ${{ inputs.deploy-to-maven }}
         continue-on-error: true
         run: git push -d origin $RELEASE_BRANCH_NAME
 
-      - name: Create release branch and switch to it
+      - name: Delete tag from timefold-solver (if exists)
```
rsynek: Why do we need to delete a tag? Is that in case we are repeating a release of the same version? Can someone already depend on it?

.github/workflows/release.yml
```diff
@@ -47,45 +57,113 @@ jobs:
...
+          # We skip tests in dry run, to make the process faster
+          ./mvnw -Dfull clean deploy -DskipTests=$DRY_RUN -DaltDeploymentRepository=local::default::file://`pwd`/target/staging-deploy
+          # Temporally disabled
+          #./mvnw -Dfull verify --file model/test-model/pom.xml
           cp docs/target/antora.yml docs/src/antora.yml
           git add docs/src/antora.yml
           find . -name 'pom.xml' | xargs git add
+          # Create a compressed file containing all the artifacts
+          tar cvzf artifacts.tar.gz target/staging-deploy
+          git add artifacts.tar.gz
```
rsynek: Is this the archive with all the JAR files? If so, why do we push it to the tag? Binaries usually don't belong to SCM. Why not pushing it to jFrog?

.github/workflows/release.yml
```diff
@@ -47,45 +57,113 @@ jobs:
...
+      ######
+      # 5 - Restore the artifacts
+      ######
+      - name: Restore the artifacts for deploying timefold-solver
+        if: ${{ inputs.deploy-to-maven }}
+        run: |
+          # Restore all the artifacts
+          tar xvzf artifacts.tar.gz
+          # Remove the artifact file from the release branch
+          git rm artifacts.tar.gz
```
rsynek: Here we are removing the archive, previously added to the git repository. This commit will be visible in the commit history. Even if we remove it, it will stay a part of the git repository's history and as such will be transferred every time the repository is cloned.

rsynek (changes requested):
Thanks @zepfred ; I left there some comments and questions. If I understood correctly we push the archive with binaries to the github repository, I think we shouldn't. More details inline.

.github/workflows/release.yml
triceo: The tag needs to remain; if we've built something and published it, the tag must never go away.

I'm assuming this exists in case we ever need to re-run CI perhaps due to a later failure? The better solution here is to just fail and delete the tag manually; tags in general should not be deleted. (Unlike branches.)

.github/workflows/release.yml
triceo: Yes, this will quickly make the repo blow up in size.

jreleaser-jfrog.yml
rsynek, line 1: If everything moves to the enterprise repository, do we actually need to keep the `jreleaser-*.yml` config files here?

Can we move them to the enterprise repo as well?

.github/scripts/change_versions.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Expects the following environment variables to be set:
+#   $NEW_VERSION                 (Example: "1.2.0")
+# This will fail the Maven build if the version is not available.
+# Thankfully, this is not the case (yet) in this project.
+# If/when it happens, this needs to be replaced by a manually provided version,
+# as scanning the text of the POM would be unreliable.
+echo "     New version: $NEW_VERSION"
```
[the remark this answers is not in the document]
zepfred: Ignored

rsynek (approved):

PR #2323, opened 2026-05-29
chore: merge model parents and disconnect it from build parent

rsynek:
Notable changes:

- merged the `model-base-parent` and `model-parent` as the idea of layered models is no longer relevant
- disconnected the `model-parent` from `build-parent` to avoid solver build process impact models
- added some forgotten dependency renames in the `enterprise` profile


model/facade/model-parent/pom.xml
```diff
@@ -1,13 +1,9 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <parent>
-    <groupId>ai.timefold.solver</groupId>
-    <artifactId>timefold-solver-model-base-parent</artifactId>
-    <version>${revision}</version>
-    <relativePath>../model-base-parent/pom.xml</relativePath>
-  </parent>
   <modelVersion>4.0.0</modelVersion>
+  <groupId>ai.timefold.solver</groupId>
```
winklerm, line 4: Q: Is this the right `groupId`? I would expect something like `ai.timefold.solver.model` so that the artifacts are clearly distinguishable from the "core" solver artifacts, but fine if it was decided to name it this way.

model/facade/model-parent/pom.xml
```diff
@@ -60,6 +59,10 @@
     <!-- ************************************************************************ -->
     <!-- Dependencies -->
     <!-- ************************************************************************ -->
+    <version.ai.timefold.solver>${revision}</version.ai.timefold.solver>
+    <version.io.quarkus>3.35.3</version.io.quarkus>
```
winklerm: PP: Would it make sense to add a comment to the build-parent mentioning that for instance Quarkus version is declared also in this file and should be kept in sync?

model/facade/model-parent/pom.xml
```diff
@@ -1,13 +1,9 @@
...
-  <parent>
-    <groupId>ai.timefold.solver</groupId>
-    <artifactId>timefold-solver-model-base-parent</artifactId>
-    <version>${revision}</version>
-    <relativePath>../model-base-parent/pom.xml</relativePath>
-  </parent>
   <modelVersion>4.0.0</modelVersion>
+  <groupId>ai.timefold.solver</groupId>
   <artifactId>timefold-solver-model-parent</artifactId>
+  <version>${revision}</version>
```
winklerm, line 6: I am concerned a bit that we are not using the `maven-flatten-plugin` because Maven docs explicitely says it has to be used when installing/deploying poms using properties like `revision`: https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html#install-deploy

But I am not sure how to test that, apart from the fact that we try to build a model with this model parent (which already seems to work). I am wondering if we could get some pushback from Maven central when trying to deploy there..

winklerm (commented):
Thanks, the changes look good to me!

I have left a question and one concern about the flatten plugin, which I find potentially problematic, but I do not have any proof that it will not work without it, so leaving that up to your judgment.

model/facade/model-parent/pom.xml
rsynek: That has been already decided for all the artifacts to share the same groupId. The difference is in the artifactId prefix, which for the "sdk" artifacts contains `model`, like `timefold-solver-model-definition`, etc.

winklerm: Sounds good!

model/facade/model-parent/pom.xml
triceo: FYI Dependabot happened earlier today. Some of these dependencies were upgraded.

model/facade/model-parent/pom.xml
```diff
@@ -218,36 +259,49 @@
...
       <activation>
         <property>
           <name>enterprise</name>
           <value>true</value>
         </property>
       </activation>
       <properties>
+        <ai.timefold.sdk.enterprise>true</ai.timefold.sdk.enterprise>
+        <!-- enable copy of Timefold diagnostic tools when building container image - see container profile -->
         <ai.timefold.sdk.enterprise.skipTools>false</ai.timefold.sdk.enterprise.skipTools>
```
triceo: Is there an issue anywhere to refactor these properties to remove the "sdk" keyword?

rsynek: Now there is: https://github.com/TimefoldAI/timefold-solver-enterprise/issues/655

model/facade/model-parent/pom.xml
rsynek: Very good point. Retested and indeed the flatten maven plugin is needed to avoid the `revision` property value propagate from the model.

model/facade/model-parent/pom.xml
```diff
@@ -287,6 +366,16 @@
...
+      <repositories>
+        <repository>
+          <id>timefold</id>
+          <name>Timefold Enterprise Artifacts Repository</name>
+          <url>https://timefold.jfrog.io/artifactory/releases/</url>
+          <snapshots>
+            <enabled>false</enabled>
+          </snapshots>
+        </repository>
+      </repositories>
```
winklerm: Q: Do we need also `<pluginRepositories>` because of `timefold-maven-plugin`?

winklerm (commented):
Thank you! One more question inline, otherwise looks good to me!

model/facade/model-parent/pom.xml
rsynek: thanks; added a note and upgraded the Quarkus to keep it in sync with the build-parent.

model/facade/model-parent/pom.xml
rsynek: Good point, but I wouldn't introduce it yet. Models can declare it themselves.

rsynek: Removed also this repository, as all artifacts should be deployed to Maven Central.

model/facade/model-parent/pom.xml
```diff
@@ -60,6 +59,10 @@
     <!-- ************************************************************************ -->
     <!-- Dependencies -->
     <!-- ************************************************************************ -->
+    <version.ai.timefold.solver>${revision}</version.ai.timefold.solver>
+    <version.io.quarkus>3.35.4</version.io.quarkus>
+    <version.org.assertj>3.27.7</version.org.assertj>
+    <version.org.awaitility>4.3.0</version.org.awaitility>
```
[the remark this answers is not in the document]
rsynek: Late to the party; already upgraded.

model/facade/model-parent/pom.xml
```diff
@@ -147,6 +189,71 @@
   </dependencies>
 
   <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>flatten-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
rsynek: Fixed.

model/facade/model-parent/pom.xml
```diff
@@ -177,77 +285,59 @@
...
+        <dependencies>
+          <dependency>
+            <groupId>ai.timefold.solver.enterprise</groupId>
+            <artifactId>timefold-solver-enterprise-bom</artifactId>
+            <version>${version.ai.timefold.solver}</version>
+            <type>pom</type>
+            <scope>import</scope>
+          </dependency>
+        </dependencies>
+      </dependencyManagement>
```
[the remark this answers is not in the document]
rsynek: Nope; enterprise artifacts will be deployed to Maven Central in the end.

winklerm (approved):

model/facade/model-parent/pom.xml
```diff
@@ -147,6 +201,127 @@
   </dependencies>
 
   <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>io.quarkus</groupId>
+          <artifactId>quarkus-maven-plugin</artifactId>
+          <version>${version.io.quarkus}</version>
```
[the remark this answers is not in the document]
rsynek: Should be the other way around.

model/facade/model-parent/pom.xml
```diff
@@ -147,6 +201,127 @@
...
+            <!-- Currently not running on the module path, despite including JPMS -->
+            <useModulePath>false</useModulePath>
+            <redirectTestOutputToFile>true</redirectTestOutputToFile>
+            <skipTests>${skipUTs}</skipTests>
+            <!-- Avoid defining the argLine; it's configured in a profile for code coverage -->
+          </configuration>
+        </plugin>
+        <plugin>
+          <artifactId>maven-failsafe-plugin</artifactId>
+          <version>${version.surefire.plugin}</version>
```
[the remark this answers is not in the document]
rsynek: Fixed

model/facade/model-parent/pom.xml
```diff
@@ -147,6 +201,127 @@
...
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>exec-maven-plugin</artifactId>
+          <version>${version.exec.plugin}</version>
+          <configuration>
+            <mainClass>ai.timefold.solver.model.config.ConfigDumperMain</mainClass>
+          </configuration>
+        </plugin>
+        <plugin>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>${version.surefire.plugin}</version>
```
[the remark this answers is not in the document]
rsynek: Fixed

model/facade/model-parent/pom.xml
```diff
@@ -17,6 +13,16 @@
   </description>
 
   <properties>
+    <revision>999-SNAPSHOT</revision>
+    <ai.timefold.sdk.enterprise>false</ai.timefold.sdk.enterprise>
```
triceo: Since we are introducing this property, and the property is for end users only, why not name it properly now (no "sdk") and not create work for later?

triceo (approved):
Small comment; LGTM when resolved.

model/facade/model-parent/pom.xml
rsynek: Fair; it seems unused by models directly, let me change it.

rsynek: Done in the last commit.

PR #2324, opened 2026-05-29
feat: Add sliding window and minimum improvement ratio to termination confiig

diogodanielsoaresferreira:


cristianonicolai (approved):

mswiderski (approved):

triceo:
I really dislike the fact that we're opening up these configuration options and I would like to have one last change to prevent that.
Is there any product requirement for this work, should I talk to Jurriaan?

diogodanielsoaresferreira:
> I really dislike the fact that we're opening up these configuration options and I would like to have one last change to prevent that. Is there any product requirement for this work, should I talk to Jurriaan?

Yes, it's better to talk with @oemebamo, as this feature comes from a product request (https://github.com/TimefoldAI/timefold-platform/issues/4439).

triceo:
Discussed with Jurriaan, feature stays, I go away.

model/model-defaults/src/main/resources/application.properties
```diff
@@ -36,6 +36,11 @@ ai.timefold.platform.termination.spent-limit=PT10M
 # disabling the unimproved-spent-limit enables Diminished Returns termination
 ai.timefold.platform.termination.unimproved-spent-limit=
 
+# Override the Diminished Returns termination defaults (solver-core defaults to PT30S / 0.0001).
+# Leave unset to keep the built-in defaults.
+#ai.timefold.platform.termination.diminished-returns.sliding-window-duration=
+#ai.timefold.platform.termination.diminished-returns.minimum-improvement-ratio=
```
rsynek: Not sure we need this. The `model-defaults` just gets applied and eventually overridden by the model application.properties, but the user does not see it.

I doubt we should have two sets for defaults: one for solver-core and other for models.

model/worker/src/main/java/ai/timefold/solver/model/worker/impl/termination/TerminationService.java
```diff
@@ -29,24 +31,44 @@ public class TerminationService {
...
     @Inject
     TerminationService(
             @ConfigProperty(name = TerminationConfigParams.TERMINATION_SPENT_LIMIT) String spentLimit,
             @ConfigProperty(
                     name = TerminationConfigParams.TERMINATION_UNIMPROVED_SPENT_LIMIT) Optional<String> unimprovedSpentLimit,
             @ConfigProperty(name = TerminationConfigParams.TERMINATION_BEST_SCORE_LIMIT) Optional<String> bestScoreLimit,
-            @ConfigProperty(name = TerminationConfigParams.TERMINATION_STEP_COUNT_LIMIT) Optional<Integer> stepCountLimit) {
+            @ConfigProperty(name = TerminationConfigParams.TERMINATION_STEP_COUNT_LIMIT) Optional<Integer> stepCountLimit,
+            @ConfigProperty(
+                    name = TerminationConfigParams.TERMINATION_DIMINISHED_RETURNS_SLIDING_WINDOW_DURATION) Optional<String> diminishedReturnsSlidingWindowDuration,
```
rsynek: The method parameter name hides the field, which has a different type (Optional vs. the target type in the Optional).

model/worker/src/main/java/ai/timefold/solver/model/worker/impl/termination/TerminationService.java
```diff
@@ -29,24 +31,44 @@ public class TerminationService {
...
         this.spentLimit = parseDurationFromConfig(TerminationConfigParams.TERMINATION_SPENT_LIMIT, spentLimit);
         this.unimprovedSpentLimit = unimprovedSpentLimit
                 .map(s -> parseDurationFromConfig(TerminationConfigParams.TERMINATION_UNIMPROVED_SPENT_LIMIT, s)).orElse(null);
         this.bestScoreLimit = bestScoreLimit.orElse(null);
         this.stepCountLimit = stepCountLimit.orElse(null);
+        this.diminishedReturnsSlidingWindowDuration = diminishedReturnsSlidingWindowDuration
+                .map(s -> parseDurationFromConfig(
+                        TerminationConfigParams.TERMINATION_DIMINISHED_RETURNS_SLIDING_WINDOW_DURATION, s))
+                .orElse(null);
+        Double minimumImprovementRatio = diminishedReturnsMinimumImprovementRatio.orElse(null);
```
rsynek: In the solver repository, there is a rule to use `var` local variables where possible.

rsynek (commented):
Thanks @diogodanielsoaresferreira !

I left a couple of comments inline. Plus, there are no tests. At least, we should check if we create the expected Solver TerminationConfig.

rsynek:
The failure on windows-latest is a flaky test. Rerunning.

model/model-defaults/src/main/resources/application.properties
diogodanielsoaresferreira: I removed these two properties from the configurations. Now we only read the properties from the termination service.

model/worker/src/main/java/ai/timefold/solver/model/worker/impl/termination/TerminationService.java
diogodanielsoaresferreira: Updated 👍 changed lovar properties to var

diogodanielsoaresferreira:
> Thanks @diogodanielsoaresferreira !
> 
> I left a couple of comments inline. Plus, there are no tests. At least, we should check if we create the expected Solver TerminationConfig.

Added some unit tests to TerminationService and SolverTerminationConfig.

rsynek (approved):
LGTM, thank you.

PR #2332, opened 2026-06-01
perf: Support reusing the best solution instead of cloning in LS

Christopher-Chianelli:
Currently the working solution is cloned when ever a new best solution is found. This is slow and create additional memory pressure on giant datasets, since each entity must be cloned.

This adds a new SolverConfig property, `reuseBestSolution`, which will reuse the initial best solution clone in local search. When enabled, local search keep track of moves used in each step, which are then used to update the best solution instance.

A read-write lock is used to allow reading the best solution safely outside the solver thread. This read-write lock is automatically used for SolverManager events.

Since the same instance is reused, when this feature is enabled, users should ensure the solution and entities are not saved outside the event handler.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
+++ b/core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
@@ -73,6 +73,7 @@ import org.jspecify.annotations.Nullable;
         "terminationConfig",
         "nearbyDistanceMeterClass",
         "phaseConfigList",
+        "reuseBestSolution",
 })
 public final class SolverConfig extends AbstractConfig<SolverConfig> {
 
@@ -247,6 +248,8 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
 
     @XmlElement(name = "monitoring")
     private MonitoringConfig monitoringConfig = null;
+
+    private Boolean reuseBestSolution = null;
 
     // ************************************************************************
     // Constructors and simple getters/setters
@@ -448,6 +451,14 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         this.monitoringConfig = monitoringConfig;
     }
 
+    public @Nullable Boolean getReuseBestSolution() {
+        return reuseBestSolution;
+    }
+
+    public void setReuseBestSolution(@Nullable Boolean reuseBestSolution) {
+        this.reuseBestSolution = reuseBestSolution;
+    }
+
     // ************************************************************************
     // With methods
     // ************************************************************************
@@ -600,6 +611,11 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
         return this;
     }
 
+    public @NonNull SolverConfig withReuseBestSolution(@NonNull Boolean reuseBestSolution) {
+        this.reuseBestSolution = reuseBestSolution;
+        return this;
+    }
+
     // ************************************************************************
     // Smart getters
     // ************************************************************************
@@ -677,6 +693,7 @@ public final class SolverConfig extends AbstractConfig<SolverConfig> {
                 inheritedConfig.nearbyDistanceMeterClass);
         phaseConfigList = ConfigUtils.inheritMergeableListConfig(phaseConfigList, inheritedConfig.getPhaseConfigList());
         monitoringConfig = ConfigUtils.inheritConfig(monitoringConfig, inheritedConfig.getMonitoringConfig());
+        reuseBestSolution = ConfigUtils.inheritOverwritableProperty(reuseBestSolution, inheritedConfig.getReuseBestSolution());
         return this;
     }
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
@@ -5,6 +5,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
@@ -66,33 +68,36 @@ final class ConsumerSupport<Solution_, ProblemId_> implements AutoCloseable {
     }
 
     void consumeIntermediateBestSolution(Solution_ solution, EventProducerId producerId,
-            BooleanSupplier isEveryProblemChangeProcessed) {
+            BooleanSupplier isEveryProblemChangeProcessed, Lock lock) {
         /*
          * If the bestSolutionConsumer is not provided, the best solution is still set for the purpose of recording
          * problem changes.
          */
         bestSolutionHolder.set(solution, producerId, isEveryProblemChangeProcessed);
         if (bestSolutionConsumer != null) {
-            tryConsumeWaitingIntermediateBestSolution();
+            tryConsumeWaitingIntermediateBestSolution(lock);
         }
     }
 
     // Called both on the Solver thread and the Consumer thread.
-    private void tryConsumeWaitingIntermediateBestSolution() {
+    private void tryConsumeWaitingIntermediateBestSolution(Lock lock) {
         if (bestSolutionHolder.isEmpty()) {
             return; // There is no best solution to consume.
         }
         if (activeConsumption.tryAcquire()) {
-            scheduleIntermediateBestSolutionConsumption()
+            scheduleIntermediateBestSolutionConsumption(lock)
                     .whenCompleteAsync((solution, throwable) -> {
                         activeConsumption.release();
-                        tryConsumeWaitingIntermediateBestSolution();
+                        tryConsumeWaitingIntermediateBestSolution(lock);
                     }, consumerExecutor);
         }
     }
 
-    private CompletableFuture<Void> scheduleIntermediateBestSolutionConsumption() {
+    private CompletableFuture<Void> scheduleIntermediateBestSolutionConsumption(Lock lock) {
         return CompletableFuture.runAsync(() -> {
+            // If reuse best solution is enabled, this will block if the best solution
+            // is currently being updated.
+            lock.lock();
             var bestSolutionContainingProblemChanges = bestSolutionHolder.take();
             if (bestSolutionContainingProblemChanges != null) {
                 try {
@@ -107,6 +112,7 @@ final class ConsumerSupport<Solution_, ProblemId_> implements AutoCloseable {
                     bestSolutionContainingProblemChanges.completeProblemChangesExceptionally(throwable);
                 }
             }
+            lock.unlock();
         }, consumerExecutor);
     }
 
@@ -175,7 +181,7 @@ final class ConsumerSupport<Solution_, ProblemId_> implements AutoCloseable {
         // Situation:
         // The consumer is consuming the last but one best solution. The final best solution is waiting for the consumer.
         if (bestSolutionConsumer != null) {
-            scheduleIntermediateBestSolutionConsumption();
+            scheduleIntermediateBestSolutionConsumption(new ReentrantLock());
         }
         scheduleFinalBestSolutionConsumption(solution)
                 .whenComplete((unused, throwable) -> releaseAll());
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/event/LockableSolverEventListener.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/event/LockableSolverEventListener.java
@@ -1,0 +1,16 @@
+package ai.timefold.solver.core.impl.solver.event;
+
+import java.util.concurrent.locks.Lock;
+
+import ai.timefold.solver.core.api.solver.event.BestSolutionChangedEvent;
+import ai.timefold.solver.core.api.solver.event.SolverEventListener;
+
+import org.jspecify.annotations.NonNull;
+
+public interface LockableSolverEventListener<Solution_> extends SolverEventListener<Solution_> {
+    default void bestSolutionChanged(@NonNull BestSolutionChangedEvent<Solution_> event) {
+        throw new UnsupportedOperationException();
+    }
+
+    void bestSolutionChanged(@NonNull BestSolutionChangedEvent<Solution_> event, @NonNull Lock lock);
+}
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
@@ -1,5 +1,7 @@
 package ai.timefold.solver.core.impl.solver.event;
 
+import java.util.concurrent.locks.Lock;
+
 import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
 import ai.timefold.solver.core.api.solver.Solver;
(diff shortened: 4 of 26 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
```diff
@@ -20,15 +22,20 @@ public SolverEventSupport(Solver<Solution_> solver) {
     }
 
     public void fireBestSolutionChanged(SolverScope<Solution_> solverScope, EventProducerId eventProducerId,
-            Solution_ newBestSolution) {
+            Solution_ newBestSolution, Lock readLock) {
```
zepfred, line 25: Why are we including the lock as a parameter? Considering the principle of separation of concerns, wouldn't it be better for the lock to be part of the `LockableSolverEventListener` instead?

core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
```diff
@@ -20,15 +22,20 @@ public SolverEventSupport(Solver<Solution_> solver) {
...
         var it = getEventListeners().iterator();
         var timeMillisSpent = solverScope.getBestSolutionTimeMillisSpent();
         var bestScore = solverScope.getBestScore();
         if (it.hasNext()) {
             var event =
                     new DefaultBestSolutionChangedEvent<>(solver, eventProducerId, timeMillisSpent, newBestSolution, bestScore);
             do {
-                it.next().bestSolutionChanged(event);
+                var eventListener = it.next();
```
zepfred, line 33: If the listener contract remains unchanged, there will be no need to update the block.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
```diff
@@ -73,6 +73,7 @@
         "terminationConfig",
         "nearbyDistanceMeterClass",
         "phaseConfigList",
+        "reuseBestSolution",
```
zepfred, line 76: I'm not sure if we need a new property. Do we always want to use the new related logic? Are there scenarios where enabling this feature by default may not be recommended?

core/src/main/java/ai/timefold/solver/core/impl/solver/event/LockableSolverEventListener.java
```diff
@@ -0,0 +1,16 @@
+package ai.timefold.solver.core.impl.solver.event;
+
+import java.util.concurrent.locks.Lock;
+
+import ai.timefold.solver.core.api.solver.event.BestSolutionChangedEvent;
+import ai.timefold.solver.core.api.solver.event.SolverEventListener;
+
+import org.jspecify.annotations.NonNull;
+
+public interface LockableSolverEventListener<Solution_> extends SolverEventListener<Solution_> {
```
zepfred, line 10: I wonder if we really need a new contract. Can we provide a new event listener implementation that incorporates the required logic with locking without changing the contract?

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -36,6 +46,19 @@ public void setAssertBestScoreIsUnmodified(boolean assertBestScoreIsUnmodified)
         this.assertBestScoreIsUnmodified = assertBestScoreIsUnmodified;
     }
 
+    public void setReuseBestSolution(boolean reuseBestSolution) {
+        this.reuseBestSolution = reuseBestSolution;
+        if (reuseBestSolution) {
+            reusingBestSolutionUpdater = TimefoldSolverEnterpriseService
```
zepfred, line 52: Does this mean that if a community user sets `reuseBestSolution` to true, the solving process will fail?

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -79,65 +102,98 @@ public void processWorkingSolutionDuringConstructionHeuristicsStep(AbstractStepS
...
-    public <Score_ extends Score<Score_>> void processWorkingSolutionDuringStep(AbstractStepScope<Solution_> stepScope) {
+    public void processWorkingSolutionDuringStep(AbstractStepScope<Solution_> stepScope) {
+        processWorkingSolutionDuringStep(stepScope, null);
+    }
+
+    /**
+     * Return true if best solution was updated
+     */
+    public <Score_ extends Score<Score_>> void processWorkingSolutionDuringStep(AbstractStepScope<Solution_> stepScope,
+            @Nullable List<Move<Solution_>> acceptedMoveList) {
```
zepfred, line 113: Can we fetch the list from stepScope::getPhaseScope? This way, we won't need to change the contract of `BestSolutionRecaller`.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -21,8 +29,10 @@ public class BestSolutionRecaller<Solution_> extends PhaseLifecycleListenerAdapt
     protected boolean assertInitialScoreFromScratch = false;
     protected boolean assertShadowVariablesAreNotStale = false;
     protected boolean assertBestScoreIsUnmodified = false;
-
+    protected boolean reuseBestSolution = false;
     protected SolverEventSupport<Solution_> solverEventSupport;
+    protected ReusingBestSolutionUpdater<Solution_> reusingBestSolutionUpdater;
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
```
zepfred, line 35: If we can create the `BestSolutionRecaller` instance and pass the lock `readWriteLock` as a constructor parameter. We could share it with other components that need it, such as the `LockableSolverEventListener/SolverEventListener` implementation.

My main argument against adding the new arguments for lock and accepted moves is that we could achieve the same without changing the existing contract.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecallerFactory.java
```diff
@@ -8,13 +8,17 @@ public static BestSolutionRecallerFactory create() {
...
+    public <Solution_> BestSolutionRecaller<Solution_> buildBestSolutionRecaller(EnvironmentMode environmentMode,
+            Boolean reuseBestSolution) {
         BestSolutionRecaller<Solution_> bestSolutionRecaller = new BestSolutionRecaller<>();
         if (environmentMode.isFullyAsserted()) {
             bestSolutionRecaller.setAssertInitialScoreFromScratch(true);
             bestSolutionRecaller.setAssertShadowVariablesAreNotStale(true);
             bestSolutionRecaller.setAssertBestScoreIsUnmodified(true);
         }
+        if (reuseBestSolution != null && reuseBestSolution) {
+            bestSolutionRecaller.setReuseBestSolution(true);
```
zepfred, line 20:
```suggestion
            bestSolutionRecaller.setReuseBestSolution(reuseBestSolution != null && reuseBestSolution);
```

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecallerFactory.java
```diff
@@ -8,13 +8,17 @@ public static BestSolutionRecallerFactory create() {
...
-    public <Solution_> BestSolutionRecaller<Solution_> buildBestSolutionRecaller(EnvironmentMode environmentMode) {
+    public <Solution_> BestSolutionRecaller<Solution_> buildBestSolutionRecaller(EnvironmentMode environmentMode,
+            Boolean reuseBestSolution) {
         BestSolutionRecaller<Solution_> bestSolutionRecaller = new BestSolutionRecaller<>();
         if (environmentMode.isFullyAsserted()) {
             bestSolutionRecaller.setAssertInitialScoreFromScratch(true);
             bestSolutionRecaller.setAssertShadowVariablesAreNotStale(true);
             bestSolutionRecaller.setAssertBestScoreIsUnmodified(true);
         }
+        if (reuseBestSolution != null && reuseBestSolution) {
```
zepfred, line 19:
```suggestion
```

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecallerFactory.java
```diff
@@ -8,13 +8,17 @@ public static BestSolutionRecallerFactory create() {
...
+            bestSolutionRecaller.setReuseBestSolution(true);
+        }
```
zepfred, line 21:
```suggestion
```

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecallerFactory.java
```diff
@@ -8,13 +8,17 @@ public static BestSolutionRecallerFactory create() {
         return new BestSolutionRecallerFactory();
     }
 
-    public <Solution_> BestSolutionRecaller<Solution_> buildBestSolutionRecaller(EnvironmentMode environmentMode) {
+    public <Solution_> BestSolutionRecaller<Solution_> buildBestSolutionRecaller(EnvironmentMode environmentMode,
```
zepfred, line 11: Please refer to my comment on `BestSolutionRecaller`. It appears we could pass the lock instance as a parameter here.

core/src/main/java/ai/timefold/solver/core/impl/solver/ConsumerSupport.java
```diff
@@ -66,33 +68,36 @@ public ConsumerSupport(ProblemId_ problemId, @Nullable Consumer<NewBestSolutionE
...
                     .whenCompleteAsync((solution, throwable) -> {
                         activeConsumption.release();
-                        tryConsumeWaitingIntermediateBestSolution();
+                        tryConsumeWaitingIntermediateBestSolution(lock);
                     }, consumerExecutor);
         }
     }
 
-    private CompletableFuture<Void> scheduleIntermediateBestSolutionConsumption() {
+    private CompletableFuture<Void> scheduleIntermediateBestSolutionConsumption(Lock lock) {
```
zepfred, line 96: I believe here is another place where we should consider keeping the existing contract unchanged. Let's say we can create `LockableConsumerSupport` that receives or creates its internal lock instance. We could extend `ConsumerSupport` and override `scheduleIntermediateBestSolutionConsumption` without changing the existing contract. Does it make sense?

zepfred (changes requested):
Reusing the best solution is a great idea! I have some questions and suggestions about not changing the existing contract to achieve the goal of this PR.

core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
Christopher-Chianelli: This fires events for all listeners. `SolverManager` fire off its events in a seperate thread, which makes a lock neccessary for it but not other event listeners. This is not a part of the API; it just propagate events.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
triceo: +1. Already discussed on Slack with Chris; no external configuration, we do the right thing.

core/src/main/java/ai/timefold/solver/core/impl/solver/event/SolverEventSupport.java
Christopher-Chianelli: `LockableSolverEventListener` is a `SolverEventListener`; `SolverEventListener` is not a `LockableSolverEventListener`. The block changed so a lock can be passed to `LockableSolverEventListener` without adapting all `SolverEventListener`.

core/src/main/java/ai/timefold/solver/core/config/solver/SolverConfig.java
Christopher-Chianelli: The primary reason for the switch is that this reuses the existing solution; users might rely on getting new best solution instances if they save them outside the event handler (ex: store them in a `MutableReference`).

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
Christopher-Chianelli: Yes.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
Christopher-Chianelli: The list is only applicable to (and only available in) `LocalSearchStepScope`. Custom phases and exhaustive search do not have the concept of steps, and CH does not update the best solution. IMO, this is cleaner.

triceo (changes requested):
Folks, we will not be merging this now.

I am not convinced that the problem we're solving warrants this level of complexity. Specifically:

- Growing list of changes that need to be applied, effectively a controlled memory leak.
- Heuristics for the solution sizes where this should kick in, and heuristics for the list size where this should be disabled.
- External configuration for something so low-level, because we need to be able to deal with a situation that the user modifies the best solution while we still need to apply moves on it.
- _Lockable event listeners_?!

This shows me that the approach selected here is not the right one. Considering that we currently do not feel this pain, and that we have other priorities to tackle, I think we should not be spending more time right now to get this right.

triceo:
I was thinking about this more, and I really think that this should not be merged. My previous objections still stand, but I want to add another one - even if the user is aware of the fact that the best solutions will be updated, there is still no sane way to work with best solution events. 

If two events can share the same solution, that means the user possibly has two events which influence each other; this behavior is impossible to deal with. If the new event already exists, the old event has the new event's data; this tells me that if we want to keep this mechanism instead of cloning, we need to refactor how best solutions are consumed.

I consider this mechanism fundamentally and unfixably incompatible with best solution events.

Christopher-Chianelli:
> I was thinking about this more, and I really think that this should not be merged. My previous objections still stand, but I want to add another one - even if the user is aware of the fact that the best solutions will be updated, there is still no sane way to work with best solution events.
> 
> If two events can share the same solution, that means the user possibly has two events which influence each other; this behavior is impossible to deal with. If the new event already exists, the old event has the new event's data; this tells me that if we want to keep this mechanism instead of cloning, we need to refactor how best solutions are consumed.
> 
> I consider this mechanism fundamentally and unfixably incompatible with best solution events.

Hence why event listeners are lockable; via SolverFactory, nothing changes since event listeners are synchronous (and if they will spawn a new thread, they can implement LockableEventListener instead). Via SolverManager, the event listener threads locks the best solution so the Solver cannot update it (and thus cannot fire additional events) until the event listener finishes.

triceo:
> Hence why event listeners are lockable; via SolverFactory, nothing changes since event listeners are synchronous (and if they will spawn a new thread, they can implement LockableEventListener instead). Via SolverManager, the event listener threads locks the best solution so the Solver cannot update it (and thus cannot fire additional events) until the event listener finishes.

But all of this is wrong! With `SolverFactory`, the event listeners are synchronously triggered, but who knows how they're processed by the user; my objection still stands. 

The approach is more obviously wrong on the `SolverManager` side. Once the new best solution is available, arguably the old one becomes garbage; at the very least, it becomes much less useful. Blocking the new, valuable solutions so that the stale one can finish being processed is counterproductive.

triceo:
I'm afraid this can't be salvaged. Valuable experiment, but the mechanism of interacting with best solutions makes this not feasible.

PR #2333, opened 2026-06-02
perf: remove indirection from ElementAwareArrayList

triceo:
Only a small optimization, but generally helpful.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
@@ -1,7 +1,8 @@
 package ai.timefold.solver.core.impl.util;
 
+import java.lang.reflect.Array;
 import java.util.AbstractList;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


triceo:
Ping @Christopher-Chianelli

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -49,23 +52,41 @@ public final class ElementAwareArrayList<T extends @Nullable Object> extends Abs
...
-        entryList.add(newEntry);
+        resize(lastElementPosition + 1);
+        entries[lastElementPosition] = newEntry;
         return newEntry;
     }
 
+    @SuppressWarnings("unchecked")
+    private void resize(int minCapacity) {
+        if (entries == null) {
+            entries = (Entry[]) Array.newInstance(Entry.class, Math.max(DEFAULT_CAPACITY, minCapacity));
```
Christopher-Chianelli, line 70: Why not `entries = new Entry[Math.max(DEFAULT_CAPACITY, minCapacity)]`? Entry is not generic, given you can do `Entry.class`?

Christopher-Chianelli (approved):
LGTM

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: image

PR #2334, opened 2026-06-03
feat(neighborhoods): add groupBy() support

triceo:


core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/bi/AbstractBiEnumeratingStream.java
```diff
@@ -36,15 +36,41 @@ public final BiEnumeratingStream<Solution_, A, B> filter(BiNeighborhoodsPredicat
...
 
-    protected <GroupKeyA_, GroupKeyB_> AbstractBiEnumeratingStream<Solution_, GroupKeyA_, GroupKeyB_>
-            groupBy(BiFunction<A, B, GroupKeyA_> groupKeyAMapping, BiFunction<A, B, GroupKeyB_> groupKeyBMapping) {
-        GroupNodeConstructor<BiTuple<GroupKeyA_, GroupKeyB_>> nodeConstructor =
-                GroupNodeConstructor.twoKeysGroupBy(groupKeyAMapping, groupKeyBMapping, Group2Mapping0CollectorBiNode::new);
-        return buildBiGroupBy(nodeConstructor);
+    @Override
+    public <GroupKey_> AbstractUniEnumeratingStream<Solution_, GroupKey_> groupBy(
+            BiNeighborhoodsMapper<Solution_, A, B, GroupKey_> key) {
+        return buildUniGroupBy(NeighborhoodsGroupNodeConstructor.biOneKeyGroupBy(key));
```
Christopher-Chianelli, line 42: This looks to be a pain to maintain like `ConstraintStream`'s `groupBy`. Any reason why the node constructor cannot just take two lists (one for keys, one for collectors)?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/NeighborhoodsGroupNodeConstructor.java
```diff
@@ -0,0 +1,117 @@
...
+    static <Solution_, A, B, GroupKey_, Result_>
+            NeighborhoodsGroupNodeConstructor<Solution_, BiTuple<GroupKey_, Result_>>
+            biOneKeyAndCollectorGroupBy(BiNeighborhoodsMapper<Solution_, A, B, GroupKey_> key,
+                    BiNeighborhoodsCollector<Solution_, A, B, ?, Result_> collector) {
+        return new NeighborhoodsGroupNodeConstructorWithAccumulate<>(new Pair<>(key, collector),
+                view -> GroupNodeConstructor.oneKeyGroupBy(key.toBiFunction(view),
+                        NeighborhoodsCollectorUtils.toConstraintCollector(
+                                (BiNeighborhoodsCollector<Solution_, A, B, Object, Result_>) collector, view),
+                        Group1Mapping1CollectorBiNode::new));
+    }
```
Christopher-Chianelli, line 106: Is only a single collector supported? There no methods for two collectors, but are methods for two keys.

core/src/main/java/ai/timefold/solver/core/preview/api/neighborhood/stream/collector/NeighborhoodsCollectors.java
```diff
@@ -0,0 +1,41 @@
...
+@NullMarked
+public final class NeighborhoodsCollectors {
+
+    private NeighborhoodsCollectors() {
+    }
+
+    /**
+     * Collects all facts from a {@link UniEnumeratingStream} group into a {@link List}.
+     */
+    public static <Solution_, A> UniNeighborhoodsCollector<Solution_, A, ?, List<A>> toList() {
```
Christopher-Chianelli: No variant for mapping via `SolutionView` for `Uni`?

core/src/main/java/ai/timefold/solver/core/preview/api/neighborhood/stream/collector/UniNeighborhoodsCollectorAccumulator.java
```diff
@@ -0,0 +1,31 @@
+package ai.timefold.solver.core.preview.api.neighborhood.stream.collector;
+
+import ai.timefold.solver.core.preview.api.move.SolutionView;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Accumulates values into a group container for a {@link UniNeighborhoodsCollector}.
+ * Called once per group to obtain a {@link UniNeighborhoodsCollectorValueHandle}
```
Christopher-Chianelli: Bad wording; maybe "Created once per each group, then called for each value to obtain ..."

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -532,7 +532,7 @@ the `MONDAY_MORNING` timeslot is a valid value for the `timeslotVariable` of the
 If that's not the case, we filter out this lesson from the enumeration
 and therefore will not generate moves which use this lesson.
 
-The same pattern applies to other building blocks as well, such as `join` and `ifExists`;
+The same pattern applies to other building blocks as well, such as `join`, `ifExists` or `groupBy`;
```
Christopher-Chianelli:
```suggestion
The same pattern applies to other building blocks as well, such as `join`, `ifExists` and `groupBy`;
```

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/bi/AbstractBiEnumeratingStream.java
triceo: It intentionally looks the same - I was going for consistency. I think it's better if the same things look the same; that way, people familiar with CS will be familiar with Neighborhoods as well.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/NeighborhoodsGroupNodeConstructor.java
triceo: My thinking is that in CS, you need a general-purpose data manipulation language. In Neighborhoods, not so much - you only need to produce arbitrary groups (therefore support for multiple keys) and one collector to provide some aggregation. (You don't need count+average; you will generally only need a list of elements.)

But now that you bring it up, I think we may do _a different departure_ from CS. We need only provide overloads for 0-1 keys and 0-1 collectors; the rest can be made up by the user with aggregation - `a -> Pair.of(..., ...)` and `Collectors.compose(...)`. (That's what CS does under the hood anyway.)

I'll do that here. If it ever becomes insufficient for any reason, we can always introduce all the extra overloads.

core/src/main/java/ai/timefold/solver/core/preview/api/neighborhood/stream/collector/NeighborhoodsCollectors.java
triceo: Well spotted; should be there.

triceo:
Since I made significant changes (mostly adding new collectors), I'm re-requesting a review, @Christopher-Chianelli.

Christopher-Chianelli (approved):

PR #2337, opened 2026-06-07
perf(constraint-streams): implement runtime node deactivation

triceo:
While I was there, I moved the hidden network visualization feature to Enterprise, and exposed it publicly as part of constraint profiling.

Enterprise counterpart: https://github.com/TimefoldAI/timefold-solver-enterprise/pull/662


core/src/main/java/ai/timefold/solver/core/impl/bavet/uni/ForEachFilteredUniNode.java
```diff
@@ -14,16 +14,34 @@
...
+        // For this reason, this node must be considered active if it saw at least one input;
+        // only with zero tuples can it be considered inactive, as the filter has nothing to propagate.
+        return tupleCountWithoutFiltering > 0 && nextNodesTupleLifecycle.isActive();
+    }
+
     @Override
     public void insert(@Nullable A a) {
+        tupleCountWithoutFiltering++; // This is safe; each element is only inserted once, guaranteed by a fail-fast in the parent.
         if (!filter.test(a)) { // Skip inserting the tuple as it does not pass the filter.
             return;
```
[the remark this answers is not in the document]
triceo: False positive; the insert counts everything before filtering, therefore update cannot break the decision.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/AggregatedTupleLifecycle.java
```diff
@@ -1,51 +1,90 @@
...
 import java.util.Arrays;
-import java.util.Objects;
 
-public record AggregatedTupleLifecycle<Tuple_ extends Tuple>(TupleLifecycle<Tuple_>... lifecycles)
-        implements
-            TupleLifecycle<Tuple_> {
+public final class AggregatedTupleLifecycle<Tuple_ extends Tuple>
+        implements TupleLifecycle<Tuple_> {
+
+    private boolean upstreamCanProduceTuples;
```
Christopher-Chianelli: Inconsistent with the other boolean; although both booleans default to false, the other boolean is set to false explictly while this relies on the Java Lanaguage Specification. Additionally, I wouid prefer upstreamCanProduceTuples be declared below downstreamFinal, since the fields are the same type and it puts them in alphabetical order.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/AggregatedTupleLifecycle.java
```diff
@@ -1,51 +1,90 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public boolean isActive() {
+        if (downstreamFinal) {
+            return downstream.length > 0;
+        }
+        if (upstreamCanProduceTuples) {
+            // Iterating a list in update() was measurably slower in micro benchmarks, so we deal with arrays.
```
Christopher-Chianelli: Iterating a list was slower than a stream done on an array? Or was the list also streamed?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/ConditionalTupleLifecycle.java
```diff
@@ -3,14 +3,31 @@
 import java.util.Objects;
 import java.util.function.Predicate;
 
-public record ConditionalTupleLifecycle<Tuple_ extends Tuple>(TupleLifecycle<Tuple_> downstreamLifecycle,
-        TuplePredicate<Tuple_> predicate)
-        implements
-            TupleLifecycle<Tuple_> {
+public final class ConditionalTupleLifecycle<Tuple_ extends Tuple>
```
Christopher-Chianelli, line 9: Maybe add `@NullMarked`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/LeftBridgeTupleLifecycle.java
```diff
@@ -0,0 +1,62 @@
...
+        leftTupleLifecycle.retractLeft(tuple);
+    }
+
+    public LeftTupleLifecycle<Tuple_> leftTupleLifecycle() {
+        return leftTupleLifecycle;
+    }
+
+    @Override
+    public String toString() {
+        return "left " + leftTupleLifecycle;
```
Christopher-Chianelli:
```suggestion
        return "left %s".formatted(leftTupleLifecycle);
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/ProfilingTupleLifecycle.java
```diff
@@ -28,4 +49,42 @@ public void retract(Tuple_ tuple) {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(constraintProfiler, profileId, delegate);
+    }
+
+    @Override
+    public String toString() {
+        return "ProfilingTupleLifecycle[" +
```
Christopher-Chianelli: Multi-line string + `.formatted()`

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/ProfilingTupleLifecycle.java
```diff
@@ -1,14 +1,35 @@
...
+    public ProfilingTupleLifecycle(InnerConstraintProfiler constraintProfiler, ConstraintNodeProfileId profileId,
+            TupleLifecycle<Tuple_> delegate) {
         constraintProfiler.register(profileId);
+        this.constraintProfiler = constraintProfiler;
+        this.profileId = profileId;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void afterAllFactsInserted(boolean upstreamCanProduceTuples) {
```
Christopher-Chianelli, line 27: TBH, I find these implementations to be odd for `ProfilingTupleLifecycle`; `ProfilingTupleLifecycle` never exists on their own; they always wraps an existing `TupleLifecycle`. Thus, the extra fields should be not be necessary; `afterAllFactsInserted` and `isActive` should just delegate to the `delegate`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/RightBridgeTupleLifecycle.java
```diff
@@ -0,0 +1,62 @@
...
+        rightTupleLifecycle.retractRight(tuple);
+    }
+
+    public RightTupleLifecycle<Tuple_> rightTupleLifecycle() {
+        return rightTupleLifecycle;
+    }
+
+    @Override
+    public String toString() {
+        return "right " + rightTupleLifecycle;
```
Christopher-Chianelli:
```suggestion
        return "right %s".formatted(rightTupleLifecycle);
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
```diff
@@ -75,6 +76,11 @@ public RecordAndReplayPropagator(Supplier<BavetPrecomputeBuildHelper<Tuple_>> pr
         this(precomputeBuildHelperSupplier, internalTupleToOutputTupleMapper, nextNodesTupleLifecycle, 1000);
     }
 
+    public boolean canProduceTuples() {
+        return !objectToOutputTuplesMap.isEmpty() // Tuples were produced.
+                || !insertQueue.isEmpty(); // Tuples will be produced.
```
Christopher-Chianelli: A precompute might not produce any output tuples even if it has objects in its insert queue. If the insert queue is non-empty, it simply means the cache is stale, nothing more, nothing less.

```suggestion
                || !insertQueue.isEmpty(); // Tuples might be produced.
```

core/src/main/java/ai/timefold/solver/core/impl/bavet/AbstractBavetNodeNetwork.java
```diff
@@ -0,0 +1,177 @@
...
+     *         otherwise returns all root nodes.
+     *         This means that if this information was ever read before activation checks were complete,
+     *         it should be re-read after to make sure no inactive nodes are included.
+     */
+    public Stream<AbstractRootNode<?>> getRootNodesAcceptingType(Class<?> factClass) {
+        // The node needs to match the fact, or the node needs to be applicable to the entire solution.
+        // The latter is for FromSolution nodes.
+        return declaredClassToNodeMap.entrySet().stream()
+                .flatMap(entry -> entry.getValue().stream())
+                .filter(tupleSourceRoot -> factClass == PlanningSolution.class || tupleSourceRoot.allowsInstancesOf(factClass))
```
Christopher-Chianelli: Shouldn't this be `factClass.hasAnnotation(PlanningSolution.class)`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/AbstractSession.java
```diff
@@ -68,11 +70,23 @@ public final void settle() {
...
+        }
         settled = true;
     }
 
-    public final void summarizeProfileIfPresent() {
-        nodeNetwork.summarizeProfileIfPresent();
+    private void removeInactiveRootNodes(Map<Class<?>, AbstractRootNode<Object>[]> effectiveClassToNodeArrayMap) {
+        // Use getActiveNodes() for this, to not rerun the activity checking logic again.
+        effectiveClassToNodeArrayMap.replaceAll((k, v) -> Arrays.stream(v)
+                .filter(n -> nodeNetwork.getActiveNodes().contains(n))
```
Christopher-Chianelli:
```suggestion
                .filter(nodeNetwork.getActiveNodes()::contains)
```

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/SingleInputNodeActivityTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @Test
+    void inactiveWhenDownstreamInactive() {
+        var downstream = mockDownstream(false);
+        var node = node(downstream);
+        node.afterAllFactsInserted(true);
+        assertThat(node.isActive()).isFalse();
+    }
+
+    @Test
+    void forwardsUpstreamCapabilityDownstream() {
```
Christopher-Chianelli: Might want to parameterize this so it tests both `true` and `false`.

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/AggregatedTupleLifecycle.java
triceo: I generally order fields by their order of use, unless something else makes more sense in the given context. (Just like methods.) But good point with the initialization.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/AggregatedTupleLifecycle.java
triceo: This refers to the _resulting array_, not to the process of obtaining it. I'll move the comment to where it makes more sense.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/RecordAndReplayPropagator.java
triceo: Is there a better way how to detect if tuples will be produced then?
If we can improve this, we can deactivate this node more often, leading to better performance.

PR #2346, opened 2026-06-10
perf(constraint-streams): fused equal indexer

triceo:
Optimizes equal-join-heavy workloads by cutting `HashMap` lookups by 50+ percent. For increased efficiency, reorders joiners so that all `equal()` joiners come before any others.


core/src/main/java/ai/timefold/solver/core/impl/bavet/bi/joiner/DefaultBiJoiner.java
```diff
@@ -28,40 +31,64 @@ public DefaultBiJoiner(Function<A, ?>[] leftMappings, JoinerType[] joinerTypes,
...
         return new DefaultBiJoiner<>(newLeftMappings, newJoinerTypes, newRightMappings);
     }
 
+    /**
+     * @return this if already equal-first (or single joiner); otherwise a copy with all
+     *         {@link JoinerType#EQUAL} joiners moved to the front (stable, see
+     *         {@link AbstractJoiner#equalsFirstOrder}).
+     */
+    public DefaultBiJoiner<A, B> reorderedEqualsFirst() {
+        var order = equalsFirstOrder(joinerTypes);
```
Christopher-Chianelli: It was hard to tell `order` was an `int` array. Maybe `equalFirstPermutation` or `equalFirstSortedPositions?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FusedEqualIndex.java
```diff
@@ -0,0 +1,200 @@
...
+     * <p>
+     * Only meaningful for {@link #hasSuffix()} indexes (the node gates on it); for a pure-equal index the equal key
+     * is the whole composite key, so this coincides with the full-key equality the caller has already ruled out.
+     * {@code Objects.equals} because the equal-prefix component may be null (a nullable planning variable feeding
+     * the equal joiner; the bucket map permits a null key).
+     */
+    public boolean isSameBucket(Object oldCompositeKey, Object newCompositeKey) {
+        var a = topEqualKeyUnpacker.apply(oldCompositeKey);
+        var b = topEqualKeyUnpacker.apply(newCompositeKey);
+        return (a == b) || (a != null && a.equals(b));
```
Christopher-Chianelli, line 76: Does this have an advantage over `Objects.equals(a, b)`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FusedEqualIndex.java
```diff
@@ -0,0 +1,200 @@
...
+        public void forEachRight(Object compositeKey, Consumer<R> consumer) {
+            if (rightDownstream != null) {
+                rightDownstream.forEach(compositeKey, consumer);
+            }
+        }
+
+        /**
+         * Returns 0 without allocating when the right side has never received a tuple.
+         */
+        public int sizeRight(Object compositeKey) {
```
Christopher-Chianelli, line 186: No `sizeLeft`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIndexedIfExistsNode.java
```diff
@@ -118,22 +163,41 @@ public final void retractLeft(LeftTuple_ leftTuple) {
...
-        indexerLeft.remove(compositeKey, counterEntry);
+    private void clearLeftTrackerList(LeftTuple_ leftTuple) {
         if (isFiltering) {
             ElementAwareLinkedList<FilteringTracker<LeftTuple_>> leftTrackerList =
                     leftTuple.getStore(inputStoreIndexLeftTrackerList);
             leftTrackerList.clear(FilteringTracker::removeByLeft);
         }
     }
 
+    private void updateIndexerLeft(Object compositeKey, ListEntry<ExistsCounter<LeftTuple_>> counterEntry, LeftTuple_ leftTuple,
```
Christopher-Chianelli, line 196: Duplicates quite a bit of code from `retractLeft`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/joiner/QuadJoinerComber.java
```diff
@@ -6,86 +6,93 @@
...
             if (joiner instanceof FilteringQuadJoiner) {
                 // From now on, only allow filtering joiners.
                 indexOfFirstFilter = i;
                 filteringList.add(((FilteringQuadJoiner<A, B, C, D>) joiner).getFilter());
             } else if (joiner instanceof DefaultQuadJoiner) {
                 if (indexOfFirstFilter >= 0) {
-                    throw new IllegalStateException("Indexing joiner (" + joiner + ") must not follow " +
-                            "a filtering joiner (" + joiners[indexOfFirstFilter] + ").\n" +
-                            "Maybe reorder the joiners such that filtering() joiners are later in the parameter list.");
+                    throw new IllegalStateException("""
```
Christopher-Chianelli, line 36: Does this exception make much sense now; given we reorder equals joiners before all other joiners, we can also reorder filtering joiners after all other joiners.

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FusedEqualIndex.java
triceo: You and I have discussed this before. JDK seems to have an issue with handling the `Objects.equals()` megamorphic site and cannot inline these calls; since this is on the hot path, I have decided to inline it myself.

That said, I have not benchmarked the impact of this.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/FusedEqualIndex.java
triceo: The code did not require it. (`ifExists` is not symmetrical.)

core/src/main/java/ai/timefold/solver/core/impl/bavet/quad/joiner/QuadJoinerComber.java
triceo: I was considering this.
The original intention here was to show that `filtering` joiners are different from other joiners; in fact, they are kind-of not joiner at all. The special handling IMO makes sense; if we force the user to put them in the back manually, it will be implied by the code already that these joiners happen last.

PR #2351, opened 2026-06-10
ci: remove `1.x` release workflow

zepfred:


rsynek:
I am a bit confused. Why do we have a 1.x release workflow on both the `main` branch and the `1.x` branch?

I understand we have to have it on the `main` branch as otherwise, it won't show up in the UI. Why do we need it on 1.x?

rsynek:
@zepfred since the 1.x release will be only ever started from the `main` branch, the 1.x workflow is effectively a dead code. Can we rather remove it?

rsynek (approved):

zepfred:
We cannot delete it from the `1.x` branch, or we will not be able to trigger it.

image

PR #2352, opened 2026-06-10
fix: validate the effective configuration

rsynek:
Fixes an issue that we validate the original configuration coming from a request, but not the effective configuration that is created by merging the configuration profile with the request configuration.

Tested with a custom version of field-service-routing in the sandbox environment.

### DIFF
--- a/model/rest/src/main/java/ai/timefold/solver/model/rest/impl/AbstractModelAPIResource.java
+++ b/model/rest/src/main/java/ai/timefold/solver/model/rest/impl/AbstractModelAPIResource.java
@@ -656,12 +656,13 @@ public abstract class AbstractModelAPIResource<ModelInput_ extends ModelInput, M
     @Produces(MediaType.APPLICATION_JSON)
     public ValidationResult<ValidationIssue_> getValidationResult(
             @Parameter(description = "Unique identifier of the schedule", required = true) @PathParam("id") String id) {
-        ValidationBuilder validationBuilder = new ValidationBuilder();
-        ModelRequest<ModelInput_, ModelConfigurationOverrides_> modelRequest = storageService.getModelRequest(id);
-        if (modelRequest.modelInput() == null) {
+        ModelInput_ modelInput = storageService.getModelInput(id);
+        if (modelInput == null) {
             throw new ItemNotFoundException(ErrorCodes.STORAGE_NO_JOB_FOUND, id);
         }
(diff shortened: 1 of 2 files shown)


model/rest/src/main/java/ai/timefold/solver/model/rest/impl/AbstractModelAPIResource.java
```diff
@@ -656,12 +656,13 @@ public Multi<Metadata<Score_>> streamMetadata(
     @Produces(MediaType.APPLICATION_JSON)
     public ValidationResult<ValidationIssue_> getValidationResult(
             @Parameter(description = "Unique identifier of the schedule", required = true) @PathParam("id") String id) {
-        ValidationBuilder validationBuilder = new ValidationBuilder();
-        ModelRequest<ModelInput_, ModelConfigurationOverrides_> modelRequest = storageService.getModelRequest(id);
-        if (modelRequest.modelInput() == null) {
+        ModelInput_ modelInput = storageService.getModelInput(id);
+        var modelConfig = Configuration.getSafeModelConfig(storageService.getConfiguration(id));
+        if (modelInput == null) {
```
cristianonicolai, line 660: maybe check input before retrieving the config?

model/rest/src/main/java/ai/timefold/solver/model/rest/impl/AbstractModelAPIResource.java
```diff
@@ -656,12 +656,13 @@ public Multi<Metadata<Score_>> streamMetadata(
...
             throw new ItemNotFoundException(ErrorCodes.STORAGE_NO_JOB_FOUND, id);
         }
```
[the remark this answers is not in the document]
rsynek: Fixed.

model/rest/src/main/java/ai/timefold/solver/model/rest/impl/AbstractModelAPIResource.java
rsynek: Thanks, fixed.

cristianonicolai (approved):

winklerm (approved):
Thank you, looks good to me!

triceo (approved):
Please use var consistently; both files have `var` only selectively, yet the code introduces lines where `var` is an obvious win.

I recommend configuring IntelliJ to show that as soft-warning, and then mass-fixing that on the entire file; there is a keyboard shortcut for it which I currently don't remember. :-)

LGTM when that's done.

PR #2355, opened 2026-06-10
feat: add model post processors results to FinalBestSolutionEvent

cristianonicolai:


service/definition/src/main/java/ai/timefold/solver/service/definition/api/ModelPostProcessingResult.java
```diff
@@ -0,0 +1,8 @@
+package ai.timefold.solver.service.definition.api;
+
+/**
+ * Marker interface for an optional output produced by a {@link ModelPostProcessor} for a completed model run.
+ */
+public interface ModelPostProcessingResult {
```
rsynek, line 6: What exactly are we trying to send for the `FinalBestSolutionEvent`?

cristianonicolai: Idea here is that _ScoreAnalysisModelPostProcessor_ could return some data structure that implements this interface. Something like _ScoreAnalysisConstraintDetail_ is what we actually need to send to Insights. I'm adding this as interface and generic solution so this can potentially be extended to other post processors data that we might need in the future.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/log/LogsPostProcessor.java
```diff
@@ -46,13 +47,15 @@ public void processComputed(ModelOutput modelOutput, SolverModel<?> solverModel,
     }
 
     @Override
-    public void process(ModelOutput modelOutput, SolverModel<?> solverModel, String id) {
+    public ModelPostProcessingResult processSolved(ModelOutput modelOutput, SolverModel<?> solverModel, String id) {
         if (logsStored.get()) {
-            return;
+            return null;
         }
         logsStored.set(true);
```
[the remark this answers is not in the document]
cristianonicolai: not in scope

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -710,15 +714,20 @@ private void postProcessOutput(String id, ModelOutput modelOutput, SolverModel s
...
-                processor.process(modelOutput, solverModel, id);
+                ModelPostProcessingResult result = processor.processSolved(modelOutput, solverModel, id);
+                if (result != null) {
+                    results.add(result);
+                }
             } catch (Throwable e) {
-                LOGGER.warn("Unexpected error while invoking post processor {} that returned {}",
-                        processor.getClass().getSimpleName(), e.getMessage());
+                LOGGER.warn("Unexpected error while invoking post processor {}", processor.getClass().getSimpleName(), e);
             }
```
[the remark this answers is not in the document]
cristianonicolai: not in scope

service/definition/src/main/java/ai/timefold/solver/service/definition/api/ModelPostProcessor.java
```diff
@@ -1,16 +1,21 @@
 package ai.timefold.solver.service.definition.api;
 
+import org.jspecify.annotations.Nullable;
+
 public interface ModelPostProcessor {
 
     /**
-     * Process successfully solved model run.
+     * Process a successfully solved model run and optionally return a result to attach to the final best solution event.
```
rsynek: I don't think the `ModelPostProcessor` should mention the event; it's not its business.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -681,10 +683,12 @@ protected void notifyOnComplete(String id, SolverModel solverModel) {
                     extractOutputMetrics(solverModel));
 
             processor.onNext(metadata);
-            sendEvent(finalSolutionEmitter, new FinalBestSolutionEvent(metadata, solverModel,
-                    new SolverWorkerJobState(SolverStatus.NOT_SOLVING, solverJob), planName, tenantName));
 
-            postProcessCompleteOutput(id, modelOutput, solverModel);
+            List<ModelPostProcessingResult> postProcessingResults = postProcessCompleteOutput(id, modelOutput, solverModel);
```
rsynek, line 687: I think changing the order has impacts:

- we complete solving in the `metadata` object, then start post processing and after it's done, we send the final event
- post processing involves score calculation (how long does it take for large datasets?) and for routing models also computing waypoints (the same; this time involves calling the map service and OSRM, so the delay may be longer)

Won't this create a discrepancy between reported solving completed, logs, and the score chart?

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -710,15 +714,20 @@ private void postProcessOutput(String id, ModelOutput modelOutput, SolverModel s
...
     }
 
-    private void postProcessCompleteOutput(String id, ModelOutput modelOutput, SolverModel solverModel) {
+    private List<ModelPostProcessingResult> postProcessCompleteOutput(String id, ModelOutput modelOutput,
+            SolverModel solverModel) {
+        List<ModelPostProcessingResult> results = new ArrayList<>();
         for (var processor : modelPostProcessors) {
             try {
-                processor.process(modelOutput, solverModel, id);
+                ModelPostProcessingResult result = processor.process(modelOutput, solverModel, id);
```
rsynek, line 722: As I said, the post-processing is optional, which is best seen in the try-catch that only logs a warning. If the score analysis fails for whatever reason, what do we send over to Insights?

Proposal: maybe we should take the score analysis out of post-processing and compute it right after the solving finishes, before we actually change the state to SOLVING_COMPLETED.

What do you think?

rsynek (commented):
Thanks @cristianonicolai. I left there a couple of comments that crossed my mind only when I saw the last changes.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
cristianonicolai: I guess what you're proposing is to make only score analysis mandatory for completing a solve, instead of all post processors. I think that's fine if we still want to keep that flexibility, otherwise I would say to consider solving completed only when all processors are done.
Another challenge I see is how do we move the processing of score analysis as required, so before SolverWorker can update the status, and bring back the required data to be attached to the FinalBestSolutionEvent?

service/definition/src/main/java/ai/timefold/solver/service/definition/api/ModelPostProcessor.java
cristianonicolai: indeed, removed.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
rsynek: We could call the `ScoreAnalysisService` directly from the `SolverWorker`, then store it, update the status, and fire the `FinalBestSolutionEvent` event.

PR #2356, opened 2026-06-10
fix: seperate move and solver RandomGenerator usage

Christopher-Chianelli:
Before, some foragers and acceptors used the same RandomGenerator as moves, which can affect reproducibility when multithreading is used.

Now, foragers and acceptors use their own independent random.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseFactory.java
@@ -162,7 +162,7 @@ public class DefaultExhaustiveSearchPhaseFactory<Solution_>
             boolean scoreBounderEnabled, boolean isListVariable) {
         var manualEntityMimicRecorder = new ManualEntityMimicRecorder<>(sourceEntitySelector);
         var entityClassName = sourceEntitySelector.getEntityDescriptor().getEntityClass().getName();
-        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom());
+        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom().moveUsage());
         configPolicy.addEntityMimicRecorder(mimicSelectorId, manualEntityMimicRecorder);
         var variableDescriptorList = getGenuineVariableDescriptorList(sourceEntitySelector, isListVariable);
         MoveSelectorConfig<?> moveSelectorConfig = phaseConfig.getMoveSelectorConfig();
--- a/core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseFactory.java
@@ -162,7 +162,7 @@ public class DefaultExhaustiveSearchPhaseFactory<Solution_>
             boolean scoreBounderEnabled, boolean isListVariable) {
         var manualEntityMimicRecorder = new ManualEntityMimicRecorder<>(sourceEntitySelector);
         var entityClassName = sourceEntitySelector.getEntityDescriptor().getEntityClass().getName();
-        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom().moveUsage());
+        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom().factoryUsage());
         configPolicy.addEntityMimicRecorder(mimicSelectorId, manualEntityMimicRecorder);
         var variableDescriptorList = getGenuineVariableDescriptorList(sourceEntitySelector, isListVariable);
         MoveSelectorConfig<?> moveSelectorConfig = phaseConfig.getMoveSelectorConfig();
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
@@ -27,18 +27,21 @@ public final class SelectorBasedRuinRecreateMove<Solution_> extends AbstractSele
     private final SolverScope<Solution_> solverScope;
     private final List<Object> ruinedEntityList;
     private final SequencedSet<Object> affectedValueSet;
+    private final long randomSeed;
 
     private Object @Nullable [] recordedNewValues;
 
     public SelectorBasedRuinRecreateMove(GenuineVariableDescriptor<Solution_> genuineVariableDescriptor,
             RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> constructionHeuristicPhaseBuilder,
-            SolverScope<Solution_> solverScope, List<Object> ruinedEntityList, SequencedSet<Object> affectedValueSet) {
+            SolverScope<Solution_> solverScope, List<Object> ruinedEntityList, SequencedSet<Object> affectedValueSet,
+            long randomSeed) {
         this.genuineVariableDescriptor = genuineVariableDescriptor;
         this.ruinedEntityList = ruinedEntityList;
         this.affectedValueSet = affectedValueSet;
         this.constructionHeuristicPhaseBuilder = constructionHeuristicPhaseBuilder;
         this.solverScope = solverScope;
         this.recordedNewValues = null;
+        this.randomSeed = randomSeed;
     }
 
     @Override
@@ -66,7 +69,7 @@ public final class SelectorBasedRuinRecreateMove<Solution_> extends AbstractSele
         var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
         nestedSolverScope.setSolver(solverScope.getSolver());
         nestedSolverScope.setScoreDirector(innerScoreDirector);
-        nestedSolverScope.setWorkingRandom(DefaultRandomSource.seeded(0L));
+        nestedSolverScope.setWorkingRandom(DefaultRandomSource.seeded(randomSeed));
         constructionHeuristicPhase.solvingStarted(nestedSolverScope);
         constructionHeuristicPhase.solve(nestedSolverScope);
         constructionHeuristicPhase.solvingEnded(nestedSolverScope);
@@ -92,7 +95,7 @@ public final class SelectorBasedRuinRecreateMove<Solution_> extends AbstractSele
         var rebasedRuinedEntityList = rebaseList(ruinedEntityList, lookup);
         var rebasedAffectedValueSet = rebaseSet(affectedValueSet, lookup);
         return new SelectorBasedRuinRecreateMove<>(genuineVariableDescriptor, constructionHeuristicPhaseBuilder, solverScope,
-                rebasedRuinedEntityList, rebasedAffectedValueSet);
+                rebasedRuinedEntityList, rebasedAffectedValueSet, randomSeed);
     }
 
     @Override
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
@@ -12,6 +12,7 @@ import ai.timefold.solver.core.impl.heuristic.move.AbstractSelectorBasedMove;
 import ai.timefold.solver.core.impl.move.VariableChangeRecordingScoreDirector;
 import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
 import ai.timefold.solver.core.impl.score.director.VariableDescriptorAwareScoreDirector;
+import ai.timefold.solver.core.impl.solver.random.DefaultRandomSource;
 import ai.timefold.solver.core.impl.solver.scope.SolverScope;
 import ai.timefold.solver.core.preview.api.move.Move;
 
@@ -65,6 +66,7 @@ public final class SelectorBasedRuinRecreateMove<Solution_> extends AbstractSele
         var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
         nestedSolverScope.setSolver(solverScope.getSolver());
         nestedSolverScope.setScoreDirector(innerScoreDirector);
+        nestedSolverScope.setWorkingRandom(DefaultRandomSource.seeded(0L));
         constructionHeuristicPhase.solvingStarted(nestedSolverScope);
         constructionHeuristicPhase.solve(nestedSolverScope);
         constructionHeuristicPhase.solvingEnded(nestedSolverScope);
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
@@ -31,6 +31,23 @@ public final class SelectorBasedRuinRecreateMove<Solution_> extends AbstractSele
 
     private Object @Nullable [] recordedNewValues;
 
+    /**
+     * Prefer
+     * {@link #SelectorBasedRuinRecreateMove(GenuineVariableDescriptor, RuinRecreateConstructionHeuristicPhaseBuilder, SolverScope, List, SequencedSet, long)}
+     * with explicitly provided random seed.
+     *
+     * @param genuineVariableDescriptor
+     * @param constructionHeuristicPhaseBuilder
+     * @param solverScope
+     * @param ruinedEntityList
+     * @param affectedValueSet
+     */
+    public SelectorBasedRuinRecreateMove(GenuineVariableDescriptor<Solution_> genuineVariableDescriptor,
+            RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> constructionHeuristicPhaseBuilder,
+            SolverScope<Solution_> solverScope, List<Object> ruinedEntityList, SequencedSet<Object> affectedValueSet) {
+        this(genuineVariableDescriptor, constructionHeuristicPhaseBuilder, solverScope, ruinedEntityList, affectedValueSet, 0);
+    }
+
     public SelectorBasedRuinRecreateMove(GenuineVariableDescriptor<Solution_> genuineVariableDescriptor,
             RuinRecreateConstructionHeuristicPhaseBuilder<Solution_> constructionHeuristicPhaseBuilder,
             SolverScope<Solution_> solverScope, List<Object> ruinedEntityList, SequencedSet<Object> affectedValueSet,
--- a/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMoveSelectorFactory.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMoveSelectorFactory.java
@@ -51,7 +51,7 @@ public class SwapMoveSelectorFactory<Solution_>
             if (entitySelectorConfig.getId() == null && entitySelectorConfig.getMimicSelectorRef() == null) {
                 var entityName = Objects.requireNonNull(entityDescriptor.getEntityClass().getSimpleName());
                 // We set the id to make sure the value selector will use the mimic recorder
-                entityRecorderId = ConfigUtils.addRandomSuffix(entityName, configPolicy.getRandom());
+                entityRecorderId = ConfigUtils.addRandomSuffix(entityName, configPolicy.getRandom().moveUsage());
                 entitySelectorConfig.setId(entityRecorderId);
             } else {
                 entityRecorderId = entitySelectorConfig.getId() != null ? entitySelectorConfig.getId()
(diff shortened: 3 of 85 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -143,12 +144,25 @@ public void setStartingSolverCount(int startingSolverCount) {
...
+        return solverWorkingRandom;
+    }
+
+    /**
+     * Only used for tests
+     */
+    public void setTestWorkingRandom(RandomGenerator workingRandom) {
+        this.moveGeneratorWorkingRandom = workingRandom;
+        this.solverWorkingRandom = workingRandom;
+    }
```
triceo: Please don't. Find another way. Public methods on types as prominent as this must not be only used for testing.

core/src/main/java/ai/timefold/solver/core/impl/solver/scope/SolverScope.java
```diff
@@ -143,12 +144,25 @@ public void setStartingSolverCount(int startingSolverCount) {
...
-        return workingRandom;
+    public RandomGenerator getMoveGeneratorWorkingRandom() {
+        return moveGeneratorWorkingRandom;
     }
 
-    public void setWorkingRandom(RandomGenerator workingRandom) {
-        this.workingRandom = workingRandom;
+    public RandomGenerator getSolverWorkingRandom() {
+        return solverWorkingRandom;
+    }
```
triceo: These two need to get very good Javadocs.
Who's supposed to use them? What for? What not for? Why?

The whole "using extra random calls affects multi-threading" is not obvious at all. It needs to be clearly explained.

core/src/main/java/ai/timefold/solver/core/impl/phase/scope/AbstractMoveScope.java
```diff
@@ -66,8 +66,12 @@ public Solution_ getWorkingSolution() {
...
 
-    public RandomGenerator getWorkingRandom() {
-        return getStepScope().getWorkingRandom();
+    public RandomGenerator getMoveGeneratorWorkingRandom() {
+        return getStepScope().getMoveGeneratorWorkingRandom();
+    }
+
+    public RandomGenerator getSolverWorkingRandom() {
+        return getStepScope().getSolverWorkingRandom();
     }
```
triceo: Why would move scope need to expose this?
Generally I'd think twice about which of these randoms we make accessible in which circumstances - what a user (or a developer in this case) can't see cannot confuse them and can't be misused.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -48,7 +48,7 @@ public abstract class AbstractSolver<Solution_> implements Solver<Solution_> {
     protected final UniversalTermination<Solution_> globalTermination;
     protected final List<Phase<Solution_>> phaseList;
 
-    private RandomGenerator.@Nullable SplittableGenerator savedRandom;
+    private RandomGenerator.@Nullable SplittableGenerator savedMoveGeneratorRandom;
```
triceo: Why does solver save the _move_ random, and not the _solver_ random? This is confusing; at the very least we have a naming problem here.

triceo (commented):
I think we need more clarity around this concept; as it stands, it is confusing and I smell future problems here.

Is it a naming issue? Is it a documentation issue? Is it a scope/visibility issue? Hard to say. I'd start by figuring out the right naming, and very clearly defining the demarcation line - what do I use one random for, and what the other?

One problem I have is that - once you have the random - you don't really know anymore _what_ random you have. There's no telling. Maybe instead of having two random instances, we update the internal type to carry the two randoms internally and user would have to specifically request the one which they need? 

Something like `var number = ourCustomRandom.forMoves(r -> r.nextInt());`. It's not impossible to store `r` and reuse it later, but it makes it sufficiently difficult plus it's always very clear that `r` came from `forMoves` because the call site will likely be close in code to the store site.

Exploring that idea further, the lambda there could be a problem, as very often it would be capturing and we don't like that. So, assuming that we only have one instance of `ourCustomRandom`, the best we can do is to have `ourCustomRandom.forMoves().nextInt(bound)`; users can easily store the child random and reuse it, but we control the code and we can kill that idea in code review.

This to me seems like a better approach than to have two RNG instances floating around the solver without having any way to identify which is which. This way, we will be able to immediately verify and look up in the IDE which pieces of code are using which RNG, simply because each call site would be using `forMoves()` or `forSolver()` (TODO naming).

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhaseFactory.java
```diff
@@ -162,7 +162,7 @@ protected EntityDescriptor<Solution_> deduceEntityDescriptor(SolutionDescriptor<
             boolean scoreBounderEnabled, boolean isListVariable) {
         var manualEntityMimicRecorder = new ManualEntityMimicRecorder<>(sourceEntitySelector);
         var entityClassName = sourceEntitySelector.getEntityDescriptor().getEntityClass().getName();
-        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom());
+        var mimicSelectorId = ConfigUtils.addRandomSuffix(entityClassName, configPolicy.getRandom().moveUsage());
```
triceo: IMO this is not move usage. I'd like to completely isolate direct move usage from anything else.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
```diff
@@ -96,6 +97,7 @@ protected void execute(VariableDescriptorAwareScoreDirector<Solution_> scoreDire
             var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
             nestedSolverScope.setSolver(solverScope.getSolver());
             nestedSolverScope.setScoreDirector(nonRecordingScoreDirector);
+            nestedSolverScope.setWorkingRandom(DefaultRandomSource.seeded(0L));
```
triceo: Why is this here now, when previously we did not feel the need to set a new random?
And why is it always seeded to `0`? Shouldn't it be seeded from the parent random? This way, the nested code will always do the same thing, regardless of reproducible/non reproducible.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
```diff
@@ -65,6 +66,7 @@ protected void execute(VariableDescriptorAwareScoreDirector<Solution_> scoreDire
         var nestedSolverScope = new SolverScope<Solution_>(solverScope.getClock());
         nestedSolverScope.setSolver(solverScope.getSolver());
         nestedSolverScope.setScoreDirector(innerScoreDirector);
+        nestedSolverScope.setWorkingRandom(DefaultRandomSource.seeded(0L));
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SwapMoveSelectorFactory.java
```diff
@@ -51,7 +51,7 @@ protected MoveSelector<Solution_> buildBaseMoveSelector(HeuristicConfigPolicy<So
             if (entitySelectorConfig.getId() == null && entitySelectorConfig.getMimicSelectorRef() == null) {
                 var entityName = Objects.requireNonNull(entityDescriptor.getEntityClass().getSimpleName());
                 // We set the id to make sure the value selector will use the mimic recorder
-                entityRecorderId = ConfigUtils.addRandomSuffix(entityName, configPolicy.getRandom());
+                entityRecorderId = ConfigUtils.addRandomSuffix(entityName, configPolicy.getRandom().moveUsage());
```
triceo: Arguably also not move usage.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodTester.java
```diff
@@ -39,6 +40,7 @@ public NeighborhoodTestContext<Solution_> using(Solution_ solution) {
         var moveTestContext = (DefaultMoveTestContext<Solution_>) moveTester.using(solution);
         var scoreDirector = moveTestContext.getScoreDirector();
         var solverScope = new SolverScope<Solution_>();
+        solverScope.setWorkingRandom(RandomSource.seeded(0L));
```
triceo, line 43: Similar comment as above - why newly introduced? Why a fixed seed?

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomSource.java
```diff
@@ -0,0 +1,64 @@
...
+     * seed with the seed included in {@link Object#toString()}.
+     * <p>
+     * Required so the {@link RandomGenerator} is created in the thread {@link AbstractSolver#solve}
+     * is called.
+     *
+     * @param seed The seed to use for the {@link RandomGenerator}.
+     * @return A supplier include the seed in its {@link Object#toString()}
+     */
+    public static Supplier<RandomSource> seededSupplier(long seed) {
+        return new Supplier<>() {
```
triceo, line 22: Thinking out loud... wouldn't `IntFunction` work better here? No need for multiple instances in that case.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Used by {@link ai.timefold.solver.core.impl.heuristic.selector.move.MoveSelector}
+     * and other components.
+     *
+     * @return A {@link RandomGenerator} instance that advances the source
+     *         {@link RandomGenerator} state.
+     */
+    RandomGenerator moveUsage();
```
triceo: `inMoveUsage()` perhaps?

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Used by {@link ai.timefold.solver.core.impl.localsearch.decider.acceptor.Acceptor}
+     * and other components inbetween move generation.
+     *
+     * @return A {@link RandomGenerator} instance that does not advance the source
+     *         {@link RandomGenerator} state (and hence, any usage does not affect
+     *         reproducibility).
+     */
+    RandomGenerator acceptorUsage();
```
triceo, line 55: I'd be looking for a different name. This is a general purpose random, as opposed to `moveUsage()`.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
```diff
@@ -0,0 +1,35 @@
...
+    static RandomSource seeded(long seed) {
+        return DefaultRandomSource.seeded(seed);
+    }
+
+    /**
+     * Used by {@link ai.timefold.solver.core.impl.heuristic.selector.move.MoveSelector}
+     * and other components.
+     *
+     * @return A {@link RandomGenerator} instance that advances the source
+     *         {@link RandomGenerator} state.
```
triceo: The "advancing" of the source needs to be explain in the class comments. What is it, what it's for.

Arguably, the reason for why we have these two different random should also be explained. (How non-moves influence move selection.)

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param seed Controls the random sequence generated.
+     *
+     * @return A new {@link RandomSource} from a given seed.
+     */
+    static RandomSource seeded(long seed) {
+        return DefaultRandomSource.seeded(seed);
+    }
+
+    /**
+     * Used by {@link ai.timefold.solver.core.impl.heuristic.selector.move.MoveSelector}
```
triceo: I'd prefer avoiding FQNs in Javadoc - it makes them hard to read in the source code.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -1408,12 +1409,15 @@ void solveStaleBuiltinShadows() {
     }
 
     @Test
+    @Timeout(value = 10, unit = TimeUnit.SECONDS)
```
triceo: Why is this suddenly necessary?

core/src/test/java/ai/timefold/solver/core/testutil/PlannerTestUtils.java
```diff
@@ -216,6 +217,16 @@ public static <X extends Comparable<X>, Y> SortedMap<X, Y> asSortedMap(X x1, Y y
     // Scope helpers
     // ************************************************************************
 
+    /**
+     * Returns a mock {@link SolverScope} with {@link SolverScope#getWorkingRandom()} returning a mock random.
+     */
+    @SuppressWarnings("unchecked")
+    public static <Solution_> SolverScope<Solution_> mockSolverScope() {
+        var out = mock(SolverScope.class);
+        when(out.getWorkingRandom()).thenReturn(mock(RandomSource.class));
```
triceo, line 226: Since you already have `MockRandomSource`, wouldn't it make sense to return that here?

triceo (commented):
Comments mostly on the API.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
Christopher-Chianelli: These moves do not even access Random IIRC, but because all MoveSelectors saves Random via `RandomSource.moveUsage()`, `RandomSource` cannot be null.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/SelectorBasedRuinRecreateMove.java
Christopher-Chianelli: Same as above

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodTester.java
Christopher-Chianelli: Same as above; a fixed seed for reproducibility since this is used in tests.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
Christopher-Chianelli: You cannot call the Solver's random here, because this move may be executed in a different thread.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
Christopher-Chianelli: I changed the termination to `bestScoreFeasible`, since the changes to Random change the step count required. Thus, if we somehow made a change that make LS never find a feasible solution, this test would run forever.

core/src/test/java/ai/timefold/solver/core/testutil/PlannerTestUtils.java
Christopher-Chianelli: `MockRandomSource` is more for using a given `Random` or `RandomGenerator` as a `RandomSource`. This is mostly to make sure `getWorkingRandom()` does not return null to avoid a NPE since every MoveSelector saves their Random via `RandomSource#moveUsage`.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
triceo: A good opportunity to add a comment.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
Christopher-Chianelli: All usage should have their own RandomGenerator; there no general purpose `RandomGenerator`. What matter is how random calls are woven together. If components share the same random generator, they cannot conditionally consume random based on ex: if a move is available. This `RandomGenerator` is exclusively for acceptors since they might do that. If another component could do that, then they would also need their own generator.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
Christopher-Chianelli: Removed; there no advancing done, since the split occurs at step start (and thus number of random calls does not matter, what does matter is how they are interwoven (i.e. does an acceptor consume random then the iterator, or does the iterator consume random then the acceptor?)

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomSource.java
Christopher-Chianelli: The Solver constructs it and does not have access to the seed, so the supplier would need to be somewhere. Cannot just return a `RandomSource` instance, since the thread that created is the only thread allowed to access it.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
triceo: I accept that logic, but seeding this with zero is still wrong.

None of the code may use this random now, but that's no guarantee for the future. If something suddenly starts using the random, this PR silently introduces a determinism where the user may not want any based on the environment mode.

Please figure out a way to seed this correctly.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/DefaultNeighborhoodTester.java
triceo: In this case, zero seed makes sense. Resolving comment.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
triceo: Aren't we being a little paranoid here?
We had a real problem with moves - since we isolated moves, IMO we can treat all other randomness as a single class, all together. It should not affect the solver's reproducibility if all codebase - other than moves - is using the same random.

Am I wrong? Give me an example.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomSource.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static DefaultRandomSource seeded(long seed) {
+        var moveRandom = new DelegatingSplittableRandomGenerator(seed);
+        var factoryRandom = new DelegatingSplittableRandomGenerator(seed, moveRandom.split());
+        var acceptorRandom = new DelegatingSplittableRandomGenerator(seed, moveRandom.split());
+        return new DefaultRandomSource(moveRandom, factoryRandom, acceptorRandom);
+    }
+
+    public DefaultRandomSource split() {
```
triceo, line 42: I think that, somewhere along the way, a comment regarding the logic of and need for splitting was lost. Please add it here, in one centralized place; describe why we're doing this.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/RandomSource.java
Christopher-Chianelli: Say a new component X and acceptor share the same RandomGenerator; say the component runs right before a step end and calls Random; since acceptors are called conditionally based on if a move result is available,  the number of calls for the acceptor's RNG is nondeterministic. Likewise, the number of moves generated is nondeterministic, so for that component X to have a deterministic RNG, it cannot share either the acceptor nor move's RNG.

>  Moves and acceptors use different `{@link RandomGenerator}` since
>  an acceptor is called when a move has been evaluated, and
>  in a multithreaded context, we can generate moves while waiting
>  for a prior move to be evaluated. Thus, if they used the same
>  `{@link RandomGenerator}`, the moves generated would not be deterministic,
>  since sometimes there will be another random call since an earlier move was
>  evaluated, and sometimes there won't be.

core/src/main/java/ai/timefold/solver/core/impl/solver/random/DefaultRandomSource.java
Christopher-Chianelli: It in the Javadoc of `RandomSource`:
```
Moves and acceptors use different {@link RandomGenerator} since
an acceptor is called when a move has been evaluated, and
in a multithreaded context, we can generate moves while waiting
for a prior move to be evaluated. Thus, if they used the same
{@link RandomGenerator}, the moves generated would not be deterministic,
since sometimes there will be another random call since an earlier move was
evaluated, and sometimes there won't be.
```

triceo (approved):

PR #2360, opened 2026-06-11
chore: rename model to service

rsynek:
Renames the `timefold-solver-model` to `timefold-solver-service`.

Successfully migrated locally:
- FSR compiles and all tests pass
- platform compiles; test failures due to missing local configuration (same failures even before the migration)


triceo (approved):
I'm approving pro-forma. If the models migrate to this, this is correct.

PR #2366, opened 2026-06-12
chore: update 1.x workflow

zepfred:


zepfred:
Thinking about this workflow, I believe we won't need the file on the main branch, since we will check out the `1.x` branch, which contains it.

PR #2373, opened 2026-06-15
perf: further improve performance of ElementAwareArrayList

triceo:
The goal here is to minimize the overhead of add/remove methods, as they showed up the highest in profiling in comparison to the original ElementAwareLinkedList.


core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -108,14 +121,15 @@ private Entry partialCompact(int rightBoundaryPosition) {
...
                 }
                 if (lastNonNullPosition == rightBoundaryPosition) {
-                    // Invariant: positions [0, index] are all non-null,
-                    // so all gapCount nulls lie in [index+1, lastElementPosition].
-                    // If that suffix is entirely nulls (equivalent to index == size()-1), trim it now.
+                    // Invariant: positions [0, rightBoundaryPosition] are all non-null,
+                    // so all gapCount nulls lie in [rightBoundaryPosition+1, lastElementPosition].
+                    // If that suffix is entirely nulls (equivalent to rightBoundaryPosition == size()-1), trim it now.
                     if (gapCount == lastElementPosition - rightBoundaryPosition) {
                         truncateTo(rightBoundaryPosition);
```
zepfred, line 131: I wonder if we should a parameter to `truncateTo` method not to load the range with null values. For this specific call, it seems not necessary to set the positions with null.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -174,27 +189,27 @@ public void add(int index, T element) {
     }
 
     private void addWithoutGaps(int index, T element) {
-        modCount++;
         var newEntry = new Entry(element, index);
         resize(lastElementPosition + 2);
```
zepfred, line 194: Since the resize operation can be costly, why not simply request the necessary space of `lastElementPosition + 1`?

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -324,8 +345,14 @@ public Iterator<T> iterator() {
...
     }
 
+    @Override
+    public ListIterator<T> listIterator() {
+        return this.listIterator(0);
+    }
+
     @Override
     public ListIterator<T> listIterator(int index) {
+        compact(); // Ensure fast-path iteration; remove all gaps at once.
```
zepfred, line 356: If the `index` is smaller than the first gap position, we could skip the compact operation and instruct the iterator to bypass null values.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -389,35 +414,35 @@ public int previousIndex() {
 
         @Override
         public T next() {
-            checkModCount();
             if (logicalPosition >= size()) {
                 throw new NoSuchElementException();
             }
-            var entry = entries[currentPosition];
+            var entry = (Entry) entries[currentPosition];
             while (entry == null) {
```
zepfred, line 419: We are now compacting the values. Is it still possible to find any null values?

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -389,35 +414,35 @@ public int previousIndex() {
...
         }
 
         @Override
         public T previous() {
-            checkModCount();
             if (logicalPosition <= 0) {
                 throw new NoSuchElementException();
             }
             Entry entry = null;
             while (entry == null) {
```
zepfred, line 436: Ditto.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
```diff
@@ -440,26 +463,17 @@ public void set(T element) {
...
                 throw new IllegalStateException("set() called without a preceding next() or previous().");
             }
-            checkModCount();
             lastEntry.replaceElement(element);
         }
 
         @Override
         public void add(T element) {
-            checkModCount();
             ElementAwareArrayList.this.add(logicalPosition, element);
```
zepfred, line 470: Should we allow the iterator to modify the state of the list?

zepfred (commented):
Looks good to me. I have a few minor comments and some questions.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: Since the nulling out happens with `Arrays.fill` which is a CPU intrinsic, the few extra bytes won't make too much of a difference.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: We will actually resize to `Math.max(entries.length * 2, minCapacity)` - doubling the array - and therefore whether `minCapacity` is +1 or +2 is pretty much irrelevant. We double the size (much like `ArrayList` does) precisely because resizing is expensive and we want to minimize how often it's done.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: We have to compact at some point, the list can't keep growing indefinitely. This is a logical place to compact; iterating over an ever-growing list full of gaps costs time too.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: Yes, the iterator has a `remove()` method and allows iterating both backward and forward. This means we can possibly remove something and then iterate to the new gap.

core/src/main/java/ai/timefold/solver/core/impl/util/ElementAwareArrayList.java
triceo: Yes. It is the standard contract of a `List`. This class is intended as a replacement for `ArrayList` where we have to.

zepfred (approved):

PR #2378, opened 2026-06-17
fix: remove the resolvedMapLocation from API

rsynek:
The `resolvedMapLocation` leaked into the API, where it does not belog.

### DIFF
--- a/service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/Metadata.java
+++ b/service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/Metadata.java
@@ -70,8 +70,7 @@ public final class Metadata<Score_> implements Status<Score_> {
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private String failureMessage;
 
-    @Schema(nullable = true,
-            description = "The map-service region resolved when location is auto-select.")
+    @Schema(hidden = true)
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private String resolvedMapLocation;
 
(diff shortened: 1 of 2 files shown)


service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/Metadata.java
```diff
@@ -70,8 +70,7 @@ public final class Metadata<Score_> implements Status<Score_> {
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private String failureMessage;
 
-    @Schema(nullable = true,
-            description = "The map-service region resolved when location is auto-select.")
+    @Schema(hidden = true)
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
```
rsynek, line 74: This is only a partial hotfix, since for routing models, the field will be non-empty and thus it will be sent in the JSON payload to the client side.

Doing the full split means handling this in the REST API, both in the solver-service and the platform.

Another option is reverting the original change.

mswiderski (approved):
I'd go with this as quick fix to make sure this is not "advertised" via open api and then be refactored down the line.

diogodanielsoaresferreira (approved):

diogodanielsoaresferreira:
I am ok with hiding it or removing it completely for now, as this is not an urgent feature.

rsynek:
Closing in favor of https://github.com/TimefoldAI/timefold-solver/pull/2380 and https://github.com/TimefoldAI/timefold-solver-enterprise/pull/702.

PR #2385, opened 2026-06-18
feat: Add effective map location in AbstractDatesetEvents

diogodanielsoaresferreira:
Add effective map location in AbstractDatasetEvents, so that we know which effective location was used when auto-select location was enabled.
Fixes: https://github.com/TimefoldAI/timefold-platform/issues/4452

### DIFF
--- a/service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
+++ b/service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
@@ -53,7 +53,6 @@ import ai.timefold.solver.service.definition.internal.MapEnrichmentContext;
 import ai.timefold.solver.service.definition.internal.error.ErrorCodes;
 import ai.timefold.solver.service.definition.internal.error.ItemNotFoundException;
 import ai.timefold.solver.service.definition.internal.error.TimefoldRuntimeException;
-import ai.timefold.solver.service.definition.internal.events.AbstractDatasetEvent;
 import ai.timefold.solver.service.definition.internal.events.AbstractEvent;
 import ai.timefold.solver.service.definition.internal.events.BestSolutionEvent;
 import ai.timefold.solver.service.definition.internal.events.DatasetComputedEvent;
@@ -161,8 +160,6 @@ public class SolverWorker {
 
     private AtomicBoolean shuttingDown = new AtomicBoolean(false);
 
-    private final AtomicBoolean resolvedMapLocationLogged = new AtomicBoolean(false);
-
     private BroadcastProcessor<Metadata<?>> processor = BroadcastProcessor.create();
 
     @Inject
@@ -249,7 +246,6 @@ public class SolverWorker {
 
     private void sendEvent(Emitter emitter, AbstractEvent event) {
         try {
-            enrichResolvedMapLocation(event);
             emitter.send(event).toCompletableFuture().get(EMITTER_TIMEOUT, TimeUnit.SECONDS);
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
             if (e instanceof InterruptedException) {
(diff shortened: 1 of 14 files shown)


mswiderski (approved):

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -776,6 +785,19 @@ public void notifyOnFailure(Object id, Throwable throwable) {
         }
     }
 
+    private void enrichResolvedMapLocation(AbstractEvent event) {
```
rsynek: I don't understand the architecture of how we transfer the resolved map location to the platform.

This code effectively sends the resolved location in every single instance of `AbstractDatasetEvent`, which includes all intermediate best solution events, and more.

This is a piece of information that gets resolved at enrichment time, but since then, it does not change. Why do we need to keep sending it over and over again?

Which component in the platform needs to know the resolved map location and why?

diogodanielsoaresferreira: Currently we store all events in insights, with mapLocation as part of the metadata. By default, we fetch the location from the env properties, which sets the location to auto-select.

With this PR and https://github.com/TimefoldAI/timefold-solver-enterprise/pull/705, if there is any effective location, we override the location fetched from the env vars.

Currently the location is only exposed via the GraphQL API, but I can see this being used by the UI to show the map region. In that case, only the last event is important. We can send the effective map location only for terminal states, but I decided to send for all events if available because 1) it's relatively cheap, and 2) if the run fails and the failed events are emitted by the gateway, we know which map it used.

diogodanielsoaresferreira: @cristianonicolai since this is mainly insights data, I'm tagging Cristiano for input. Currently we set the effective map location in all model events, so that this is always set in the Run metadata. I believe this populates the field in the collections field-service-routing_v1 (for each tenant) and also RunStatistics.

Can we send it only in the first or only in the last event, and mark the field as null for other events?
The drawback of sending it only in the first event is that we need to keep track of it. If later events are send with null, what happens to the field? Does it stay as null or does the initial value still holds?
And if we consider to send it only in the final events, what happens if the pod fails mid-run and we never send the map region?

cristianonicolai: That can be done in Insgihts but needs special handling. Basically you dont want to override the value if we send null. I believe we have other properties that behave like that already. Like these: https://github.com/TimefoldAI/timefold-platform/blob/main/insights/insights-service/src/main/java/ai/timefold/platform/insights/mapper/PlanningSolutionMapper.java#L58-L61

rsynek: Don't we have different even types attached to individual state transitions?

rsynek: I am looking for attaching this piece of information to the right event type and sending it over the wire hopefully just once.

Alternatively, we can consider some internal storage of metadata, different than the API metadata, where we can store additional information about the dataset.

diogodanielsoaresferreira: Makes sense to me. I added the field to the DatasetComputed event and removed from all other events.

diogodanielsoaresferreira: solver-enterprise and platform repos were also updated so that the effective map location is stored in the collection even if other events do not send it.

rsynek (approved):

PR #2395, opened 2026-06-19
chore: improve maven plugin documentation

oemebamo:
- Made the page structure better (it currently doesn't render properly on Github);
- Included more details on what permissions are needed for the PAT (and how to generate them);

Apart from that, I'm wondering about these things:
- Is the `groupId` in the POM example correct? `ai.timefold.sdk.tools` should be `ai.timefold.solver`?
- Wrong package path in "Implementation notes": `src/main/java/ai/timefold/sdk/tools/maven` should probably be `src/main/java/ai/timefold/solver/tools/maven`?
- `timefold.model.key` is marked as required for deploy & undeploy, but probably is also required for configure?


rsynek (approved):
Thanks for improving the format and adding the details about the token.

PR #2396, opened 2026-06-19
perf: add ability to ignore inconsistent solutions

Christopher-Chianelli:
Added a new preview feature `IGNORE_INCONSISTENT_SOLUTIONS`.
When enabled, when a move causes the solution to be in an inconsistent state, that move is effectively ignored; `calculateScore` will return a score with the minimum init value possible (i.e. a score that is less than all other scores) and no shadow variables will be updated.

Technically, this potentially introduces a score trap: all inconsistent solutions get the same -infinity score. However, this doesn't seem to be a problem in the benchmarks I performed.

`TopologicalGraph#commitChanges` signature has changed to return a boolean which is true iff the graph has loops (and thus the solution inconsistent.

Benchmarks show significant improvements for Machine Job Scheduling on Enterprise.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
@@ -21,12 +21,16 @@ final class AffectedEntitiesUpdater<Solution_>
     // Internal state; expensive to create, therefore we reuse.
     private final LoopedTracker loopedTracker;
     private final BitSet visited;
+    private final boolean ignoreInconsistentSolutions;
     private final PriorityQueue<BaseTopologicalOrderGraph.NodeTopologicalOrder> changeQueue;
+    private boolean consistencyProcessed;
+
 
     AffectedEntitiesUpdater(BaseTopologicalOrderGraph graph, List<GraphNode<Solution_>> nodeList,
             BaseTopologicalOrderGraph.NodeTopologicalOrder[] nodeTopologicalOrders,
             Function<Object, List<GraphNode<Solution_>>> entityToContainingNode,
-            int entityCount, ChangedVariableNotifier<Solution_> changedVariableNotifier) {
+            int entityCount, ChangedVariableNotifier<Solution_> changedVariableNotifier,
+            boolean ignoreInconsistentSolutions) {
         this.graph = graph;
         this.nodeList = nodeList;
         this.nodeTopologicalOrders = nodeTopologicalOrders;
@@ -36,6 +40,8 @@ final class AffectedEntitiesUpdater<Solution_>
                 createNodeToEntityNodes(entityCount, nodeList, entityToContainingNode));
         this.visited = new BitSet(instanceCount);
         this.changeQueue = new PriorityQueue<>(instanceCount);
+        this.ignoreInconsistentSolutions = ignoreInconsistentSolutions;
+        this.consistencyProcessed = false;
     }
 
     static <Solution_> int[][] createNodeToEntityNodes(int entityCount,
@@ -98,6 +104,7 @@ final class AffectedEntitiesUpdater<Solution_>
 
         // Prepare for the next time updateChanged() is called.
         // No need to clear changeQueue, as that already finishes empty.
+        consistencyProcessed = true;
         loopedTracker.clear();
         visited.clear();
     }
@@ -129,19 +136,22 @@ final class AffectedEntitiesUpdater<Solution_>
 
         // Do not need to update anyChanged here; the graph already marked
         // all nodes whose looped status changed for us
-        var groupEntities = shadowVariableReferences.get(0).groupEntities();
-        var groupEntityIds = entityVariable.groupEntityIds();
 
-        if (groupEntities != null) {
-            for (var i = 0; i < groupEntityIds.length; i++) {
-                var groupEntity = groupEntities[i];
-                var groupEntityId = groupEntityIds[i];
+        if (!ignoreInconsistentSolutions || !consistencyProcessed) {
+            var groupEntities = shadowVariableReferences.get(0).groupEntities();
+            var groupEntityIds = entityVariable.groupEntityIds();
+
+            if (groupEntities != null) {
+                for (var i = 0; i < groupEntityIds.length; i++) {
+                    var groupEntity = groupEntities[i];
+                    var groupEntityId = groupEntityIds[i];
+                    anyChanged |=
+                            updateLoopedStatusOfEntity(groupEntity, groupEntityId, entityConsistencyState);
+                }
+            } else {
                 anyChanged |=
-                        updateLoopedStatusOfEntity(groupEntity, groupEntityId, entityConsistencyState);
+                        updateLoopedStatusOfEntity(entity, entityVariable.entityId(), entityConsistencyState);
             }
-        } else {
-            anyChanged |=
-                    updateLoopedStatusOfEntity(entity, entityVariable.entityId(), entityConsistencyState);
         }
 
         for (var shadowVariableReference : shadowVariableReferences) {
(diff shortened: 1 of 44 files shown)


triceo:
What backwards incompatible behavior we are guarding against here using `IGNORE_INCONSISTENT_SOLUTIONS`? This feature is harmless; or can be made harmless if we ensure that a move which causes an invalid score never actually interacts with the user in any way. If the changes the move has caused are immediately undone, which they will be, you are already 90 % of the way there.

1. Should that move count towards move speed?
2. Should that move show up in the benchmarker?
3. Should that move be logged any differently than productive moves?

I want to argue 1-YES, 2-YES, 3-YES. It is a move like any other, and it does not leave the solver in any inconsistent state. The user should see it. But the score is pretty much pointless, which means the logging should - instead of showing a score - show some other message.

Score traps should not be a problem either. If this move never influences any of the solver's algorithms, it is as if it never existed, and therefore it can not cause a score trap. It will only reduce move speed, a vampiric loss.

Should the move be counted towards tabu search? IMO NO.
Should the move be counted towards late list? IMO NO.
We should create a separate category of "invalid moves" and ensure that our internal algorithms are not misguided by them.
(Although maybe not - if the move is never accepted, it will never enter these structures anyway.)

IMO we should work towards silently enabling this in the solver. We can do this backwards compatibly.

triceo:
Observation (JS lta21, first 5 minutes):

image

Notice that the one method at the top takes 8x time of the second worst. This deserves optimization, not necessarily in this PR.

Christopher-Chianelli:
> What backwards incompatible behavior we are guarding against here using IGNORE_INCONSISTENT_SOLUTIONS? This feature is harmless; or can be made harmless if we ensure that a move which causes an invalid score never actually interacts with the user in any way. If the changes the move has caused are immediately undone, which they will be, you are already 90 % of the way there.

Some users might want to get the score of a planning solution with some inconsistent variables; that is currently allowed and is handled via `@ShadowVariablesInconsistent` (as one solution can be more inconsistent than another). Thing change effectively makes `@ShadowVariablesInconsistent` deprecated. 

> Observation (JS lta21, first 5 minutes): ...

TBH I am shocked that method is even being called; Entities can never be looped with how this is designed. My guess is it might still be iterating all the shadow variables on the entities to check if they are looped.

triceo:
> Some users might want to get the score of a planning solution with some inconsistent variables; that is currently allowed and is handled via `@ShadowVariablesInconsistent` (as one solution can be more inconsistent than another). Thing change effectively makes `@ShadowVariablesInconsistent` deprecated.

Fair point. Even though I'm not worried about this scenario, we can solve this without introducing any preview features:

- We deprecate the annotation.
- But if it is present, we activate the old code.
- If the annotation is not present, we use the new mechanism.
- The documentation will avoid mentioning the annotation prominently and over time, this will result in most users never even starting to use it.

Unfortunately, this means we will have to keep the MNR complexity, but so be it. We will also have to investigate all the things I mentioned above - these invalid moves must never interact with the solver's logic around how future moves are generated, or how new best solutions are selected.

triceo:
More consequences to investigate:

- What should happen when the solver is given an inconsistent solution from the outside?
- What happens when you run score analysis on an inconsistent solution?
- What should happen when `SolutionManager.updateShadowVariables()` is called on an inconsistent problem?
- Any impacts on `ConstraintVerifier`? 
- Any impacts on neighborhoods testing?

I'm starting to think that the solution should carry an inconsistency flag. `@ShadowVariablesInconsistent`, but on solution level. Optionally. That way, many problems can be avoided.

Christopher-Chianelli:
> More consequences to investigate:
> 
>     * What should happen when the solver is given an inconsistent solution from the outside?

Currently, if the feature is enabled, it fail-fast on solve start.

>     * What happens when you run score analysis on an inconsistent solution?

I think it uses the old behaviour.

>     * What should happen when `SolutionManager.updateShadowVariables()` is called on an inconsistent problem?

Ditto.

>     * Any impacts on `ConstraintVerifier`?

None (unless we remove the ability to solve/calculate the score of inconsistent solutions

>     * Any impacts on neighborhoods testing?
> 
> 
> I'm starting to think that the solution should carry an inconsistency flag. `@ShadowVariablesInconsistent`, but on solution level. Optionally. That way, many problems can be avoided.

How would that help? People cannot check the solution in their constraints; I would rather throw an exception in that case so users don't forget to check a flag.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -129,19 +135,22 @@ private boolean updateEntityShadowVariables(GraphNode<Solution_> entityVariable,
...
         // Do not need to update anyChanged here; the graph already marked
         // all nodes whose looped status changed for us
-        var groupEntities = shadowVariableReferences.get(0).groupEntities();
-        var groupEntityIds = entityVariable.groupEntityIds();
 
-        if (groupEntities != null) {
-            for (var i = 0; i < groupEntityIds.length; i++) {
-                var groupEntity = groupEntities[i];
-                var groupEntityId = groupEntityIds[i];
+        if (!ignoreInconsistentSolutions || !consistencyProcessed) {
```
triceo, line 139: This is strange logic.

If I want to ignore inconsistent solutions, why would I also want to check if consistency was processed? I'd expect that, if inconsistencies are ignored, that no further check for consistency processing would be necessary.

Maybe a better naming of variables would clear this?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
```diff
@@ -49,12 +52,16 @@ void markChanged(@NonNull GraphNode<Solution_> node) {
     }
 
     @Override
-    void innerUpdateChanged() {
+    boolean innerUpdateChanged() {
         if (changeTracker.isEmpty()) {
-            return;
+            return true;
+        }
+        if (graph.commitChanges(changeTracker) && ignoreInconsistentSolutions) {
```
triceo, line 59: Wouldn't it make more changes or `ignoreInconsistentSolutions` to come first to possibly avoid the `commitChanges()` work?

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
```diff
@@ -1449,6 +1453,83 @@ void solveStaleDeclarativeShadows() {
         assertThat(solution.getScore()).isEqualTo(HardSoftScore.of(0, -240));
     }
 
+    @Test
+    void solveWhenIgnoringInconsistentSolutionsThrowsIfInitialSolutionInconsistent() {
```
triceo, line 1457: Is this among the agreed list of behaviors?
I recall that inconsistent initial solutions would be solved by unassigning.

triceo (approved):
Good to go to the feature branch when comments resolved.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
Christopher-Chianelli: It for the ConsistencyTracker that ConstraintStream uses; this initializes it on the first update/run. Without it, there are score corruptions.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
Christopher-Chianelli: `commitChanges` return true if there is a loop. This is saying "If there are loops and we can ignore inconsistent solutions, exit early, otherwise process shadow variable update". `commitChanges` MUST be called before the shadow variable update occurs, and hence why it must come first in a short-circuiting boolean expression.

core/src/test/java/ai/timefold/solver/core/impl/solver/DefaultSolverTest.java
Christopher-Chianelli: That is done in a seperate branch to not overwhelm you with changes (otherwise why are we doing feature branches?); see the TODO inside the test.

triceo: I get that, but I was wondering why are we testing this at all, if it's not on the list of things to have.

Christopher-Chianelli: Largely because it already existed before the doc and give me the location where I need to do the unassigning.

PR #2420, opened 2026-06-25
perf: avoid boxing in externalized list var cases

triceo:
6-7 % in CVRP profiling.


Christopher-Chianelli (approved):
Did we not check if `index == null` anywhere in this code base when accessing a `listVariableState`? Changes LGTM, but it seems a bit small.

PR #2425, opened 2026-06-26
perf: simplify how value ranges are built

triceo:


core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -35,7 +36,6 @@ final class ValueRangeState<Solution_, Entity_, Value_> {
 
     // Solution related fields
     private @Nullable ValueRangeItem<Solution_, Entity_, ValueRange<Value_>, Value_> fromSolutionItem;
-    private @Nullable Map<Value_, Integer> fromSolutionValueIndexMap;
```
triceo, line 38: Was only returned by a method that was never called...

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeState.java
```diff
@@ -103,14 +99,6 @@ public ValueRange<Value_> getFromSolution(Solution_ solution, @Nullable Selectio
         return null;
     }
 
-    private Map<Value_, Integer> getIndexMapFromSolution() {
```
triceo, line 106: ... this method.

zepfred (approved):

PR #2427, opened 2026-06-26
springboot2.7改造

wyyxaaaaa:


triceo:
Hello @wyyxaaaaa!
We will not be accepting this PR, as it does way more than it is necessary.
If you want your code change to be considered, please provide a minimalistic pull request with only the code changes that are necessary to achieve it.

PR #2429, opened 2026-06-27
Fail build when timefold.application.version is missing

### LINKED ISSUE #2393: Service: Should fail build when mandatory information is missing.
**Describe the bug**
I was able to build the service, without build error, despite not filling in the application properties (timefold.application.version=v1), which led to a vague error reported here: https://github.com/TimefoldAI/timefold-solver/issues/2391

**Expected behavior**
If required information is not available at build time, fail the build. This includes this property as well as potential other properties.

**Actual behavior**
Doesn't fail the build, leading into unexpected errors.

**To Reproduce**
See reproducer for https://github.com/TimefoldAI/timefold-solver/issues/2391

mvanhorn:
## Summary

Fixes #2393. The Timefold service currently builds successfully even when the mandatory `timefold.application.version` property is not provided. The resolved version silently falls through to a null/blank value at the `descriptor.setVersion(...)` call site in the Quarkus build-time processor, and the missing value only surfaces later as a vague runtime error.

This change adds a static `validateApplicationVersion(String)` check in `TimefoldModelDescriptorProcessor`, mirroring the existing `validateModelId(String)`. The resolved version is now validated inside `generateModelDescriptor` before it is set on the descriptor, so the build fails fast with a clear, actionable message naming the missing `timefold.application.version` property.

## Why this matters

It aligns with the project Constitution's "Fail Fast" and "Understandable Error Messages" principles, and matches the maintainer's note on the issue that validation should be improved so a missing mandatory property does not slip through to runtime. The scope is intentionally limited to the named `timefold.application.version` property rather than the open-ended set of other properties mentioned in passing on the issue.

## Testing

Added `ApplicationVersionValidationTest`, mirroring `ModelNameValidationTest` and using AssertJ assertions per the Constitution. It covers the happy path (non-blank version passes) and the null, empty, and blank/whitespace cases (each throws `IllegalArgumentException` whose message names `timefold.application.version`). Run locally:

```
./mvnw -pl service/quarkus/deployment -am -Dtest=ApplicationVersionValidationTest test
```

All 4 tests pass.

Fixes #2393

### DIFF
--- a/service/quarkus/deployment/src/main/java/ai/timefold/solver/service/quarkus/deployment/TimefoldModelDescriptorProcessor.java
+++ b/service/quarkus/deployment/src/main/java/ai/timefold/solver/service/quarkus/deployment/TimefoldModelDescriptorProcessor.java
@@ -511,6 +511,14 @@ class TimefoldModelDescriptorProcessor {
         }
     }
 
+    protected static void validateApplicationVersion(String version) {
+        if (version == null || version.trim().isEmpty()) {
+            throw new IllegalArgumentException(
+                    "The application version is missing. Set the '" + APPLICATION_VERSION_PROPERTY
+                            + "' property (e.g. in application.properties) so the build can proceed.");
+        }
+    }
+
     private void generateModelDescriptor(String modelId, String model, OpenAPI openAPI,
             Path outputDirectory,
             Optional<ClassInfo> restResource,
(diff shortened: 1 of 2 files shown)


rsynek (approved):
Thank you!

rsynek:
Thanks for the contribution. This PR is replaced by https://github.com/TimefoldAI/timefold-solver/pull/2500 (including your commit).

PR #2437, opened 2026-06-29
fix: write generated service build artifacts to target instead of sources

### LINKED ISSUE #2410: Service: Move generated files jib diagonotic-tools.jar and .flattened-pom.xml and enhanced-openapi.json to the target directory
If I remove those two files and rerun the build, they appear automatically.

Image

- They must not be checked in into Git.
- They should not live among the sources, but in be in the target directory

For both cases:

Proposal A) Put it under the target directory instead (recommended)

Proposal B) .gitignore it in all quickstarts, templates, models, etc

mvanhorn:
## Summary

Keep the service build's generated artifacts out of the source tree. The flatten-maven-plugin now writes `.flattened-pom.xml` under `target/` instead of each module basedir, and the generated `diagnostic-tools.jar` is git-ignored and removed at the start of every build so it never gets committed or shipped stale.

## Why this matters

Issue #2410 reports that `diagnostic-tools.jar`, `.flattened-pom.xml`, and the enhanced OpenAPI spec reappear among the sources on every build and risk being checked into Git across models and quickstarts.

- In `service/facade/service-parent/pom.xml` the `flatten-maven-plugin` config now sets `outputDirectory` to `${project.build.directory}` with `flattenedPomFilename` `.flattened-pom.xml`, so the flattened pom lands in `target/`.
- The `timefold-model-enhanced-openapi.json` is already written to the build output directory by `TimefoldModelDescriptorProcessor` (`out.getOutputDirectory()`), so no producer change was needed.
- The `copy-timefold-diagnostic-tools` execution still writes to `src/main/jib/timefold` because Quarkus `container-image-jib` only includes files from that fixed directory and it cannot be pointed at `target/`. To satisfy the "must not be checked into Git" requirement, that jar is git-ignored, and `maven-clean-plugin` deletes it on `clean` and at the `initialize` phase (before the copy runs) so a stale jar from a previous enterprise build can never be packaged by a later build that skips the copy.
- A defensive `.gitignore` entry backs up all three artifact patterns.

## Testing

`mvn validate` cannot run here because it requires the `999-SNAPSHOT` solver artifacts that are not published. Validated that both edited XML files are well-formed and that the flatten-maven-plugin (`outputDirectory`, `flattenedPomFilename`) and maven-clean-plugin (`filesets`, `excludeDefaultDirectories`) options used are the documented parameters for those plugins.

Fixes #2410

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ local
 
 # Repository wide ignore mac DS_Store files
 .DS_Store
+
+# Generated service build artifacts that must never be committed (issue #2410)
+**/src/main/jib/timefold/diagnostic-tools.jar
+**/.flattened-pom.xml
+**/timefold-model-enhanced-openapi.json
--- a/service/facade/service-parent/pom.xml
+++ b/service/facade/service-parent/pom.xml
@@ -385,6 +385,24 @@
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>flatten-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <!-- The diagnostic-tools jar is copied into src/main/jib (above) because Quarkus
+             container-image-jib only includes files from that fixed directory and it cannot be
+             relocated to target/. Remove it on clean so a stale jar from a previous enterprise build
+             is never shipped by a later non-enterprise build that skips the copy (issue #2410). -->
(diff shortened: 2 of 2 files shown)


rsynek:
Thank you!

I am closing this PR in favor of https://github.com/TimefoldAI/timefold-solver/pull/2495, with your commits included.

PR #2442, opened 2026-06-29
perf(constraint-streams): improve joins by removing list entries

triceo:
Turns tuples into list entries without creating an entry middleman.

This eliminates many `ElementAwareLinkedList` instances, leading to a significant GC pressure drop and resulting performance. +20 % was observed in machine reassignment.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractGroupNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractGroupNode.java
@@ -182,9 +182,12 @@ public abstract class AbstractGroupNode<InTuple_ extends Tuple, OutTuple_ extend
             updateGroup(tuple, oldGroup);
             return;
         }
-        var oldUserSuppliedGroupKey = extractUserSuppliedKey(oldGroup.getGroupKey());
         var newUserSuppliedGroupKey = groupKeyFunction.apply(tuple);
-        if (Objects.equals(oldUserSuppliedGroupKey, newUserSuppliedGroupKey)) {
+        var storedKey = oldGroup.getGroupKey();
+        var sameKey = useAssertingGroupKey
+                ? Objects.equals(extractUserSuppliedKey(storedKey), newUserSuppliedGroupKey)
+                : Objects.equals(storedKey, newUserSuppliedGroupKey);
+        if (sameKey) {
             updateGroup(tuple, oldGroup);
         } else {
             if (hasCollector) {
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
@@ -5,9 +5,9 @@ import ai.timefold.solver.core.impl.bavet.common.tuple.Tuple;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleLifecycle;
 import ai.timefold.solver.core.impl.bavet.common.tuple.TupleState;
 import ai.timefold.solver.core.impl.bavet.common.tuple.UniTuple;
-import ai.timefold.solver.core.impl.util.ElementAwareLinkedList;
 
 import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
 
 /**
  * This class has two direct children: {@link AbstractIndexedIfExistsNode} and {@link AbstractUnindexedIfExistsNode}.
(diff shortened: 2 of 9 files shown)


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractGroupNode.java
```diff
@@ -182,9 +182,12 @@ public final void update(InTuple_ tuple) {
             updateGroup(tuple, oldGroup);
             return;
         }
-        var oldUserSuppliedGroupKey = extractUserSuppliedKey(oldGroup.getGroupKey());
         var newUserSuppliedGroupKey = groupKeyFunction.apply(tuple);
-        if (Objects.equals(oldUserSuppliedGroupKey, newUserSuppliedGroupKey)) {
+        var storedKey = oldGroup.getGroupKey();
+        var sameKey = useAssertingGroupKey
```
triceo, line 187: Unrelated and not intended for this PR, but if it's already here, it doesn't hurt anyway.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -123,26 +126,104 @@ protected void decrementCounterRight(ExistsCounter<LeftTuple_> counter) {
...
     }
 
-    protected ElementAwareLinkedList<FilteringTracker<LeftTuple_>> clearRightTrackerList(UniTuple<Right_> rightTuple) {
-        ElementAwareLinkedList<FilteringTracker<LeftTuple_>> rightTrackerList =
-                rightTuple.getStore(inputStoreIndexRightTrackerList);
-        rightTrackerList.clear(tracker -> {
+    // Prepends tracker into the left tuple's hidden intrusive tracker list.
+    // The left tuple's store at inputStoreIndexLeftTrackerList holds the list head (null = empty).
+    private void linkLeft(FilteringTracker<LeftTuple_> tracker) {
+        var leftTuple = tracker.counter.leftTuple;
```
Christopher-Chianelli, line 132: I wonder if you can abstract the code somehow in `FilteringTracker` to remove the code duplication here between `left` and `right` tuples; perhaps with a view interface? Or put the combined `(next, prev)` for left/right into a new `Node` object?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -163,7 +244,9 @@ protected void updateCounterFromRight(ExistsCounter<LeftTuple_> counter, UniTupl
         }
         if (testFiltering(leftTuple, rightTuple)) {
             incrementCounterRight(counter);
-            new FilteringTracker<>(counter, leftTuple.getStore(inputStoreIndexLeftTrackerList), rightTrackerList);
+            var tracker = new FilteringTracker<>(counter, rightTuple);
```
Christopher-Chianelli, line 247: `rightTuple` is used for both `updateCounterFromLeft` and `updateCounterFromRight`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
Christopher-Chianelli: Maybe have a `link` method that takes `tracker, tracker.leftPrev, tracker.leftNext` or `tracker, tracker.rightPrev, tracker.rightNext` and have `linkLeft`/`linkRight` delegate to those (will avoid allocations and duplication at the cost of some extra stack frames).

Christopher-Chianelli (approved):

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
triceo: I don't see how to do that without some extra allocations (capturing lambdas at the very least), and avoiding allocations was the goal here. But maybe I'm just not seeing what you mean.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
triceo: Yes, the left tuple is in the counter and we need both.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
Christopher-Chianelli: Looking at it deeper, probably impossible unless you store fields as a mutable reference so you can change the value without needing to change the field in the code. Might be possible in Vahalla one day though with value types allowing us to create views without allocations.

PR #2443, opened 2026-06-29
chore: unassign if solution is inconsistent

Christopher-Chianelli:
Depends on https://github.com/TimefoldAI/timefold-solver/pull/2396

Unassigns any entities/values involved in dependency loops at solve start. Throws if any of those entities/values are pinned.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
@@ -6,6 +6,8 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.function.IntFunction;
 
+import ai.timefold.solver.core.impl.util.LinkedIdentityHashSet;
+
 import org.jspecify.annotations.NonNull;
 
 final class DefaultVariableReferenceGraph<Solution_> extends AbstractVariableReferenceGraph<Solution_, BitSet> {
@@ -70,5 +72,21 @@ final class DefaultVariableReferenceGraph<Solution_> extends AbstractVariableRef
     public void setUnknownInconsistencyValues() {
         graph.commitChanges(changeTracker);
         affectedEntitiesUpdater.setUnknownInconsistencyValues();
+    }
+
+    @Override
+    public List<Object> getInconsistentEntities() {
+        var out = new LinkedIdentityHashSet<>();
+        var graphTrackingInconsistentEntities = new DefaultTopologicalOrderGraph(this.nodeTopologicalOrders.length);
+        graph.forEachEdge(graphTrackingInconsistentEntities::addEdge);
+        graphTrackingInconsistentEntities.commitChanges(new BitSet());
+        var loopedComponentList = graphTrackingInconsistentEntities.getLoopedComponentList();
+        for (var loopedComponent : loopedComponentList) {
+            for (var nodeId : loopedComponent) {
+                var node = this.nodeList.get(nodeId);
+                out.add(node.entity());
+            }
+        }
+        return new ArrayList<>(out);
     }
 }
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
@@ -374,6 +374,14 @@ public final class VariableListenerSupport<Solution_> implements SupplyManager {
         return true;
     }
 
+    public List<Object> getInconsistentEntities() {
+        if (shadowVariableSession == null) {
+            throw new IllegalStateException(
+                    "The shadowVariableSession is null. A solution without shadow variables cannot be inconsistent.");
+        }
+        return shadowVariableSession.getInconsistentEntities();
+    }
+
     /**
      * Triggers all cascading update shadow variable user-logic.
      */
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
@@ -377,7 +377,7 @@ public final class VariableListenerSupport<Solution_> implements SupplyManager {
     public List<Object> getInconsistentEntities() {
         if (shadowVariableSession == null) {
             throw new IllegalStateException(
-                    "The shadowVariableSession is null. A solution without shadow variables cannot be inconsistent.");
+                    "Impossible state: The shadowVariableSession is null. A solution without shadow variables cannot be inconsistent.");
         }
         return shadowVariableSession.getInconsistentEntities();
     }
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
@@ -344,6 +344,45 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
         // Do nothing
     }
 
+    public void unassignInconsistentEntities() {
+        var inconsistentEntities = variableListenerSupport.getInconsistentEntities();
+        if (listVariableStateSupply != null) {
+            var listVariableDescriptor = listVariableStateSupply.getSourceVariableDescriptor();
+            var listElementClass = listVariableStateSupply.getSourceVariableDescriptor().getElementType();
+            for (var inconsistentEntity : inconsistentEntities) {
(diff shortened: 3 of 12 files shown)


triceo:
Please rebase.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
```diff
@@ -374,6 +374,14 @@ public boolean triggerVariableListenersInNotificationQueues() {
         return true;
     }
 
+    public List<Object> getInconsistentEntities() {
+        if (shadowVariableSession == null) {
+            throw new IllegalStateException(
```
triceo, line 379: I wonder if this should just return empty instead.
Isn't failing here a bit too harsh?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +344,56 @@ protected void afterSetWorkingSolution() {
...
+            var listVariableDescriptor = listVariableStateSupply.getSourceVariableDescriptor();
+            var listElementClass = listVariableStateSupply.getSourceVariableDescriptor().getElementType();
+            for (var inconsistentEntity : inconsistentEntities) {
+                if (listElementClass.isInstance(inconsistentEntity)) {
+                    var inverse = Objects.requireNonNull(listVariableStateSupply.getInverseSingleton(inconsistentEntity));
+                    int index = Objects.requireNonNull(listVariableStateSupply.getIndex(inconsistentEntity));
+
+                    if (listVariableDescriptor.isElementPinned(Objects.requireNonNull(workingSolution), inverse, index)) {
+                        throw new IllegalStateException(
+                                "Entity (%s) is pinned but is involved in a dependency loop".formatted(inconsistentEntity));
```
triceo:
```suggestion
                                "Entity (%s) is pinned while involved in a dependency loop. This creates an unresolvable inconsistency.".formatted(inconsistentEntity));
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +344,56 @@ protected void afterSetWorkingSolution() {
...
+
+                    beforeListVariableElementUnassigned(listVariableDescriptor, inconsistentEntity);
+                    beforeListVariableChanged(listVariableDescriptor, inverse, index, index + 1);
+                    listVariableDescriptor.removeElement(inverse, index);
+                    afterListVariableChanged(listVariableDescriptor, inverse, index, index);
+                    afterListVariableElementUnassigned(listVariableDescriptor, inconsistentEntity);
+                    triggerVariableListeners();
+                }
+                // Unassign any normal @PlanningVariable on the entity too
+                unassignPlainEntity(inconsistentEntity);
```
triceo, line 371: I would not do this. IMO we should only do the bare minimum necessary updates.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +344,56 @@ protected void afterSetWorkingSolution() {
...
+            for (var inconsistentEntity : inconsistentEntities) {
+                unassignPlainEntity(inconsistentEntity);
+            }
+        }
+    }
+
+    private void unassignPlainEntity(Object inconsistentEntity) {
+        var entityDescriptor = getSolutionDescriptor().findEntityDescriptor(inconsistentEntity.getClass());
+        if (entityDescriptor == null) {
+            throw new IllegalStateException("Object (%s) is not an entity but is inconsistent".formatted(inconsistentEntity));
```
triceo: How would this happen?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +344,56 @@ protected void afterSetWorkingSolution() {
...
+        }
+        if (entityDescriptor.isGenuine()
+                && !entityDescriptor.isMovable(Objects.requireNonNull(workingSolution), inconsistentEntity)) {
+            throw new IllegalStateException(
+                    "Entity (%s) is pinned but is involved in a dependency loop".formatted(inconsistentEntity));
```
triceo: Same message suggestion as above.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +344,56 @@ protected void afterSetWorkingSolution() {
...
+        }
+        for (var genuineVariableDescriptor : entityDescriptor.getGenuineVariableDescriptorList()) {
```
triceo, line 392: Can we only unassign the vars that actually cause the inconsistency?
We should go for bare minimum changes.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
```diff
@@ -71,4 +73,20 @@ public void setUnknownInconsistencyValues() {
         graph.commitChanges(changeTracker);
         affectedEntitiesUpdater.setUnknownInconsistencyValues();
     }
+
+    @Override
+    public List<Object> getInconsistentEntities() {
```
triceo, line 78: I wonder if the signature of this method should change.
You are building quite a large collection only to then iterate it later.
How about making this a `forEach` pattern instead? 
The collection will not have to be created.

triceo (commented):
IMO we can improve here.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
Christopher-Chianelli: Well, it is an impossible state for an internal component; if you are trying to get inconsistent entities on a problem that by definition cannot have inconsistent entities, you are probably doing something wrong. Although returning an empty list does make sense for the method if it were to ever be exposed to a user, it does not alert you to the fact it is impossible to have inconsistent entities and maybe you used the wrong testdomain. (Moreover, this method is pretty expensive since it needs to rebuild the graph). I am not too fixated on this; can go either way.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: It shouldn't happen; it catches an impossible state to catch bugs within our code.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: We do not know what variables are causing the inconsistency; it could be either from list variables or from a plain variable (i.e. if someone were to reinvent chained). To ensure the resulting solution is consistent, all genuine variables of an entity involved in a dependency loop must be unassigned.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: Not really; from the edges we can only infer the shadows involved, and there no way to go from shadows -> geninue variable descriptors for declarative shadow variables. Adding that infrastructure would be maintenance heavy and error-prone should we do any changes like adding a new builtin shadow variable type. The user submitted an invalid problem that doesn't make sense; it a miracle we attempt to solve it at all.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
triceo: In those cases, we use the "Impossible state: " prefix.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/listener/support/VariableListenerSupport.java
triceo: In that case, the "Impossible state: " prefix solves this.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
Christopher-Chianelli: A collection will still need to be created to collect the unique entities (since the same entity can have many nodes in the graph). Moreover, this is used to create the custom exception that tell the user what entities are inconsistent inside the next PR in this series.

PR #2444, opened 2026-06-29
chore: make score analysis/recommendations/update fail fast on inconsistent solution

Christopher-Chianelli:
Depends on https://github.com/TimefoldAI/timefold-solver/pull/2443

Fail fast on inconsistent solution for `SolutionManager`'s `update` and `analyze` methods.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
@@ -343,8 +343,12 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
         // Do nothing
     }
 
+    public List<Object> getInconsistentEntities() {
+        return shadowVariableSupport.getInconsistentEntities();
+    }
+
     public void unassignInconsistentEntities() {
-        var inconsistentEntities = shadowVariableSupport.getInconsistentEntities();
+        var inconsistentEntities = getInconsistentEntities();
         if (listVariableStateSupply != null) {
             var listVariableDescriptor = listVariableStateSupply.getSourceVariableDescriptor();
             var listElementClass = listVariableStateSupply.getSourceVariableDescriptor().getElementType();
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
@@ -489,6 +489,10 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
         lastVariableUpdateWasSuccessful = shadowVariableSupport.updateShadowVariables();
     }
 
+    public boolean isLastVariableUpdateWasSuccessful() {
+        return lastVariableUpdateWasSuccessful;
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,8 +344,12 @@ protected void afterSetWorkingSolution() {
         // Do nothing
     }
 
+    public List<Object> getInconsistentEntities() {
```
triceo: I think a more fitting name would be `computeInconsistentEntities` to make it clear that this is expensive.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -490,6 +494,10 @@ public void triggerVariableListeners() {
         lastVariableUpdateWasSuccessful = variableListenerSupport.triggerVariableListenersInNotificationQueues();
     }
 
+    public boolean isLastVariableUpdateWasSuccessful() {
```
triceo: IMO works equally as well, even better, without the "was" there. `lastVariableUpdateSuccessful` perhaps best.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolutionManager.java
```diff
@@ -51,14 +52,15 @@ public ScoreDirectorFactory<Solution_, Score_> getScoreDirectorFactory() {
     @Override
     public Score_ update(Solution_ solution, SolutionUpdatePolicy solutionUpdatePolicy) {
         if (solutionUpdatePolicy == SolutionUpdatePolicy.NO_UPDATE) {
-            throw new IllegalArgumentException("Can not call " + this.getClass().getSimpleName()
-                    + ".update() with this solutionUpdatePolicy (" + solutionUpdatePolicy + ").");
+            throw new IllegalArgumentException(
+                    "Can not call %s.update() with this solutionUpdatePolicy (%s)."
```
triceo: I know this is not actually new, but maybe we add a hint as to why they cannot call this? Like this, the exception message is not very helpful.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolutionManager.java
```diff
@@ -112,7 +121,7 @@ public ScoreAnalysis<Score_> analyze(Solution_ solution, ScoreAnalysisFetchPolic
         var enterpriseService =
                 TimefoldSolverEnterpriseService.loadOrFail(TimefoldSolverEnterpriseService.Feature.SCORE_ANALYSIS);
         var currentScore = (Score_) scoreDirectorFactory.getSolutionDescriptor().getScore(solution);
-        var analysis = callScoreDirector(solution, solutionUpdatePolicy,
+        var analysis = callScoreDirector("Solution analysis", solution, solutionUpdatePolicy,
```
triceo: I think we call it "Score analysis", don't we?

triceo (approved):
Overall LGTM, with some comments.
We should also clearly document the cases in which inconsistency is going to result in runtime failures, but that may wait for a later PR.

triceo (approved):

PR #2445, opened 2026-06-29
chore: fail fast on custom phase if it returns an inconsistent solution

Christopher-Chianelli:
Depends on #2444


triceo (approved):
We'll refactor the invalid score detection, but not in this PR.

PR #2447, opened 2026-06-30
chore: identify RC builds by their Git ref

triceo:
@rsynek's review will suffice.


core/src/main/java/ai/timefold/solver/core/enterprise/TimefoldSolverEnterpriseService.java
```diff
@@ -64,26 +65,56 @@ final class InstanceCarrier {
...
+        TimefoldSolverEnterpriseService service = null;
+        try {
+            service = load();
+        } catch (Exception e) {
+            // No enterprise edition on the classpath.
+        }
+        if (service == null) {
+            return SolverVersionUtils.communityBannerWithGitRef(SolverFactory.class, coreRef);
+        } else {
+            return SolverVersionUtils.enterpriseBannerWithGitRef(SolverFactory.class, coreRef, service.getGitRef());
```
rsynek: Why does only the enterprise provide the git ref?

triceo: If enterprise is detected, it shows an enterprise banner with both refs.
If enterprise is not detected, it shows a community banner with community ref.
I don't understand the question.

rsynek: Ok, so the `coreRef` is defined already here and the enterprise git ref comes from the enterprise service.

rsynek (approved):

PR #2452, opened 2026-07-01
feat: add nullable to model config params

diogodanielsoaresferreira:
Fixes: https://github.com/TimefoldAI/timefold-platform/issues/4992


cristianonicolai (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelConfigParameter.java
```diff
@@ -18,6 +18,7 @@
  * @param arrayItemType the OpenAPI schema type for array items
  * @param format the data format which further specifies the schema type
  * @param schemaTypeRef the OpenAPI schema type reference (used when the schema type is not a primitive type)
+ * @param nullable whether the parameter accepts {@code null} as a value, as defined by the OpenAPI schema
```
mswiderski, line 21: `nullable` has been removed from OpenAPI in 3.1 so not sure we should rely on it https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

mswiderski: just noticed that platform uses 3.1 and models use still 3.0.3 ... might be worth to address this at some point, but we might be ok to keep it as is for now. @rsynek FYI

mswiderski (approved):
approving as this is valid for the OpenAPI version used by models

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelConfigParameter.java
diogodanielsoaresferreira: Thanks, I didn't know that! I took a look at the migration guide and it seems that there are other things that will also have to be changed when we migrate to 3.1. I would say to implement now as is and when we migrate, we can change everything (type arrays, examples).

rsynek (approved):

rsynek:
LGTM; let's wait for the CI to finish, as there were some failures related to the changes.

rsynek:
ai.timefold.solver.service.quarkus.deployment.ModelConfigDescriptorSchemaTest.detectModelConfigOverrides keeps failing. @diogodanielsoaresferreira could you please have a look?

diogodanielsoaresferreira:
> ai.timefold.solver.service.quarkus.deployment.ModelConfigDescriptorSchemaTest.detectModelConfigOverrides keeps failing. @diogodanielsoaresferreira could you please have a look?

@rsynek should be fixed, the current error in ESS seems unrelated.

rsynek (approved):

PR #2454, opened 2026-07-02
perf: improve comparison indexer

triceo:
Introduces an alternative to `TreeMap` which, for a small amount of keys, uses a sorted array. This is a significant performance improvement (~10 % in score director benchmarks), if the number of keys stays low.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
@@ -1,5 +1,6 @@
 package ai.timefold.solver.core.impl.bavet.common.index;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -20,21 +17,44 @@
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
+/**
```
zepfred, line 20: The class clearly distinguishes between comparisons using trees or arrays. My initial observation is to ask why we shouldn't make `ComparisonIndexer` an abstract class and provide two different implementations.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -221,22 +377,38 @@ public String toString() {
         return "size = " + comparisonMap.size();
     }
 
+    /**
```
zepfred, line 381: The same general comment about having different implementations for each data structure applies here too.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
triceo: Because the choice (and restructure) will be made as the indexer grows. It makes no sense to have to switch the indexer based on what happens externally - it is an internal implementation detail of the indexer to figure out the most efficient was of using its data. This is how you hide complexity; complexity should never escape further than it absolutely has to.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ScalingNavigableMap.java
```diff
@@ -0,0 +1,220 @@
...
+ * <p>
+ * {@link ComparisonIndexer} branches on {@link #arrayBased} and calls {@link #keyAt}/{@link #valueAt} for range scans
+ * (plain, final-class, trivially inlined methods; benchmarked to be optimal);
+ * only the get/put/remove/treeify machinery is actually encapsulated here.
+ *
+ * @param <K> the key type
+ * @param <V> the value type
+ */
+@NullMarked
+final class ScalingNavigableMap<K extends Comparable<K>, V> {
```
zepfred, line 51: Switching between data structures based on the size of the elements seems a good idea. The code would benefit from implementing a strategy pattern. This approach would provide clear and consistent strategies for each data structure, allowing us to switch between them accordingly.

triceo: This was implemented at one point, and performed horribly.
The JDK can handle bimorphic sites, but it can do so efficiently if only one of the 2 types is there. What you are proposing here is that there would be either of those types, and they would switch at runtime based on dataset size.
This effectively eliminates several of JVM's optimizations, and with it, all improvements provided by this PR.

I have benchmarked this for 2 days. The choices here have data behind them.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ScalingNavigableMap.java
```diff
@@ -0,0 +1,220 @@
...
+        return value;
+    }
+
+    private V getOrCreateArray(K key, Supplier<V> valueSupplier) {
+        var index = indexOf(key);
+        if (index >= 0) {
+            return valueAt(index);
+        }
+        var value = valueSupplier.get();
+        insertIntoArray(-(index + 1), key, value);
```
zepfred: We can avoid inserting data into the array by checking the size+1 first.
```
if (size() +1 > ARRAY_THRESHOLD) {
    treeify();
    return getOrCreateTree(key, valueSupplier);
}
```

zepfred (approved):
LGTM!

PR #2461, opened 2026-07-03
perf: improve filtered joins

triceo:
This PR does two things:

- Improves joins in the filtered part on unchanged key (= update). The time complexity goes down from `O(bucketSize x fanOut)` to `O(bucketSize + fanOut)`. One example improved by 10 %, none worsened.
- Fixes a long-standing bug on unindexed joins which was uncovered during this work.


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
```diff
@@ -206,20 +223,18 @@ private void processOutTupleUpdate(LeftTuple_ leftTuple, UniTuple<Right_> rightT
         }
     }
 
-    private static <Tuple_ extends Tuple> Tuple_ findOutTuple(TupleList<Tuple_> sourceList,
```
triceo, line 209: This code was effectively doing `M*N`. Now the code does `M+N`.

core/src/test/java/ai/timefold/solver/core/impl/bavet/bi/UnindexedJoinBiNodeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        verify(downstream).insert(argThat(t -> t.getA().equals("L1") && t.getB().equals("R1")));
+
+        left.setA("L2");
+        node.updateLeft(left);
+        node.getPropagator().propagateEverything();
+        verify(downstream).update(argThat(t -> t.getA().equals("L2") && t.getB().equals("R1")));
+    }
+
+    @Test
+    void updateRight_refreshesRightFact() {
```
triceo, line 76: This test would fail without these changes.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/TupleList.java
```diff
@@ -98,4 +100,28 @@ public int size() {
...
+        this.markVersion = version;
+    }
+
+    /**
+     * Returns the tuple stamped by {@link #mark(Tuple, long)} with exactly this {@code version},
+     * or {@code null} on a version mismatch, which means no mark.
+     * A stale mark may reference a tuple no longer in this list —
+     * callers must never read a mark with an old version.
+     */
+    public @Nullable T getMark(long version) {
```
zepfred, line 118: Should we remove the mark when we call the `clear(...)` method?

zepfred: Could the call to `remove(...)` possibly result in the removal of the marked tuple? If so, should we also clear it?

core/src/test/java/ai/timefold/solver/core/impl/bavet/bi/UnindexedJoinBiNodeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        var right = createInputTuple("R1");
+        node.insertLeft(left);
+        node.insertRight(right);
+        node.getPropagator().propagateEverything();
+        verify(downstream).insert(argThat(t -> t.getA().equals("L1") && t.getB().equals("R1")));
+
+        right.setA("R2");
+        node.updateRight(right);
+        node.getPropagator().propagateEverything();
+        verify(downstream).update(argThat(t -> t.getB().equals("R2") && t.getA().equals("L1")));
```
zepfred:
```suggestion
        verify(downstream).update(argThat(t -> t.getA().equals("L1") && t.getB().equals("R2")));
```

zepfred (approved):

PR #2473, opened 2026-07-07
perf: improve performance of ConnectedRange collector

triceo:
This contains two separate commits:

- First, it significantly improves the performance of the ConnectedRange collector based on the latest ESS dataset. This shows 50 % of CPU time was spent in range's hash code computations, and also a lot of memory was wasted allocating arrays.
- Second, a move iterator bug was discovered during this session that should lead to faster solving overall.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeImpl.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeImpl.java
@@ -21,6 +21,8 @@ final class ConnectedRangeImpl<Range_, Point_ extends Comparable<Point_>, Differ
     private int minimumOverlap;
     private int maximumOverlap;
     private boolean hasOverlap;
+    private int cachedHashCode;
+    private boolean hashCodeDirty = true;
 
     ConnectedRangeImpl(NavigableSet<RangeSplitPoint<Range_, Point_>> splitPointSet,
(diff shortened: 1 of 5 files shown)


triceo:
Adding @Christopher-Chianelli to review the range work.
Also CC @LDWNS who can check how faster this is in practice. I wouldn't be surprised if it was 2x or more.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/collector/connected_ranges/ConnectedRangeImpl.java
```diff
@@ -153,22 +157,41 @@ public int getMaximumOverlap() {
...
+        // (Objects.equals short-circuits on ==),
+        // whereas hashCode() has no such shortcut
+        // and would have to walk and hash every split point in the whole problem,
+        // not just this range's.
+        // The checks are ordered from cheapest to most expensive.
+        return o instanceof ConnectedRangeImpl<?, ?, ?> that
+                && hasOverlap == that.hasOverlap
+                && count == that.count
+                && getMinimumOverlap() == that.getMinimumOverlap()
                 && getMaximumOverlap() == that.getMaximumOverlap()
```
Christopher-Chianelli: Do we need `minimum` and `maximum` overlap in `hashCode` and `equals`? They are derived from `splitPointSet`, `startSplitPoint` and `endSplitPoint`, so if those three are equal, so would minimum and maximum overlap.

Christopher-Chianelli (approved):
LGTM; can probably simplify the `hashCode`/`equals` even more.

PR #2494, opened 2026-07-14
fix: make sure test outputs for models are visible in the console

rsynek:
Test outputs were hidden for models; likely overlooked during merging build parents pom.xml


winklerm (approved):
Thanks!

PR #2495, opened 2026-07-14
fix: clean up the solver service parent build

### LINKED ISSUE #2410: Service: Move generated files jib diagonotic-tools.jar and .flattened-pom.xml and enhanced-openapi.json to the target directory

rsynek:
Simplifies and replaces https://github.com/TimefoldAI/timefold-solver/pull/2437.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ local
 
 # Repository wide ignore mac DS_Store files
 .DS_Store
+
+# Generated service build artifacts that must never be committed (issue #2410)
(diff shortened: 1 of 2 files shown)


triceo (approved):
LGTM when comments resolved.

PR #2497, opened 2026-07-15
feat: Add validated summary data to dataset validated event

diogodanielsoaresferreira:


cristianonicolai (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/api/validation/dto/ValidationSummary.java
```diff
@@ -0,0 +1,19 @@
+package ai.timefold.solver.service.definition.api.validation.dto;
```
rsynek: Please move the record to the internal package. This should not be introduced as a public API.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -835,4 +842,23 @@ private ModelOutputMetrics extractOutputMetrics(SolverModel solverModel) {
...
+        Map<String, Integer> errorCountsByCode = new HashMap<>();
+        Map<String, Integer> warningCountsByCode = new HashMap<>();
+        for (Issue issue : validationResponse.issues()) {
+            var countsByCode = issue.getSeverity() == IssueSeverity.ERROR ? errorCountsByCode : warningCountsByCode;
+            countsByCode.merge(issue.getCode().value(), 1, Integer::sum);
+        }
+        int errorCount = errorCountsByCode.values().stream().mapToInt(Integer::intValue).sum();
+        int warningCount = warningCountsByCode.values().stream().mapToInt(Integer::intValue).sum();
+        return new ValidationSummary(status, errorCount, warningCount, errorCountsByCode, warningCountsByCode);
+    }
```
rsynek: This method can become a factory method in the `ValidationSummary`:

- `SolverWorker` does not handle the conversion (that is good, it should only coordinate the workflow)
- the code becomes easily testable (see the coverage report)
- The summary should be ever only created from the validation results

rsynek (changes requested):
Thanks for adding the more structured data!

service/definition/src/main/java/ai/timefold/solver/service/definition/api/validation/dto/ValidationSummary.java
diogodanielsoaresferreira: Moved to the internal package.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
diogodanielsoaresferreira: Good idea, makes sense. Moved to the ValidationSummary

diogodanielsoaresferreira:
Replaced by https://github.com/TimefoldAI/timefold-solver/pull/2502

PR #2514, opened 2026-07-16
fix: create copy DecimalFormat

zepfred:
This PR addresses a [comment](https://github.com/TimefoldAI/timefold-solver/pull/2512#discussion_r3597917015) regarding a concurrent access issue and fixes the class's mutability.


triceo (changes requested):
First of all - this does not solve the problem.
Second - not sure this is even something we need to solve; this thing exists for testing. All other code will already return from the first branch.

IMO not an issue.

zepfred:
Unnecessary fix.

PR #2525, opened 2026-07-20
refactor: remove variable listeners

triceo:


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -544,49 +542,33 @@ private void afterAnnotationsProcessed(DescriptorPolicy descriptorPolicy) {
     }
 
     private void determineGlobalShadowOrder() {
-        // Topological sorting with Kahn's algorithm
-        var pairList = new ArrayList<MutablePair<ShadowVariableDescriptor<Solution_>, Integer>>();
-        var shadowToPairMap =
-                new HashMap<ShadowVariableDescriptor<Solution_>, MutablePair<ShadowVariableDescriptor<Solution_>, Integer>>();
+        // Every shadow variable has at most 1 source, so this is a forest: walk each one up to its root and number
```
Christopher-Chianelli: Does shadow variable order even matter now? Their sources must be a geninue variable now, considering the two that remain are inverse and list variable state.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/violation/BasicVariableTracker.java
```diff
@@ -14,16 +13,16 @@
 import org.jspecify.annotations.NonNull;
 
 /**
- * Tracks variable listener events for a given genuine or shadow variable
+ * Tracks shadow variable update events for a given genuine or shadow variable
```
Christopher-Chianelli: It tracks variable change events, not specific to shadow variables. It basically ensures if variable `a` has changed for entity `e` in the solution, there was a `beforeVariableChanged(e, "a")` and `afterVariableChanged(e, "a")`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/BasicVariableChangeHandler.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Flattened, event-free dispatch shape for shadow variable updates that need to be notified
+ * eagerly of basic variable changes, without allocating a change event for every notification.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
+ */
+@NullMarked
+public interface BasicVariableChangeHandler<Solution_> {
```
Christopher-Chianelli, line 16: ...Isn't this `VariableListener` but by a different name?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
```diff
@@ -53,117 +44,129 @@
...
+        } else if (listVariableStateSupply != null
+                && (descriptor instanceof IndexShadowVariableDescriptor<Solution_>
+                        || descriptor instanceof PreviousElementShadowVariableDescriptor<Solution_>
+                        || descriptor instanceof NextElementShadowVariableDescriptor<Solution_>)) {
+            // When multiple variable types are used,
+            // the shadow variable process needs to account for each variable
+            // and process them according to their types.
+            processShadowVariableDescriptorWithListVariable(descriptor, listVariableStateSupply);
+        }
+    }
```
Christopher-Chianelli, line 163: I would add early returns to the `if` and throw an `exception` if it reaches the end of the method without an early return (so we remember to update this when we add a new builtin shadow variable type)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
```diff
@@ -53,117 +44,129 @@
...
+
+    private @Nullable ListVariableStateSupply<Solution_, Object, Object> getListVariableStateSupply() {
+        if (listVariableChangeHandlerList.isEmpty()) {
+            return null;
+        }
+        return (ListVariableStateSupply<Solution_, Object, Object>) listVariableChangeHandlerList.getFirst();
```
Christopher-Chianelli, line 169: This will probably break if we ever add a 2nd list variable listener; maybe change `listVariableChangeHandlerList` to `listVariableChangeHandler` (that way, we are forced to change this code if we ever add a 2nd one).

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
```diff
@@ -191,19 +194,40 @@ public <Supply_ extends Supply> Supply_ demand(Demand<Supply_> demand) {
...
-            // An external ScoreDirector can be created before the working solution is set
-            if (scoreDirector.getWorkingSolution() != null) {
-                variableListener.resetWorkingSolution(scoreDirector);
-            }
-            notifiableRegistry.registerNotifiable(
-                    variableListener.getSourceVariableDescriptor(),
-                    AbstractNotifiable.buildNotifiable(scoreDirector, variableListener, nextGlobalOrder++));
+        if (supply instanceof BasicVariableChangeHandler<?> handler) {
+            var basicVariableChangeHandler = (BasicVariableChangeHandler<Solution_>) handler;
+            resetWorkingSolutionIfSet(() -> basicVariableChangeHandler.resetWorkingSolution(scoreDirector));
```
Christopher-Chianelli, line 217: Why is this required? Isn't `resetWorkingSolution` is called for the initial solution?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
```diff
@@ -106,7 +106,7 @@ protected void execute(VariableDescriptorAwareScoreDirector<Solution_> scoreDire
                         listVariableDescriptor.getFirstUnpinnedIndex(entity),
                         listVariableDescriptor.getListSize(entity));
             }
-            scoreDirector.triggerVariableListeners();
```
Christopher-Chianelli, line 109: How isolated is `scoreDirector`? I fear this rename will break some users that still use legacy custom moves.

core/src/main/java/ai/timefold/solver/core/impl/move/RecordedUndoMove.java
```diff
@@ -21,8 +21,12 @@ record RecordedUndoMove<Solution_>(List<ChangeAction<Solution_>> variableChangeA
     @Override
     public void execute(MutableSolutionView<Solution_> solutionView) {
         var scoreDirector = ((InnerMutableSolutionView<Solution_>) solutionView).getScoreDirector();
-        for (var changeAction : variableChangeActionList) {
-            changeAction.undo(scoreDirector);
+        // Undo actions must be replayed in reverse recording order,
+        // otherwise repeated changes to the same variable restore a stale value
+        // and multi-action list windows replay out of order.
+        var listIterator = variableChangeActionList.listIterator(variableChangeActionList.size());
```
Christopher-Chianelli, line 27: IIRC, `variableChangeActionList` is already in reversed order?

Christopher-Chianelli: It isn't, but I swear we did reverse iteration at one point for this.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/ruin/SelectorBasedListRuinRecreateMove.java
Christopher-Chianelli: I see that it was added as an alternative (i.e. `triggerVariableListener` still exists and is deprecated); nevermind then.

pom.xml
```diff
@@ -68,6 +68,7 @@
     <module>quarkus-integration</module>
     <module>tools</module>
     <module>service</module>
+    <module>docs</module>
```
Christopher-Chianelli: ...when was this removed from the pom!?!

Christopher-Chianelli (approved):
LGTM; although I personally would call it a simplification of shadow variables. Should we do the backport of the fix to UndoRecordedMove in this PR to 1.x?

core/src/main/java/ai/timefold/solver/core/impl/move/RecordedUndoMove.java
triceo: Yeah, I was surprised by this too. Not exactly sure why it worked before.

pom.xml
triceo: I wish I knew. :-(

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/BasicVariableChangeHandler.java
triceo: You could say that. I wanted to move away from that concept entirely, so that we are not limited by the memories of what that API used to be every time we see the name.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
triceo: The entire codebase assumes there is only one list variable descriptor, incl. in this class - if we ever change that, a massive refactor will have to happen.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
triceo: Code that is calling this method makes sure this is only called when a new supply is guaranteed. Therefore the reset is part of the contract, indicating that the thing being reset can now perform any initialization that it wishes to do.

PR #2528, opened 2026-07-23
chore: add indictment support to Bavet's Node Network

Christopher-Chianelli:
Each tuple now have an `IndictmentSource` field. This field is set to `IndictmentSource.DISABLED` when indictments are disabled, which is used to exit early and avoid work only required for indictments. `ConstraintMatchPolicy` have two new values:

- `ENABLED_WITHOUT_JUSTIFICATIONS_AND_INDICTMENTS` this has the previous behaviour of `ENABLED_WITHOUT_JUSTIFICATIONS`; has constraint matches but no justifications nor indictments
- `ENABLED_WITHOUT_INDICTMENTS` this includes justifications but not indictments

`ENABLED` was changed to include both justifications and indictments instead of just justifications. `ENABLED_WITHOUT_JUSTIFICATIONS` was changed to include indictments.

The actual work each node does is pretty simple. Do note that the attach IndictmentSource does not change after an insert:

- `forEach`: if indictments are enabled, set the source of the tuple to a `RootIndictmentSource` (perhaps `LeafIndictmentSource` would be a better name) on insert. Retract and update do no extra work; `retract` removes the tuple (and hence the indictment source and potential constraint match), and `update` does not change the instance tracked.
- `filter`: not a node, and does not affect what object is indicted.
- `map`/`expand`/flatten`: these creates new tuples from a source tuple, and simply copy the source's indictment source on insert. Update and retract do not need to do anything following the same reasoning as `forEach`.
- `join`: on `insert`, set the indictment source of the joined tuple to a `JoinedIndictmentSource` that have a reference to both the left and right tuples' indictment sources.
- `groupBy`: on `insert`, set the indictment source of the group's tuple to an aggregation containing the indictment sources of the tuples inside the group. On `update`, remove the tuple's indictment source from the old group aggregation and add it to the new group aggregation. On `retract`, remove the tuple's indictment source from the aggregation.
- `distinct`: this is `groupBy` in a trenchcoat and behaves similarly.
- `concat`: behaves similarly to `map`/`expand`/`flatten`, since the left side and right side act independently and don't affect each other.
- `ifExists`/`ifNotExists`: The most complex of the bunch. When indictments are enabled, it takes a drastically different path than normal. In particular, right tuple will propagate updates even if the corresponding left tuple is not update. This is to prevent constraint matches with stale indictments referring to removed/retracted entities/problem facts. Additional tuple iteration occurs for indexed `ifExists` nodes so the tuple's indictments are up to date when the left tuple changes. Moreover, since `ifExists` does not create a new tuple but still adds to the indictment, the same indictment source is used, but the indictment source's support map for the key corresponding to the `ifExists` node's id is updated.
- `precompute`: Since it uses an independent node network, its node's ids do not correspond to the outer node network's node ids. As such, the indictment support map of the produced tuples must be aggregated along with the original indictment source.

Each constraint is now aware of the ids of the nodes that affect the constraint. It uses it when creating constraint matches to iterate through the support map so it does not indict objects from other constraints with `ifExists` that so happen to share the same tuple. Each `IndictmentSource` is iterated in a tree like matter to collect a set of unique indicted objects which is converted to a list inside the actual ConstraintMatch.

The vast majority of the changes were making all existing ConstraintStream tests aware of indictments. `assertMatch` was changed so it sets the indictment list to the justifications, which work for some but not all tests. For the other tests, `withIndictedObjects` is used to set the expected indicted objects for each match.

Note: I am aware of the stale "NOTE: By not propagating here, ..." comment in `IfExistsNode` and will remove it.
Note: I am aware that the code under `if (testFiltering(...))` in `IfExistsNode` should do a branch depending on if indictments are enabled and call the right method (was written before the change to keep the indictments up to date and the two methods were not too different from each other).
Note: I am aware the `carnality` and `storeSize` methods added to `Tuple` can be removed; they are an artifact from when tuple cloning were used for `ifExists`'s indictment enabled path.


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenNode.java
```diff
@@ -73,6 +73,7 @@ private void addTuple(InTuple_ originalTuple, FlattenedItem_ item,
         var reuse = bag.reuseOrAdvance();
         if (reuse == null) {
             var created = createTuple(originalTuple, bag.value);
+            created.setIndictmentSource(originalTuple.getIndictmentSource());
```
triceo, line 76: Cases like this would benefit from inheriting the indictment source at construction time; the mutability aspect will probably be necessary in some places (haven't yet read the entire thing), but I think the default pattern here should not happen post-construction.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/TupleRecorder.java
```diff
@@ -13,6 +16,18 @@ public record TupleRecorder<Tuple_ extends Tuple>(List<Tuple_> recordedTupleList
         UnaryOperator<Tuple_> mapper,
         IdentityHashMap<Tuple_, Tuple_> inputTupleToOutputTuple) {
     public void recordTuple(Tuple_ tuple) {
-        recordedTupleList.add(inputTupleToOutputTuple.computeIfAbsent(tuple, mapper));
+        var outTuple = inputTupleToOutputTuple.computeIfAbsent(tuple, mapper);
+        if (tuple.getIndictmentSource() != IndictmentSource.DISABLED) {
+            if (outTuple.getIndictmentSource() == IndictmentSource.DISABLED) {
+                outTuple.setIndictmentSource(
+                        new IndictmentSource.AggregateIndictmentSource(new ArrayList<>(), new LinkedHashMap<>()));
```
triceo: This is inconsistent with the otherwise seemingly standard pattern of having a named static factory for these operations.

core/src/main/java/ai/timefold/solver/core/impl/score/constraint/ConstraintMatch.java
```diff
@@ -28,16 +30,19 @@ public final class ConstraintMatch<Score_ extends Score<Score_>> implements Comp
...
     private final Score_ score;
 
     /**
      * @param constraintRef unique identifier of the constraint
      * @param justification only null if justifications are disabled
      * @param score penalty or reward associated with the constraint match
      */
-    public ConstraintMatch(ConstraintRef constraintRef, @Nullable ConstraintJustification justification, Score_ score) {
+    public ConstraintMatch(ConstraintRef constraintRef, @Nullable ConstraintJustification justification,
+            @Nullable List<Object> indictedObjects, Score_ score) {
```
triceo: I wonder if - instead of dealing with null here - we instead want to use an empty immutable collection here.

core/src/main/java/ai/timefold/solver/core/impl/score/constraint/ConstraintMatchPolicy.java
```diff
@@ -13,33 +13,37 @@
...
+     * we can configure the solver to not produce justifications ({@link #ENABLED_WITHOUT_JUSTIFICATIONS_AND_INDICTMENTS}).
      *
      * @param scoreAnalysisFetchPolicy
      * @return Match policy best suited for the given fetch policy.
      */
     public static ConstraintMatchPolicy match(ScoreAnalysisFetchPolicy scoreAnalysisFetchPolicy) {
         return switch (scoreAnalysisFetchPolicy) {
-            case FETCH_MATCH_COUNT, FETCH_SHALLOW -> ENABLED_WITHOUT_JUSTIFICATIONS;
+            case FETCH_MATCH_COUNT, FETCH_SHALLOW -> ENABLED_WITHOUT_JUSTIFICATIONS_AND_INDICTMENTS;
             case FETCH_ALL -> ENABLED;
```
triceo, line 35: You could argue this is not backwards compatible. People will get indictments where previously they did not.

Not sure what I think about this yet.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/Scorer.java
```diff
@@ -78,6 +78,10 @@ public void retract(Tuple_ tuple) {
         }
     }
 
+    public void setNodeIds(long[] scorerNodeIds) {
```
triceo, line 81: Not excited that we're effectively creating more and more shared mutable state. Let's try to keep these things `final` and supplied at construction time. These "half-initialized" lazy objects make reasoning about the functionality/interactions significantly more difficult.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/common/inliner/ConstraintMatchSupplier.java
```diff
@@ -31,6 +43,17 @@
...
 
+    static @Nullable List<Object> collectIndictments(Constraint constraint, Tuple tuple) {
+        if (tuple.getIndictmentSource() == IndictmentSource.DISABLED) {
+            return null;
+        }
+        var out = new LinkedHashSet<>();
+        var abstractConstraint = (AbstractConstraint<?, ?, ?>) constraint;
+        var involvedNodeIds = Objects.requireNonNull(abstractConstraint.getInvolvedNodeIds());
+        tuple.getIndictmentSource().visitSources(involvedNodeIds, out::add);
+        return new ArrayList<>(out);
```
triceo: I wonder if we need this to be a list. (It used to be one in the old days.) But what would be the actual benefit/need? A `SequencedSet` works equally well IMO, and avoids this inefficient copy. 

`SequencedCollection`? I'm thinking that `List.of(...)` may be useful for testing, but `new LinkedHashSet<>()` is best for performance.

triceo (commented):
First read-through. An elegant solution, was expecting this to be more code; not happy with some bits, though. Let's try to adjust if we can.

Before we move further though, I believe we need to have some clarity on how stable this is. We have the [turtle tests](https://github.com/TimefoldAI/timefold-solver-benchmarks/actions/workflows/turtle.yml) - let's run them on this. (They may need some ad-hoc adjusting to be able to run with indictments.) If at the same time we can gather move speed data from them, we can compare to non-indictment runs and have a decent idea as to how much slower this will be.

core/src/main/java/ai/timefold/solver/core/impl/score/constraint/ConstraintMatch.java
Christopher-Chianelli: I was debating that; issue is, with an empty collection, we cannot differentiate between a constraint match having no indictments (ex: if we support user custom indictments later) and indictments being disabled.

core/src/main/java/ai/timefold/solver/core/impl/score/stream/bavet/common/Scorer.java
Christopher-Chianelli: Node ids are set in reverse order, so we do not know them until after construction of the Scorer.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractFlattenNode.java
Christopher-Chianelli: Strong disagree here. It would severely complicate the construction of tuples, and not all nodes want to inherit indictment source (ex: `join`, `groupBy`, `precompute`).

core/src/main/java/ai/timefold/solver/core/impl/score/constraint/ConstraintMatchPolicy.java
Christopher-Chianelli: Poor naming of the enum in the past; by using `ENABLED_WITHOUT_JUSTIFICATIONS`, backward compatability was bound to break the moment we added more features other than justifications. Perhaps `ENABLED` should have been called `ENABLED_ALL` and `ENABLED_WITHOUT_JUSTIFICATIONS` been called `ENABLED_CONSTRAINT_MATCH_ONLY`.

triceo: That is all probably true; but the question is how do we handle it now. IMO we can not just enable this out of a sudden.

Christopher-Chianelli:
For benchmarking purposes, I added `PreviewFeature.SOLVE_WITH_INDICTMENTS`, which can be used to solve with indictments enabled.

core/src/main/java/ai/timefold/solver/core/impl/score/constraint/ConstraintMatchPolicy.java
Christopher-Chianelli: Remove the new enum values; `ENABLED_WITHOUT_JUSTIFICATIONS` also disables indictments, `DISABLED` disable all, `ENABLED` enable all.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/IndictmentAnalysis.java
```diff
@@ -0,0 +1,15 @@
+package ai.timefold.solver.core.api.score.analysis;
+
+import java.util.List;
+
+import ai.timefold.solver.core.api.score.Score;
+
+public interface IndictmentAnalysis<Score_ extends Score<Score_>> {
```
triceo, line 13: Doesn't this need Javadoc?

core/src/main/java/ai/timefold/solver/core/api/score/analysis/Indictable.java
```diff
@@ -0,0 +1,37 @@
...
+ * and {@link IndictmentAnalysis#indictee()} is the same instance.
+ * <p>
+ * The primary use case for implementing {@link Indictable} is if you have a separate
+ * solver and rest model, and don't want to expose your solver types in your REST API.
+ * <p>
+ * Note: {@link ScoreAnalysis} in general and indictments in particular
+ * are exclusive to Timefold Solver Enterprise Edition.
+ */
+@NullMarked
+public interface Indictable {
```
triceo, line 19: I am not aware of any such model requirement.
I'm assuming that this class came out of the conversation with Geoff?

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
```diff
@@ -154,6 +155,15 @@ public interface ScoreAnalysis<Score_ extends Score<Score_>> {
...
     Collection<ConstraintAnalysis<Score_>> constraintAnalyses();
 
+    /**
+     * Returns a list of {@link IndictmentAnalysis} instances that make up this {@link ScoreAnalysis},
+     * sorted by the absolute value of their total score contribution.
+     *
+     * @return a list of {@link IndictmentAnalysis} linking planning entities and problem facts to their impact
+     *         in this {@link ScoreAnalysis}
+     */
+    List<IndictmentAnalysis<Score_>> indictmentAnalyses();
```
triceo, line 165: What is the value when indictments are disabled?

core/src/main/java/ai/timefold/solver/core/config/solver/PreviewFeature.java
```diff
@@ -26,6 +26,10 @@ public enum PreviewFeature {
      * It is intended to simplify the creation of custom moves, eventually replacing move selectors.
      * The component is under development, and many key features are yet to be delivered.
      */
-    NEIGHBORHOODS
+    NEIGHBORHOODS,
+    /**
+     * Not really a preview feature; allows solving with indictments enabled for benchmark purposes.
+     */
+    SOLVE_WITH_INDICTMENTS
```
triceo: Do we still need this?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -143,9 +158,28 @@ protected void decrementCounterRight(ExistsCounter<LeftTuple_> counter) {
             } else {
                 doInsertCounter(counter);
             }
+        } else {
+            // count != 0, so only propagate if we are in an `ifExists`
+            if (shouldExist) {
+                doUpdateCounter(counter);
+            }
```
triceo: Why are we changing `ifExists` behavior? Has it been broken?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/tuple/indictment/IndictmentSource.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.Set;
+import java.util.function.Consumer;
+
+import ai.timefold.solver.core.impl.bavet.common.tuple.Tuple;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public sealed interface IndictmentSource {
```
triceo, line 26: Please document this interface and its purpose.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIndexedIfExistsNode.java
```diff
@@ -97,6 +98,11 @@ public final void insertLeft(LeftTuple_ leftTuple) {
         if (!isFiltering) {
             counter.countRight = rightSize(leftTuple, compositeKey);
             initCounterLeft(counter);
+            if (leftTuple.getIndictmentSource() != IndictmentSource.DISABLED) {
```
triceo, line 101: IMO it should be documented somewhere what the behavior of indictments is supposed to be in each node. In ifExists, it is the least obvious.

triceo (commented):
First read.

Other than the questions inline, I think we should think about what happens with other score directors (easy, incremental) when indictments are enabled. Easy probably fails with justifications already, but incremental does have some support for justifications AFAIK, so we should consider adding support for indictments there as well.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/Indictable.java
Christopher-Chianelli: It is not a requirement for a model to have seperate solver and rest types, but I know a quite a few cases where they do use seperate solver and rest types. This class was primary created so the `type` field (which I later realized was neccessary so a frontend can identify the object that was serialized without the user manually adding it to their object) can be customized if they do not want it to be the canonical name of the class.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
Christopher-Chianelli: Currently a empty list.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
Christopher-Chianelli: Must of been a duplicate of the change in `decrementCounterRightUpdatingIndictment` when resolving merge conflicts; will remove

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
triceo: As long as it's the same behavior with justifications, I'm good.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
triceo: I am a bit surprised/worried that no test caught this.

Christopher-Chianelli: Well, it is an additional update, so it should do nothing assuming the update idempotent. The only way to check this would if scoring was retriggered or not.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/Indictable.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * An identifier that uniquely identifies the type returned by
+     * {@link #getIndictedObject()}.
+     * 
+     * @return a string that can be used to identify the class of {@link #getIndictedObject()}
+     */
+    IndictableTypeRef getIndictableTypeRef();
```
triceo, line 26: Personally, I'd just turn this to a `String` - the extra level of indirection doesn't seem to be necessary. People can easily convert class name to a string themselves, or choose a different ID entirely.

Avoids introducing a new type, and with it a naming conversation.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/IndictmentAnalysis.java
```diff
@@ -0,0 +1,36 @@
...
+ * @param <Score_>
+ * @see ScoreAnalysis Description of score analysis and the purpose of this class.
+ */
+public interface IndictmentAnalysis<Score_ extends Score<Score_>> {
+    /**
+     * The object that was indicted by constraints.
+     * 
+     * @return the indicted object
+     */
+    Object indictee();
```
triceo, line 19: I would explicitly mention that the identification of the object sits one level above. (It's not immediately obvious here.)

core/src/main/java/ai/timefold/solver/core/api/solver/ScoreAnalysisFetchPolicy.java
```diff
@@ -12,20 +12,23 @@ public enum ScoreAnalysisFetchPolicy {
 
     /**
      * {@link ScoreAnalysis} is fully initialized.
-     * All included {@link ConstraintAnalysis} objects include full {@link ConstraintAnalysis#matches() match analysis}.
+     * All included {@link ConstraintAnalysis} objects include full {@link ConstraintAnalysis#matches() match analysis},
+     * and {@link ScoreAnalysis#indictmentMap()} will be populated.
```
triceo, line 16: So we've decided to throw indictments even at people who never asked for them before?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -298,9 +339,12 @@ protected void updateCounterRight(ExistsCounter<LeftTuple_> counter, UniTuple<Ri
         }
         if (testFiltering(leftTuple, rightTuple)) {
             incrementCounterRight(counter);
+            IndictmentSource.addSupport(getId(), counter.getTuple(), rightTuple);
```
triceo, line 342: Now that I see it used, I question the name `addSupport`/`removeSupport` - it seems as if we're somehow adding or removing support for indictments, which is not true.

I'd consider renaming this concept. Auxilliary, corroborating, ancillary, aiding, ...

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -154,7 +155,6 @@ public Solver<Solution_> buildSolver(SolverConfigOverride configOverride) {
         solverScope.setScoreDirector(scoreDirector);
         solverScope.setProblemChangeDirector(new DefaultProblemChangeDirector<>(scoreDirector));
         var moveThreadCount = resolveMoveThreadCount(true);
-        var previewFeaturesEnabled = solverConfig.getEnablePreviewFeatureSet();
```
triceo, line 157: Seems like unnecessary change.

triceo (commented):
- Final points for discussion.
- Documentation is missing.
- Sonar is worth checking.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/Indictable.java
Christopher-Chianelli: I was debating this with myself when implementing it. One advantage of this is that it is easier/more intuitive for the user to get the correct key when not implementing Indictable; do `IndictableTypeRef.of(MyClass.class)` instead of `MyClass.class.getCanonicalName()`; and it matches `ConstraintRef`, which is also a record of a single String component. No strong feelings for it though.

core/src/main/java/ai/timefold/solver/core/api/solver/ScoreAnalysisFetchPolicy.java
Christopher-Chianelli: It is `FETCH_ALL`; it would be confusing to have a `FETCH_ALL_INCLUDING_INDICTMENTS`, and would make more sense to have `FETCH_ALL_EXCLUDING_INDICTMENTS` which we could make the new default used for SolutionManager.analyze()`.

triceo: Looked at from one side, entirely correct reasoning. From the point of backwards compatibility, an unexpected addition that _could_ break people if they iterate the JSON fields as opposed to accessing them as keys in a map.

The 100 % correct choice would have been to deprecate `FETCH_ALL` and introduce something else to take its place. Worth it though? Not entirely sure.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/Indictable.java
triceo: IMO just the name `IndictableTypeRef` is something that turns a simple concept into rocket science. I would avoid it.

If we want to stay with the concept of this being any random ID, then naming is an issue. However, if we decide that this is just a class name, then we can simply make this `String getClassName()` and we have something everyone will immediately understand.

Christopher-Chianelli: It can be any id, since people may not want to leak class names in their REST API (ex: they want their API to use 'Shift' as the type and not 'my.very.long.company.name.soiver.domain.TimefoldShift'). It is still related to classes though; indictees with the same `IndictableTypeRef` should be of the same type.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
Christopher-Chianelli: Was necessary when benchmarking indictments in score director (as it needed preview feature info before then). No longer necessary.

triceo:
Still missing the docs, plus Javadocs build appears to be broken.

Christopher-Chianelli:
Docs were added in commit https://github.com/TimefoldAI/timefold-solver/pull/2528/changes/5263dcb896369194326a03a0a9f101f3bda38395 ; so automation comment is stale.

PR #2529, opened 2026-07-23
fix: race condition in SolverWorker job registration

diogodanielsoaresferreira:
This PR aims to fix a very rare race condition in the SolverWorker. If the dataset fails right after it has started (eg. when creating a solver model), it may be the case that the job was not stored yet in the jobs map. That will make the sending of the failed event failed, because we have no way of fetching the job's metadata.

This PR makes sure that if we need to access the job metadata, we will wait for the job to be there to fetch all the metadata needed.

Fixes: github.com/TimefoldAI/timefold-platform/issues/5172


rsynek (approved):
Nice finding!

PR #2532, opened 2026-07-26
fix: prevent stale tuples after a join

triceo:
I fixed a left input variant of this a long time ago.
With the help of Claude, I was finally able to reproduce this from the right as well, and fixed it.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
@@ -69,18 +69,7 @@ public abstract class AbstractJoinNode<LeftTuple_ extends Tuple, Right_, OutTupl
 
     protected abstract boolean testFiltering(LeftTuple_ leftTuple, UniTuple<Right_> rightTuple);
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractJoinNode.java
```diff
@@ -92,15 +92,28 @@ protected final void insertOutTupleFilteredFromLeft(LeftTuple_ leftTuple, UniTup
...
     }
 
-    protected final void insertOutTupleFiltered(LeftTuple_ leftTuple, UniTuple<Right_> rightTuple) {
+    /**
+     * The mirror image of {@link #insertOutTupleFilteredFromLeft}:
+     * the right tuple is the one read out of storage, and can therefore be the retracting one.
+     * Reachable when the right input's node sits in a higher layer than the left input's,
+     * which delivers the left's inserts before the right's retracts.
+     */
+    protected final void insertOutTupleFilteredFromRight(LeftTuple_ leftTuple, UniTuple<Right_> rightTuple) {
```
Christopher-Chianelli: TBH not in love with this name; maybe `insertOutTupleIfRightActiveFiltered`?

Christopher-Chianelli (approved):

PR #2538, opened 2026-07-27
fix: change graalvm version to latest tag with a corresponding release

Christopher-Chianelli:
This is techically a downgrade. This seems to be the only relevant commit that changed in the past week (GraalVM was bumped from `1.6.2` (which also does not have a release) to `1.6.3` here: https://github.com/TimefoldAI/timefold-solver/pull/2521.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -89,7 +89,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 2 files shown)


Christopher-Chianelli:
(and since actions are run from main, any dependabot PR that touch actions will not show failures until after they are merged).

triceo (approved):
Well, let's give it a try.

PR #2540, opened 2026-07-27
fix: downgrade native-image-plugin

Christopher-Chianelli:
Downgrading to 1.1.5 fixes the native image failure locally.

### DIFF
--- a/spring-integration/spring-boot-integration-test/pom.xml
+++ b/spring-integration/spring-boot-integration-test/pom.xml
@@ -100,8 +100,6 @@
           <plugin>
             <groupId>org.graalvm.buildtools</groupId>
             <artifactId>native-maven-plugin</artifactId>
-            <!-- The version inherited by Spring (1.1.6) has a dependency issue causing a failure at runtime -->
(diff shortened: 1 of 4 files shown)


Christopher-Chianelli:
The action is still using the `native:1.1.6` version despite the override in the `pom.xml` inside the action. Running `mvn clean install -Dnative` works locally.

spring-integration/spring-boot-integration-test/pom.xml
```diff
@@ -100,6 +100,8 @@
           <plugin>
             <groupId>org.graalvm.buildtools</groupId>
             <artifactId>native-maven-plugin</artifactId>
+            <!-- The version inherited by Spring (1.1.6) has a dependency issue causing a failure at runtime -->
+            <version>1.1.5</version>
```
triceo: Let's declare this is the build parent, please.
Also, let's bring back the newer setup-graalvm action, since that was a red herring.

triceo (approved):
Good that we have the fix!
LGTM after comments resolved.

PR #2542, opened 2026-07-28
fix: remove model.api.version

### LINKED ISSUE #2516: Possible duplication of model version properties
`model.api.version` and `timefold.application.version` express the same concept, but since they are defined in different places, might get out of sync.

service-parent pom.xml
    <model.api.version>v1</model.api.version>          ← Maven property
      │
      └─► quarkus-maven-plugin <properties>
            <model.api.version>${model.api.version}</model.api.version>   ← passed to Quarkus at runtime
                │
                └─► service-defaults/application.properties
                      quarkus.rest.path=${model.api.version}              ← becomes the URL prefix

  timefold.application.version=${model.api.version} (also in service-defaults) just happens to read the same Maven property for its metadata default — they're coincidentally the
  same value but completely independent.

  One side effect worth noting: the guide telling users to explicitly set timefold.application.version=v1 in their own application.properties (ordinal 250, higher than
  service-defaults' 230) locks that metadata to v1 even if someone later bumps model.api.version in the POM. They'd get /v2/timetables served but the OpenAPI spec still saying
  version v1. That's a minor inconsistency risk in the guide's example, but not related to the URL doubling bug. The docs are otherwise correct as written.

rsynek:
Fixes https://github.com/TimefoldAI/timefold-solver/issues/2516.

Also, removed the SDK version from metadata (and keep only the Solver version).

See also: https://github.com/TimefoldAI/timefold-solver-enterprise/pull/791

The API version is now defined by the `timefold.application.version` in the `pom.xml`. The maven property is used to define a model descriptor classifier, e.g. `timefold-pickup-delivery-routing-descriptor-1.2.0-v1.zip`. Thus, it must be defined in `pom.xml` and not in the `application.properties`, where it gets injected by the Quarkus maven plugin.

The service-parent `pom.xml` defines the default value `v1`, which can be overridden in the model `pom.xml`.

The user must not override the value in `application.properties`, which may lead to the model being updated to the new version, but its model descriptor classifier still being `v1`.

### DIFF
--- a/docs/src/modules/ROOT/pages/quickstart/service/getting-started.adoc
+++ b/docs/src/modules/ROOT/pages/quickstart/service/getting-started.adoc
@@ -250,7 +250,6 @@ Create the `src/main/resources/application.properties` file:
 ai.timefold.platform.termination.spent-limit=PT30S
 
 # Application metadata — included in the generated OpenAPI specification
-model.api.version=v1
 timefold.application.name=my-model
 timefold.application.contact.email=info@example.com
 timefold.application.contact.name=Your Name
@@ -264,7 +263,6 @@ For production use, consider at least five minutes (`PT5M`) to allow the solver 
 NOTE: The per-request `config.run.termination.spentLimit` field in the POST body (shown later) overrides this default for individual runs.
 
 `timefold.application.name` and the contact fields are *required* metadata.
-`model.api.version` is required (for example `v1`) and is used as the default for `timefold.application.version` and `quarkus.rest.path`.
 They are used to identify your service, validate the JSON schema of requests, and populate the generated OpenAPI specification.
 
 == REST API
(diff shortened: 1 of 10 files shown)


triceo (approved):
LGTM. Although you may want to rebase to see if the CI failures went away.

cristianonicolai (approved):

mswiderski (approved):
I assume it is still possible to change the version from command line as it used to be, right? I used it quite often when working with maven plugin deployments.

rsynek:
> I assume it is still possible to change the version from command line as it used to be, right? I used it quite often when working with maven plugin deployments.

Yes, it just needs to be changed as a maven property, so e.g.:

`mvn clean install -Dtimefold.application.version=v2`

PR #2543, opened 2026-07-28
chore: new no-loops branch

Christopher-Chianelli:
The no-loops branch has a ton of conflicts with main; to have a meaniful diff, the PR is based on main (although it will replace no-loops via a force push). If approved, it will replace the existing no-loops branch which cannot be cleanly merged currently.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
@@ -57,6 +57,7 @@ import ai.timefold.solver.core.impl.domain.solution.cloner.FieldAccessingSolutio
 import ai.timefold.solver.core.impl.domain.solution.cloner.gizmo.GizmoSolutionCloner;
 import ai.timefold.solver.core.impl.domain.solution.cloner.gizmo.GizmoSolutionClonerFactory;
 import ai.timefold.solver.core.impl.domain.variable.declarative.DeclarativeShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.declarative.ShadowVariablesInconsistentVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.BasicVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.GenuineVariableDescriptor;
 import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
@@ -986,6 +987,13 @@ public final class SolutionDescriptor<Solution_> {
         return declarativeShadowVariableDescriptorList;
     }
 
+    public boolean hasAnyShadowVariablesInconsistentMember() {
+        return entityDescriptorMap.values().stream()
+                .flatMap(entityDescriptor -> entityDescriptor.getShadowVariableDescriptors().stream())
+                .anyMatch(
+                        shadowVariableDescriptor -> shadowVariableDescriptor instanceof ShadowVariablesInconsistentVariableDescriptor<Solution_>);
+    }
+
     public Stream<Object> extractAllEntitiesStream(Solution_ solution) {
         var stream = Stream.empty();
         for (var memberAccessor : entityMemberAccessorMap.values()) {
@@ -1051,5 +1059,4 @@ public final class SolutionDescriptor<Solution_> {
     public String toString() {
         return "%s(%s)".formatted(getClass().getSimpleName(), solutionClass.getName());
     }
-
 }
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
@@ -85,6 +85,8 @@ public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeTra
      *           so {@link #beforeVariableChanged(VariableMetaModel, Object)}
      *           and {@link #afterVariableChanged(VariableMetaModel, Object)}
      *           can short circuit.
+     *
+     * @return true if the update successful; false otherwise
      */
     abstract boolean innerUpdateChanged();
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
@@ -86,7 +86,7 @@ public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeTra
      *           and {@link #afterVariableChanged(VariableMetaModel, Object)}
      *           can short circuit.
      */
-    abstract void innerUpdateChanged();
+    abstract boolean innerUpdateChanged();
 
     /**
      * Called when any non-declarative source variable for the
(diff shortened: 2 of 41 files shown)


core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -986,6 +987,13 @@ public List<DeclarativeShadowVariableDescriptor<Solution_>> getDeclarativeShadow
         return declarativeShadowVariableDescriptorList;
     }
 
+    public boolean hasAnyShadowVariablesInconsistentMember() {
```
triceo, line 990: Arguably we can make this a field that is precomputed as the entity descriptors are being built. (Not sure how often this is called and from where; may be a moot point.)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AbstractVariableReferenceGraph.java
```diff
@@ -86,7 +86,7 @@ public abstract sealed class AbstractVariableReferenceGraph<Solution_, ChangeTra
      *           and {@link #afterVariableChanged(VariableMetaModel, Object)}
      *           can short circuit.
      */
-    abstract void innerUpdateChanged();
+    abstract boolean innerUpdateChanged();
```
triceo, line 91: Since this method has a Javadoc, it should document the newly added return.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
```diff
@@ -129,19 +135,22 @@ private boolean updateEntityShadowVariables(GraphNode<Solution_> entityVariable,
...
         // Do not need to update anyChanged here; the graph already marked
         // all nodes whose looped status changed for us
-        var groupEntities = shadowVariableReferences.get(0).groupEntities();
-        var groupEntityIds = entityVariable.groupEntityIds();
 
-        if (groupEntities != null) {
-            for (var i = 0; i < groupEntityIds.length; i++) {
-                var groupEntity = groupEntities[i];
-                var groupEntityId = groupEntityIds[i];
+        if (!ignoreInconsistentSolutions || !consistencyProcessed) {
```
triceo: I'd flip the condition and start with the other branch. Easier to read that way, without all that negation.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -162,29 +162,31 @@ public ChangedVariableNotifier<Solution_> changedVariableNotifier() {
         }
     }
 
-    public static <Solution_> VariableReferenceGraph buildGraph(GraphDescriptor<Solution_> graphDescriptor) {
+    public static <Solution_> VariableReferenceGraph buildGraph(GraphDescriptor<Solution_> graphDescriptor,
+            boolean ignoreInconsistentSolutions) {
```
triceo: I wonder why the boolean is not part of the graph descriptor. As opposed to having to be propagated everywhere alongside.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
```diff
@@ -64,4 +73,20 @@ public void setUnknownInconsistencyValues() {
         graph.commitChanges(changeTracker);
         affectedEntitiesUpdater.setUnknownInconsistencyValues();
     }
+
+    @Override
+    public List<Object> getInconsistentEntities() {
+        var out = new LinkedIdentityHashSet<>();
```
triceo: We do not need to track entities by their identity.
Entities are not allowed to equal each other; in Solver 2.0, we made sure this is documented and at some places even enforced by a fail-fast.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultVariableReferenceGraph.java
```diff
@@ -64,4 +73,20 @@ public void setUnknownInconsistencyValues() {
...
+        graph.forEachEdge(graphTrackingInconsistentEntities::addEdge);
+        graphTrackingInconsistentEntities.commitChanges(new BitSet());
+        var loopedComponentList = graphTrackingInconsistentEntities.getLoopedComponentList();
+        for (var loopedComponent : loopedComponentList) {
+            for (var nodeId : loopedComponent) {
+                var node = this.nodeList.get(nodeId);
+                out.add(node.entity());
+            }
+        }
+        return new ArrayList<>(out);
```
triceo: I wonder if we need this copy. `SequencedCollection` return type would do the same job.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -11,13 +11,17 @@
...
+import org.slf4j.LoggerFactory;
+
 /**
  * Remembers the {@link PlanningSolution best solution} that a {@link Solver} encounters.
  *
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 public class BestSolutionRecaller<Solution_> extends PhaseLifecycleListenerAdapter<Solution_> {
 
+    private static final Logger log = LoggerFactory.getLogger(BestSolutionRecaller.class);
```
triceo: `LOGGER`

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -44,12 +48,22 @@ public void setSolverEventSupport(SolverEventSupport<Solution_> solverEventSuppo
...
         var scoreDirector = solverScope.getScoreDirector();
+        @SuppressWarnings("rawtypes")
         InnerScore innerScore = scoreDirector.calculateScore();
+        if (innerScore.isInvalid()) {
+            log.warn("The initial solution passed to the solver is inconsistent. Unassigning involved entities.");
+            scoreDirector.unassignInconsistentEntities();
+            innerScore = scoreDirector.calculateScore();
+            if (innerScore.isInvalid()) {
+                throw new IllegalStateException(
+                        "The initial solution passed to the solver is inconsistent even after unassigning involved entities.");
```
triceo: On the next line, I'd say something like "Maybe provide an uninitialized solution."

Give the user a hint how to solve this problem. As it stands right now, the error message is scary, and doesn't give any way out. Which means, best case, it ends up on Discord; worst case, the user gives up.

triceo (commented):
Went over it all over again; adding some comments, likely for things I did not notice before.

I suggest we start writing docs for this; the docs need to list all the behaviors and it will be easy to forget some, if we try to write the docs at the very end of the process.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/AffectedEntitiesUpdater.java
Christopher-Chianelli: There is no other branch though; I can flip it through de morgans though.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
Christopher-Chianelli: Called once to determine if inconsistent solutions should be ignored. Result is already cached on the score director.

PR #2544, opened 2026-07-29
perf: decrease the overhead of constraint profiling

triceo:


Christopher-Chianelli (approved):
What percentage change was there from changing from a `long` to an `int` id? LGTM.

triceo:
> What percentage change was there from changing from a `long` to an `int` id? LGTM.

I only did that to facilitate the actual perf improvement, which lives in enterprise. This change is immaterial to performance.

PR #2547, opened 2026-07-30
fix: return the sdkVersion to ModelBuildInfo

rsynek:
Putting back the SDK version since it breaks the platform. Removal needs to be properly planned.

### DIFF
--- a/service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
+++ b/service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
@@ -1,7 +1,13 @@
 package ai.timefold.solver.service.definition.internal.descriptor;
 
+import io.quarkus.runtime.annotations.RecordableConstructor;
+
 public record ModelBuildInfo(String solverVersion, @Deprecated(forRemoval = true) String sdkVersion, String version,
         String buildTime, String branch, String buildCommit) {
+
+    @RecordableConstructor
+    public ModelBuildInfo {
+    }
 
     public ModelBuildInfo(String solverVersion, String version, String buildTime, String branch, String buildCommit) {
         this(solverVersion, solverVersion, version, buildTime, branch, buildCommit);
--- a/service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
+++ b/service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
@@ -16,7 +16,7 @@ import io.quarkus.runtime.annotations.RecordableConstructor;
 public record ModelBuildInfo(String solverVersion, @Deprecated(forRemoval = true) String sdkVersion, String version,
         String buildTime, String branch, String buildCommit) {
 
-    private static final ModelBuildInfo EMPTY = new ModelBuildInfo(null,null, null, null, null);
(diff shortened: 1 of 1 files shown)


service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
```diff
@@ -1,9 +1,9 @@
 package ai.timefold.solver.service.definition.internal.descriptor;
 
-public record ModelBuildInfo(String solverVersion, String version, String buildTime, String branch,
+public record ModelBuildInfo(String solverVersion, String sdkVersion, String version, String buildTime, String branch,
```
triceo: If it's going away, properly deprecate it please.

triceo: You can also create an additional constructor without this argument, to ease people's migration.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
```diff
@@ -1,9 +1,19 @@
 package ai.timefold.solver.service.definition.internal.descriptor;
 
-public record ModelBuildInfo(String solverVersion, String version, String buildTime, String branch,
-        String buildCommit) {
+import io.quarkus.runtime.annotations.RecordableConstructor;
+
+public record ModelBuildInfo(String solverVersion, @Deprecated(forRemoval = true) String sdkVersion, String version,
```
triceo, line 16: I believe Javadoc will complain here.
Proper deprecation includes `@deprecated` Javadoc tags.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
```diff
@@ -1,9 +1,19 @@
...
+        String buildTime, String branch, String buildCommit) {
+
+    @RecordableConstructor
```
triceo, line 21: What is this? Never seen this, internet isn't much help.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
```diff
@@ -1,9 +1,19 @@
...
+    public ModelBuildInfo {
+    }
+
+    public ModelBuildInfo(String solverVersion, String version, String buildTime, String branch, String buildCommit) {
+        this(solverVersion, solverVersion, version, buildTime, branch, buildCommit);
+    }
 
     public static ModelBuildInfo empty() {
-        return new ModelBuildInfo(null, null, null, null, null);
+        return new ModelBuildInfo(null, null, null, null, null, null);
```
triceo: Arguably `private static final ModelBuildInfo EMPTY`.

rsynek: Used only by tests, but ok.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/descriptor/ModelBuildInfo.java
rsynek: Quarkus Build-time recording needs to have a constructor that matches all fields.

PR #2506, opened 2026-07-15
feat: support sourcing declarative shadow variables from planning list variable elements

fodzal:
## What

Follow-up to #2469. An entity declaring a `@PlanningListVariable` can now use a
`@ShadowSources` path such as `"visits[].endTime"` to depend on a declarative shadow
variable of the list variable's elements:

```java
@PlanningListVariable
List<Visit> visits;

@ShadowVariable(supplierName = "endTimeSupplier")
LocalDateTime endTime;

@ShadowSources("visits[].endTime")
LocalDateTime endTimeSupplier() {
    return visits.isEmpty() ? departureTime : visits.getLast().getEndTime();
}
```

This is the capability discussed in the #2469 review: accessing the declarative shadow
variables of entities inside the `@PlanningListVariable`, with the edges of the graph
created and updated as the list variable changes. The fail-fast added by #2469 now points
users to this syntax.

## How

- `RootVariableSource` accepts a `[]` segment on a genuine list variable declared on the
  root entity, producing the new `ParentVariableType.LIST_ELEMENT`. The path must end on a
  declarative shadow variable of the element (optionally reached via a fact).
- Fan-in edges (`element.shadowVariable -> entity.shadowVariable`) are created from the
  list variable's contents when the graph is built. They are registered as initial dynamic
  edges, not fixed edges, so a dependency loop through them is treated as breakable by the
  solver instead of failing fast.
- `before/afterListVariableChanged` events are forwarded to the declarative shadow
  variable session, which incrementally removes/adds the fan-in edges for the changed
  range and always marks the target variable as changed (covering removals and
  reorderings). Models without list element sources skip this entirely.
- Thanks to the edges, the topological order guarantees the target variable is computed
  once per move, after all its elements' shadow variables have settled.
- `GraphStructure` classifies models with list element sources as arbitrary, since the
  graph's edges change during solving; the grouped-updater decision is static for the same
  reason.

## Testing

- Path parsing: valid `values[].shadowVar` forms and the new fail-fasts
  (non-declarative element variable, list variable reached via another member) in
  `RootVariableSourceTest`.
- `ListElementShadowVariableTest`: `SolutionManager.updateShadowVariables` (including
  empty lists and mutations between calls) and two `FULL_ASSERT` solver runs, one from an
  uninitialized solution and one from an initialized solution (exercising the initial
  edge seeding).
- A dedicated `testdomain/shadow/list_element` domain is used instead of extending
  `simple_list`, because adding a list element source there would reclassify it away from
[...]

PR #2548, opened 2026-07-30
perf: update planning list variable elements by a cascade instead of the graph

fodzal:
## Why

Follow-up to the performance discussion in #2357 and to #2506.

The arbitrary variable reference graph is often much slower than what hand-written
variable listeners achieve for the same model: it maintains a node per entity/variable
pair, dynamic edges and an incremental topological order on every move. #1659 closed that
gap for simple structures (`EMPTY`, `NO_DYNAMIC_EDGES`, `SINGLE_DIRECTIONAL_PARENT`), but
a model that falls outside them drops back to the arbitrary graph, and on large problems
that cost dominates solving. In planning list variable models this hits particularly
hard: the elements are typically counted in thousands while the other entities are
counted in hundreds, so the per-element part of the graph is usually the updater's
bottleneck.

The motivating use case is multi-trip vehicle routing that cannot be modeled by simulating
a depot return when capacity is exceeded, for example:

- an ambulance that may return to the depot at any time to change crews;
- two vehicles that must ride the same visits for part of the day (e.g. training) —
  unlike synchronized visits, where a single visit requires two technicians.

Such problems are modeled with one planning entity per trip: vehicle A becomes trips A1
and A2, and A2's first travel leg starts from A1's last visit. The trip's start must
depend on the previous trip's end, and the trip's end on its own visits — a list element
source (#2506) plus a cross-entity reference, which today classifies the whole model as
arbitrary:

```java
@PlanningEntity
public class Vehicle {

    Vehicle previousTrip; // The same physical vehicle's previous trip, or null; a fact.

    @PlanningListVariable
    List<Visit> visits;

    @ShadowVariable(supplierName = "startTimeSupplier")
    LocalDateTime startTime;

    @ShadowSources("previousTrip.endTime")
    LocalDateTime startTimeSupplier() {
        return previousTrip == null ? SHIFT_START : previousTrip.getEndTime();
    }

    @ShadowVariable(supplierName = "endTimeSupplier")
    LocalDateTime endTime;

    @ShadowSources("visits[].endServiceTime")
    LocalDateTime endTimeSupplier() {
        return visits.isEmpty() ? startTime : visits.getLast().getEndServiceTime();
    }
}

@PlanningEntity
public class Visit {

    Duration serviceDuration;

    @InverseRelationShadowVariable(sourceVariableName = "visits")
    Vehicle vehicle;

    @PreviousElementShadowVariable(sourceVariableName = "visits")
    Visit previous;

    @ShadowVariable(supplierName = "endServiceTimeSupplier")
    LocalDateTime endServiceTime;

    @ShadowSources({ "previous.endServiceTime", "vehicle.startTime" })
    LocalDateTime endServiceTimeSupplier() {
        if (vehicle == null) {
            return null;
        }
        var departureTime = previous == null ? vehicle.getStartTime() : previous.getEndServiceTime();
        return departureTime.plus(travelTime()).plus(serviceDuration);
    }
}
```

In such a model the visits vastly outnumber the vehicles, so nearly all of the arbitrary
graph's nodes, edges and topological-order maintenance is spent on them — while their
dependencies are precisely the ones that need no graph: a visit only reads its chain and
its own vehicle.

## What

A narrower fix would extend the `SINGLE_DIRECTIONAL_PARENT` detection to this shape and
add another specialized updater. This PR instead splits the model so that each half is
handled by machinery that is already good at it:

- the planning list variable's **elements** are excluded from the variable reference
  graph and updated by a **cascade** that walks each dirty entity's list in chain order,
  the way a variable listener would;
- the **remaining entities** keep the existing graph machinery, unchanged and with its
  full generality: any structure, dynamic edges, groups, several entity classes,
  dependency loops handled as inconsistency. Without per-element edges the graph is small
  and often has no dynamic edges left; the example above gets a fixed graph over the
  vehicles.

The detection (`GraphStructure.determineListElementCascade`) only inspects the element
class and the references toward it, and accepts the decomposition when:

- the elements' declarative sources stay chain-local: `previous`/`next` (a single
  direction across the model), their own members, or — through their inverse — a
  declarative variable of their entity that does not itself depend on the list
  (**pre-chain**);
- the other classes only reach the elements through `visits[].x` on their own list
  variable;
- there are no alignment keys on the element or list entity class, and the element class
  is distinct from the list entity class.

Anything else falls through to the current classification, so unsupported models behave
exactly as they do today.

## How

- The list entity's declarative variables are classified as **pre-chain** (readable by
  the elements, computed before walking a chain) or **post-chain** (depending on
  `visits[].x` directly or transitively, computed after). Cross-entity references do not
  propagate post-chain status: "B starts where A ends" keeps `startTime` pre-chain of B.
- `ListElementCascadeVariableReferenceGraph` wraps the inner graph and receives all
  events. Element source changes accumulate dirty elements; list change events mark the
  changed range dirty and the owner's post-chain variables changed, even for an empty
  range, since a removal changes the aggregate; a `ChangedVariableNotifier` wrapper
  observes pre-chain changes during updates and flags the owner for a whole-chain walk.
- `updateChanged()` alternates the cascade and the inner graph until neither has work
  left. A walk starts at the earliest dirty element and may terminate early after the
  last dirty one once values stop changing. An owner whose pre-chain variables just
  changed is deferred while other chains still need walking, so every supplier is
  computed once per update; the number of passes is bounded by the depth of inter-entity
  dependencies that traverse the lists.

  On the example above, an update that touches A's route settles in this order:

  ```text
  A.startTime --> [a1 -> a2 -> ... -> aN] --> A.endTime --> B.startTime --> [b1 -> ... -> bM] --> B.endTime --> ...
   pre-chain           cascade walk           post-chain     pre-chain          cascade walk       post-chain
  (inner graph)                              (inner graph) (inner graph)                          (inner graph)
  ```
- When the inner graph marks an owner inconsistent (a dependency loop the solver may
  break later), the cascade marks its elements inconsistent instead of computing them,
  and recomputes them when consistency returns.
- The factory changes are mechanical: the existing graph builders now take the list of
  descriptors to cover (defaulting to all of them), and under a cascade the per-element
  locator and edges are skipped while the "list changed -> target dirty" processor is
  kept.

Notes:

- The chain walk itself is the same idea as `@CascadingUpdateShadowVariable`'s listener:
  start at the first changed element, stop early once values settle. What differs is what
  triggers and surrounds it. A cascading update shadow variable only reacts to its
  element's own list position changes, so a dependency on another entity's variables is
  neither declared nor retriggered — the multi-trip model above is not expressible with
  it. Here the walk is embedded in the declarative dependency graph: a pre-chain change
  re-walks the chain, post-chain variables and downstream entities are ordered by the
  graph, and dependency loops are still detected as inconsistency.
- A model where only the elements have declarative variables keeps using
  `SINGLE_DIRECTIONAL_PARENT`. With an `EMPTY` inner graph the wrapper is behaviorally
  equivalent to it, so a later unification is possible, but out of scope here.
- The detection relies on the model having a single planning list variable, which
  `SolutionDescriptor` currently guarantees; a guard documents the assumption and falls
  back to the current behavior if that restriction is ever lifted.

## Testing

- Detection in `GraphStructureTest`: accepted shapes (fact-collection chain, plain-fact
  chain, `next` direction, a three-class model whose depot keeps its own graph) and a
  rejected shape (a fact collection of another vehicle's visits).
- Session-level tests on dedicated domains: propagation across chained vehicles, a swap
  creating non-contiguous dirty elements on the same chain, and a pre-chain change
  reaching an element that reads it directly while its predecessors are unchanged.
- Every domain solves with `FULL_ASSERT`, with constraints reading the shadow variables;
  after each scenario and each solve, the incrementally maintained values are compared
  against a from-scratch `SolutionManager.updateShadowVariables` recompute, which uses
  the arbitrary graph.
- `ListElementCascadeVariableReferenceGraphTest` counts supplier calls to pin the
  single-computation guarantee, and documents the one residual extra evaluation (the
  post-chain variable of an entity whose chain is walked again).
- A dependency loop between vehicles still hits the existing "fixed dependency loops"
  fail-fast.
- Full `core` test suite passes.

Local search throughput at the motivating scale (579 vehicle trips, 7,000 visits, same
solver configuration and machine within each row):

| Model | `main` (arbitrary graph) | This PR (cascade) | Speedup |
|---|---:|---:|---:|
| Production multi-trip VRPTW (real constraints) | 39 moves/s | 26,765 moves/s | ~686x |
| Synthetic fact-chain test domain (one trivial constraint) | 69 moves/s | ~14,100 moves/s | ~200x |

The synthetic row reproduces the gap independently of the business model, on this PR's
own fact-chain test domain, from an initialized solution and with graph construction
excluded from the measurement. The unit test domains are far too small to exhibit this
(the gap grows with the element count), which is why they focus on correctness instead.


triceo:
@fodzal Thanks for the PR! Looks interesting.

FYI This will take us a while to review. We first need to finish some other work in this area, which will likely cause conflicts for you. Once you resolve them, we should be able to review.

Please bear with us for a while; I'll let you know when we're ready to move forward.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
```diff
@@ -641,6 +641,10 @@ public ListVariableDescriptor<Solution_> getListVariableDescriptor() {
         return listVariableDescriptorList.isEmpty() ? null : listVariableDescriptorList.getFirst();
     }
 
+    public List<ListVariableDescriptor<Solution_>> getListVariableDescriptorList() {
```
Christopher-Chianelli: Why was this method added? Only a single `ListVariable` is supported; you can use `getListVariableDescriptor()` and check for `null` instead.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -207,6 +217,105 @@ static <Solution_> VariableReferenceGraph buildSingleDirectionalParentGraph(
...
+        // The pre-chain variables the elements read through their inverse:
+        // when one of them changes during an update, its entity's whole list must be walked,
+        // since any element may read it.
+        // These variables are declarative, so they only ever change through a notifier;
+        // wrapping it observes every such change, including those of the inner graph's updates.
+        var elementReadVariableSet = new HashSet<VariableMetaModel<?, ?, ?>>();
+        for (var descriptor : elementDescriptorList) {
+            for (var source : descriptor.getSources()) {
+                if (source.parentVariableType() == ParentVariableType.INVERSE) {
+                    elementReadVariableSet.add(source.variableSourceReferences().get(1).variableMetaModel());
```
Christopher-Chianelli: This looks fragile - Use `.getFirst().downstreamDeclarativeVariableMetamodel()` instead. If `downstreamDeclarativeVariableMetamodel` is null, then it referencing a non-declarative variable on the inverse (ex: `(inverse) vehicle.(planning variable) employee` in a mixed model).

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
```diff
@@ -207,6 +217,105 @@ static <Solution_> VariableReferenceGraph buildSingleDirectionalParentGraph(
...
+
+        // The topologically first chain element is the list's first element for a previous parent,
+        // but the list's last element for a next parent, whose successor function walks backwards.
+        var isPreviousDirection = graphStructureAndDirection.direction() == ParentVariableType.PREVIOUS;
+        Function<Object, @Nullable Object> ownerToFirstElement = owner -> {
+            var elementList = listVariableDescriptor.getValue(owner);
+            if (elementList.isEmpty()) {
+                return null;
+            }
+            return isPreviousDirection ? elementList.get(0) : elementList.get(elementList.size() - 1);
```
Christopher-Chianelli: Use `getFirst()`/`getLast()`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
```diff
@@ -51,9 +54,28 @@ public enum GraphStructure {
...
+    /**
+     * When present, the planning list variable's elements are excluded from the variable
+     * reference graph, which only covers the other entity classes;
+     * the elements are instead updated by a cascade that walks each dirty entity's list
+     * in the direction of {@link GraphStructureAndDirection#direction()}.
+     * This decomposition is valid because the elements only read their chain and,
+     * through their inverse, pre-chain declarative variables of their own list entity,
+     * and because the other classes only reach the elements through the list variable itself.
+     */
+    public record ListElementCascade(Class<?> elementEntityClass) {
```
Christopher-Chianelli: Do not create a record containing only a single field and not additional methods; just use the type `Class<?>` directly.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
```diff
@@ -136,6 +191,184 @@ public static <Solution_> GraphStructureAndDirection determineGraphStructure(
...
+     * Non-null if the planning list variable's elements can be excluded from the variable
+     * reference graph and updated by a cascade instead; see {@link ListElementCascade}.
+     * Only the element class's sources and the references towards the element class are
+     * checked here: the rest of the model is covered by the graph, whatever its structure.
+     */
+    private static <Solution_> @Nullable ListElementCascadeAndDirection determineListElementCascade(
+            SolutionDescriptor<Solution_> solutionDescriptor,
+            List<DeclarativeShadowVariableDescriptor<Solution_>> declarativeShadowVariableDescriptors) {
+        var listVariableDescriptorList = solutionDescriptor.getListVariableDescriptorList();
+        if (listVariableDescriptorList.size() != 1) {
```
Christopher-Chianelli: Impossible; multiple list variables are not supported; use getListVariableDescriptor and replace this with a null check.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
```diff
@@ -136,6 +191,184 @@ public static <Solution_> GraphStructureAndDirection determineGraphStructure(
...
+        ParentVariableType direction = null;
+        Class<?> elementEntityClass = null;
+        for (var descriptor : declarativeShadowVariableDescriptors) {
+            for (var source : descriptor.getSources()) {
+                var parentVariableType = source.parentVariableType();
+                if (parentVariableType == ParentVariableType.PREVIOUS || parentVariableType == ParentVariableType.NEXT) {
+                    var sourceParentMetaModel = source.variableSourceReferences().get(0).variableMetaModel();
+                    if (parentMetaModel == null) {
+                        parentMetaModel = sourceParentMetaModel;
+                        direction = parentVariableType;
```
Christopher-Chianelli, line 220: Does not properly handle when some variables depend on `PREVIOUS`, and other various depend on `NEXT`. Need to do `if (direction == null) { direction = parentVariableType; } else if (direction != parentVariableType) { return null; }`

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
```diff
@@ -136,6 +191,184 @@ public static <Solution_> GraphStructureAndDirection determineGraphStructure(
...
+                    return null;
+                }
+            }
+        }
+        if (!hasNonElementDescriptors) {
+            // A model with only element variables is covered by the existing structures.
+            return null;
+        }
+        var hasElementAlignmentKey = elementDescriptorList.stream()
+                .anyMatch(descriptor -> descriptor.getAlignmentKeyMap() != null);
```
Christopher-Chianelli: Could do this check and exit early in the `elementEntityClass.isAssignableFrom(entityClass)` of the `for`.

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/DeclarativeShadowVariableAssertions.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param shadowValueExtractors one per shadow variable to compare, each returning its value for every entity
+     */
+    @SafeVarargs
+    static <Solution_> void assertShadowsAreAtFixedPoint(Solution_ solution,
+            Function<Solution_, List<?>>... shadowValueExtractors) {
+        var incrementalValueLists = Arrays.stream(shadowValueExtractors)
+                .map(extractor -> new ArrayList<Object>(extractor.apply(solution)))
+                .toList();
+
+        SolutionManager.updateShadowVariables(solution);
```
Christopher-Chianelli: FYI, this will use the new graph type (`updateShadowVariables` creates a `ScoreDirector` when given a solution).

core/src/test/java/ai/timefold/solver/core/testdomain/shadow/multi_entity_chain/TestdataMultiEntityChainVehicle.java
```diff
@@ -0,0 +1,119 @@
...
+    int previousEndTimeCalledCount = 0;
+    int endTimeCalledCount = 0;
+
+    public TestdataMultiEntityChainVehicle() {
+    }
+
+    public TestdataMultiEntityChainVehicle(String code, int departureTime) {
+        super(code);
+        this.departureTime = departureTime;
+        // A head vehicle's only source is an empty fact collection,
```
Christopher-Chianelli: Can you also use a `null` collection to represent a head vehicle? If not, I recommend adding support for it.

core/src/test/java/ai/timefold/solver/core/testdomain/shadow/multi_entity_chain/TestdataMultiEntityChainVisit.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @ShadowSources({ "vehicle", "vehicle.previousEndTime", "previousVisit", "previousVisit.endServiceTime" })
+    public Integer endServiceTimeSupplier() {
+        calledCount++;
+        if (vehicle == null) {
+            return null;
+        }
+        Integer base;
+        if (previousVisit == null) {
+            // No unboxing: previousEndTime may be null while the vehicles' values converge.
```
Christopher-Chianelli: `converge` -- what? Is this called multiple times per update?

Christopher-Chianelli (changes requested):
Thanks for the PR @fodzal ; did you intend for this PR to be reviewed? It is still marked as draft/in-progress. Draft PR's don't typically get reviews unless explicitly requested. Comments inline.

core/src/main/java/ai/timefold/solver/core/impl/domain/solution/descriptor/SolutionDescriptor.java
fodzal: It was there to reject models with multiple list variables.

Removed; the detection now uses `getListVariableDescriptor()` with a null check.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
fodzal: Done, here and in the `INVERSE` case of the detection in `GraphStructure`, which now checks
`downstreamDeclarativeVariableMetamodel() == null` instead of `!references.get(1).isDeclarative()`.

Since the detection rejects inverse sources targeting a non-declarative variable, this spot wraps it in
`Objects.requireNonNull`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSessionFactory.java
fodzal: Done, and swept the rest of the PR's code (including the test domains) for the same pattern.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
fodzal: Done: `GraphStructureAndDirection` now carries a `@Nullable Class<?> cascadedElementClass` directly.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
fodzal: Done: replaced with `getListVariableDescriptor()` and a null check.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
fodzal: Added the explicit direction check.

> Note: the mixed case was already rejected, if only implicitly — a `PREVIOUS` source's first reference is a
> `@PreviousElementShadowVariable` and a `NEXT` source's is a `@NextElementShadowVariable`, so they can never be
> the same metamodel and `!parentMetaModel.equals(...)` already returned `null`. The explicit check is clearer,
> so it's in.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/GraphStructure.java
fodzal: Done

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/DeclarativeShadowVariableAssertions.java
fodzal: Good catch, that made the fixed-point assertion compare the cascade against itself.

The recomputation is now done in two passes:

1. the entity-based `SolutionManager.updateShadowVariables(Class, Object...)` re-derives the built-in shadow
   variables, which are outside the graph by construction;
2. a graph built with `GraphStructure.ARBITRARY` forced explicitly recomputes every declarative variable from
   scratch, overwriting whatever the first pass produced for them.

So whichever graph `SolutionManager` picks internally (today or in the future) the last pass alone decides the
reference values, independently of the graph under test.

core/src/test/java/ai/timefold/solver/core/testdomain/shadow/multi_entity_chain/TestdataMultiEntityChainVehicle.java
fodzal: Yes, the test domain now represents head vehicles with a `null` collection, so the tests exercise it.

While at it: this comment was wrong (the supplier *is* triggered at graph construction) and the
pre-initialization it justified turned out to be unnecessary, so both are removed, here and in the other test
domains that copied the pattern.

core/src/test/java/ai/timefold/solver/core/testdomain/shadow/multi_entity_chain/TestdataMultiEntityChainVisit.java
fodzal: Clarified the comment. Two situations, illustrated on chained vehicles `A -> B`, where `B` reads `A`'s `endTime`
through its own `previousEndTime`:

### During an incremental update: once per element supplier

Say a move appends a visit to `A` while `B` holds `[b1, b2]`.

- The cascade walks `A`'s chain, then one topological pass of the inner graph propagates
  `A.endTime -> B.previousEndTime -> B.endTime`.
- Since `B.previousEndTime` changed, `B` is flagged for a whole-chain walk; but that walk is deferred until the
  flagging pass is over, so `b1` and `b2` are each computed **exactly once**, with the settled `previousEndTime`.
- The one exception is `B.endTime` itself: the same topological pass already recomputed it through the
  `previousEndTime -> endTime` edge, before `B`'s chain was re-walked, and it is recomputed once more after the
  walk — **two calls**, a bounded residual cost per flagged entity.

`ListElementCascadeVariableReferenceGraphTest` pins exactly this scenario (`calledCount == 1` for the visits,
`== 2` for `B.endTime`), and the class javadoc of `ListElementCascadeVariableReferenceGraph` now spells it out.

### During the initial computation: up to one extra pass per chain level

At construction everything starts from `null`, so with `A -> B -> C`:

- the first pass computes `A`'s chain, but `B.previousEndTime` is still `null`. `A.endTime` was not known when it
  was computed, and that transient `null` is exactly what this comment and the `(Integer)` cast guard against;
- the next pass settles `B`, then `C`.

So `B`'s visits are computed twice at construction and `C`'s three times: a one-off bootstrap cost proportional to
the chain depth, not a per-move cost, and also asserted in that test.

The values are correct before the first move either way; the session constructor iterates to the fixed point when
the score director is created, and the `FULL_ASSERT` solve tests would catch any pre-move corruption.

fodzal:
Thanks a lot for the review; very helpful, and sorry for the draft confusion. A few commits since:

- **Review feedback**:  everything from the inline comments above, plus the two things they uncovered: the test domains no longer pre-initialize their shadow variables and now use null fact collections for head vehicles, and the fixed-point assertion recomputes with a forced arbitrary graph instead of comparing the cascade against itself.
- **Post-chain marking, once per list change**: it was marked on both the before and after list-change events; only the latter is needed. New test for the case that makes it necessary: unassigning a vehicle's only visit still recomputes its `endTime`, and shifts the vehicles downstream, even though no element is left to walk.
- **Dependency loop through a planning variable**: new test domain where the vehicles chain by a `@PlanningVariable`, so the inner graph has dynamic edges and can go inconsistent; a loop marks the whole route inconsistent while the visits stay cascaded.
- **Test naming**: what I called the cascade's "rebasing" elements are really its pre-chain readers.
- **Merged `main`** — the branch was behind #2630, which changed the `VariableReferenceGraph`
  contract, so the cascade had to follow. `updateChanged()` now returns a boolean: when the inner
  graph gives up on a structurally flawed solution it leaves its variables stale, so walking the
  remaining chains would only spread that staleness; the cascade gives up too and puts its pending
  chain walks back, since the caller retries the update after undoing the move. `getVariableLoops()`
  delegates to the inner graph: the elements are only excluded from it when their sources form a
  chain fed by the owner's pre-chain variables, so they can never be part of a strongly connected
  component themselves, and the cascade reports exactly the loops the arbitrary graph would.

Marking it ready for review.

triceo:
@fodzal We are preparing to merge a very large PR, which will likely cause some (small) conflicts with yours. Once we're done, I'll ask you for a final rebase, we will do a final review, and hopefully merge this not too long from now.

Thanks for the patience!

fodzal:
@triceo Ok sure, no problem!

In fact, I still have small things to do on this PR (a small fix + some improvements... and maybe an alternative way of doing by directly adding the "cascade block" in the graph, that could be better).
So, I convert the PR to draft for now :)

PR #2549, opened 2026-07-31
feat: more specific problem scale

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/api/solver/ProblemSizeStatistics.java
```diff
@@ -17,8 +18,10 @@
  */
 @NullMarked
 public record ProblemSizeStatistics(long entityCount,
```
triceo, line 23: Seeing as this is a public API, we should add a backwards-compatible constructor, and deprecate it. 
Maybe even a migration script, this one will be easy.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeStatistics.java
```diff
@@ -94,6 +114,9 @@ SolutionInitializationStatistics computeInitializationStatistics(@Nullable Consu
             var entityDescriptor = solutionDescriptor.findEntityDescriptorOrFail(entity.getClass());
             if (entityDescriptor.isGenuine()) {
                 genuineEntityCount.increment();
+                // Should we include immovable entities for the count?
+                // Problem scale ignores them.
```
triceo: That's a good question.
IMO we should include them in entity count, but they cannot affect the problem scale, so they don't.

Maybe we log that the entity count includes immovable as well? Not sure. We should definitely log not just "entity count" but "genuine entity count".

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -220,6 +220,22 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
                     problemSizeStatistics.entityCount(), problemSizeStatistics.variableCount(),
                     problemSizeStatistics.approximateValueCount(),
                     problemSizeStatistics.approximateProblemScaleAsFormattedString());
+            if (logger.isDebugEnabled()) {
+                var genuineEntityClassCountEntries = problemSizeStatistics.genuineEntityClassToEntityCount().entrySet();
+                for (var genuineEntityCountEntry : genuineEntityClassCountEntries) {
+                    var geninueEntityClass = genuineEntityCountEntry.getKey();
+                    logger.debug("    Entity ({}) count: {}",
```
triceo: `LOGGER`

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -220,6 +220,22 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
+                            geninueEntityClass.getCanonicalName(),
+                            genuineEntityCountEntry.getValue());
+                    for (var geninueVariableEntry : problemSizeStatistics
+                            .genuineEntityClassToVariableToValueCount()
+                            .get(geninueEntityClass).entrySet()) {
+                        var genuineVariable = geninueVariableEntry.getKey();
+                        logger.debug("        Variable ({}) estimated value count: {}",
```
triceo: Maybe we also want to show the variable's value type? Not sure.
I would certainly mention if basic/list.

triceo (commented):
Small comments.

core/src/main/java/ai/timefold/solver/core/api/solver/ProblemSizeStatistics.java
Christopher-Chianelli: TBH, users shouldn't ever construct this type (maybe should of been an interface).

core/src/main/java/ai/timefold/solver/core/api/solver/ProblemSizeStatistics.java
```diff
@@ -17,10 +21,25 @@
...
+        SequencedMap<Class<?>, Long> genuineEntityClassToEntityCount,
         long variableCount,
         long approximateValueCount,
+        SequencedMap<Class<?>, SequencedMap<String, Long>> genuineEntityClassToVariableToValueCount,
         double approximateProblemSizeLog) {
 
+    /**
+     * @deprecated this type isn't meant to be constructed by a user.
+     */
+    @Deprecated(since = "2.4.0", forRemoval = true)
```
triceo:
```suggestion
    @Deprecated(since = "2.6.0", forRemoval = true)
```

That is the earliest release this can be merged into.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeStatistics.java
```diff
@@ -61,6 +65,18 @@ SolutionInitializationStatistics computeInitializationStatistics(@Nullable Consu
...
         var maxValueRangeSize = new MutableLong(0L);
 
+        cachedEntityCountByEntityOrdinal = new long[solutionDescriptor.getEntityDescriptors().size()];
+        cachedValueCountByEntityAndVariableOrdinal = new long[cachedEntityCountByEntityOrdinal.length][];
+        for (var entityDescriptor : solutionDescriptor.getEntityDescriptors()) {
+            // TimefoldTestResourceTest in Quarkus has an entity class with a basic variable
+            // that was not considered a genuine entity.
+            // (i.e. solutionDescriptor.getGenuineEntityDescriptors() did not have it).
+            // Are subclass entities that do not add new basic/list variable not considered
+            // genuine entities?
```
triceo: Subclassing entities is valid, as long as it respects the documented inheritance rules. Since the parent entity is genuine, this is by definition also a genuine entity. But arguably we should not count abstract entities, only their non-abstract children.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeStatistics.java
```diff
@@ -152,10 +179,30 @@ public ProblemSizeStatistics getProblemSizeStatistics() {
             computeInitializationStatistics(null, false);
         }
         if (cachedProblemSizeStatistics == null) {
+            var entityClassToEntityCount = new LinkedHashMap<Class<?>, Long>();
+            var entityClassToVariableToValueCount = new LinkedHashMap<Class<?>, SequencedMap<String, Long>>();
```
triceo: If we can merge the `Long` and `SequencedMap` into a simple wrapper type, we can merge this into one map instance. (Not that it matters much for a one-time computation.)

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeStatistics.java
```diff
@@ -152,10 +179,30 @@ public ProblemSizeStatistics getProblemSizeStatistics() {
...
+                    variableToValueCount.put(variableDescriptor.getVariableName(),
+                            cachedValueCountByEntityAndVariableOrdinal[entityDescriptor.getOrdinal()][variableDescriptor
+                                    .getOrdinal()]);
+                }
+                if (entityDescriptor.hasAnyListVariables()) {
+                    variableToValueCount.put(entityDescriptor.getListVariableDescriptor().getVariableName(),
+                            cachedValueCountByEntityAndVariableOrdinal[entityDescriptor.getOrdinal()][entityDescriptor
+                                    .getListVariableDescriptor().getOrdinal()]);
+                }
+                entityClassToVariableToValueCount.put(entityDescriptor.getEntityClass(), variableToValueCount);
```
triceo: This is difficult to read. (Also similar cases above.) I would consider extracting a method that does this - shorter line, clearer meaning. (The only argument could be the entity descriptor.)

triceo (approved):
Small comments + Sonar rating is worse than B.
Other than that, LGTM.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ValueRangeStatistics.java
Christopher-Chianelli: I much rather not create a new public API type for this (these maps go into `ProblemSizeStatistics`, so if we create a new wrapper type, that creates a new public API).

triceo:
The rating is still a C.

Christopher-Chianelli:
Should be fixed now; if it still produces C, something is seriously wrong with Sonar's `null` analysis.

PR #2551, opened 2026-07-31
ci: do not publish Javadocs outside of public API

triceo:


rsynek (approved):
`**/api/**/*.java` should work for solver-service modules as well.

PR #2552, opened 2026-07-31
feat: emit failed solution even if there is no solverJob

diogodanielsoaresferreira:
Fixes: https://github.com/TimefoldAI/timefold-platform/issues/4984

### DIFF
--- a/service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
+++ b/service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
@@ -770,12 +770,15 @@ public class SolverWorker {
 
             // update run status only as failed
             metadata = storageService.getMetadata(problemId);
-            metadata.updateStatusOnFailure(throwable.getMessage());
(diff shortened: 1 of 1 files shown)


service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -770,12 +770,15 @@ public void notifyOnFailure(Object id, Throwable throwable) {
 
             // update run status only as failed
             metadata = storageService.getMetadata(problemId);
-            metadata.updateStatusOnFailure(throwable.getMessage());
-            storageService.storeMetadata(problemId, metadata);
+            if (metadata == null) {
```
rsynek: How it can happen that the metadata is `null`?

diogodanielsoaresferreira: That should not happen, removed. Thanks!

rsynek (approved):

PR #2559, opened 2026-08-03
feat: Maven plugin: support locally-stored PAT as fallback to TIMEFOLD_PAT env var

oemebamo:
This is a proposal to improve https://github.com/TimefoldAI/timefold-solver-enterprise/issues/806

### DIFF
--- a/service/tools/maven-plugin/README.adoc
+++ b/service/tools/maven-plugin/README.adoc
@@ -17,7 +17,7 @@ The plugin uses a goal prefix of `timefold` (see the plugin configuration in the
 === Configure goal (timefold:configure)
 
 - Fetches platform identity/config by calling GET /api/platform/v1/aboutme?includeConfig=true
-- Requires a platform personal access token available via environment variable `TIMEFOLD_PAT` (see "Authentication" below)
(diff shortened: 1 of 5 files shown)


mswiderski:
I would prefer to avoid storing secrets (PAT in this case) in clear text in user's home directory. Since we are talking about maven plugin I would expect we follow standard practice of maven https://maven.apache.org/guides/mini/guide-encryption.html that way we could rely on exactly the same mechanism as all others plugins that require secrets to communicate with remote service. Plus this can also take advantage of encryption to not keep it clear text.

oemebamo:
> I would prefer to avoid storing secrets (PAT in this case) in clear text in user's home directory. Since we are talking about maven plugin I would expect we follow standard practice of maven https://maven.apache.org/guides/mini/guide-encryption.html that way we could rely on exactly the same mechanism as all others plugins that require secrets to communicate with remote service. Plus this can also take advantage of encryption to not keep it clear text.

excellent; thanks for the pointer - didn't know about this standard practice

oemebamo:
closing in favor of the other PR from @cristianonicolai

PR #2564, opened 2026-08-04
docs: Document that PHASE_ASSERT makes dev mode drastically slower

oemebamo:
(Based on feedback by @cedricdhooge96, or their Claude.)

closes https://github.com/TimefoldAI/timefold-solver-enterprise/issues/813


triceo (changes requested):
Factually untrue.
`PHASE_ASSERT` is the default mode, and it doesn't slow down anything. `FAST_ASSERT` and `FULL_ASSERT` do, as documented here:

https://docs.timefold.ai/timefold-solver/latest/running-timefold-solver/solver-diagnostics#environmentModeAvailableModes

oemebamo:
🫡

PR #2565, opened 2026-08-04
docs: /demo-data/{id} response shape undocumented (service module)

oemebamo:
would love @TomCools and @rsynek's feedback

closes https://github.com/TimefoldAI/timefold-solver-enterprise/issues/814


winklerm (approved):
Looks good to me, thanks!

PR #2567, opened 2026-08-04
fix(bavet): close stale-activity race across all nodes

triceo:
A downstream filtering node could read an out-tuple whose
state still reported active even though it was one flush away from being
retracted, because Bavet processes its node network in strict layers and a
shallow sibling can re-test a tuple before the layer that would retract it
has run. A filtering predicate dereferencing a stale shadow variable then
could crash with an NPE.

Each producing node now records the input tuple(s) its out-tuple was
derived from and staleness guards walk that chain instead of trusting bare tuple state.


triceo:
Replaced by https://github.com/TimefoldAI/timefold-solver/pull/2576

PR #2510, opened 2026-07-16
feat: enhance the visualization metadata

rsynek:
Fixes https://github.com/TimefoldAI/timefold-solver-enterprise/issues/656.

PR #2568, opened 2026-08-04
docs: Document the model visualization contract (service-module + platform layers)

oemebamo:
closes https://github.com/TimefoldAI/timefold-solver-enterprise/issues/815


docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
```diff
@@ -0,0 +1,97 @@
...
+== Error reporting
+
+The platform automatically injects a small error-forwarding script into the served HTML.
+This script turns uncaught JavaScript errors and unhandled promise rejections into a `postMessage` call, which the platform surfaces to the user as an alert in its own UI.
+
+Don't rely solely on `console.error` to signal failure: uncaught errors and unhandled promise rejections are what actually surface to the end user.
+Anything you only log to the console stays invisible to them.
+
+[#_announcing_visualization_pages]
+== Announcing visualization pages
```
cedricdhooge96, line 79: It's the first time I'm seeing this. Do you happen to know which model we use this for? Most model visualisations currently depend on a single page with multiple tabs.

cedricdhooge96: Answering my own question after digging in:

**TL;DR:** no model uses this, and nothing reads it either. Org-wide, `timefold.model.visualization.pages` appears only in the solver's own config class and test. On the platform side the field stops at the descriptor: `ModelDescriptorDTO`/`RegisteredCatalogEntry` never carry `visualizationPages`, and the frontend renders one hardcoded iframe gated only on `uiSupport != "NONE"`. It shipped in #2510 to the shape the platform side requested (timefold-solver-enterprise#656), ahead of any consumer. The issue behind this PR (timefold-solver-enterprise#815) explicitly asked for that caveat to ship with this section ("say plainly it has no visible effect on the platform today"), but the section instead states the platform "knows what to offer users". Please add the caveat back.

**How the pieces fit** (the reason "single page with multiple tabs" is what we all see today): platform-drawn tabs need two things. First, knowing which pages exist, ideally before the iframe loads. That is this metadata: the menu card. Second, actually switching pages inside the iframe without reloading it, and knowing which page is active. That is a runtime `postMessage` channel (a `views` announce plus `navigate`), which already ships in the rebuilt FSR visualization but is not documented here: https://github.com/TimefoldAI/timefold-visualisations/blob/main/docs/EMBEDDING.md. Today the menu card is printed but nobody reads it, and the waiter works but is not on this page. Every existing model therefore still draws its own tabs inside the frame.

**One more concrete fix for this page:** the example icon `TbChartGantt` does not exist in `react-icons` 5.7.0, the version the platform frontend pins, so the doc's own example would render no icon once consumption lands (measured against the package; `TbMap` is fine). Worth picking an icon that resolves, and the "any icon name from Tabler Icons" sentence oversells slightly.

We are wiring the two halves together rather than keeping two systems, and this metadata wins the argument: the embed no longer declares its views at all (TimefoldAI/timefold-visualisations#87 removed the upward `views` message), so this build-time declaration is the **single source** of the tab set, and the runtime channel carries only what a running embed can know — readiness, the active view, and navigation. Consumption is tracked in TimefoldAI/timefold-platform#5274 (models API) and TimefoldAI/frontend#3991 (run-page menu), and the model-side declarations in TimefoldAI/timefold-field-service-routing#1385 / TimefoldAI/timefold-employee-scheduling#1387. When a model and its visualization live in the same repository, the properties file and the page keys it must match sit side by side, which is the intended end state.

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
```diff
@@ -0,0 +1,97 @@
...
+The contract described here, including iframe sizing, refresh behavior, asset paths, and page-announcement metadata, is still evolving and may change as the platform's visualization support matures, possibly without a smooth migration path.
+====
+
+[#_building_the_ui]
+== Building the UI
+
+The UI itself is the same set of static files described in xref:running-timefold-solver/service/visualization.adoc[], placed under `src/main/resources/META-INF/resources`.
+Once deployed, the platform repackages and serves these files under a `ui/` prefix, so the entry point the platform loads must be exactly `ui/index.html`.
+
+Asset paths under this prefix are single-segment only today: `ui/main.js` is servable, but `ui/assets/main.js` is not.
```
cedricdhooge96: This is already fixed, however path references in the `.html` file need to be relative, not absolute since visualisations don't run under the root domain.

cedricdhooge96: e.g. not `/assets/...` but `./assets/...`

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
```diff
@@ -0,0 +1,97 @@
...
+== How the platform embeds the UI
+
+The platform renders `ui/index.html` inside a fixed-size iframe, sized by the platform's own layout, with no auto-grow or resize mechanism.
+
+Design your UI for a fixed viewport, or handle any overflow or scrolling yourself within that fixed area.
+
+[#_calling_your_api_from_the_iframe]
+== Calling your model's API from inside the iframe
+
+The platform injects the following query parameters into the iframe's `src` URL:
```
TomCools, line 36: This is no longer the main case. Yes it works, but the new one is by using messages between the frames. @oemebamo, want me to take this PR to the finish line?

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
```diff
@@ -0,0 +1,97 @@
...
+----
+timefold.model.visualization.pages[0].key=map
+timefold.model.visualization.pages[0].icon=TbMap
+timefold.model.visualization.pages[0].label=Map
+timefold.model.visualization.pages[1].key=gantt
+timefold.model.visualization.pages[1].icon=TbChartGantt
+timefold.model.visualization.pages[1].label=Gantt chart
+----
+
+Each declared page has three required fields; omitting any of them fails the build.
```
TomCools, line 92: How does this help with routing? Are pages a new concept introduced with the specific front-end components? Is the KEY also a tab in the UI?

TomCools (changes requested):
See other comments

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
TomCools: Is this required now? Or if this is not supplied, does it use "visualization" for whatever is in the UI folder?

oemebamo:
Feedback from @muellerd that's useful to incorporate here:

> [Daniel Mueller](mailto:daniel@timefold.ai) at 2026-08-20 10:12 AM When working on the visualization, I was not sure what is feasible for the platform to handle/show. For the household pto planner, I wanted to be able to show detailed information if you click on a day of the calendar. I was thinking about either a popup or something like a sidebar where the information can be shown. Due to the fact that I didn’t know if a sidebar was possible, I went for the popup.

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
oemebamo: considering resolved, as sub-tickets are created

docs/src/modules/ROOT/pages/deploying-to-platform/visualization.adoc
oemebamo: handled in https://github.com/TimefoldAI/timefold-solver/pull/2568/commits

PR #2569, opened 2026-08-04
chore: return null when the solution is invalid in calculate score

Christopher-Chianelli:
Note: `DefaultPhaseCommandContext.executeTemporary` throws if the score is null/invalid (via `Objects.requireNonNull(score, () -> "The move (%s) failed to calculate a score.".formatted(move));`). Should that be change (i.e. what should happen when a user executes a temporary move that results in an inconsistent solution in a PhaseCommand).

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/decider/ConstructionHeuristicDecider.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/decider/ConstructionHeuristicDecider.java
@@ -159,7 +159,8 @@ public class ConstructionHeuristicDecider<Solution_> {
         }
         if (isLoggingEnabled()) {
             logger.trace("{}        Move index ({}), score ({}), move ({}).",
-                    logIndentation, moveScope.getMoveIndex(), moveScope.getScore().raw(), moveScope.getMove());
+                    logIndentation, moveScope.getMoveIndex(),
+                    (score != null) ? moveScope.getScore().raw() : "invalid", moveScope.getMove());
         }
     }
 
--- a/core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AbstractAcceptor.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AbstractAcceptor.java
@@ -1,6 +1,7 @@
 package ai.timefold.solver.core.impl.localsearch.decider.acceptor;
 
 import ai.timefold.solver.core.impl.localsearch.event.LocalSearchPhaseLifecycleListenerAdapter;
+import ai.timefold.solver.core.impl.localsearch.scope.LocalSearchMoveScope;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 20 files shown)


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/decider/ConstructionHeuristicDecider.java
```diff
@@ -159,7 +159,8 @@ protected void doMove(ConstructionHeuristicMoveScope<Solution_> moveScope) {
         }
         if (isLoggingEnabled()) {
             logger.trace("{}        Move index ({}), score ({}), move ({}).",
-                    logIndentation, moveScope.getMoveIndex(), moveScope.getScore().raw(), moveScope.getMove());
+                    logIndentation, moveScope.getMoveIndex(),
+                    (score != null) ? moveScope.getScore().raw() : "invalid", moveScope.getMove());
```
zepfred, line 163:
```
                    (moveScope.getScore() != null) ? moveScope.getScore().raw() : "invalid", moveScope.getMove());
```

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AbstractAcceptor.java
```diff
@@ -19,4 +20,13 @@ public abstract class AbstractAcceptor<Solution_> extends LocalSearchPhaseLifecy
...
     // ************************************************************************
 
+    public final boolean isAccepted(LocalSearchMoveScope<Solution_> moveScope) {
+        if (moveScope.getScore() == null) {
+            return false;
+        }
+        return isConsistentSolutionAccepted(moveScope);
+    }
+
+    public abstract boolean isConsistentSolutionAccepted(LocalSearchMoveScope<Solution_> moveScope);
```
zepfred, line 30: The `public` modifier may not be ideal, as this method can still be called instead of `isAccepted`. Can we change it to `protected` to prevent this?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -246,13 +246,12 @@ public NeighborhoodNotifier<Solution_> getNeighborhoodNotifier() {
     public abstract InnerScore<Score_> innerCalculateScore();
 
     @Override
+    @Nullable
     public final InnerScore<Score_> calculateScore() {
         if (lastVariableUpdateSuccessful) {
             return innerCalculateScore();
         } else {
-            var invalidScore = InnerScore.invalid(getScoreDefinition().getZeroScore());
```
zepfred: Chris, I'm not convinced that the new solution is better than the old approach. I believe returning a non-null value is preferable. What is the primary reason for changing this behavior?

zepfred (commented):
Chris, I have a few questions regarding the proposed approach.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AbstractAcceptor.java
Christopher-Chianelli: When I tried `package-private`, was getting compile errors since some implementations are in different packages. I can try `protected` though.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: @triceo wanted it, so any places where it needs to be checked become obvious. We might find a different way though.

Christopher-Chianelli:
Closed in favour of https://github.com/TimefoldAI/timefold-solver/pull/2583

PR #2545, opened 2026-07-30
chore(service): unify configuration properties

rsynek:
Fixes https://github.com/TimefoldAI/timefold-solver/issues/2496.

Renames configuration properties by the following logic:

- user-defined properties are prefixed with "ai.timefold.model"
- properties primarily injected by the platform are prefixed with "ai.timefold.platform" (this is not a clear split since user can override them e.g. for local testing)
- keep "ai.timefold.storage" as a separate category (mostly injected by the platform)

Some duplicate properties removed, `timefold.application.version` renamed to `ai.timefold.model.api-version` to better reflect what it is.

Related: https://github.com/TimefoldAI/timefold-solver-enterprise/pull/795

TODO:

- [x] update docs when changes are approved

PR #2570, opened 2026-08-05
feat: add support for moveCountLimit

cristianonicolai:
Related to  https://github.com/TimefoldAI/timefold-platform/issues/5271

### DIFF
--- a/docs/src/modules/ROOT/pages/running-timefold-solver/service/consumer-guide.adoc
+++ b/docs/src/modules/ROOT/pages/running-timefold-solver/service/consumer-guide.adoc
@@ -140,15 +140,22 @@ The `run` configuration controls how long the solver runs:
 - `spentLimit` sets the maximum duration (in ISO 8601 duration format) for solving a dataset.
 
 - `unimprovedSpentLimit` sets the maximum duration (in ISO 8601 duration format) for solving a dataset since the dataset score improved.
-If no value is provided, the default diminished returns termination will apply. If set, stepCountLimit must be empty.
+If no value is provided, the default diminished returns termination will apply. If set, stepCountLimit and moveCountLimit must be empty.
 _Warning_: using this option will disable the default diminished returns termination which is recommended for most use cases
 
 - `stepCountLimit` sets the maximum solver step count for solving a dataset. The solver will stop solving after a pre-determined amount of steps. Use when you require results independently of the hardware resources performance. Use this termination if you want to benchmark your models, not recommended for production use.
 If set, unimprovedSpentLimit must be empty. Warning: using this option will disable the default diminished returns termination which is recommended for most use cases.
 
-- `slidingWindowDuration` sets the sliding window (in ISO 8601 duration format) over which score improvement is measured by the diminished returns termination. Defaults to PT30S when omitted. Only takes effect when diminished returns termination is active (i.e. unimprovedSpentLimit and stepCountLimit are both empty).
+- `moveCountLimit` sets the maximum number of moves evaluated by the solver for solving a dataset.
+The solver will stop solving after a pre-determined amount of moves have been evaluated.
+Use when you require results independently of the hardware resources performance.
+Use this termination if you want to benchmark your models, not recommended for production use.
+If set, unimprovedSpentLimit must be empty. Warning: using this option will disable the default diminished returns termination which is recommended for most use cases.
+Can be combined with `stepCountLimit`: the solver stops as soon as either limit is reached.
 
-- `minimumImprovementRatio` sets the minimum ratio between current and initial improvement before the diminished returns termination kicks in. Must be strictly positive. Defaults to 0.0001 when omitted. Only takes effect when diminished returns termination is active (i.e. unimprovedSpentLimit and stepCountLimit are both empty).
+- `slidingWindowDuration` sets the sliding window (in ISO 8601 duration format) over which score improvement is measured by the diminished returns termination. Defaults to PT30S when omitted. Only takes effect when diminished returns termination is active (i.e. unimprovedSpentLimit, stepCountLimit and moveCountLimit are all empty).
+
+- `minimumImprovementRatio` sets the minimum ratio between current and initial improvement before the diminished returns termination kicks in. Must be strictly positive. Defaults to 0.0001 when omitted. Only takes effect when diminished returns termination is active (i.e. unimprovedSpentLimit, stepCountLimit and moveCountLimit are all empty).
 
 === Response
 
--- a/service/test-model/src/build/openapi.json
+++ b/service/test-model/src/build/openapi.json
@@ -2234,7 +2234,7 @@
             "example" : "PT1H"
           },
           "unimprovedSpentLimit" : {
-            "description" : "Maximum unimproved score duration (ISO 8601 duration format). If the score has not improved during this period (e.g. PT5M), terminate the solver. If no value is provided, the default diminished returns termination will apply. If set, stepCountLimit must be empty. Warning: using this option will disable the default diminished returns termination which is recommended for most use cases.",
+            "description" : "Maximum unimproved score duration (ISO 8601 duration format). If the score has not improved during this period (e.g. PT5M), terminate the solver. If no value is provided, the default diminished returns termination will apply. If set, stepCountLimit and moveCountLimit must be empty. Warning: using this option will disable the default diminished returns termination which is recommended for most use cases.",
             "type" : "string",
             "allOf" : [ {
               "$ref" : "#/components/schemas/Duration"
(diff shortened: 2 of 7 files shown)


service/test-model/src/build/openapi.json
```diff
@@ -8,7 +8,7 @@
       "url" : "",
       "email" : ""
     },
-    "version" : "v2-beta"
+    "version" : "v1"
```
zepfred, line 11: Are these changes correct?

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationService.java
```diff
@@ -73,9 +78,19 @@ private TerminationConfig solverTerminationConfig(Duration spentLimit, Duration
         if (unimprovedSpentLimit != null) {
             terminationConfig.withUnimprovedSpentLimit(unimprovedSpentLimit);
             LOGGER.info("Using time spent ({}) with unimproved time spent ({}) termination.", spentLimit, unimprovedSpentLimit);
-        } else if (stepCountLimit != null) {
-            terminationConfig.withStepCountLimit(stepCountLimit);
-            LOGGER.info("Using time spent ({}) with step count limit ({}) termination.", spentLimit, stepCountLimit);
+        } else if (stepCountLimit != null || moveCountLimit != null) {
+            // stepCountLimit and moveCountLimit are both hard, deterministic caps used for benchmarking;
+            // they are OR-composed and may be combined so whichever is reached first terminates the solver.
+            List<String> limits = new ArrayList<>(2);
```
zepfred, line 84: var limits = new ArrayList<>(2);

zepfred (approved):
LGTM! Thank you for implementing the feature so quickly!

service/test-model/src/build/openapi.json
cristianonicolai: yes, but not related to this pr, it only adds the new fields. It will be removed once https://github.com/TimefoldAI/timefold-solver/pull/2571 is merged.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationService.java
cristianonicolai: that would break the `String.join("` or ", limits)`

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationConfigParams.java
```diff
@@ -8,6 +8,7 @@ public class TerminationConfigParams {
     public static final String TERMINATION_MAXIMUM_UNIMPROVED_SPENT_LIMIT =
             "ai.timefold.platform.termination.maximum-unimproved-spent-limit";
     public static final String TERMINATION_STEP_COUNT_LIMIT = "ai.timefold.platform.termination.step-count-limit";
+    public static final String TERMINATION_MOVE_COUNT_LIMIT = "ai.timefold.platform.termination.move-count-limit";
```
mswiderski: isn't this being renamed to avoid platform name in the properties? https://github.com/TimefoldAI/timefold-solver/pull/2545/changes#diff-63b1e917bd1fd3d3dd458d719056346ef3d51d23086b99b84b326ee7a0bf163cR5

cristianonicolai: renamed

docs/src/modules/ROOT/pages/running-timefold-solver/service/consumer-guide.adoc
```diff
@@ -140,15 +140,19 @@ The `run` configuration controls how long the solver runs:
...
 
 - `stepCountLimit` sets the maximum solver step count for solving a dataset. The solver will stop solving after a pre-determined amount of steps. Use when you require results independently of the hardware resources performance. Use this termination if you want to benchmark your models, not recommended for production use.
 If set, unimprovedSpentLimit must be empty. Warning: using this option will disable the default diminished returns termination which is recommended for most use cases.
 
-- `slidingWindowDuration` sets the sliding window (in ISO 8601 duration format) over which score improvement is measured by the diminished returns termination. Defaults to PT30S when omitted. Only takes effect when diminished returns termination is active (i.e. unimprovedSpentLimit and stepCountLimit are both empty).
+- `moveCountLimit` sets the maximum number of moves evaluated by the solver for solving a dataset. The solver will stop solving after a pre-determined amount of moves have been evaluated. Use when you require results independently of the hardware resources performance. Use this termination if you want to benchmark your models, not recommended for production use.
```
winklerm: Please start the new sentence on a new line, thanks!

winklerm: PP: It would be nice to highlight the difference between `stepCountLimit` and `moveCountLimit`.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationConfigParams.java
```diff
@@ -8,6 +8,7 @@ public class TerminationConfigParams {
     public static final String TERMINATION_MAXIMUM_UNIMPROVED_SPENT_LIMIT =
             "ai.timefold.platform.termination.maximum-unimproved-spent-limit";
     public static final String TERMINATION_STEP_COUNT_LIMIT = "ai.timefold.platform.termination.step-count-limit";
+    public static final String TERMINATION_MOVE_COUNT_LIMIT = "timefold.model.termination.move-count-limit";
```
winklerm, line 11: Q: The property name is already in the new format introduced by https://github.com/TimefoldAI/timefold-solver/pull/2545, that's why it differs, right?

winklerm (approved):
Thanks, looks good!

I left a few nitpicks (up to you to address those), but good to merge after resolving the first comment.

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationConfigParams.java
cristianonicolai: correct

docs/src/modules/ROOT/pages/running-timefold-solver/service/consumer-guide.adoc
cristianonicolai: done

cristianonicolai: maybe a link to somewhere in the docs that could explain that?

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/termination/TerminationConfigParams.java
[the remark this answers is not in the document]
cristianonicolai: intentional since property prefix are changing.

mswiderski (commented):
as mentioned on the seed PR we should stop relying on env variables and be more based on the configuration object as this makes it more reproducible

rsynek (approved):

rsynek:
@cristianonicolai could you please resolve the conflicts?

As for environment variables vs. configuration storage, can we do it as a separate refactoring for all these config properties? Perhaps worth having a quick chat before we do.

cristianonicolai:
@rsynek addressed all comments and conflicts. Please merge with https://github.com/TimefoldAI/timefold-quickstarts/pull/1147
Yes, lets discuss env/config changes in another scope.

cristianonicolai:
@rsynek broken compatibility is due to `minimum = "1"` that was added for both stepCount and moveCount. Would that be ok, or we only validate it at runtime?

rsynek:
@triceo I checked solver-core src and we allow 0 for both step and move count limit terminations. Is that intentional? Why would the solver terminate after 0 steps or even 0 moves?

triceo:
Some time ago this was specifically requested by the platform. Just enough time ago that I no longer remember what this was for anymore.

rsynek:
@mswiderski any chance you remember the reason (see the last comment)?

rsynek:
If we are not aware of a good reason to support 0, let's keep the minimum = 1.

mswiderski:
> @mswiderski any chance you remember the reason (see the last comment)?

somehow I think it was related to some customers that want to end the run very fast like only CH phase or something similar, might be worth to check with Luciana as I think this came from her at the very beginning.

rsynek:
That's actually a good explanation. If someone wants only the initialized solution and does not want to wait any further, setting `stepCountLimit=0` simply skipts the LS completely.

rsynek (approved):

PR #2576, opened 2026-08-10
fix(bavet): prevent stale join+ifExists filtering on all cardinalities

triceo:
Introduces a small performance regression of 4 % in one case; barely statistically significant.
The turtle tests have not uncovered any score corruptions as a result of these changes.

Includes changes to groupBy which are not strictly necessary, as they are left over from a previous uncommitted version of this fix which required them; I decided to keep these changes because they nicely simplify the constructor machinery and bring it in line with what joins and ifExists do.


triceo:
@Christopher-Chianelli Please review. I recommend against reviewing commit-by-commit. 

The fix significantly evolved over time and many of the things introduced in the first commits were removed in the later ones; reviewing those changes would effectively be wasting time.

Christopher-Chianelli (approved):
LGTM; I will create a branch from this one for an alternative join solution

PR #2580, opened 2026-08-11
feat: Maven plugin progress upload for model descriptor

cristianonicolai:
Related to https://github.com/TimefoldAI/timefold-solver-enterprise/issues/807

Sample output:
```
[INFO] Uploading model descriptor archive (24.2 KiB), this can take a while on a slow connection
[INFO] Uploaded 24.2 KiB, waiting for the Timefold Platform to process it (0 s elapsed)
[INFO] Model descriptor archive transfer finished: 24.2 KiB in 0 s
```

### DIFF
--- a/service/tools/maven-plugin/README.adoc
+++ b/service/tools/maven-plugin/README.adoc
@@ -34,6 +34,7 @@ The plugin uses a goal prefix of `timefold` (see the plugin configuration in the
 - If the platform responds with 409 (conflict) and `timefold.model.overwrite=true`, the plugin will PATCH `/api/platform/v1/models/{registrationKey}` to update the existing model
 - The plugin reads the model descriptor to obtain name/id by extracting `timefold-model-descriptor.json` from the archive (or reading `target/timefold/timefold-model-descriptor.json` when present)
 - The platform validates the model version (`timefold.application.version`) on registration. It must be a release version `vN` (e.g. `v1`), a git commit hash (1–40 lowercase hex characters), or the literal `SNAPSHOT`; any other value is rejected with HTTP 400.
+- While the request is in flight the plugin reports progress. Archives of at least 1 MiB are announced up front, then at most one throttled line every 5 seconds reports the percentage uploaded, then a heartbeat reports that the upload is done and the platform is still processing, and finally a single line reports the total transferred and elapsed time. A normal deploy of an archive of a few tens of kilobytes completes before the first heartbeat and produces none of these lines. Note that `mvn -q` suppresses `[INFO]`, so it hides all progress output.
 
 === Undeploy goal (timefold:undeploy)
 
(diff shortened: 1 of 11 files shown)


mswiderski:
I believe the one that takes time is container image push and not the descriptor zip so not sure we really need that as it does seem like a lot of code for little value if at all.

cristianonicolai:
@mswiderski Indeed this is not the main issue, but didnt we see descriptor that were quite some MBs?

mswiderski:
> @mswiderski Indeed this is not the main issue, but didnt we see descriptor that were quite some MBs?

yes, there are model descriptors of 3-4 mb but compared to container image size it's nothing and if someone is really on that slow network she will never reach the point where descriptor is the issue. Just saying that the value here is questionable in my opinion.

cristianonicolai:
@mswiderski I agree in that case, I was under the impression we already seen descriptors with almost 100Mb, thats why. Happy to drop this then.

mswiderski:
> @mswiderski I agree in that case, I was under the impression we already seen descriptors with almost 100Mb, thats why. Happy to drop this then.

100MB won't fly as we have limits on the platform so that would be rejected anyway. Maybe we could actually have this limit client side too?

let's not drop it yet, I just expressed single opinion so let's see others.

service/tools/maven-plugin/README.adoc
```diff
@@ -146,6 +147,12 @@ Goals are implemented as Mojos in `src/main/java/ai/timefold/solver/tools/maven`
 - `ConfigureMojo` — writes build properties
 - `DeployModelMojo` — handles upload/register/patch flow
 - `UndeployModelMojo` — handles deletion
-- `AbstractPlatformModelMojo` — common behavior, HTTP client, descriptor reading
+- `AbstractPlatformModelMojo` — common behavior, HTTP client, descriptor reading, and `sendWithProgress`
+
+Upload progress reporting lives in `src/main/java/ai/timefold/solver/tools/maven/http`:
```
winklerm, line 154: Not sure if this is relevant to the Maven plugin user, looks to me rather as implementation details that could be just in the javadoc.

winklerm (commented):
I agree with @mswiderski that the model descriptor upload is probably not the biggest part of the upload, so the value is a little bit questionable.

On the other hand, it might be good to see some progress, especially in case when there is an issue with the artifact repository the descriptor is uploaded to.

For that reason, I am slightly more in favor of merging this than dropping it. But let's see what the others think.

Wondering if we could get a progress report also from quarkus when it is uploading the image, but that's probably another story.

cristianonicolai:
@winklerm I've reported it upstream to Quarkus and provided a PR, hopefully will get accepted. See https://github.com/quarkusio/quarkus/issues/55958

service/tools/maven-plugin/README.adoc
cristianonicolai: Added there since we already cover some other details for plugin developer.

rsynek:
+1 to what Maciej and Marek said: quite some code to maintain to show a progress-bar for a process that usually takes seconds.

I like the idea with showing the docker image upload progress by Quarkus, plus checking the descriptor size limit on the client side rather than waiting for the platform to refuse it later.

Alternatively, can we wrap the the entire process and report the progress ourselves? Perhaps hide all the logging by our dependencies by default and show it only in case of debugging or errors.

Motivation: better, unified UX; the user does not have to care about what's JIB uploading the docker image.

As for this PR, unless anybody sees a clear value, I am for closing it.

cristianonicolai:
Closing this PR since most agree is not worth it. There is still Jib progress PR submitted to Quarkus that should land in a few releases that should improve the UX. 
I wouldnt go further than that to be honest, creating yet another wrapper, since Jib is the actualy slow bit in most scenarios anyway.

PR #2581, opened 2026-08-11
feat: Ensure enterprise profile is active on Maven deploy plugin execution

cristianonicolai:
…ution

Fixes: https://github.com/TimefoldAI/timefold-platform/issues/5267

### DIFF
--- a/service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
+++ b/service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
@@ -18,16 +18,19 @@ import java.util.Properties;
 
 import ai.timefold.solver.tools.maven.client.PlatformIdentityInfo;
 
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.model.Parent;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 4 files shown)


service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
```diff
@@ -40,6 +40,13 @@ public class ConfigureMojo extends AbstractPlatformModelMojo {
 
     protected static final String PROP_MODEL_CONFIG_SKIP = "timefold.model.configuration.skip";
 
+    protected static final String PROP_MODEL_ENTERPRISE_CHECK_SKIP = "timefold.model.enterprise.check.skip";
+
+    /**
+     * Property set by the {@code enterprise} profile of {@code timefold-solver-service-parent}.
+     */
+    private static final String ENTERPRISE_PROPERTY = "ai.timefold.solver.enterprise";
```
mswiderski: this is very easy to be bypassed - why not looking at either parent pom declaration or dependencies themselves?

mswiderski (commented):
I would not allow to skip enterprise check as this can only lead to garbage being pushed to platform container registry - model images that won't execute. So if you attempt to deploy to platform it must be enterprise edition based model.

service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
cristianonicolai: refactored to validate both parent pom and enterprise artefacts in the classpath.

mswiderski (approved):

winklerm (approved):
Thanks, looks good to me!

PR #2582, opened 2026-08-11
feat(neighborhoods): iterator-based paradigm

triceo:
The branch adds a new, preview-stage way to declare custom move providers for local search; less streamy than Neighborhoods currently and more akin to move selectors on account of the exposure of iterators, but still maintaining the key winning parts: Bavet-based caching and just-in-time indexing.

## New: custom move-building via a raw iterator

Alongside the existing declarative style (`pick()` + joiners), `MoveStreamFactory` now has a lower-level escape hatch for neighborhoods that aren't expressible as a simple join — custom scoring, greedy/windowed pairing, or combining three or more datasets by hand:

- `.asCachedDataset()` turns a `UniEnumeratingStream`/`BiEnumeratingStream` into a  `UniDataset`/`BiDataset` handle.
- `factory.buildMoveStream(MoveIteratorProvider)` takes your own `Iterator<Move>` logic and resolves those handles, at solve time, into `UniDatasetInstance`/`BiDatasetInstance` objects exposing `size()` and `iterator(RandomGenerator)`.
- A registered `UniDataset` can still be joined against another dataset (`UniDataset#join(...)`, same joiners as `pick()`), computed just-in-time rather than pre-materialized.

## Clarified contract: move order is never guaranteed

The API now explicitly commits to producing moves in random order only — there is no deterministic/original-order variant, and this must never be relied on. Consistently with that:

- `NeighborhoodTester` dropped `getMovesAsList()` (which implied "all moves, in order") in favor of membership assertions: `producesAllOf(Move...)` / `producesNoneOf(Move...)`, each repeatedly drawing moves up to an iteration limit and comparing by `equals()`. The limit is overridable per-assertion via `.within(iterationLimit)`.
- `getMovesAsIterator()`/`getMovesAsStream()` are now documented as returning a bounded, non-exhaustive sample — useful for inspection, not for completeness checks.
- The `Move` interface's javadoc now spells out that `equals()`/`hashCode()` must distinguish move instances and only equate moves producing the exact same resulting solution — this is now required for the new test assertions as well as tabu search.

## Bias elimination

Previous code often suffered from selection bias, especially in cases of bi-selection. This is now removed and proved by test coverage, in some cases at the expense of performance. If/when real-world scenarios show performance issues, we will optimize.

## New restriction

Configuring the Neighborhoods API together with `LocalSearchType.VARIABLE_NEIGHBORHOOD_DESCENT` now fails fast with a clear error at solver build time, and this is documented as a known limitation. This is a consequence of no longer allowing fully exhaustive move providers; we will probably restore support later, but at this stage of the API, I don't even want to think about this niche corner case.


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultRetiringRandomIterator.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Keeps an exact track of which items were already retired,
+ * so that no item is ever returned twice unless it is picked again before being retired.
+ * It accepts a list of unique items on input, and does not copy or modify it.
+ *
+ * @param <T>
+ */
+@NullMarked
+final class DefaultRetiringRandomIterator<T> implements RetiringRandomIterator<T> {
```
zepfred: I'm not sure if I like the term "retiring", but I don't have a better name for it right now.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -472,20 +474,100 @@ public T next() {
         }
     }
 
-    private final class RandomIterator extends DefaultIterator {
+    final class RandomIterator extends DefaultIterator implements UniqueRandomIterator<T> {
```
zepfred: Is this class necessary? It seems that `DefaultIterator` could simply implement `UniqueRandomIterator`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
```diff
@@ -472,20 +474,100 @@ public T next() {
...
+    final class RepeatingIterator implements RepeatingRandomIterator<T> {
+
+        private final List<Iterator<T>> downstreamIteratorList = new ArrayList<>();
+        private final int[] distribution;
+        private final int distributionSum;
+        private final RandomGenerator workingRandom;
+
+        RepeatingIterator(Object queryCompositeKey, RandomGenerator workingRandom) {
+            this.workingRandom = workingRandom;
+            var sizeList = new ArrayList<Integer>();
```
zepfred, line 560: Should we initialize `sizeList` with `comparisonMap.size()`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainedInIndexer.java
```diff
@@ -194,20 +210,67 @@ public T next() {
 
     }
 
-    private final class RandomIterator extends DefaultIterator {
+    final class RandomIterator extends DefaultIterator implements UniqueRandomIterator<T> {
```
zepfred: Again, is this class necessary? It seems that `DefaultIterator` could simply implement `UniqueRandomIterator`.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
zepfred: I'm not sure I understand how this class randomizes selections. It appears to be an original iterator to me.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/JustInTimeBiDatasetInstance.java
```diff
@@ -0,0 +1,379 @@
...
+        public @Nullable B next() {
+            if (!hasNext()) {
+                throw new NoSuchElementException();
+            }
+            return tupleIterator.next().getA();
+        }
+
+        @Override
+        public void forEachRemaining(Consumer<? super @Nullable B> action) {
+            tupleIterator.forEachRemaining(tuple -> action.accept(tuple.getA()));
```
zepfred, line 136: The generic type B is confusing for me. We actually want to return `tuple.getA()`, correct?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/JustInTimeBiDatasetInstance.java
```diff
@@ -0,0 +1,379 @@
...
+        private @Nullable UniTuple<A> currentLeftTuple;
+        private @Nullable UniTuple<B> currentRightTuple;
+
+        private RepeatingRandomBiIterator(AbstractLeftDatasetInstance<Solution_, UniTuple<A>> leftDatasetInstance,
+                UniRightDatasetInstance<Solution_, A, B> rightDatasetInstance, SolutionView<Solution_> solutionView,
+                RandomGenerator workingRandom) {
+            this.rightDatasetInstance = rightDatasetInstance;
+            this.solutionView = solutionView;
+            this.workingRandom = workingRandom;
+            this.leftTupleIterator = leftDatasetInstance.retiringRandomIterator(workingRandom);
```
zepfred, line 240: Should we retire the left tuples?

zepfred: The current strategy appears to retrieve and retire a left tuple, then create a right iterator, fetch the right tuple, and disregard the right iterator. Will this function as a repeating random iterator?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/AbstractRightDatasetInstance.java
```diff
@@ -69,17 +72,45 @@ public void retract(UniTuple<Right_> tuple) {
...
+    }
+
+    /**
+     * As defined by {@link #uniqueRandomIterator(Object, RandomGenerator)},
+     * but only returning elements matching the given predicate.
+     */
+    public Iterator<UniTuple<Right_>> uniqueRandomIterator(Object compositeKey, RandomGenerator workingRandom,
             Predicate<UniTuple<Right_>> predicate) {
-        return indexer.randomIterator(compositeKey, workingRandom, predicate);
+        return new FilteringIterator<>(indexer.uniqueRandomIterator(compositeKey, workingRandom), predicate);
```
zepfred, line 132: No bailout?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -31,108 +30,70 @@
...
-    private final Iterator<UniTuple<A>> leftTupleIterator;
+    private final RetiringRandomIterator<UniTuple<A>> leftTupleIterator;
     private @Nullable Move<Solution_> nextMove;
 
     public BiRandomMoveIterator(BiMoveStreamContext<Solution_, A, B> context, RandomGenerator workingRandom) {
         this.context = Objects.requireNonNull(context);
         this.workingRandom = Objects.requireNonNull(workingRandom);
         this.leftTupleIterator = context.getLeftDatasetInstance()
-                .randomIterator(workingRandom);
+                .retiringRandomIterator(workingRandom);
```
zepfred, line 90: This pattern is similar to one of my previous comments. Do we want to retire left tuples if the iterator is never-ending?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/FilteringIterator.java
```diff
@@ -3,69 +3,80 @@
...
-    private final SolutionView<Solution_> solutionView;
-    private final BiNeighborhoodsPredicate<Solution_, A, B> filter;
-    private final UniTuple<A> leftTuple;
-    private final Iterator<UniTuple<B>> rightTupleIterator;
+    private final Iterator<T> delegate;
+    private final Predicate<T> filter;
+    private final long bailOutSize;
 
-    // Required for iteration.
     private boolean hasNext = false;
```
zepfred, line 49: There is no need for the variable `hasNext` since we can rely solely on `next`.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/NeighborhoodsBasedMoveRepository.java
```diff
@@ -102,11 +100,11 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
 
     @Override
     public Iterator<Move<Solution_>> iterator() {
-        if (random) {
-            return new RandomOrderNeighborhoodIterator<>(moveIterableList, Objects.requireNonNull(workingRandom));
-        } else {
-            return new OriginalOrderNeighborhoodIterator<>(moveIterableList);
-        }
+        return iterator(Objects.requireNonNull(workingRandom));
```
zepfred, line 103: Should we consider failing instead?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexer.java
triceo: If profiling shows a problem, I will do that.
Right now, I would not blindly do this, because the list will very likely be considerably smaller than the full map, based on the key.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/JustInTimeBiDatasetInstance.java
triceo: No. For a collection of pairs of A+B, this iterator will return all known Bs for any given A. Returning B is the intended behavior here.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/JustInTimeBiDatasetInstance.java
triceo: This only retires left tuples when there is no right tuple available for it. And that will only happen if all of them are filtered out - in that case, having A really makes no sense. In all other cases, A will not be retired, and the iterator is fully repeating. 

You could argue that this complexity could be removed and empty B would just result in another pick of A. I am open to it, although it feels like that would only make move selection take longer.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/enumerating/common/AbstractRightDatasetInstance.java
triceo: The filtering iterator has its own bailout inside.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
triceo: As explained above.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/FilteringIterator.java
triceo: I disagree. `next` can independently be `null` - `hasNext` exists to distinguish "null because that is the return" from "null because no more values".

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/NeighborhoodsBasedMoveRepository.java
triceo: This _will_ fail on null; nothing else to do here.

triceo:
@zepfred I have pushed a whole bunch of new code. I recommend not reading individual commits - one builds on another, and much of the code you would read is removed/changed in subsequent commits.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultRepeatingRandomIterator.java
```diff
@@ -0,0 +1,49 @@
...
+ * It accepts a list of items on input, and does not copy or modify it.
+ *
+ * @param <T>
+ */
+@NullMarked
+final class DefaultRepeatingRandomIterator<T> implements RepeatingRandomIterator<T> {
+
+    private static final DefaultRepeatingRandomIterator<Object> EMPTY =
+            new DefaultRepeatingRandomIterator<>(new ElementAwareArrayList<>(), () -> 0);
+
```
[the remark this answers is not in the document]
triceo: Clearly wrong, because it compiles.

triceo:
Taking this back to a draft. I introduced new tests locally, and found new bugs.

triceo:
I think I've got them all now. Ready for review.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/MultiBucketUniqueRandomIterator.java
```diff
@@ -0,0 +1,70 @@
...
+    public boolean hasNext() {
+        return distributionSum > 0;
+    }
+
+    @Override
+    public T next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        var selectedIndex = RandomUtils.sampleWithDistribution(workingRandom, distributionSum, distribution);
```
zepfred, line 64: I see it is fairest now. Cool!

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/ComparisonIndexerTest.java
```diff
@@ -218,6 +223,111 @@ void boundaryComparisonHandlesExtremeCompareToWithoutOverflow() {
...
+        indexer.put(10, newTuple("age10"));
+        indexer.put(20, newTuple("age20"));
+
+        var iterator = indexer.uniqueRandomIterator(5, new Random(0));
+        assertThat(iterator.hasNext()).isFalse();
+        assertThatExceptionOfType(NoSuchElementException.class).isThrownBy(iterator::next);
+    }
+
+    @Test
+    void randomIteratorEmptyRange() {
```
zepfred, line 305: I don't see a test for the happy path of never-ending random iterators. Should we consider adding one?

zepfred: If so, please double-check the other test classes, as some of them, such as `ContainedInIndexerTest`, already includes it.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/ContainingAnyOfIndexer.java
zepfred: `HashSet.newHashSet()`?

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultRetiringRandomIterator.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public boolean hasNext() {
+        if (activeCount <= 0) {
+            return false;
+        }
+        if (nextSlot != -1) {
+            return true;
+        }
+        nextSlot = workingRandom.nextInt(activeCount);
+        next = source.get(resolveSlot(nextSlot));
```
zepfred, line 89: Does this mean that every time we select a retired element, we return the corresponding last active element? I can't relate the logic to the Fisher-Yates shuffle strategy. Could you please explain it to me?

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/JustInTimeBiDatasetInstance.java
zepfred: Okay. I see it now.

zepfred: My main concern was retiring the left tuple, which could cause the iterator to stop behaving like a never-ending iterator. Now I understand it will only retire if there is no right tuple.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
```diff
@@ -31,108 +32,113 @@
...
         var solutionView = context.neighborhoodSession().getSolutionView();
-        return rightDatasetInstance.randomIterator(compositeKey, workingRandom,
-                rightTuple -> filter.test(solutionView, leftTuple.getA(), rightTuple.getA()));
+        var leftFact = leftTuple.getA();
+        // Random draws with replacement can never prove that no right tuple matches;
+        // bail out after many consecutive rejections.
+        // RetiringBiWalk.advance() retries this call up to PROBE_ATTEMPT_COUNT times before retiring the left,
+        // since a single bail-out is a false negative, not proof of emptiness.
+        var bailOutSize = rightDatasetInstance.size(compositeKey) * FilteringIterator.BAIL_OUT_SAFETY_MULTIPLIER;
+        return new FilteringIterator<>(rightTupleIterator,
```
zepfred, line 129: ~~Is there a risk of retiring a left tuple if the random selection does not choose the correct index within the specified bailout size?~~

zepfred: My previous question was addressed in your comment, and it got me thinking about whether we should retire tuples when working with never-ending iterators. We have the option to bail out, which allows us to stop trying to fetch an element. I believe we should only retire the left tuples if the right iterator is empty.

zepfred: In move generation, this could prevent the solver from attempting some potentially good moves in a step.

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultUniqueRandomIteratorTest.java
```diff
@@ -9,88 +9,49 @@
...
-
-import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.Test;
 
+/**
+ * Unlike {@link DefaultRetiringRandomIteratorTest}, this iterator needs no cooperation from the
+ * caller: it retires every element it returns by itself, so it ends on its own once every element has
+ * been drawn exactly once.
+ */
 class DefaultUniqueRandomIteratorTest {
```
zepfred, line 20: Let’s increase the coverage of this class.

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/EqualIndexerTest.java
```diff
@@ -128,7 +129,7 @@ void iteratorNoMatchingKey() {
     void randomIteratorEmpty() {
         Indexer<UniTuple<String>> indexer = new IndexerFactory<>(neighborhoodsJoiner).buildIndexer(true);
         var random = new Random(0);
-        var iterator = indexer.randomIterator(CompositeKey.ofMany("F", 40), random);
+        var iterator = indexer.uniqueRandomIterator(CompositeKey.ofMany("F", 40), random);
```
zepfred, line 132: Are these changes correct? I would expect to keep all tests related to `RepeatingRandomIterator` and add new ones for `RetiringRandomIterator`.

core/src/test/java/ai/timefold/solver/core/impl/bavet/common/index/EqualsAndComparisonIndexerTest.java
```diff
@@ -178,7 +179,7 @@ void iteratorDifferentGender() {
     void randomIteratorEmpty() {
         Indexer<UniTuple<String>> indexer = new IndexerFactory<>(neighborhoodsJoiner).buildIndexer(true);
         var random = new Random(0);
-        var iterator = indexer.randomIterator(CompositeKey.ofMany("F", 40), random);
+        var iterator = indexer.uniqueRandomIterator(CompositeKey.ofMany("F", 40), random);
```
zepfred, line 182: Same comment as `EqualIndexerTest`.

core/src/test/java/ai/timefold/solver/core/impl/neighborhood/bias/DatasetBucketBiasIT.java
```diff
@@ -0,0 +1,101 @@
...
+        probe.setValue(valueList.getFirst());
+        var solution = new TestdataSolution("solution");
+        solution.setEntityList(List.of(probe));
+        solution.setValueList(valueList);
+
+        var session = session(moveStreamFactory, solution);
+        var instance = session.getInstance(justInTimeDataset);
+
+        var root = new Random(0);
+        tally("just-in-time per-A unique, draw #" + drawIndex, TRIAL_COUNT, trial -> {
```
zepfred, line 55: Cool!

core/src/test/java/ai/timefold/solver/core/impl/neighborhood/bias/DatasetBucketBiasIT.java
```diff
@@ -0,0 +1,101 @@
...
+import ai.timefold.solver.core.preview.api.neighborhood.stream.joiner.BiNeighborhoodsJoiner;
+import ai.timefold.solver.core.preview.api.neighborhood.stream.joiner.NeighborhoodsJoiners;
+import ai.timefold.solver.core.testdomain.TestdataEntity;
+import ai.timefold.solver.core.testdomain.TestdataSolution;
+import ai.timefold.solver.core.testdomain.TestdataValue;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class DatasetBucketBiasIT extends AbstractBiasIT {
```
zepfred, line 21: The idea of validating the statistical distribution is quite clever!

docs/src/modules/ROOT/pages/optimization-algorithms/neighborhoods.adoc
```diff
@@ -667,17 +727,25 @@ The API requires:
 3. **Entity classes**: One or more classes annotated with `@PlanningEntity`
 4. **Solution instance**: The working solution to execute the move on
 
-The API returns all moves that the move provider can generate for the given solution,
-allowing you to write assertions to verify that the generated moves are correct.
+Because the order in which a move provider generates moves is never part of the Move Streams contract,
```
zepfred: "Move Streams" or "Neighborhood"?

zepfred (approved):
I reviewed the new code, and these new features look great to me. It unblocks the Neighborhoods for additional use cases while keeping the benefits of the CS feature. I have few more questions and suggestions, but overall, you have addressed my previous comments and concerns.

I'm not entirely sure about the reasoning behind removing determinism in move generation. I wonder if this will be an issue we need to revisit in the near future. Currently, it appears that we cannot use Neighborhood for methods like CH or metaheuristics, which rely on evaluating the entire neighborhood.

Additionally, I would like to understand the main challenge of not being able to maintain determinism, which led to the design of the current solution.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/BiRandomMoveIterator.java
triceo: You are right, and I am open to making that change.
Currently code exists there which makes sure this retirement only happens when the likelihood of such move still existing is super-low. (Notice the bailout of `10x size` further multiplied by three.)

Although this is not certainty, it is very close to it and I doubt any code will actually get to exercise this. The code will be simplified if we remove this and simply never retire, but I am not sure it is the better approach - iterating endlessly over something that never finds anything.

triceo:
I have made changes to the PR - specifically, I have increased coverage where asked, and I have also changed the APIs slightly following up on feedback from some other people.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultRetiringRandomIterator.java
triceo: I have added comments directly to the code, as explaining to everyone there is better than explaining to one person here. Let me know if this helped.

triceo:
> I'm not entirely sure about the reasoning behind removing determinism in move generation. I wonder if this will be an issue we need to revisit in the near future. Currently, it appears that we cannot use Neighborhood for methods like CH or metaheuristics, which rely on evaluating the entire neighborhood.

Neighborhoods was never ment for CH. We need a completely different approach to CH - because currently every model is using custom phases instead of our CH. Our CH is not flexible enough.

Wrt. evaluating entire neighborhoods - I do not believe that to be practically possible. Either the neighborhood is so small that it can be fully evaluated and therefore you might as well just use brute force, or the neighborhood is too large and therefore there is no need to support it.

I am leaving this out for now, because I do not want to complicate the API right now for things which we may need long way down the line. If it ever turns out we need this, we will figure it out then.

core/src/main/java/ai/timefold/solver/core/impl/bavet/common/index/DefaultRetiringRandomIterator.java
zepfred: Makes sense to me now

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/CachedBiDatasetInstance.java
```diff
@@ -0,0 +1,124 @@
...
+            this.tupleIterator = tupleIterator;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return tupleIterator.hasNext();
+        }
+
+        @Override
+        public void next() {
```
zepfred, line 84: Revisiting this code makes me consider returning the `BiTuple` directly and extracting `a` and `b` from it. I don't remember seeing this pattern in other iterator classes.

core/src/test/java/ai/timefold/solver/core/preview/api/neighborhood/example/SwapEntityTailsMoveProvider.java
```diff
@@ -0,0 +1,83 @@
...
+            var pairInstance = session.getInstance(entityPairs);
+            if (pairInstance.size() == 0) {
+                return Collections.emptyIterator(); // Fewer than 2 entities.
+            }
+            var destinationInstance = session.getInstance(entityDestinations);
+            var solutionView = session.getSolutionView();
+
+            var moveList = new ArrayList<Move<TestdataListSolution>>();
+            var pairIterator = pairInstance.iterator();
+            while (pairIterator.hasNext()) {
```
zepfred, line 62: The providers `RelocateValueBlockMoveProvider`, `SwapEntityTailsMoveProvider`, and `SwapValuesInCodeRangeMoveProvider` have the same pattern and don't use the random generator. I'm not sure if this is intentional, but these providers are generating all moves upfront and returning a complete list of moves without any randomization.

core/src/main/java/ai/timefold/solver/core/impl/neighborhood/stream/dataset/CachedBiDatasetInstance.java
triceo: I will never expose `BiTuple` as public API.
And returning a new instance of some other pair type would create significant allocation pressure, affecting performance.
Therefore this is the only way.

core/src/test/java/ai/timefold/solver/core/preview/api/neighborhood/example/SwapEntityTailsMoveProvider.java
triceo: Yes, I noticed that earlier today while we were reviewing the code with Marek and Radek. The moves were refactored, and the original iterator is no longer available.

core/src/main/java/ai/timefold/solver/core/preview/api/neighborhood/stream/enumerating/UniEnumeratingStream.java
```diff
@@ -476,4 +481,18 @@ <GroupKey_, Result_> BiEnumeratingStream<Solution_, GroupKey_, Result_> groupBy(
...
+     * Use this instead of {@link MoveStreamFactory#pick(UniEnumeratingStream)}
+     * when the dataset is to be consumed from a custom {@link MoveIteratorProvider}.
+     * Resolve the returned handle against a {@link MoveIteratorSession}
+     * inside {@link MoveStreamFactory#buildMoveStream(MoveIteratorProvider)}.
+     * <p>
+     * Repeated calls on the same stream return an equal handle, and the rows are materialized only once.
+     *
+     * @see UniSamplingStream For the declarative alternative, which reads from this stream directly.
+     */
+    UniDataset<Solution_, A> asCachedDataset();
```
rsynek, line 496: Perhaps worth mentioning that after this call all following operations are non-incremental?

PR #2583, opened 2026-08-11
chore: introduce structural component to score

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
```diff
@@ -22,9 +22,14 @@
  * @see Score
  */
 @NullMarked
-public record BendableBigDecimalScore(BigDecimal[] hardScores,
+public record BendableBigDecimalScore(long structuralScore, BigDecimal[] hardScores,
```
triceo, line 26: Maybe `int`? (Probably irrelevant.)

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
```diff
@@ -22,9 +22,14 @@
...
         BigDecimal[] softScores) implements IBendableScore<BendableBigDecimalScore> {
 
+    public BendableBigDecimalScore(BigDecimal[] hardScores,
```
triceo, line 29: Let's deprecate these convenience constructors to make it clear that the canonical constructor is still the way to go.

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
```diff
@@ -274,6 +282,9 @@ public boolean equals(Object o) {
                     || softLevelsSize() != other.softLevelsSize()) {
                 return false;
             }
+            if (structuralScore != other.structuralScore) {
```
triceo: I would start with this. It is the first component, and if this one is different, nothing else matters anyway.

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
```diff
@@ -323,6 +337,9 @@ public String toShortString() {
 
     @Override
     public String toString() {
+        if (structuralScore < 0) {
+            return "invalid";
```
triceo: I'd much rather spell this out fully. (`-Nstructural/...`)
"invalid" says nothing, if you see it written out in a random place. Whereas the full string will tell you that this is a score.

core/src/main/java/ai/timefold/solver/core/api/score/BendableScore.java
```diff
@@ -270,12 +281,15 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(Arrays.hashCode(hardScores), Arrays.hashCode(softScores));
+        return Objects.hash(structuralScore, Arrays.hashCode(hardScores), Arrays.hashCode(softScores));
```
triceo: I know this is status quo, but I would consider manually spelling this out. `Objects.hash(...)` creates arrays on the hot path.

core/src/main/java/ai/timefold/solver/core/api/score/HardMediumSoftBigDecimalScore.java
```diff
@@ -218,7 +230,8 @@ public String toShortString() {
 
     @Override
     public String toString() {
-        return hardScore + HARD_LABEL + "/" + mediumScore + MEDIUM_LABEL + "/" + softScore + SOFT_LABEL;
+        return (structuralScore < 0) ? "invalid"
+                : hardScore + HARD_LABEL + "/" + mediumScore + MEDIUM_LABEL + "/" + softScore + SOFT_LABEL;
```
triceo: Maybe `formatted(...)` here? I never liked these long string concats.

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
```diff
@@ -19,6 +19,13 @@ public sealed interface Score<Score_ extends Score<Score_>>
         extends Comparable<Score_>
         permits HardMediumSoftBigDecimalScore, HardMediumSoftScore, HardSoftBigDecimalScore, HardSoftScore, IBendableScore,
         SimpleBigDecimalScore, SimpleScore {
+    /**
+     * The structural component of a Score. 0 if the solution is structurally sound
+     * (i.e. no inconsistent variables), -1 otherwise.
```
triceo: I would specifically mention that this is effectively a superhard score - it is the most important component of any score.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AbstractAcceptor.java
```diff
@@ -18,5 +19,14 @@ public abstract class AbstractAcceptor<Solution_> extends LocalSearchPhaseLifecy
     // ************************************************************************
     // Worker methods
     // ************************************************************************
+    @Override
+    public final boolean isAccepted(LocalSearchMoveScope<Solution_> moveScope) {
+        if (moveScope.getScore().isInvalid()) {
```
triceo, line 24: Should it be `isInvalid` or rather `isStructurallyFlawed`?

core/src/main/java/ai/timefold/solver/core/impl/score/definition/ScoreDefinition.java
```diff
@@ -47,6 +50,13 @@ public interface ScoreDefinition<Score_ extends Score<Score_>> {
      */
     Class<Score_> getScoreClass();
 
+    /**
+     * The score that represents a solution that is structurally invalid.
+     *
+     * @return never null
+     */
+    Score_ getInvalidStructureScore();
```
triceo: `getStructurallyFlawedScore`?

I'm not sure how long this pattern will hold though. If (when) we add more things to structural score, the value will no longer be just -1, and therefore there will not be just one such score.

I don't know if this future will come, but I would prepare for it by not publishing `public static final INVALID` on the public score types.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScore.java
```diff
@@ -51,11 +47,14 @@ public boolean isFullyAssigned() {
     }
 
     public boolean isInvalid() {
-        return unassignedCount == Integer.MAX_VALUE;
+        return raw.structuralScore() < 0;
```
triceo, line 50: We spoke yesterday about having fail-fasts for when people unknowingly access an invalid score. I do not see them (yet).

core/src/main/java/ai/timefold/solver/core/impl/score/ScoreUtil.java
```diff
@@ -94,6 +94,9 @@ public static <Score_ extends Score<Score_>> String buildShortString(Score<Score
             String... levelLabels) {
         var shortString = new StringBuilder();
         var i = 0;
+        if (score.structuralScore() != 0L) {
+            shortString.append(score.structuralScore()).append("struct");
```
triceo:
```suggestion
            shortString.append(score.structuralScore()).append("structural");
```

IMO no need for the shortening. Also when I see `struct`, I smell `C`.

triceo (commented):
Leaving comments. Comments on any score type should be considered to apply to every other score type as well - I didn't want to duplicate every comment 10 times.

Score documentation needs updating; we need to explain what is a structural score, what it's for, and how it's handled. (If you prefer, can wait for a later PR.)

Consider adding migration recipes as much as you can. Going forward, we will try to make every change to happen automatically. However, automated or not, please update the recipes in the docs to mention all the changes.

Also, I would like to see tests added; structural score is an important part of the score types, we should test for it. The existing coverage should be easily extendable.

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
Christopher-Chianelli: debated, but choose long since we no longer use int for score types, and it odd to see `int` when everything else is `long`.

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
Christopher-Chianelli: Not sure about that; the solver sets the structural score, not score calculation, so for the vast majority, if not all, cases, users will do `Score(0L, ...)`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScore.java
Christopher-Chianelli: The only place a user could encounter one would be in score analysis and maybe custom phases; so those changes would likely be in enterprise as users cannot touch `InnerScore`.

core/src/main/java/ai/timefold/solver/core/api/score/HardMediumSoftBigDecimalScore.java
```diff
@@ -218,7 +238,10 @@ public String toShortString() {
 
     @Override
     public String toString() {
-        return hardScore + HARD_LABEL + "/" + mediumScore + MEDIUM_LABEL + "/" + softScore + SOFT_LABEL;
+        return (structuralScore < 0)
+                ? "%dstructural/%s%s/%s%s/%s%s".formatted(structuralScore, hardScore, HARD_LABEL, mediumScore, MEDIUM_LABEL,
```
triceo: We have String constants for hard/medium/soft, let's introduce a constant for "structural" as well.

core/src/main/java/ai/timefold/solver/core/api/score/SimpleScore.java
```diff
@@ -106,7 +114,7 @@ public String toShortString() {
 
     @Override
     public String toString() {
-        return Long.toString(score);
+        return (structuralScore < 0) ? "%dstructural/%d".formatted(structuralScore, score) : Long.toString(score);
```
triceo: Similarly here.

core/src/main/java/ai/timefold/solver/core/impl/score/ScoreUtil.java
```diff
@@ -111,8 +121,13 @@ public static <Score_ extends Score<Score_>> String buildShortString(Score<Score
     }
 
     public static String[][] parseBendableScoreTokens(Class<? extends IBendableScore<?>> scoreClass, String scoreString) {
-        var scoreTokens = new String[2][];
+        var scoreTokens = new String[3][];
         var startIndex = 0;
+        var structuralSlashIndex = scoreString.indexOf("structural/");
```
triceo: And here.

core/src/main/java/ai/timefold/solver/core/impl/score/ScoreUtil.java
```diff
@@ -111,8 +121,13 @@ public static <Score_ extends Score<Score_>> String buildShortString(Score<Score
...
+        if (structuralSlashIndex >= 0) {
+            scoreTokens[0] = new String[] { scoreString.substring(0, structuralSlashIndex) };
+            startIndex = structuralSlashIndex + "structural/".length();
```
triceo: And here.

triceo (approved):
LGTM, with comments.
If you want to leave docs for later, that's your call, but in that case I recommend keeping a list of things to document; it's growing long.

core/src/main/java/ai/timefold/solver/core/api/score/BendableBigDecimalScore.java
triceo: Does the user need to create a score? Should they? IMO no.
I'd even go as far as to write in the Javadoc that the constructor is not considered public API, as the user has no need to create `Score` instances.

Christopher-Chianelli:
I would leave docs for later to not leave the PR open for too long.

triceo:
One last thing I just remembered - if my memory serves, we have a whole bunch of score (de)serializers in the persistence module, and in Quarkus. Let's make sure they work with the structural score too.

PR #2586, opened 2026-08-12
feat: Add configured cores and memory to solver worker event

diogodanielsoaresferreira:
Related with https://github.com/TimefoldAI/timefold-platform/pull/5343


cristianonicolai (approved):

mswiderski (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/events/SolverWorkerEvent.java
```diff
@@ -20,14 +20,25 @@ public abstract sealed class SolverWorkerEvent extends AbstractDatasetEvent perm
 
     private final String eventProducerId;
 
+    private final Integer configuredCores;
+
+    private final Long configuredMemoryMi;
```
rsynek: Given we configure the memory in mebibytes, isn't Integer enough? If a model needs more memory than that, we have a bigger problem.

rsynek:
@diogodanielsoaresferreira Why do we need this piece of information for the `FinalBestSolutionEvent` and the `FailedSolutionEvent` ?

The previous attributes we needed to send to Insights, the `solveRequested` and `resolvedMapLocation`, we included in the `DatasetComputedEvent`, because it's the first event sent that can have all the necessary information.

How are the configured memory and CPU cores different?

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
```diff
@@ -827,6 +831,29 @@ protected ModelOutput loadModelOutput(String id) {
...
+    private Configuration<?> getConfiguration(String id) {
+        try {
+            return storageService.getConfiguration(id);
+        } catch (Exception e) {
+            LOGGER.warn("Unable to load configuration for id {} to report configured resources: {}", id, e.getMessage());
+            return null;
+        }
+    }
+
+    private Integer configuredCores(Configuration<?> configuration) {
```
rsynek: A fair point from SonarCloud; this and the other method can be static.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/events/SolverWorkerEvent.java
diogodanielsoaresferreira: True, changed to integer

service/worker/src/main/java/ai/timefold/solver/service/worker/impl/SolverWorker.java
diogodanielsoaresferreira: Updated methods to be static

diogodanielsoaresferreira:
@rsynek makes sense. our idea was to replace in-place the run previous cores and memory that was added in solver-enterprise, but it makes more sense to add to dataset computed, since it's defined before the run starts. Refactored 👍

rsynek (approved):
Thank you!

I will merge once the CI passed; the downstream builds were failing due to the solver-enterprise feature branch not being rebased on the latest main.

PR #2589, opened 2026-08-13
chore: Validate accountId can be resolved

cristianonicolai:


service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
```diff
@@ -159,6 +158,45 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+     */
+    protected String resolveAccountId(PlatformIdentityInfo info) throws MojoFailureException {
+        String configuredAccountId = getPropertyOrParameter(PROP_ACCOUNT_ID, this.accountId);
+        if (configuredAccountId != null && !configuredAccountId.isBlank()) {
+            return configuredAccountId.trim();
+        }
+
+        Set<String> accountIds = info.accountIds() == null ? Set.of() : info.accountIds();
+        if (accountIds.size() == 1) {
+            return accountIds.iterator().next();
```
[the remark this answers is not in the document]
cristianonicolai: its ok to handle both cases the same way.

mswiderski (commented):
actually this PR might be better to rename `accountId` to `namespace` I don't think any of our projects used it explicitly so should be safe to do so

cristianonicolai:
> actually this PR might be better to rename `accountId` to `namespace` I don't think any of our projects used it explicitly so should be safe to do so

namespace migration will be handled with a follow up fix https://github.com/TimefoldAI/timefold-platform/issues/5360

mswiderski (approved):

service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
```diff
@@ -100,12 +102,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 if (info == null || !info.hasPushAccessRights()) {
                     throw new RuntimeException("No access to deploy model on Timefold Platform");
                 }
-                String accountId = getPropertyOrParameter(PROP_ACCOUNT_ID, this.accountId);
-                if (accountId == null && info.accountIds().size() == 1) {
-                    accountId = info.accountIds().iterator().next();
-                }
+                String accountId = resolveAccountId(info);
```
rsynek: To avoid hiding the field:

```suggestion
                var resolvedAccountId = resolveAccountId(info);
```

service/tools/maven-plugin/src/test/java/ai/timefold/solver/tools/maven/ConfigureMojoTest.java
```diff
@@ -266,6 +319,79 @@ public void testConfigureNotAuthorizaedForAccountId(ConfigureMojo mojo) throws E
...
+    public void testConfigureFailsWhenTokenHasNoAccount(ConfigureMojo mojo) {
+
+        session.getRequest().setGoals(List.of("timefold:deploy"));
+        setEnterpriseModel(mojo);
+
+        mojo.setAccessTokenProvider(new TestAccessTokenProvider("noaccounts"));
+        mojo.setLog(log);
+        mojo.platformUrl = wm1.getRuntimeInfo().getHttpBaseUrl();
+
+        assertThatThrownBy(() -> mojo.execute()).isInstanceOf(MojoFailureException.class)
```
rsynek: A relevant comment by SonarCloud: could be turned into a method reference (plus two more occurrences below).

rsynek (approved):
Thanks @cristianonicolai ! I left some nitpicks to look at, but already approving.

service/tools/maven-plugin/src/test/java/ai/timefold/solver/tools/maven/ConfigureMojoTest.java
cristianonicolai: fixed

service/tools/maven-plugin/src/main/java/ai/timefold/solver/tools/maven/ConfigureMojo.java
cristianonicolai: applied

PR #2590, opened 2026-08-13
feat: Allow custom options in the RunConfiguration

cristianonicolai:
Related to https://github.com/TimefoldAI/timefold-platform/issues/5335

### DIFF
--- a/service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
+++ b/service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
@@ -1,5 +1,6 @@
 package ai.timefold.solver.service.definition.api.domain;
 
+import java.util.Map;
 import java.util.Set;
 
 import jakarta.validation.constraints.Positive;
@@ -11,7 +12,6 @@ import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 
-@Schema(additionalProperties = Schema.False.class)
 public record RunConfiguration(
         @Schema(nullable = true,
                 description = "Optional name to be given to the dataset. If not provided, the name will be generated.") @Size(
@@ -21,13 +21,11 @@ public record RunConfiguration(
                 description = "Optional maximum number of threads to be used for solving.",
                 minimum = "1") @JsonInclude(JsonInclude.Include.NON_EMPTY) @Positive Integer maxThreadCount,
         @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(
-                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags) {
+                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(hidden = true) Map<String, String> options) {
 
     public RunConfiguration(String name, SolverTerminationConfig termination, Integer maxThreadCount, Set<String> tags) {
-        this.name = name;
-        this.termination = termination;
-        this.tags = tags;
-        this.maxThreadCount = maxThreadCount;
+        this(name, termination, maxThreadCount, tags, null);
     }
 
     public RunConfiguration(String name, SolverTerminationConfig termination) {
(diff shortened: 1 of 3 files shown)


service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
```diff
@@ -21,13 +22,21 @@ public record RunConfiguration(
                 description = "Optional maximum number of threads to be used for solving.",
                 minimum = "1") @JsonInclude(JsonInclude.Include.NON_EMPTY) @Positive Integer maxThreadCount,
         @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(
-                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags) {
+                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(
```
mswiderski: I would add `@Schema` annotation that would hide it from OpenAPI spec. It should not really be advertised and by that none of the models api definitions would change.

cristianonicolai: updated

mswiderski (approved):

diogodanielsoaresferreira (approved):
Looks good to me 👍

triceo:
I would appreciate if the Sonar/Copilot comments were resolved before we merge this. If the comment is wrong, it's fine to close it without any action - but very often, these comments are right.

cristianonicolai:
@triceo I dont see much option here on what could be done to satisfy it both ways. Idea is that `options` is hidden/not visible since these are intended only for Platform use. Since `options` is hidden, schema validation would fail if it is inlucded in the payload, hence the decision to remove `additionalProperties`.  I think its either like this or we accept to have it shown in the schema as a raw options map, which means we could include `@Schema(additionalProperties = Schema.False.class)` back. @mswiderski @rsynek wdyt?

mswiderski:
I am more in favour of allowing additional properties rather than make the field visible to all. The whole idea with this was to not promote this to end consumers.

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
```diff
@@ -21,13 +21,11 @@ public record RunConfiguration(
                 description = "Optional maximum number of threads to be used for solving.",
                 minimum = "1") @JsonInclude(JsonInclude.Include.NON_EMPTY) @Positive Integer maxThreadCount,
         @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(
-                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags) {
+                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(hidden = true) Map<String, String> options) {
```
rsynek, line 25: If it's hidden from the schema, it cannot be sent from the client, since the platform gateway will consider it an invalid parameter, correct?

So we allow these options to only be sent inside the cluster.

rsynek, line 25: Do we want to keep it open for any parameters?

If we want to configure a custom seed, it will probably originate from the platform configuration profile, but we will need to set `quarkus.timefold.solver.random-seed={some-unknown-key}` to propagate it to the solver.

If we make change it to a `Map<InternalOptions, String>`, where `InternalOptions` is an enum, we have a good control over what we send to the model pods as hidden options.

cristianonicolai: the idea is that it can still be part of the POST body, so it wont fail, but only the Platform would know how to handle it. Sort of hidden parameter to be used.

mswiderski: I have slightly different view, shall we have a short meeting tomorrow or on Monday three of us (@cristianonicolai @rsynek )?

cristianonicolai: Idea was to keep it open to allow for other customization in the future. Other options that wouldnt be public available. I think if we add a `InternalOptions` enum, it would still show up in the OpenAPI as type no?
Regarding the `quarkus.timefold.solver.random-seed` that would be set as env var by the Platform, at least thats how I understood this need to be done. So despite changing the `RunConfiguration` here, thats only because we need to extend it to accept these other parameters. Do you see another way this should be done?

diogodanielsoaresferreira:
Implemented in https://github.com/diogodanielsoaresferreira/timefold-solver/pull/1 so that the Seed Execution Profile can be tested. We can close this one when https://github.com/diogodanielsoaresferreira/timefold-solver/pull/1 is merged.

PR #2595, opened 2026-08-14
feat: add environment mode per phase

zepfred:
Allows each solver phase (Construction Heuristic, Local Search, Exhaustive Search, Partitioned Search, Custom) to override the solver's `environmentMode` with a stricter mode of its own via `PhaseConfig.withEnvironmentMode(...)`. This makes it possible to run a suspect phase under `FULL_ASSERT` (or another stricter mode) for debugging, without paying that performance cost for the whole solving run.

- A phase's environment mode must be at least as strict as the solver's; it can never be looser.
- If the solver's mode is `NON_REPRODUCIBLE`, no phase can override it (every other mode is reproducible, hence stricter).
- Any number of phases may override the mode, including all of them. The solver's own mode always stands: it governs everything outside the phases (`SolutionManager`, the integrations), and the score director factory built for it is needed regardless.
- **A benchmark accepts no phase-level override at all.** A `<solverBenchmark>` whose solver config sets `<environmentMode>` on a phase is rejected, so that every solver benchmark in a report runs under exactly one environment mode and stays comparable with its peers.

All three rules are enforced up front — the first two when the `SolverFactory` is built, the third when the `PlannerBenchmark` is built — so a misconfiguration fails there rather than during solving.

### Key changes

- **`PhaseConfig`**: new `environmentMode` field/getter/setter/`withEnvironmentMode(...)`, added to the `@XmlType` `propOrder`, and wired into `solver.xsd` / `benchmark.xsd`. Includes a `revapi-differences.json` ignore entry for the resulting `@XmlType.propOrder` change. The accessors are `@Nullable` (null means "run in the solver's mode") and carry javadoc stating the rules above.

- **`EnvironmentModeUtil` (new, core)**: the phase-override rules in one place, shared by the solver and the benchmark so neither re-derives them. `validate(..)` throws on a config that breaks the rules and is the only method that throws; `resolve(..)` gives the solver's own mode; `resolvePhases(config, useGlobalMode)` gives the phases' — with `useGlobalMode` false it lists only the phases that actually override, which is exactly what the benchmark rule tests for.

- **`ScoreDirectorFactoryFactory`**: stays what it was — a builder (not a `ScoreDirectorFactory`) that validates the config, picks the one score calculation implementation it selected (easy / incremental / Constraint Streams), and applies the initializing score trend and the optional assertion score director factory. It now also decides whether a solve needs more than one environment mode: it does as soon as a phase runs in a mode other than the solver's, including when every phase overrides, because the solver's own mode still has to be served to everything outside the phases.

- **`MultiEnvironmentScoreDirectorFactory` (new)**: the decorator returned in that case. A score director factory is built for exactly one environment mode, and while easy and incremental calculators just pass the mode on to the score director, `BavetConstraintStreamScoreDirectorFactory` builds its constraint network from the mode in its constructor — so serving a phase from the solver's factory would give it a score director reporting one mode while running the network of another. The decorator builds a concrete factory per mode instead and caches it, since building a constraint network is expensive and building a score director on top of one is not. The solver's own factory is not in the cache; requests for that mode go straight through. `DefaultSolverFactory.getScoreDirectorFactory()` unwraps it, because callers outside the solving life cycle (`SolverManager`, Quarkus DI injecting `ConstraintMetaModel`) expect the concrete implementation.

- **Constraint match policy**: `ScoreDirectorFactory.decideConstraintMatchPolicy(EnvironmentMode)` (an interface default that only knew about the environment mode) is replaced by the static `ScoreDirectorFactoryFactory.decideConstraintMatchPolicy(SolverScope, EnvironmentMode)`, which also honours constraint-match-based metrics via the new `SolverScope.isAnyMetricConstraintMatchBased()`. `SolverContextManager` applies it on every phase swap, so a metric that requires constraint matching keeps working after a phase changes the environment mode.

- **`SolverContextManager` (new)**: owns the `InnerScoreDirector` the solver is working with and swaps it when a phase requires a different environment mode. It compares the phase's mode to the current one at `phaseStarted` and only builds a replacement when they differ, carrying the working solution, the score calculation count, the `SolverScope`'s view of both directors, and the `BestSolutionRecaller`'s assertion level across the swap. Score directors themselves are not cached — caching lives one level down in the factory, which is the expensive part to build. Also fixes score director closing on the failure path, where `outerSolvingEnded` never runs.

- **`AbstractSolver` / `DefaultSolver` / phase factories**: restructured so each phase is built with, and runs under, its own config policy and score director rather than always reusing the solver-level one.

- **Bundled fix** — stale `ListVariableStateSupply` references in list move selectors (`ElementDestinationSelector`, `RandomSubListSelector`, `ListChangeMoveSelector`, `ListSwapMoveSelector`, `KOptListMoveSelector`, `ListRuinRecreateMoveSelector`). These previously cached the supply once at selector construction, which breaks as soon as a phase runs on a different score director. They now acquire it per phase in `phaseStarted` from the phase's own score director — which owns the supply and allocates it once — and drop it in `phaseEnded`.

- **Benchmark**: `DefaultPlannerBenchmarkFactory.validate()` rejects a solver benchmark whose config sets `<environmentMode>` on any phase, naming the offending benchmark. A benchmark report states one environment mode per solver benchmark; a phase-level override would make that statement false and would silently handicap one config against the others, so the override is refused rather than described. Nothing else in `tools/benchmark` changes.

- **Docs**: new "Using different environment modes per phase" section in `solver-diagnostics.adoc` with a config example, the rules, and the cost note below; plus a note in each of `solver-diagnostics.adoc` and `benchmarking-and-tweaking.adoc` recording that the feature is not available inside a benchmark.

### User-visible notes

- A phase running in a mode other than the solver's gets its own score director. With Constraint Streams that means a second constraint network is built for that mode — built once and reused, but not free, so override only the phases you actually want to inspect.
- The feature is not available inside a benchmark: `<environmentMode>` on a phase of a `<solverBenchmark>` is rejected when the benchmark is built. The schema still permits it, because the benchmark XSD's phase types are generated from the same `PhaseConfig`, so this surfaces as a runtime error rather than a schema violation.
- The benchmark report itself is unchanged; its "Environment mode" row still holds a single mode per report.
- **Behavioural change worth watching:** a Constraint Streams config with a phase-level override now genuinely gets a constraint network built for that phase's mode. Until this PR such a phase could run a lenient network while reporting a strict mode, so its assertions were not really asserting. Configs that appeared to pass under `FULL_ASSERT` may now surface score corruption they were previously hiding — which is the point of the mode, but it can look like a regression.


core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
```diff
@@ -232,17 +233,17 @@ public static class DefaultConstructionHeuristicPhaseBuilder<Solution_>
...
+                EntityPlacer<Solution_> entityPlacer, ConstructionHeuristicDecider<Solution_> decider) {
+            super(phaseIndex, lastInitializingPhase, environmentMode, logIndentation, phaseTermination);
             this.entityPlacer = entityPlacer;
             this.decider = decider;
         }
 
         @Override
-        public DefaultConstructionHeuristicPhaseBuilder<Solution_> enableAssertions(EnvironmentMode environmentMode) {
-            super.enableAssertions(environmentMode);
+        public DefaultConstructionHeuristicPhaseBuilder<Solution_> enableAssertions() {
```
triceo: Why does this exist? It doesn't do anything but delegate to the parent, which means the parent method could have been called instead.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupplyHolder.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Objects;
+
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
+
+/**
+ * Demands a {@link ListVariableStateSupply} on {@link #phaseStarted(AbstractPhaseScope)} and releases it on
+ * {@link #phaseEnded(AbstractPhaseScope)}, re-demanding on every phase start because a phase may run under a
+ * different {@link ai.timefold.solver.core.config.solver.EnvironmentMode}, which swaps in a new score director
+ * (and thus a new {@link ai.timefold.solver.core.impl.domain.variable.supply.SupplyManager}).
```
triceo: IMO the dropping and re-demanding should happen in phaseStarted(), and only when the next phase actually needs a new supply. These things are expensive - let's not rebuild them needlessly.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
```diff
@@ -141,17 +142,17 @@ public static class Builder<Solution_> extends AbstractPhaseBuilder<Solution_> {
...
             this.decider = decider;
         }
 
         @Override
-        public Builder<Solution_> enableAssertions(EnvironmentMode environmentMode) {
-            super.enableAssertions(environmentMode);
+        public Builder<Solution_> enableAssertions() {
+            super.enableAssertions();
             assertWorkingSolutionScoreFromScratch = environmentMode.isFullyAsserted();
             assertExpectedWorkingSolutionScore = environmentMode.isIntrusivelyAsserted();
```
triceo, line 159: The decider no longer needs these fields; are they still necessary here?

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/ElementDestinationSelector.java
```diff
@@ -89,20 +85,21 @@ private IterableValueSelector<Solution_> filterUnassignedValues(
...
     @Override
-    public void solvingStarted(SolverScope<Solution_> solverScope) {
-        super.solvingStarted(solverScope);
-        var supplyManager = solverScope.getScoreDirector().getSupplyManager();
-        listVariableStateSupply = supplyManager.demand(listVariableDescriptor.getStateDemand());
+    public void phaseStarted(AbstractPhaseScope<Solution_> phaseScope) {
+        super.phaseStarted(phaseScope);
+        // The phase may operate in a different environment mode, which uses a new score director.
+        // We must ensure that the list variable state supply remains up to date.
+        listVariableStateSupplyHolder.phaseStarted(phaseScope);
```
triceo: I wonder if it would be cleaner for the supply to be coming from the current score director. That way, this code can be entirely eliminated, and it becomes _impossible_ for the supply and the score director to be incompatible.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/list/RandomSubListSelector.java
```diff
@@ -22,15 +21,16 @@ public class RandomSubListSelector<Solution_> extends AbstractSelector<Solution_
...
+    private final ListVariableStateSupplyHolder<Solution_> listVariableStateSupplyHolder;
 
     public RandomSubListSelector(
             EntitySelector<Solution_> entitySelector,
             IterableValueSelector<Solution_> valueSelector,
             int minimumSubListSize, int maximumSubListSize) {
         this.entitySelector = entitySelector;
-        this.valueSelector = filterPinnedListPlanningVariableValuesWithIndex(valueSelector, this::getListVariableStateSupply);
         this.listVariableDescriptor = (ListVariableDescriptor<Solution_>) valueSelector.getVariableDescriptor();
+        this.listVariableStateSupplyHolder = new ListVariableStateSupplyHolder<>(listVariableDescriptor);
```
triceo: This is super-weird. We create our own supply here? I wouldn't.
Another reason for the supply to be on the score director directly.

I won't mention it anymore, even though it shows up on other selectors as well.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/HeuristicConfigPolicy.java
```diff
@@ -130,15 +130,15 @@ public Builder<Solution_> cloneBuilder() {
...
                 .withMoveThreadBufferSize(moveThreadBufferSize)
                 .withThreadFactoryClass(threadFactoryClass)
-                .withNearbyDistanceMeterClass(nearbyDistanceMeterClass)
-                .withRandom(random)
                 .withInitializingScoreTrend(initializingScoreTrend)
                 .withSolutionDescriptor(solutionDescriptor)
                 .withClassInstanceCache(classInstanceCache)
-                .withLogIndentation(logIndentation);
+                .withNearbyDistanceMeterClass(nearbyDistanceMeterClass)
+                .withRandom(random);
```
triceo, line 141: Why is this code changing?
There is no actual change here.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AcceptorFactory.java
```diff
@@ -93,15 +94,16 @@ private Optional<StepCountingHillClimbingAcceptor<Solution_>> buildStepCountingH
         return Optional.empty();
     }
 
-    private Optional<EntityTabuAcceptor<Solution_>> buildEntityTabuAcceptor(HeuristicConfigPolicy<Solution_> configPolicy) {
+    private Optional<EntityTabuAcceptor<Solution_>> buildEntityTabuAcceptor(EnvironmentMode environmentMode,
```
triceo: Since we're touching this, let's remove the `Optional`. We do not use `Optional` - null will be enough.

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AcceptorFactory.java
```diff
@@ -39,13 +40,13 @@ public AcceptorFactory(LocalSearchAcceptorConfig acceptorConfig) {
         this.acceptorConfig = acceptorConfig;
     }
 
-    public Acceptor<Solution_> buildAcceptor(HeuristicConfigPolicy<Solution_> configPolicy) {
+    public Acceptor<Solution_> buildAcceptor(HeuristicConfigPolicy<Solution_> configPolicy, EnvironmentMode environmentMode) {
```
triceo: I wonder why the env mode is not part of the config policy.
This way, it requires us to propagate it everywhere separately, causing much larger diffs...

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/acceptor/AcceptorFactory.java
```diff
@@ -39,13 +40,13 @@ public AcceptorFactory(LocalSearchAcceptorConfig acceptorConfig) {
...
         List<Acceptor<Solution_>> acceptorList = Stream.of(
                 buildHillClimbingAcceptor(),
                 buildStepCountingHillClimbingAcceptor(),
-                buildEntityTabuAcceptor(configPolicy),
-                buildValueTabuAcceptor(configPolicy),
-                buildMoveTabuAcceptor(configPolicy),
+                buildEntityTabuAcceptor(environmentMode, configPolicy.getLogIndentation()),
```
triceo: ... and things like this.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -111,7 +114,10 @@ protected AbstractScoreDirector(AbstractScoreDirectorBuilder<Solution_, Score_,
         this.variableDescriptorCache = new VariableDescriptorCache<>(solutionDescriptor);
         this.shadowVariableSupport = ShadowVariableSupport.create(this);
         this.shadowVariableSupport.linkShadowVariables();
-        this.solutionTracker = this.scoreDirectorFactory.isTrackingWorkingSolution()
+        //  When true, a snapshot of the solution is created before, after and after the undo of a move.
```
triceo: "before, after" what?

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -206,8 +213,8 @@ public void resetCalculationCount() {
     }
 
     @Override
-    public void incrementCalculationCount() {
-        this.calculationCount++;
+    public void incrementCalculationCount(long count) {
```
triceo, line 231: Why this change? Seems entirely unrelated.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -438,20 +445,23 @@ protected void setCalculatedScore(Score_ score) {
...
+                childThreadScoreDirector.calculationCount = calculationCount;
+                return childThreadScoreDirector;
+            }
+            case MOVE_THREAD -> {
+                var childThreadScoreDirector = scoreDirectorFactory.createScoreDirectorBuilder().withLookUpEnabled(true)
+                        .withConstraintMatchPolicy(constraintMatchPolicy).buildDerived();
+                childThreadScoreDirector.setWorkingSolution(cloneWorkingSolution());
+                return childThreadScoreDirector;
+            }
+            default -> throw new IllegalStateException("The childThreadType (" + childThreadType + ") is not implemented.");
```
triceo: `formatted(...)`

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -665,6 +675,32 @@ public void afterProblemFactRemoved(Object problemFact) {
     // Assert methods
     // ************************************************************************
 
+    /**
+     * Asserts that if the {@link Score} is calculated for the parameter solution,
+     * it would be equal to the score of that parameter.
+     *
+     * @param solution never null
+     * @see InnerScoreDirector#assertWorkingScoreFromScratch(InnerScore, Object)
```
triceo: Unnecessary - the `@Override` makes it obvious.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -7,19 +7,43 @@
...
 
-    public ScoreDirectorFactoryFactory(ScoreDirectorFactoryConfig config) {
+    public DelegateScoreDirectorFactory(ScoreDirectorFactoryConfig config, boolean hasMetricRequiringConstraintMatch) {
         this.config = config;
+        this.hasMetricRequiringConstraintMatch = hasMetricRequiringConstraintMatch;
+        assertCorrectDirectorFactory(config);
     }
 
+    /**
+     * Build a score director factory according to the given environment mode.
```
triceo: I wonder... if the factory already produced a director for this environment mode, why would it build another one? Does it make sense to cache and reuse?

Score directors are super-expensive. I worry that this PR will introduce a lot of overhead in cases where it is not necessary.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -37,27 +61,35 @@ public ScoreDirectorFactory<Solution_, Score_> buildScoreDirectorFactory(Environ
...
 
-    protected AbstractScoreDirectorFactory<Solution_, Score_, ?> decideMultipleScoreDirectorFactories(
-            SolutionDescriptor<Solution_> solutionDescriptor, EnvironmentMode environmentMode) {
-        assertCorrectDirectorFactory(config);
+    /**
+     * Creates a new instance of the score director.
+     *
+     * @param scoreScoreDirectorFactory the factory to be used to create the score director instance.
+     */
+    public InnerScoreDirector<Solution_, Score_>
```
triceo: Strange signature. Delegate factory requires some other factory to build a score director? I would expect to build a score director out of a factory without requiring any other factory.

core/src/main/java/ai/timefold/solver/core/impl/score/director/ScoreDirectorFactory.java
```diff
@@ -22,24 +24,24 @@ public interface ScoreDirectorFactory<Solution_, Score_ extends Score<Score_>> {
...
-    InitializingScoreTrend getInitializingScoreTrend();
+    EnvironmentMode getEnvironmentMode();
 
     /**
-     * Asserts that if the {@link Score} is calculated for the parameter solution,
-     * it would be equal to the score of that parameter.
-     *
-     * @param solution never null
-     * @see InnerScoreDirector#assertWorkingScoreFromScratch(InnerScore, Object)
+     * @return never null
```
triceo: Can be avoided with `@NullMarked`.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -95,20 +110,69 @@ protected void runPhases(SolverScope<Solution_> solverScope) {
...
+        // because the required environment mode differs from both the current and the default modes.
+        ScoreDirectorFactory newScoreDirectorFactory = delegateScoreDirectorFactory
+                .buildScoreDirectorFactory(phase.getEnvironmentMode(), solverScope.getSolutionDescriptor());
+        var newScoreDirector = delegateScoreDirectorFactory.createScoreDirector(newScoreDirectorFactory);
+        var newSolverContext = new SolverContext<>(phase.getEnvironmentMode(), newScoreDirector,
+                new DefaultProblemChangeDirector<>(newScoreDirector));
+        loadContext(currentContext, newSolverContext, solverScope);
+    }
+
+    private void loadContext(SolverContext<Solution_, ?> oldSolverContext, SolverContext<Solution_, ?> newSolverContext,
```
triceo: This is a strange method. The way the context mutation is handled is not particularly nice. It needs two contexts, does not return anything. IMO all of this should be hidden within the method, and the method should return something; it should hide more of the implementation that it is hiding. 

The solver scope mutation prevents it from being a pure function, but we can certainly get it closer to that.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -95,20 +110,69 @@ protected void runPhases(SolverScope<Solution_> solverScope) {
...
 
     public void solvingError(SolverScope<Solution_> solverScope, Exception exception) {
         phaseLifecycleSupport.fireSolvingError(solverScope, exception);
         for (Phase<Solution_> phase : phaseList) {
             phase.solvingError(solverScope, exception);
         }
+        if (currentContext != defaultSolverContext) {
+            // A phase may have failed while operating under a non-default environment mode,
+            // and we need to restore the default context
+            loadContext(currentContext, defaultSolverContext, solverScope);
```
triceo: Why? There should be a fail-fast here. A phase has failed - something somewhere will throw an exception and kill the solver.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -95,20 +110,69 @@ protected void runPhases(SolverScope<Solution_> solverScope) {
...
         for (Phase<Solution_> phase : phaseList) {
             phase.solvingEnded(solverScope);
         }
         bestSolutionRecaller.solvingEnded(solverScope);
         globalTermination.solvingEnded(solverScope);
         phaseLifecycleSupport.fireSolvingEnded(solverScope);
+        if (currentContext != defaultSolverContext) {
+            // Restore the default context
+            // so solverScope operate on the original score director
+            loadContext(currentContext, defaultSolverContext, solverScope);
```
triceo: Why do we care? Solving ended. Can you reproduce in a test why this should be a problem?

If we're doing this because of release()`, then this should be done much earlier anyway. Score directors are very heavy - they must not survive phase end, if they are no longer necessary for subsequent phases.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -272,6 +291,59 @@ public void ensurePreviewFeature(PreviewFeature previewFeature) {
...
+                phaseConfigList.stream()
+                        .map(phaseConfig -> Objects.requireNonNullElse(phaseConfig.getEnvironmentMode(),
+                                defaultEnvironmentMode))
+                        .toList();
+        if (defaultEnvironmentMode == EnvironmentMode.NON_REPRODUCIBLE
+                && phaseEnvironmentList.stream().anyMatch(environmentMode -> environmentMode != defaultEnvironmentMode)) {
+            // If the default environment is non-reproducible,
+            // then all phase environment modes must also be non-reproducible
+            throw new IllegalStateException(
+                    "The default environment mode is (%s), and all phase environments [%s] must also be non-reproducible."
```
triceo: This message is confusing.
"Phase-level environmentMode override is only possible when global environmentMode is reproducible, but was %s."

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -272,6 +291,59 @@ public void ensurePreviewFeature(PreviewFeature previewFeature) {
...
+            if (phaseEnvironment == defaultEnvironmentMode) {
+                checkDefaultEnvironment = true;
+                break;
+            }
+        }
+        if (!checkDefaultEnvironment) {
+            throw new IllegalStateException("""
+                    The default environment mode (%s) is not used in any of the defined phases environment modes [%s].
+                    Maybe adjust the solver config's default environment mode.
+                    Maybe adjust at least one of the phase environment modes to match the default environment mode (%s)"""
```
triceo: I do not understand what this message is telling me.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -272,6 +291,59 @@ public void ensurePreviewFeature(PreviewFeature previewFeature) {
...
+        var invalidPhaseEnvironmentList = new ArrayList<String>(phaseConfigList.size());
+        for (var phaseEnvironment : phaseEnvironmentList) {
+            if (phaseEnvironment.ordinal() > defaultEnvironmentMode.ordinal()) {
+                invalidPhaseEnvironmentList.add(phaseEnvironment.name());
+            }
+        }
+        if (!invalidPhaseEnvironmentList.isEmpty()) {
+            // The phase environments must have an assertion level greater than or equal to the default environment level
+            throw new IllegalStateException(
+                    "The phase environments must have an assertion level higher than or equal to the default environment level (%s). The following phase environment modes are not valid: [%s]."
```
triceo: IMO calling the default level "global" makes more sense.
Let's consistently change that throughout the entire PR.

triceo: Also, wouldn't it make more sense to list which modes ARE valid?

triceo (commented):
Good start.

The list var state holder arguably should live on the score director; that will make many things more transparent.

I do not like the state mutation in `AbstractSolver` - it is poorly encapsulated, and does not make it at all clear what happens when. 

The interaction between score director factories is unclear for the same reason - the complexity of creating score factories on-demand should be wrapped _inside_ some score director factory implementation, and not exposed to everyone via a wrapper.

Generally, it should be super-obvious when a new score director is being created, because that is a very heavy operation and does have major performance impact on the solver.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/HeuristicConfigPolicy.java
zepfred: The only reason is to make it match the class field definition order.

core/src/main/java/ai/timefold/solver/core/impl/constructionheuristic/DefaultConstructionHeuristicPhase.java
zepfred: This is necessary to keep the API fluent and to avoid requiring any casts during the build phase, but I believe it can be improved and the method removed.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateSupplyHolder.java
zepfred: The demand already occurs at `phaseStarted`, but I prefer the dropping to happen at `phaseEnded` because it makes more sense when thinking about initializing and releasing resources. If I understand correctly, moving the holder to the score director will result in a single instance of it, and we can initialize the state lazily and allocate the resource only when requested.

I believe these two changes will address your concern about resource usage.

core/src/main/java/ai/timefold/solver/core/impl/exhaustivesearch/DefaultExhaustiveSearchPhase.java
zepfred: These fields are still used by the phase in the `solve` method:

```
decider.restoreWorkingSolution(stepScope, assertWorkingSolutionScoreFromScratch,
                    assertExpectedWorkingSolutionScore);
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: The change is required when switching between contexts to update the count and ensure the related termination settings are triggered consistently.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
zepfred: > I wonder... if the factory already produced a director for this environment mode, why would it build another one? Does it make sense to cache and reuse?

I've chosen not to cache to avoid increasing memory usage. That was a design decision. Otherwise, the application may fail due to OOM instead of spending more time creating the score directors.

> Score directors are super-expensive. I worry that this PR will introduce a lot of overhead in cases where it is not necessary.

Using different environments for each phase should be an exception, and this PR won't introduce any overhead if only one environment mode is active. Can you explain why you think there will be overhead?

I'm not against caching the environment. If you have a strong opinion on that, we can change the logic.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
zepfred: My goal is to ensure that at least one phase's environment mode matches the global one.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
zepfred: I have added a cache layer to simplify the state change in the `AbstractSolver`and will conduct a performance evaluation once the changes are complete.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
zepfred: The state management has been updated, but it's important to note that the solver does not necessarily terminate if it fails during solving. The score director remains open if it fails, potentially causing a leak. I have added a test to verify this, located in `TimefoldProcessorFailedSolveTest#twoFailedSolvesLeaveTheSolverManagerUsable`.

I'm not sure if I'll keep the test, because it's not possible to verify it when running in Quarkus. However, you can debug it using the main branch to confirm that the score director is not closed, since `outerSolvingEnded` is never called.

zepfred:
Here are the benchmark warnings for the following configuration:

```
<plannerBenchmark xmlns="https://timefold.ai/xsd/benchmark" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="https://timefold.ai/xsd/benchmark https://timefold.ai/xsd/benchmark/benchmark.xsd">
    <parallelBenchmarkCount>18</parallelBenchmarkCount>
    <inheritedSolverBenchmark>
        <solver>
            <termination>
                <secondsSpentLimit>10</secondsSpentLimit>
            </termination>
        </solver>
        <problemBenchmarks>
            <problemStatisticType>BEST_SCORE</problemStatisticType>
        </problemBenchmarks>
    </inheritedSolverBenchmark>
    <solverBenchmark>
        <name>Solver1</name>
        <solver>
            <constructionHeuristic/>
            <localSearch/>
        </solver>
    </solverBenchmark>
    <solverBenchmark>
        <name>Solver2</name>
        <solver>
            <constructionHeuristic/>
            <localSearch>
                <environmentMode>FULL_ASSERT</environmentMode>
            </localSearch>
        </solver>
    </solverBenchmark>
</plannerBenchmark>
```

image

tools/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/BenchmarkReport.java
```diff
@@ -302,35 +303,62 @@ public List<String> getWarningList() {
         List<String> warningList = new ArrayList<>();
         String javaVmName = System.getProperty("java.vm.name");
         if (javaVmName != null && javaVmName.contains("Client VM")) {
```
triceo: Client VMs have not existed for probably a decade now. This can be completely removed.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/GenericListMoveSelector.java
```diff
@@ -0,0 +1,59 @@
...
+import ai.timefold.solver.core.impl.domain.variable.ListVariableStateSupply;
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.heuristic.selector.move.generic.GenericMoveSelector;
+import ai.timefold.solver.core.impl.phase.scope.AbstractPhaseScope;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public abstract class GenericListMoveSelector<Solution_> extends GenericMoveSelector<Solution_> {
```
triceo: Abstract classed, by convention, start by `Abstract`.
It appears we broke that convention already with the parent, fix that too.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -0,0 +1,293 @@
...
+ * the {@link InitializingScoreTrend},
+ * the optional assertion score director factory,
+ * and the {@link ConstraintMatchPolicy} implied by the environment mode and the enabled metrics.
+ *
+ * @param <Solution_> the solution type, the class with the {@link PlanningSolution}
+ *        annotation
+ * @param <Score_> the score type to go with the solution
+ */
+@NullMarked
+public class DelegateScoreDirectorFactory<Solution_, Score_ extends Score<Score_>>
```
triceo: I do not understand why this exists; seems over-engineered.
What is the difference between passing this, and passing `AbstractScoreDirectorFactory` instead?

If you even need the specific type of the factory to get the specific builder out of it, you can always do an exhaustive `switch` over it.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -0,0 +1,293 @@
...
+    public DelegateScoreDirectorFactory(ScoreDirectorFactoryConfig config, SolutionDescriptor<Solution_> solutionDescriptor,
+            EnvironmentMode environmentMode, List<SolverMetric> metricsRequiringConstraintMatchList) {
+        this.config = Objects.requireNonNull(config);
+        assertCorrectDirectorFactory(config);
+        this.solutionDescriptor = solutionDescriptor;
+        this.globalEnvironmentMode = environmentMode;
+        this.metricsRequiringConstraintMatchList = metricsRequiringConstraintMatchList;
+        this.scoreDirectorFactory = internalBuildScoreDirectorFactory(solutionDescriptor, environmentMode);
+        // Constraint Stream factory requires a new factory if the environment changes
+        this.requireNewFactoryOnDifferentEnvironment = config.getConstraintProviderClass() != null;
```
triceo: This is very brittle. You are making the assumption that only CS will ever need this, and when something in the other score directors changes, this immediately introduces a silent bug which nobody notices.

IMO it is safer for this to always be `true`, and therefore it doesn't need to exist at all.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -0,0 +1,293 @@
...
+ * and then reused for every score director requested for that mode.
+ * <p>
+ * Since a solver phase may override the solver's environment mode,
+ * {@link #createScoreDirectorBuilder(EnvironmentMode)} may be called with a different mode than the default one.
+ * Most delegates only pass the environment mode on to the score director they build,
+ * so they can serve any mode and are reused as they are.
+ * The exception is {@link BavetConstraintStreamScoreDirectorFactory},
+ * which builds its constraint network from the environment mode up front;
+ * for it, a separate delegate is built for the requested mode,
+ * then cached and shared like the default one, as building it is expensive.
```
triceo: See comment below. IMO this should not be the case.

core/src/main/java/ai/timefold/solver/core/impl/score/director/DelegateScoreDirectorFactory.java
```diff
@@ -0,0 +1,293 @@
...
+     * @param metricsRequiringConstraintMatchList the enabled metrics which can only be computed
+     *        when the score directors track constraint matches
+     */
+    public DelegateScoreDirectorFactory(ScoreDirectorFactoryConfig config, SolutionDescriptor<Solution_> solutionDescriptor,
+            EnvironmentMode environmentMode, List<SolverMetric> metricsRequiringConstraintMatchList) {
+        this.config = Objects.requireNonNull(config);
+        assertCorrectDirectorFactory(config);
+        this.solutionDescriptor = solutionDescriptor;
+        this.globalEnvironmentMode = environmentMode;
+        this.metricsRequiringConstraintMatchList = metricsRequiringConstraintMatchList;
```
triceo: The class should not be dealing with this. That is the delegate's problem.
Everything but the environment mode is the delegate's problem.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -67,7 +89,8 @@ public final class DefaultSolverFactory<Solution_> implements SolverFactory<Solu
     private final Clock clock;
     private final SolverConfig solverConfig;
     private final SolutionDescriptor<Solution_> solutionDescriptor;
-    private final ScoreDirectorFactory<Solution_, ?> scoreDirectorFactory;
+    private final EnvironmentMode globalEnvironmentMode;
+    private final DelegateScoreDirectorFactory<Solution_, ?> delegateScoreDirectorFactory;
```
triceo: Does this need to be `DelegateSDF` always? Why can't this be just `AbstractSDF`? 

IMO we should only pass the complexity of `DelegateSDF` when we actually need it. Most of the time, we will not need it (= most of the time, the env mode is consistent).

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -91,9 +118,15 @@ public SolutionDescriptor<Solution_> getSolutionDescriptor() {
...
+    /**
+     * @return the factory built for the default environment mode;
+     *         the delegate and not its {@link DelegateScoreDirectorFactory} wrapper,
+     *         as callers outside the solving life cycle expect the concrete implementation,
+     *         such as {@code BeanUtil#buildConstraintMetaModel} which needs a constraint stream factory
+     */
     @SuppressWarnings("unchecked")
     public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> getScoreDirectorFactory() {
-        return (ScoreDirectorFactory<Solution_, Score_>) scoreDirectorFactory;
+        return (ScoreDirectorFactory<Solution_, Score_>) delegateScoreDirectorFactory.getDelegate();
```
triceo: This only reinforces my previous comment.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeResolver.java
```diff
@@ -0,0 +1,142 @@
...
+        var invalidPhaseEnvironmentList = new ArrayList<String>(phaseEnvironmentModeList.size());
+        for (var phaseEnvironmentMode : phaseEnvironmentModeList) {
+            if (phaseEnvironmentMode.ordinal() > globalEnvironmentMode.ordinal()) {
+                invalidPhaseEnvironmentList.add(phaseEnvironmentMode.name());
+            }
+        }
+        if (!invalidPhaseEnvironmentList.isEmpty()) {
+            // The phase environments must have an assertion level greater than or equal to the global environment level
+            throw new IllegalStateException(
+                    "The phase environments must have an assertion level higher than or equal to the global environment level (%s). The following phase environment modes are not valid: [%s]."
```
triceo: Deserves a multi-line comment, line too long for read both here and in the actual log.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeResolver.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * The environment mode the solver as a whole runs in,
+     * which is the declared {@link SolverConfig#determineEnvironmentMode()}
+     * unless every phase agrees on one mode, in which case that mode is adopted as the global one.
+     * There is then nothing to swap away from mid-solve, which spares the solver a second score director
+     * factory — and, with Constraint Streams, a second constraint network — for a mode no phase ever runs in.
+     * <p>
+     * This is not the strictest mode the solve runs in; see {@link #resolveStrictest(SolverConfig)} for that.
+     *
+     * @see #validate(SolverConfig) never throws, unlike the validation
```
triceo: Doesn't make sense.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeResolver.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public static EnvironmentMode resolve(SolverConfig solverConfig) {
+        var globalEnvironmentMode = solverConfig.determineEnvironmentMode();
+        var phaseEnvironmentModeList = determinePhaseEnvironmentModeList(solverConfig, globalEnvironmentMode);
+        if (phaseEnvironmentModeList.isEmpty()) {
+            return globalEnvironmentMode;
+        }
+        var distinctPhaseEnvironmentModeList = phaseEnvironmentModeList.stream().distinct().toList();
```
triceo: `EnumSet`

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeResolver.java
```diff
@@ -0,0 +1,142 @@
...
+        return distinctPhaseEnvironmentModeList.size() == 1
```
triceo: What's the logic behind picking the first phase as global?
I would argue the global is still the default global; even if global is undefined in the solver config, it exists - it is the default one.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeResolver.java
```diff
@@ -0,0 +1,142 @@
...
+ * and must not be able to fail on validation.</li>
+ * </ul>
+ * A resolution of a config which never passed {@link #validate(SolverConfig)} is therefore best-effort:
+ * it applies the rules to a config the solver would have refused to build.
+ * <p>
+ * Note that {@link EnvironmentMode} is declared from strictest to most lenient,
+ * so a lower {@link Enum#ordinal()} means a stricter mode.
+ */
+@NullMarked
+public final class EnvironmentModeResolver {
```
triceo: Considering this class has no actual instances, `EnvironmentModeUtil` would be more suitable.

core/src/main/java/ai/timefold/solver/core/impl/solver/SolverContextManager.java
```diff
@@ -0,0 +1,189 @@
...
+ * A consequence is that the solver ends holding the *last* phase's score director,
+ * not the one it started with.
+ * <p>
+ * Everything a swap has to keep continuous lives in {@link #loadContext}:
+ * the working solution, the score calculation count, the {@link SolverScope}'s view of both directors,
+ * and the assertion level of the {@link BestSolutionRecaller}.
+ * <p>
+ * <b>Ownership of closing.</b> This class closes only the score directors it replaces.
+ * The last one standing is closed by the solver
+ * ({@code DefaultSolver.outerSolvingEnded}) on the normal path.
```
triceo: IMO this is the wrong design.
Either the class owns the score directors, or it does not.
If it owns them, it closes them all - if it doesn't own them, somebody else closes them all.
I would prefer if this class controlled all score director closing.

Also, if something needs to be closed at the end, it typically `implements AutoCloseable`.

core/src/main/java/ai/timefold/solver/core/impl/solver/SolverContextManager.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Closes the score director in use, as the solver's normal cleanup does not run on the failure path.
+     * <p>
+     * Solving can fail before {@link #solvingStarted(SolverScope)} has adopted a context — for instance in
+     * {@code DefaultSolver.assertCorrectSolutionState()}, or in any listener notified earlier in solving start.
+     * The score director the solver was built with still has to be closed in that case,
+     * or a long-lived {@link SolverManager} would accumulate one per failed job.
+     * Whatever happens here must not throw: the caller is on its way to rethrowing the real failure.
```
triceo: See, this is a mess. If you always close score directors here, you never need to explain who closes what and when.

This is how complexity becomes unbearable:
- shared mutable state,
- exposing resource management to external callers.

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+By default, every phase of the solver - such as xref:optimization-algorithms/construction-heuristics.adoc#constructionHeuristicsOverview[Construction Heuristic] and xref:optimization-algorithms/local-search.adoc#localSearchOverview[Local Search] - runs in the solver's environment mode.
+If the solver's environment mode is not explicitly configured, that is the default `<<environmentModePhaseAssert,PHASE_ASSERT>>` mode.
+
+Each phase can override the solver's environment mode with a stricter mode of its own.
+This is useful when you suspect that a bug is introduced during a specific phase:
+instead of paying the performance cost of a stricter mode (such as `<<environmentModeFullAssert,FULL_ASSERT>>`) for the entire solver,
+you enable it for only the phase under suspicion, while the other phases keep running at a faster mode.
+
+A phase's environment mode must be at least as strict as the solver's environment mode; it can never be less strict.
+Any number of phases can override it, including all of them:
```
triceo:
```suggestion
Any number of phases can override it while
```

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+the solver's environment mode still applies outside the phases.
```
triceo:
```suggestion
the solver's global environment mode still applies outside these phases.
```

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+If the solver's environment mode is `<<environmentModeNonReproducible,NON_REPRODUCIBLE>>`, no phase can override it,
```
triceo:
```suggestion
If the solver's global environment mode is `<<environmentModeNonReproducible,NON_REPRODUCIBLE>>`, no phase can override it,
```

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+because every other mode is <<environmentModeReproducibility,reproducible>> and therefore stricter.
+
+If every phase ends up in the same environment mode, that mode becomes the solver's environment mode too,
+since no phase is left running in the configured one.
```
triceo: IMO wrong and unexpected, as explained above.

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+that means a second constraint network is built for that mode.
+It is built once and reused, but it is not free: prefer overriding the phases you actually want to inspect.
+====
+
+[NOTE]
+====
+The solver's environment mode also applies outside the phases,
+including to xref:using-timefold-solver/modeling-planning-problems.adoc[`SolutionManager`] operations,
+so a stricter solver-level mode makes those slower as well.
+====
```
triceo: What does this mean? How is `SolutionManager` affected by solver config phases?

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,56 @@ If your production environment doesn't care about reproducibility, use this mode
...
+If every phase ends up in the same environment mode, that mode becomes the solver's environment mode too,
+since no phase is left running in the configured one.
+
+[NOTE]
+====
+A phase that runs in a different environment mode than the solver gets its own score director.
+With the xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams] API,
+that means a second constraint network is built for that mode.
+It is built once and reused, but it is not free: prefer overriding the phases you actually want to inspect.
+====
```
triceo:
```suggestion
NOTE: A phase that runs in a different environment mode than the solver gets its own score director.
With the xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams] API,
that means a second constraint network is built for that mode.
It is built once and reused, but it is not free: prefer overriding the phases you actually want to inspect.
```

triceo (changes requested):
- The changes to the benchmarker should go away; benchmarker should fail fast when this is detected. (Considering how tiny benchmarker test coverage is, we should not be touching it unless we absolutely have to.
- The Delegate factory is IMO overly complex, and overused.
- The SolverContextManager is poorly designed; its resource management is leaking to the rest of the solver.
- The global env mode should not be decided by the phases.
- Plus usual smaller comments.

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
zepfred: `SolverManager` uses the default score director factory to retrieve the constraint model as an example. Following our discussion about the global not being switched to the phase environment, this comment will be removed.

tools/benchmark/src/main/java/ai/timefold/solver/benchmark/impl/report/BenchmarkReport.java
zepfred: All changes to the benchmark module have been reverted, and we are now touching only `DefaultPlannerBenchmarkFactory` to update the configuration validation.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeUtil.java
```diff
@@ -0,0 +1,107 @@
...
+        var phaseConfigList = solverConfig.getPhaseConfigList();
+        if (phaseConfigList == null || phaseConfigList.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return phaseConfigList.stream()
+                .map(phaseConfig -> useGlobalEnvironmentMode
+                        ? Objects.requireNonNullElse(phaseConfig.getEnvironmentMode(), globalEnvironmentMode)
+                        : phaseConfig.getEnvironmentMode())
+                .filter(Objects::nonNull)
+                .toList();
```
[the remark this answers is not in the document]
zepfred: No, overriding the phase environment with the global environment is a mistake.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
```diff
@@ -102,16 +114,27 @@ public void solvingEnded(SolverScope<Solution_> solverScope) {
...
         globalTermination.solvingEnded(solverScope);
         phaseLifecycleSupport.fireSolvingEnded(solverScope);
+        // The score director is deliberately not closed here; a real-time problem change restarts solving and
+        // runs more phases on it. DefaultSolver.outerSolvingEnded closes it, through closeSolverContext().
+    }
+
+    public void outerSolvingEnded(SolverScope<Solution_> solverScope) {
+        // Closes the score director after the problem change loop,
+        // so the score director is finally done
+        solverContextManager.close();
```
triceo: This is inconsistent.
You initialize the context by `solvingStarted` - why not close it with `solvingEnded`? I understand that this is forced by `Closeable`, but in this case, IMO the asymmetry it creates is worse.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolver.java
```diff
@@ -207,11 +197,11 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
                 EventProducerId.solvingStarted());
 
         LOGGER.info("Solving {}: time spent ({}), best score ({}), "
```
triceo: Since we already have the `INFO` check on line 207, let's move this statement there.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
```diff
@@ -91,8 +120,18 @@ public SolutionDescriptor<Solution_> getSolutionDescriptor() {
...
 
+    /**
+     * @return the factory built for the global environment mode, never a
+     *         {@link MultiEnvironmentScoreDirectorFactory} wrapper,
+     *         as callers outside the solving life cycle expect the concrete implementation —
+     *         such as {@code BeanUtil#buildConstraintMetaModel}, which needs a constraint stream factory
+     */
     @SuppressWarnings("unchecked")
     public <Score_ extends Score<Score_>> ScoreDirectorFactory<Solution_, Score_> getScoreDirectorFactory() {
+        if (scoreDirectorFactory instanceof MultiEnvironmentScoreDirectorFactory<Solution_, ?, ?> multiEnvironmentScoreDirectorFactory) {
```
triceo: I don't think this should be necessary.
Multi-env should be an implementation detail, it should not proliferate throughout the codebase.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeUtil.java
```diff
@@ -0,0 +1,107 @@
...
+        var invalidPhaseEnvironmentList = new ArrayList<String>(phaseEnvironmentModeList.size());
+        for (var phaseEnvironmentMode : phaseEnvironmentModeList) {
+            if (phaseEnvironmentMode.ordinal() > globalEnvironmentMode.ordinal()) {
+                invalidPhaseEnvironmentList.add(phaseEnvironmentMode.name());
+            }
+        }
+        if (!invalidPhaseEnvironmentList.isEmpty()) {
+            // The phase environments must have an assertion level greater than or equal to the global environment level
+            throw new IllegalStateException(
+                    """
```
triceo, line 61:
```suggestion
            throw new IllegalStateException("""
```

Save some lines.

docs/src/modules/ROOT/pages/running-timefold-solver/benchmarking-and-tweaking.adoc
```diff
@@ -136,6 +136,12 @@ It runs the solver configuration on each of those unsolved solution files.
 The element `name` is optional, because it is generated if absent.
 The `inputSolutionFile` is read by a <<solutionFileIO,SolutionFileIO>>, relative to the working directory.
 
+[IMPORTANT]
+====
+A solver configuration in a benchmark cannot override `environmentMode` in any of its phases,
+even though xref:running-timefold-solver/solver-diagnostics.adoc#environmentMode[the solver] allows it.
```
triceo:
```suggestion
A solver configuration in a benchmark must not override `environmentMode` in any of its phases.
```

docs/src/modules/ROOT/pages/running-timefold-solver/solver-diagnostics.adoc
```diff
@@ -210,6 +210,43 @@ If your production environment doesn't care about reproducibility, use this mode
...
+Any number of phases can override it while
+the solver's global environment mode still applies outside these phases.
+If the solver's global environment mode is `<<environmentModeNonReproducible,NON_REPRODUCIBLE>>`,
+no phase can override it,
+because every other mode is <<environmentModeReproducibility,reproducible>> and therefore stricter.
+
+NOTE: A phase that runs in a different environment mode than the solver gets its own score director.
+With the xref:constraints-and-score/score-calculation.adoc#constraintStreams[Constraint Streams] API,
+that means a second constraint network is built for that mode.
+It is built once and reused, but it is not free: prefer overriding the phases you actually want to inspect.
```
triceo:
```suggestion
It is built once and reused, but it is not free: only override the phases you actually want to inspect.
```

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -96,34 +97,50 @@ public abstract class AbstractScoreDirector<Solution_, Score_ extends Score<Scor
...
-                : null;
-        this.valueRangeManager = new ValueRangeManager<>(solutionDescriptor);
+        // When using a list variable,
+        // we ensure that the listVariableStateSupply is initialized,
+        // as it will serve as the single source of truth for all other classes.
         var listVariableDescriptor = solutionDescriptor.getListVariableDescriptor();
         if (listVariableDescriptor == null) {
             this.listVariableStateSupply = null;
         } else {
             this.listVariableStateSupply = getSupplyManager().demand(listVariableDescriptor.getStateDemand());
```
triceo, line 127: Remember to remove the supply-demand mechanism from there, as agreed.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
zepfred: `solvingEnded` is called after the solver completes all phases. However, the solver can still be restarted if the user submits a change to the problem. In such cases, the score director must remain open. This is why the `outerSolvingEnded` method in `DefaultSolver` exists: to release resources and update the solver status when a restart isn't needed.

```
    public final Solution_ solve(Solution_ problem) {
        // No tags for these metrics; they are global
        var solveLengthTimer = Metrics.more().longTaskTimer(SolverMetric.SOLVE_DURATION.getMeterId());
        var errorCounter = Metrics.counter(SolverMetric.ERROR_COUNT.getMeterId());

        solverScope.setInitialSolution(Objects.requireNonNull(problem, "The problem must not be null."));
        solverScope.setSolver(this);
        outerSolvingStarted(solverScope);

        var restartSolver = true;
        while (restartSolver) {
            var sample = solveLengthTimer.start();
            try {
                // solvingStarted will call registerSolverSpecificMetrics(), since
                // the solverScope need to be fully initialized to calculate the
                // problem's scale metrics
                solvingStarted(solverScope);
                runPhases(solverScope);
                solvingEnded(solverScope);
            } catch (Exception e) {
                errorCounter.increment();
                solvingError(solverScope, e);
                throw e;
            } finally {
                sample.stop();
                unregisterSolverSpecificMetrics();
            }
            restartSolver = checkProblemChanges();
        }
        outerSolvingEnded(solverScope);
        return solverScope.getBestSolution();
    }
```

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
zepfred: Other components fetch the score director factory through the `DefaultSolverFactory`, including `TimefoldProcessor`, `DefaultTimefoldBeanProvider`, and `DefaultSolutionManager`. Some of them expect the concrete factories and will fail if we return the multi-environment factory instead. `DefaultSolverFactory` is closer to the multi-environment factory, and that's why I'm applying the casting here.

We have two options: 
1. Apply the casting to the other components that expect the concrete factory.
2. Create a new multi-env contract and have the concrete factories implement it. This way, there will be no multi-env decorator.

core/src/main/java/ai/timefold/solver/core/impl/solver/EnvironmentModeUtil.java
zepfred: The formatter adds the line break

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
zepfred: Yes, that's going to be my next task right after I finish the PR and the pending review.

core/src/main/java/ai/timefold/solver/core/impl/solver/DefaultSolverFactory.java
triceo: Once again, you should look at this from the point of complexity.

Does the solver layer need to know and think about the fact that score directors (one level below) need to perhaps do something special in terms of environment modes? This is effectively a leaky abstraction.

core/src/main/java/ai/timefold/solver/core/impl/solver/AbstractSolver.java
triceo: This hidden solver restart thing is very messy, but that's not caused by this PR, nor should it be fixed here.

That said, there is a real inconsistency in the solver context. If it's has to be stopped by `close`, then it must not be started by `solvingStarted`. _Everywhere_ in the solver excepts those things to come in pairs - started/ended. Breaking convention has to have a good reason; I do not see one here.

core/src/main/java/ai/timefold/solver/core/impl/score/director/stream/BavetConstraintStreamScoreDirectorFactory.java
```diff
@@ -99,13 +103,7 @@ public ConstraintMetaModel getConstraintMetaModel() {
     }
 
     @Override
-    public BavetConstraintStreamScoreDirector.Builder<Solution_, Score_> createScoreDirectorBuilder() {
-        return new BavetConstraintStreamScoreDirector.Builder<>(this);
+    public Builder<Solution_, Score_> createScoreDirectorBuilder(EnvironmentMode environmentMode) {
+        return new Builder<>(this, environmentMode);
     }
```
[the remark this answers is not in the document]
triceo: Why would it even allow that? Now that we are building entire new factories...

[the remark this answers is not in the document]
zepfred: The `BavetConstraintStreamScoreDirectorFactory` factory is used only when there isn't more than one phase, so this check is unnecessary.

triceo (approved):
Made some finishing touches myself. (Sonar etc.)

PR #2609, opened 2026-08-21
chore: add docs, fail fast on execute for structural flawed solutions, make ScoreAnalysis not fail fact on structural flawed input

Christopher-Chianelli:
For the record, I strongly dislike this change to `ScoreAnalysis` (returning with structural flaws instead of throwing an exception), and it will force the old implementation of the graph to live forever in some ways.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
@@ -2,19 +2,19 @@ package ai.timefold.solver.core.api.domain.variable;
 
 import java.util.List;
 
-import ai.timefold.solver.core.api.score.analysis.LoopedVariableInfo;
+import ai.timefold.solver.core.api.score.analysis.VariableLoop;
 
 import org.jspecify.annotations.NullMarked;
 
 @NullMarked
 public class InconsistentSolutionException extends RuntimeException {
     private final Object solution;
-    private final List<LoopedVariableInfo> inconsistentGroups;
+    private final List<VariableLoop> variableLoops;
 
-    public InconsistentSolutionException(String feature, Object solution, List<LoopedVariableInfo> inconsistentGroups) {
+    public InconsistentSolutionException(String feature, Object solution, List<VariableLoop> variableLoops) {
         super("The solution (%s) is inconsistent. %s requires a consistent solution.".formatted(solution, feature));
         this.solution = solution;
-        this.inconsistentGroups = inconsistentGroups;
+        this.variableLoops = variableLoops;
     }
 
     @SuppressWarnings("unchecked")
@@ -22,7 +22,7 @@ public class InconsistentSolutionException extends RuntimeException {
         return (T) solution;
     }
 
-    public List<LoopedVariableInfo> getInconsistentGroups() {
-        return inconsistentGroups;
+    public List<VariableLoop> getVariableLoops() {
+        return variableLoops;
     }
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
@@ -1,19 +1,20 @@
 package ai.timefold.solver.core.api.domain.variable;
 
-import java.util.Collection;
 import java.util.List;
+
+import ai.timefold.solver.core.api.score.analysis.LoopedVariableInfo;
 
 import org.jspecify.annotations.NullMarked;
 
 @NullMarked
 public class InconsistentSolutionException extends RuntimeException {
     private final Object solution;
-    private final Collection<Object> involvedEntityCollection;
+    private final List<LoopedVariableInfo> inconsistentGroups;
 
-    public InconsistentSolutionException(String feature, Object solution, Collection<Object> involvedEntityCollection) {
+    public InconsistentSolutionException(String feature, Object solution, List<LoopedVariableInfo> inconsistentGroups) {
         super("The solution (%s) is inconsistent. %s requires a consistent solution.".formatted(solution, feature));
         this.solution = solution;
-        this.involvedEntityCollection = involvedEntityCollection;
+        this.inconsistentGroups = inconsistentGroups;
     }
 
     @SuppressWarnings("unchecked")
@@ -21,8 +22,7 @@ public class InconsistentSolutionException extends RuntimeException {
         return (T) solution;
     }
 
-    @SuppressWarnings("unchecked")
-    public <T> List<T> getInvolvedEntityCollection() {
-        return (List<T>) involvedEntityCollection;
+    public List<LoopedVariableInfo> getInconsistentGroups() {
+        return inconsistentGroups;
     }
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
@@ -7,7 +7,7 @@ import ai.timefold.solver.core.api.score.analysis.VariableLoop;
 import org.jspecify.annotations.NullMarked;
 
 @NullMarked
-public class InconsistentSolutionException extends RuntimeException {
+public final class InconsistentSolutionException extends RuntimeException {
     private final Object solution;
     private final List<VariableLoop> variableLoops;
 
--- a/core/src/main/java/ai/timefold/solver/core/api/domain/variable/ShadowVariablesInconsistent.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/domain/variable/ShadowVariablesInconsistent.java
@@ -9,6 +9,7 @@ import java.lang.annotation.Target;
 
 import ai.timefold.solver.core.api.domain.entity.PlanningEntity;
 import ai.timefold.solver.core.api.domain.solution.PlanningSolution;
+import ai.timefold.solver.core.api.score.Score;
 import ai.timefold.solver.core.api.score.stream.Constraint;
 
 /**
@@ -129,8 +130,13 @@ import ai.timefold.solver.core.api.score.stream.Constraint;
  * Do not use a {@link ShadowVariablesInconsistent} property in a method annotated with {@link ShadowSources}.
  * {@link ShadowSources} marked methods do not need to check {@link ShadowVariablesInconsistent} properties,
  * since they are only called if all their dependencies are consistent.
+ *
+ * @deprecated The introduction of {@link Score#structuralScore()} removed the need for this annotation.
+ *             If you currently have this annotation on a property, you are encouraged to remove
+ *             it to have simpler constraints and faster solve speeds.
  */
 @Target({ METHOD, FIELD })
 @Retention(RUNTIME)
+@Deprecated(since = "2.6.0")
 public @interface ShadowVariablesInconsistent {
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/domain/variable/ShadowVariablesInconsistent.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/domain/variable/ShadowVariablesInconsistent.java
@@ -137,6 +137,6 @@ import ai.timefold.solver.core.api.score.stream.Constraint;
  */
 @Target({ METHOD, FIELD })
 @Retention(RUNTIME)
-@Deprecated(since = "2.6.0")
+@Deprecated(since = "2.7.0")
 public @interface ShadowVariablesInconsistent {
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/score/analysis/EntityVariablePair.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/analysis/EntityVariablePair.java
@@ -1,0 +1,17 @@
+package ai.timefold.solver.core.api.score.analysis;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A pair of an entity and a variable on it.
+ * 
+ * @param entity The entity.
+ * @param variableName The variable on the entity.
+ */
+@NullMarked
+public record EntityVariablePair(Object entity, String variableName) {
+    @Override
+    public String toString() {
+        return "%s.%s".formatted(entity, variableName);
+    }
+}
--- a/core/src/main/java/ai/timefold/solver/core/api/score/analysis/LoopedVariableInfo.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/analysis/LoopedVariableInfo.java
@@ -1,0 +1,31 @@
+package ai.timefold.solver.core.api.score.analysis;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A set of entity-variable pairs that form a cycle.
+ * 
+ * @param involvedVariableSet
+ */
+@NullMarked
+public record LoopedVariableInfo(Set<EntityVariablePair> involvedVariableSet) {
+    /**
+     * Get the set of involved entities in the cycle
+     */
+    @SuppressWarnings("unchecked")
+    public <T> Set<T> getEntitySet() {
+        return (Set<T>) involvedVariableSet.stream()
+                .map(EntityVariablePair::entity)
+                .collect(Collectors.toSet());
+    }
+
+    @Override
+    public String toString() {
+        return involvedVariableSet.stream()
+                .map(EntityVariablePair::toString)
+                .collect(Collectors.joining(", ", "[", "]"));
+    }
+}
--- a/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
@@ -1,0 +1,17 @@
+package ai.timefold.solver.core.api.score.analysis;
+
+import java.util.Collection;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Represents a breakdown of the structural flaws of a solution.
+ */
+@NullMarked
+public interface StructuralFlawAnalysis {
+    /**
+     * Return a collection of {@link ai.timefold.solver.core.api.domain.entity.PlanningEntity}
+     * that have inconsistent shadow variables.
+     */
+    Collection<Object> getInconsistentEntities();
+}
--- a/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
@@ -10,8 +10,8 @@ import org.jspecify.annotations.NullMarked;
 @NullMarked
 public interface StructuralFlawAnalysis {
     /**
-     * Return a list of independent {@link LoopedVariableInfo}
+     * Return a list of independent {@link VariableLoop}
      * that form cycles and thus cause inconsistencies in the solution.
      */
-    List<LoopedVariableInfo> getInconsistentGroups();
+    List<VariableLoop> getVariableLoops();
 }
--- a/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
@@ -1,6 +1,6 @@
 package ai.timefold.solver.core.api.score.analysis;
 
-import java.util.Collection;
+import java.util.List;
 
 import org.jspecify.annotations.NullMarked;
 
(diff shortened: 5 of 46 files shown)


core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * Represents a breakdown of the structural flaws of a solution.
+ */
+@NullMarked
+public interface StructuralFlawAnalysis {
+    /**
+     * Return a collection of {@link ai.timefold.solver.core.api.domain.entity.PlanningEntity}
```
triceo: No FQNs please.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
```diff
@@ -0,0 +1,17 @@
...
+     * that have inconsistent shadow variables.
+     */
+    Collection<Object> getInconsistentEntities();
```
triceo: Can we separate this into individual loops?
IMO that would make some sense - it would allow for better debugging of what is looping.

triceo: Also, I would prefer if this returned a `SequencedSet` - clearer semantics, without having to explain.

core/src/main/java/ai/timefold/solver/core/api/solver/phase/PhaseCommandContext.java
```diff
@@ -53,6 +53,7 @@ public interface PhaseCommandContext<Solution_>
      * without recalculating the score for performance reasons.
      *
      * @param move the move to execute
+     * @throws IllegalArgumentException if the move causes the solution to have a negative {@link Score#structuralScore()}.
```
triceo: IMO it should by explained somewhere _why_ this exception is there and what the user can do to avoid it.

core/src/main/java/ai/timefold/solver/core/enterprise/TimefoldSolverEnterpriseService.java
```diff
@@ -226,7 +228,10 @@ <Solution_> DestinationSelector<Solution_> applyNearbySelection(DestinationSelec
     InnerConstraintProfiler buildConstraintProfiler();
 
     <Score_ extends Score<Score_>> ScoreAnalysis<Score_> analyze(InnerScore<Score_> state,
-            Map<ConstraintRef, ConstraintMatchTotal<Score_>> constraintMatchTotalMap, ScoreAnalysisFetchPolicy fetchPolicy);
+            Map<ConstraintRef, ConstraintMatchTotal<Score_>> constraintMatchTotalMap,
+            Collection<Object> inconsistentEntities,
```
triceo: Since I'd like this to be separated into individual loops, arguably an internal wrapping type would work better.

core/src/main/java/ai/timefold/solver/core/api/solver/phase/PhaseCommandContext.java
```diff
@@ -74,10 +76,24 @@ public interface PhaseCommandContext<Solution_>
...
+    /**
+     * As defined by {@link #executeTemporarily(Move, Function)},
+     * except having a separate consumer to handle solutions with a negative {@link Score#structuralScore()}.
+     *
+     * @param move the move to execute temporarily
+     * @param temporarySolutionConsumer the consumer to execute with the temporarily modified solution;
+     *        this solution must not be modified any further.
+     * @return the result of the consumer
+     */
+    <Result_> @Nullable Result_ executeTemporarilyHandlingStructurallyFlawedSolutions(Move<Solution_> move,
```
triceo: The Javadoc here is missing `param structurallyFlawedSolutionConsumer` - what should the consumer do? How does it interact with the other consumer? IMO this method is significantly under-documented.

core/src/main/java/ai/timefold/solver/core/impl/phase/custom/DefaultPhaseCommandContext.java
```diff
@@ -75,6 +84,15 @@ public <Score_ extends Score<Score_>> Score_ executeTemporarily(Move<Solution_>
         return moveDirector.executeTemporary(move, temporarySolutionConsumer, true);
     }
 
+    @Override
+    public @Nullable <Result_> Result_ executeTemporarilyAndCalculateScoreHandlingStructurallyFlawedSolutions(
```
triceo: This name is unacceptably long. (You could argue the others already are as well, but this one is clearly extreme.)

core/src/main/java/ai/timefold/solver/core/impl/score/director/InnerScoreDirector.java
```diff
@@ -331,6 +331,11 @@ default void forceTriggerVariableListeners() {
         forceUpdateShadowVariables();
     }
 
+    /**
+     * @return true if the last {@link #updateShadowVariables()} was successful, false otherwise.
```
triceo: Define "successful".

core/src/main/java/ai/timefold/solver/core/enterprise/TimefoldSolverEnterpriseService.java
```diff
@@ -226,7 +228,10 @@ <Solution_> DestinationSelector<Solution_> applyNearbySelection(DestinationSelec
...
+            @Nullable ScoreDefinition<Score_> scoreDefinition,
```
triceo, line 237: Why nullable? When? IMO should be documented.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
```diff
@@ -55,11 +55,11 @@ public void solvingStarted(SolverScope<Solution_> solverScope) {
...
-        if (innerScore.isInvalid()) {
+        if (innerScore.isStructurallyFlawed()) {
             LOGGER.warn("The initial solution passed to the solver is inconsistent. Unassigning involved entities.");
             scoreDirector.unassignInconsistentEntities();
             innerScore = scoreDirector.calculateScore();
-            if (innerScore.isInvalid()) {
+            if (innerScore.isStructurallyFlawed()) {
                 // If there were a fixed dependency loop, the shadow variable session would fail fast before here
                 throw new IllegalStateException(
                         "Impossible state: The initial solution passed to the solver is inconsistent even after unassigning involved entities.");
```
triceo, line 65: Is this actually "impossible state"? If this could be user error, then the "impossible state" prefix should not be there - that is reserved for unforeseen flaws in our own logic.

docs/src/modules/ROOT/pages/constraints-and-score/understanding-the-score.adoc
```diff
@@ -91,6 +91,13 @@ For performance reasons and especially with large datasets that you'll later nee
 In that case, use `ScoreAnalysisFetchPolicy.FETCH_MATCH_COUNT` instead of the default `ScoreAnalysisFetchPolicy.FETCH_ALL` when calling `SolutionManager.analyze(...)`.
 ====
 
+[NOTE]
+====
+The score analysis of structurally flawed solutions has a structural flaw analysis,
+which can be used to get inconsistent entities.
+Constraint analyses are still available for structurally flawed solutions, but will exclude any matches sourced from an inconsistent entity.
+====
```
triceo:
```suggestion
NOTE: The score analysis of structurally flawed solutions has a structural flaw analysis,
which can be used to get inconsistent entities.
Constraint analyses are still available for structurally flawed solutions, but will exclude any matches sourced from an inconsistent entity.
```

A more concise alternative where you do not need to remember the closing statement, therefore avoiding bugs from not closing the block.

For these simple blocks, it works well. For the latter blocks of multiple paragraphs, the verbose syntax is much better.

docs/src/modules/ROOT/pages/domain-modeling/modeling-planning-problems.adoc
```diff
@@ -2035,7 +2035,35 @@ For example, if variable `a` depends on `b` and `b` depends on `a`, both are in
 - It depends on another variable that is inconsistent.
 For example, if `c` depends on `a`, and `a` is in a loop with `b`, then `c` is also considered part of the loop.
 
-When a declarative shadow variable is inconsistent, it will be set to `null`.
+When a declarative shadow variable is inconsistent, the solver's reaction depends on whether any entity in the model declares a `@ShadowVariablesInconsistent` field.
```
triceo, line 2038: IMO there should be a place (where exactly?) in which we discuss:

- The recommendations around this. (In other words: drop this, get a significantly faster solve.)
- The consequences. (If you have the annotation, you need to have a constraint based on the field. If you do not have the annotation, you do not need any constraint, because you get a structural score.)

Methinks we should slightly restructure the section; guide people from the start towards the annotation-less approach, and only mention the annotation once in a corner somewhere, with the explicit mention of it being deprecated. (In other words: the structural approach is _the default_, anything else is a legacy behavior which should only be mentioned, not promoted.)

triceo (commented):
Leaving some comments.

Your comment about keeping the graph forever is noted. It is an implementation downside of a requirement which makes sense for users; we take the pain. Going forward, maybe we refactor the graph algorithm; it is fast becoming marginal, maybe we do not need all the incrementality anymore. Anyway, not for now.

core/src/main/java/ai/timefold/solver/core/impl/solver/recaller/BestSolutionRecaller.java
Christopher-Chianelli: Is impossible state; should fail-fast earlier inside the variable session creator

core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
Christopher-Chianelli: It cannot be sequenced; loops are strongly connected components, and are not always of the form `x1 -> x2 -> ... -> x1`. There may be shortcuts:

- `x1 -> x2 -> x3 -> x4 -> x1`,  `x1 -> y -> x1` and `x2 -> y`. `Sequenced` implies an ordering which is not possible for strongly connected components -- it is literally an "impossible to order" situlation.

triceo: I am not asking for any _meaningful_ ordering. I am asking for _reproducible_ ordering - the difference between `HashSet` and `LinkedHashSet`. Is that not possible?

core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
```diff
@@ -1,28 +1,28 @@
...
 
-import java.util.Collection;
 import java.util.List;
 
+import ai.timefold.solver.core.api.score.analysis.LoopedVariableInfo;
+
 import org.jspecify.annotations.NullMarked;
 
 @NullMarked
 public class InconsistentSolutionException extends RuntimeException {
```
triceo: Do we need to expose this exception?
I'm not excited about putting exception types in our public API.
If we absolutely have to expose it, at least make it `final`.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/ShadowVariablesInconsistent.java
```diff
@@ -129,8 +130,13 @@
...
  * {@link ShadowSources} marked methods do not need to check {@link ShadowVariablesInconsistent} properties,
  * since they are only called if all their dependencies are consistent.
+ *
+ * @deprecated The introduction of {@link Score#structuralScore()} removed the need for this annotation.
+ *             If you currently have this annotation on a property, you are encouraged to remove
+ *             it to have simpler constraints and faster solve speeds.
  */
 @Target({ METHOD, FIELD })
 @Retention(RUNTIME)
+@Deprecated(since = "2.6.0")
```
triceo:
```suggestion
@Deprecated(since = "2.7.0")
```

Time moves on.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/LoopedVariableInfo.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A set of entity-variable pairs that form a cycle.
+ * 
+ * @param involvedVariableSet
+ */
+@NullMarked
+public record LoopedVariableInfo(Set<EntityVariablePair> involvedVariableSet) {
```
triceo: We need a better name for this. "Info" is one of those useless words that doesn't really say anything.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/EntityVariablePair.java
```diff
@@ -0,0 +1,17 @@
...
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A pair of an entity and a variable on it.
+ * 
+ * @param entity The entity.
+ * @param variableName The variable on the entity.
+ */
+@NullMarked
+public record EntityVariablePair(Object entity, String variableName) {
```
triceo, line 12: This will lead to a very verbose JSON.
I question if we need the variable information at all.
If we do, then arguably we can list entities per variable, as opposed to listing the variable with every entity.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Represents a breakdown of the structural flaws of a solution.
+ */
+@NullMarked
+public interface StructuralFlawAnalysis {
+    /**
+     * Return a list of independent {@link LoopedVariableInfo}
+     * that form cycles and thus cause inconsistencies in the solution.
+     */
+    List<LoopedVariableInfo> getInconsistentGroups();
```
triceo: See, this is the naming issue right here. `getInconsistentGroups` returns `LoopedVariableInfo`s. We're using many names for the same concept.

How about `List<EntityLoop> getEntityLoops()` - short and clear.

core/src/main/java/ai/timefold/solver/core/api/solver/phase/PhaseCommandContext.java
```diff
@@ -53,6 +53,10 @@ public interface PhaseCommandContext<Solution_>
      * without recalculating the score for performance reasons.
      *
      * @param move the move to execute
+     * @throws IllegalArgumentException if the move causes the solution to have a negative
+     *         {@link Score#structuralScore()}. If you are unsure if a move will result in a structural
+     *         solution, use {@link #executeTemporarily(Move)} to check
```
triceo: What is a "structural solution"? I think you're missing a word there.
(Also below.)

core/src/main/java/ai/timefold/solver/core/enterprise/TimefoldSolverEnterpriseService.java
```diff
@@ -225,8 +227,15 @@ <Solution_> DestinationSelector<Solution_> applyNearbySelection(DestinationSelec
 
     InnerConstraintProfiler buildConstraintProfiler();
 
+    /**
+     * @param inconsistentEntities the entities that are inconsistent
```
triceo:
```suggestion
     * @param inconsistentEntities the entities that are in a loop
```

I'm still not sure why we're saying "loop" in some places and "inconsistent" in others. We should make a choice, and stick to it.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/DefaultShadowVariableSession.java
```diff
@@ -53,7 +54,7 @@ public boolean updateVariables() {
         return graph.updateChanged();
     }
 
-    public Collection<Object> getInconsistentEntities() {
-        return graph.getInconsistentEntities();
+    public List<LoopedVariableInfo> getInconsistentGroups() {
```
triceo: Same naming question as above.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/VariableReferenceGraph.java
```diff
@@ -71,5 +71,5 @@ default void afterListVariableChanged(VariableMetaModel<?, ?, ?> variableReferen
         // Most graphs do not have edges that depend on a list variable's contents.
     }
 
-    Collection<Object> getInconsistentEntities();
+    List<LoopedVariableInfo> getInconsistentGroups();
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ShadowVariableSupport.java
```diff
@@ -423,12 +425,11 @@ public boolean updateShadowVariables() {
         return true;
     }
 
-    public Collection<Object> getInconsistentEntities() {
+    public List<LoopedVariableInfo> getInconsistentGroups() {
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -345,12 +347,16 @@ protected void afterSetWorkingSolution() {
         // Do nothing
     }
 
-    public Collection<Object> computeInconsistentEntities() {
-        return shadowVariableSupport.getInconsistentEntities();
+    public List<LoopedVariableInfo> computeInconsistentGroups() {
```
triceo: Dtto.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -1028,6 +1039,12 @@ public Builder_ withExpectShadowVariablesInCorrectState(boolean expectShadowVari
             return (Builder_) this;
         }
 
+        @SuppressWarnings("unchecked")
+        public Builder_ withForceAllowInconsistentSolutions(boolean forceAllowInconsistentSolutions) {
```
triceo, line 1043: Here's what we have now, when it comes to solutions:

- "inconsistent"
- "looped"
- "structurally flawed"

They all mean the same thing.
Thinking out loud here:

"Structural flaw" is a generic concept which we should keep; it may even be extended in the future by adding new kinds of flaws.
"Inconsistent" and "looped" are different names for the same thing - IMO we should only use one of them, and _only_ if we specifically want to talk about loops.
In all other cases - such as here - we should IMO talk strictly and only about "structurally flawed".

triceo (approved):
We have a naming issue to solve.
(Also, a bit of a confusion in terms.)

Other than that, I think we're good.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/StructuralFlawAnalysis.java
Christopher-Chianelli: You can have `reproducible` ordering without exposing the type as `Sequenced`; the actual instance is `LinkedHashSet`.

core/src/main/java/ai/timefold/solver/core/api/domain/variable/InconsistentSolutionException.java
Christopher-Chianelli: Yes, for the platform so they can send a more meaningful error message/read the inconsistent entities.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/EntityVariablePair.java
Christopher-Chianelli: If you want the individual loops, you need the variables, and the loop might be `entity1:a -> entity2:b -> entity1:c`.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
Christopher-Chianelli: `Inconsistent` more aptly describes the flaw and has prior usage in the solver API (`@ShadowVariablesInconsistent`).

Christopher-Chianelli: Although we also use `loops` a lot, and the existing exception messages mention looped. So if anything, the `@ShadowVariablesInconsistent` annotation (which is the only existing/currently published public API) is inconsistent with everything else.

Christopher-Chianelli: Choose `VariableLoop` in the end, since ultimately the loop is formed by variables pointing to each others, not entities. Although Some usage of inconsistent have to remain in the codebase to deal with `@ShadowVariablesInconsistent` (which is now deprecated, so those usage of inconsistent instead of looped can be removed when that annotation is removed).

PR #2617, opened 2026-08-25
fix: false positive warnings for new properties

### LINKED ISSUE #2616: False positive warnings for new property names
The property renaming introduced [TimefoldRenamedPropertiesFallbackInterceptor](https://github.com/TimefoldAI/timefold-solver/blob/main/service/quarkus/runtime/src/main/java/ai/timefold/solver/service/quarkus/deployment/config/TimefoldRenamedPropertiesFallbackInterceptor.java) which produces many false-positive warnings.

The reason is that the warning is emitted every time the property is looked up in the map of current property names to the deprecated property names, such as [here](https://github.com/TimefoldAI/timefold-solver/blob/main/service/quarkus/runtime/src/main/java/ai/timefold/solver/service/quarkus/deployment/config/TimefoldRenamedPropertiesFallbackInterceptor.java#L63). Instead, the warning should be emitted when the deprecated property has actually been resolved.

Follow up after https://github.com/TimefoldAI/timefold-solver/issues/2496.

winklerm:
The warning must only be logged when a legacy value is actually chosen over the new one, not merely because the requested name happens to have a known legacy counterpart (which may itself be unset).

Closes https://github.com/TimefoldAI/timefold-solver/issues/2616.

### DIFF
--- a/service/quarkus/runtime/src/main/java/ai/timefold/solver/service/quarkus/deployment/config/TimefoldRenamedPropertiesFallbackInterceptor.java
+++ b/service/quarkus/runtime/src/main/java/ai/timefold/solver/service/quarkus/deployment/config/TimefoldRenamedPropertiesFallbackInterceptor.java
@@ -60,14 +60,11 @@ public class TimefoldRenamedPropertiesFallbackInterceptor extends FallbackConfig
     private static String toLegacyName(String name) {
         var exactLegacyName = RENAMED_EXACT_PROPERTIES.get(name);
         if (exactLegacyName != null) {
-            logWarningForResolvedKey(name, exactLegacyName);
             return exactLegacyName;
         }
         for (Map.Entry<String, String> prefix : RENAMED_PROPERTY_PREFIXES.entrySet()) {
             if (name.startsWith(prefix.getKey())) {
-                var legacyName = prefix.getValue() + name.substring(prefix.getKey().length());
-                logWarningForResolvedKey(name, legacyName);
-                return legacyName;
+                return prefix.getValue() + name.substring(prefix.getKey().length());
             }
         }
         return name;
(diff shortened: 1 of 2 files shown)


greyhairredbear (commented):
Nice tests about the logging!

Q: Would it maybe make sense to have that parameterized (with the other new/legacy property names), or do you think this would basically just be duplicating the implementation i.e. the mapping from new name to legacy name?
(IMO, it could make sense in this instance, thus the suggestion)

winklerm:
> Nice tests about the logging!
> 
> Q: Would it maybe make sense to have that parameterized (with the other new/legacy property names), or do you think this would basically just be duplicating the implementation i.e. the mapping from new name to legacy name? (IMO, it could make sense in this instance, thus the suggestion)

@greyhairredbear Do you mean having the tests parameterized with all mapped properties in general, or just the logging tests?

greyhairredbear:
> @greyhairredbear Do you mean having the tests parameterized with all mapped properties in general, or just the logging tests?

I just thought about the logging here, but you've got a point, might as well extend it to the other tests 🤔, not really that much of a difference. But then, we might also descope it from this PR.

winklerm:
> > @greyhairredbear Do you mean having the tests parameterized with all mapped properties in general, or just the logging tests?
> 
> I just thought about the logging here, but you've got a point, might as well extend it to the other tests 🤔, not really that much of a difference. But then, we might also descope it from this PR.

Got it, updated the logging tests. The remaining tests seem to already cover different mapping types, if I have noticed correctly, so we should be good there.

winklerm:
Tested with FSR locally, looks good.

greyhairredbear (approved):
Nice work, thanks for cleaning this up!

triceo (approved):
Approving in Radek's absence. Was validated with models.

PR #2618, opened 2026-08-26
fix: false positive warnings for new properties (#2617)

winklerm:
The warning must only be logged when a legacy value is actually chosen over the new one, not merely because the requested name happens to have a known legacy counterpart (which may itself be unset).

Closes https://github.com/TimefoldAI/timefold-solver/issues/2616.

Backport of https://github.com/TimefoldAI/timefold-solver/pull/2617.


triceo (approved):
Cherry-pick from `main`.

PR #2630, opened 2026-09-01
perf: improve move evaluation speed by ignoring calculating the score for inconsistent solutions

Christopher-Chianelli:


core/src/main/java/ai/timefold/solver/core/api/score/analysis/EntityVariablePair.java
```diff
@@ -0,0 +1,17 @@
+package ai.timefold.solver.core.api.score.analysis;
+
+import org.jspecify.annotations.NullMarked;
+
+/**
+ * A pair of an entity and a variable on it.
```
triceo, line 6: I would add a `@see` to the parent type, to make it obvious what this interacts with.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/ScoreAnalysis.java
```diff
@@ -86,6 +86,21 @@ public interface ScoreAnalysis<Score_ extends Score<Score_>> {
...
+     * Indicates whether the solution was structurally flawed at the time of analysis.
+     *
+     * @return isSolutionStructurallyFlawed true if the solution was structurally flawed at the time of analysis.
+     */
+    boolean isSolutionStructurallyFlawed();
+
+    /**
+     * Returns an analysis of the structural flaws of a structurally flawed solution.
+     * 
+     * @return null if the solution was not structurally flawed at the time of analysis.
```
triceo, line 101: At this point, a reference to `#isSolutionStructurallyFlawed()` may suffice.

core/src/main/java/ai/timefold/solver/core/api/score/analysis/VariableLoop.java
```diff
@@ -0,0 +1,31 @@
...
+ * 
+ * @param involvedVariableSet
+ */
+@NullMarked
+public record VariableLoop(Set<EntityVariablePair> involvedVariableSet) {
+    /**
+     * Get the set of involved entities in the cycle
+     */
+    @SuppressWarnings("unchecked")
+    public <T> Set<T> getEntitySet() {
```
triceo: Since we are a `record`, probably `entitySet()` would be a more consistent name. (Yeah, I hate this aspect of records too.)

core/src/main/java/ai/timefold/solver/core/api/score/Score.java
```diff
@@ -19,6 +19,15 @@ public sealed interface Score<Score_ extends Score<Score_>>
         extends Comparable<Score_>
         permits HardMediumSoftBigDecimalScore, HardMediumSoftScore, HardSoftBigDecimalScore, HardSoftScore, IBendableScore,
         SimpleBigDecimalScore, SimpleScore {
+    /**
+     * The structural component of a Score. 0 if the solution is structurally sound
+     * (i.e. no inconsistent variables), -1 otherwise. This behaves as a super hard Score;
+     * a Score with a negative structural score is worse than all scores with a zero
```
triceo: In this case, it should be "score", not "Score".
Structural score is a long, not `instanceof Score`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ConsistencyTracker.java
```diff
@@ -23,9 +23,11 @@ private ConsistencyTracker(boolean isFrozen) {
     }
 
     public static <Solution_> ConsistencyTracker<Solution_> frozen(SolutionDescriptor<Solution_> solutionDescriptor,
+            boolean ignoreInconsistentSolutions,
             Object[] entityOrFacts) {
         var out = new ConsistencyTracker<Solution_>(true);
-        out.setUnknownConsistencyFromEntityShadowVariablesInconsistent(solutionDescriptor, entityOrFacts);
+        out.setUnknownConsistencyFromEntityShadowVariablesInconsistent(solutionDescriptor, ignoreInconsistentSolutions,
```
triceo: Method name overly long, and even then I still don't understand what the method will do.

core/src/main/java/ai/timefold/solver/core/impl/score/director/AbstractScoreDirector.java
```diff
@@ -344,6 +367,67 @@ protected void afterSetWorkingSolution() {
         // Do nothing
     }
 
+    public List<VariableLoop> computeVariableLoops() {
+        return shadowVariableSupport.getVariableLoops();
+    }
+
+    public void unassignInconsistentEntities() {
```
triceo, line 374: Shouldn't we log what entities we've unassigned, if any?

docs/src/modules/ROOT/pages/domain-modeling/modeling-planning-problems.adoc
```diff
@@ -2157,11 +2157,46 @@ For example, if variable `a` depends on `b` and `b` depends on `a`, both are in
 - It depends on another variable that is inconsistent.
 For example, if `c` depends on `a`, and `a` is in a loop with `b`, then `c` is also considered part of the loop.
 
-When a declarative shadow variable is inconsistent, it will be set to `null`.
+When a declarative shadow variable is inconsistent, the solver's reaction depends on whether any entity in the model declares a `@ShadowVariablesInconsistent` field.
```
triceo: IMO this deserves a mention in the upgrade recipe:

- What gets deprecated.
- What changes will the user observe once the deprecated annotation is removed. 

We currently don't have a section like we used to ("Upgrading to Solver 1.Y"), so we'll have to start it, or adapt some of the other sections in the upgrade recipe.

docs/src/modules/ROOT/pages/domain-modeling/modeling-planning-problems.adoc
```diff
@@ -2157,11 +2157,46 @@ For example, if variable `a` depends on `b` and `b` depends on `a`, both are in
...
+the `Solver` logs a warning and unassigns the involved planning entities before solving.
+If the inconsistency cannot be resolved this way,
+for example because an involved entity is xref:domain-modeling/modeling-planning-problems.adoc#pinnedPlanningEntities[pinned],
+the `Solver` throws an `Exception`.
+====
+
+If no entity have a `@ShadowVariablesInconsistent` property, you do not need to handle inconsistencies in any constraints, since the Solver will not calculate the score for any structurally flawed solutions.
+
+
+==== Deprecated: Per entity inconsistency tracking
```
triceo: I wonder if we want to even mention it here.
It is deprecated, we typically avoid speaking about deprecated things, except when explaining how to avoid them.

docs/src/modules/ROOT/pages/domain-modeling/modeling-planning-problems.adoc
```diff
@@ -2157,11 +2157,46 @@ For example, if variable `a` depends on `b` and `b` depends on `a`, both are in
 - It depends on another variable that is inconsistent.
 For example, if `c` depends on `a`, and `a` is in a loop with `b`, then `c` is also considered part of the loop.
 
-When a declarative shadow variable is inconsistent, it will be set to `null`.
+When a declarative shadow variable is inconsistent, the solver's reaction depends on whether any entity in the model declares a `@ShadowVariablesInconsistent` field.
+If at least one entity declares such a field, the inconsistent shadow variables are set to `null`
+and the `@ShadowVariablesInconsistent` field of every involved entity is set to `true`.
+If no entity declares such a field, the shadow variable update fails instead
```
triceo: Does it really fail?
IMO it doesn't - the move that caused it will be ignored; or one of the other things will happen in score analysis etc.

triceo (commented):
Leaving some comments as I see this all together for the first time. Also, please check Sonar.

Have we seen this latest version actually deliver the performance we expect? 
Have we seen it doesn't introduce regressions elsewhere?

core/src/main/java/ai/timefold/solver/core/api/score/analysis/VariableLoop.java
Christopher-Chianelli: Is that an established Java pattern for records (i.e. not using `get` for methods, even if they are **not** part of the components of the records)?

triceo: I am not 100 % sure, but I can't help but see this as unwelcome inconsistency. Why should a user of the type have to understand if something is a component or a derivative, in order to figure out the right way to call it? 

IMO mixing both of these is just undesirable. (And the fact that this completely unnecessary dilemma even exists is a part of why I consider records half-baked.)

Christopher-Chianelli:
Doesn't introduced regressions: https://github.com/TimefoldAI/timefold-solver-benchmarks/actions/runs/33667129779 (although some of the jobs were cancelled since GitHub decided to cancel them in the middle of execution).

Christopher-Chianelli:
Rerunning it locally on a Job Shop Scheduling problem (i.e. dense graph) with Enterprise enabled, my CH speed went from `209/sec` to `2394/sec`, so 10x performance increase.

docs/src/modules/ROOT/pages/upgrading-timefold-solver/migration-guides/shadow-variables-inconsistent-to-structural-score.adoc
```diff
@@ -0,0 +1,90 @@
+[#migrateShadowVariablesInconsistent]
+= `@ShadowVariablesInconsistent` to structural scores
```
triceo, line 2: Good idea to do it like this!

docs/src/modules/ROOT/pages/upgrading-timefold-solver/migration-guides/shadow-variables-inconsistent-to-structural-score.adoc
```diff
@@ -0,0 +1,90 @@
...
+                // ... other constraints
+        };
+    }
+
+    // ... other constraints
+}
+----
+
+After these changes, your domain is now using the structural score approach.
+As a result, the Solver will automatically handle consistency in your domain, avoid calculating the score of inconsistent solutions, and other performance optimizations.
```
triceo: I would also mention the behavior changes that the user can now expect.
The changes to score analysis, the default attempts at unassigned, the outright fail-fasts.

triceo (approved):

PR #2631, opened 2026-09-01
feat: store execution profile in metadata

diogodanielsoaresferreira:
- Store execution profile information in metadata
- Add method to store execution profile artifacts in Worker

Related with https://github.com/TimefoldAI/timefold-platform/pull/5469


service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/Metadata.java
```diff
@@ -70,6 +70,11 @@ public final class Metadata<Score_> implements Status<Score_> {
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private String failureMessage;
 
+    @Schema(nullable = true, description = "The name of the execution profile the run was submitted with.")
+    @JsonAlias({ "executionprofile" })
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private String executionProfile;
```
cristianonicolai: Maybe make this an enum with the known profiles?

mswiderski: I would not set this on metadata but do the same as we do for configuration profile. Allow to set it via optional query parameter and then store it as system tag on the metadata. That way we keep it consistent with config profile and don't change the api contract.

mswiderski: moreover, I also think we should support activating multiple profiles at the same time, think of debug profile that changes logging and troubleshooting profile that enables JFR and stores distance matrix

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/storage/SubModelKind.java
```diff
@@ -17,7 +17,8 @@ public enum SubModelKind {
     SCORE_ANALYSIS("score-analysis"),
     SCORE_ANALYSIS_WITH_JUSTIFICATIONS("score-analysis-with-justifications"),
     WAYPOINTS("waypoints"),
-    PATCH_REQUEST("patch-request");
+    PATCH_REQUEST("patch-request"),
+    EXECUTION_PROFILE_ARTIFACTS("execution-profile-artifacts");
```
mswiderski: maybe rename this to `EXECUTION_ARTIFACTS` only, no need to mention profile, especially if we allow to activate more profiles at the same time

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/Metadata.java
diogodanielsoaresferreira: Supporting multiple execution profiles at the same time makes sense, I updated it to use it. In this case, should we just drop the default execution profile, which basically it's empty?

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/storage/SubModelKind.java
diogodanielsoaresferreira: Makes sense, updated

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/platform/EnvironmentVars.java
```diff
@@ -22,6 +22,22 @@ public class EnvironmentVars {
...
+    /**
+     * Directory the solver worker collects execution-profile diagnostic artifacts into; its contents are zipped and
+     * uploaded when the run finishes.
+     */
+    public static final String ENV_TIMEFOLD_EXECUTION_PROFILE_DIR = "AI_TIMEFOLD_EXECUTION_PROFILE_DIR";
+
+    /**
+     * Default value for {@link #ENV_TIMEFOLD_EXECUTION_PROFILE_DIR} when the variable is not set.
+     */
+    public static final String DEFAULT_EXECUTION_PROFILE_DIR = "/tmp/timefold-diagnostics";
```
cristianonicolai:
```suggestion
    public static final String DEFAULT_EXECUTION_PROFILE_DIR = "/tmp/timefold-execution-artifacts";
```

service/definition/src/main/java/ai/timefold/solver/service/definition/api/executionprofile/DefaultExecutionProfile.java
```diff
@@ -0,0 +1,26 @@
+package ai.timefold.solver.service.definition.api.executionprofile;
+
+import jakarta.enterprise.context.ApplicationScoped;
+
+/**
+ * The always-available baseline {@link ExecutionProfile}, applied when a run selects no other profile. It contributes no
+ * runtime specification of its own.
```
mswiderski: if it does not contribute in anyway does it make sense to have it at all? Maybe execution profile will become optional and if not given then not change is applied.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfileRegistry.java
```diff
@@ -0,0 +1,36 @@
...
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+
+/**
+ * Discovers all {@link ExecutionProfile} implementations available on the classpath (as CDI beans) and exposes them by name.
+ * Consumers inject this registry rather than referencing individual profiles, so the set of profiles is extensible without
+ * touching call sites.
+ */
+@ApplicationScoped
+public class ExecutionProfileRegistry {
```
mswiderski: maybe this should be in an internal package rather than api?

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/platform/EnvironmentVars.java
diogodanielsoaresferreira: Thanks, updated 👍

service/definition/src/main/java/ai/timefold/solver/service/definition/api/executionprofile/DefaultExecutionProfile.java
diogodanielsoaresferreira: Yes, I think that this is the question I asked in https://github.com/TimefoldAI/timefold-solver/pull/2631#discussion_r3933737371. I will remove it for now, and later we can start adding the needed execution profiles.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfileRegistry.java
diogodanielsoaresferreira: Makes sense for me, updated 👍

cristianonicolai (approved):

mswiderski (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/api/executionprofile/ExecutionProfile.java
```diff
@@ -0,0 +1,21 @@
+package ai.timefold.solver.service.definition.api.executionprofile;
```
rsynek: `internal` package, please.

Motivation: model developer is not supposed to implement or call the interface. This will be used purely by the solver-service and the platform.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfileRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+
+import ai.timefold.solver.service.definition.api.executionprofile.ExecutionProfile;
+
+/**
+ * Discovers all {@link ExecutionProfile} implementations available on the classpath (as CDI beans) and exposes them by name.
+ * Consumers inject this registry rather than referencing individual profiles, so the set of profiles is extensible without
+ * touching call sites.
+ */
+@ApplicationScoped
+public class ExecutionProfileRegistry {
```
rsynek: No CDI beans in the definition module, please.

The module contains public and internal API classes shared across various places in the solver-service, platform, and models. Thus, it must be safe to add on the classpath.

The moment we start adding beans that are automatically bootstrapped we cannot use the module everywhere safely anymore.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/platform/EnvironmentVars.java
```diff
@@ -22,6 +22,22 @@ public class EnvironmentVars {
...
+    /**
+     * Directory the solver worker collects execution-profile diagnostic artifacts into; its contents are zipped and
+     * uploaded when the run finishes.
+     */
+    public static final String ENV_TIMEFOLD_EXECUTION_PROFILE_DIR = "AI_TIMEFOLD_EXECUTION_PROFILE_DIR";
+
+    /**
+     * Default value for {@link #ENV_TIMEFOLD_EXECUTION_PROFILE_DIR} when the variable is not set.
+     */
+    public static final String DEFAULT_EXECUTION_PROFILE_DIR = "/tmp/timefold-execution-artifacts";
```
rsynek, line 39: I suppose this is inside the model container?

service/definition/src/main/java/ai/timefold/solver/service/definition/api/executionprofile/ExecutionProfile.java
diogodanielsoaresferreira: Done, moved to internal package

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfileRegistry.java
diogodanielsoaresferreira: Makes sense, we now load the profiles with ServiceLoader and we create the registry bean in api-gateway manually

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/platform/EnvironmentVars.java
diogodanielsoaresferreira: Yes, this is used in solver-enterprise in the ExecutionProfileModelPostProcessor to store data related with the execution profile

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfile.java
```diff
@@ -0,0 +1,33 @@
...
+ * model developers are not expected to implement or reference it. Implementations are provided by the solver service and
+ * the platform, and are discovered via {@link java.util.ServiceLoader}, so adding a new profile does not require editing
+ * any central registry. A run may activate several profiles at once.
+ */
+public interface ExecutionProfile {
+
+    /**
+     * Stable identifier of the profile, used in APIs and permissions. Must be unique across all implementations.
+     */
+    String name();
```
rsynek, line 24: Will this ever be shown in the platform UI?

If so, we need both `id` and `name` (human-readable). The same as with constraints where we learned that just the `name` is not enough to show and possibly rename in the platform UI.

rsynek (approved):
One last question inline; good to go after resolved.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfile.java
diogodanielsoaresferreira: That's a good idea, updated 👍 Thanks!

PR #2642, opened 2026-09-07
perf: decrease allocations in shadow variable processing

triceo:
In NearbyMixedSingleEntitySolverTest in the Enterprise repo, this removed 300 GB of allocations of a list iterator.

The next biggest source of allocations was Object[] with 40 GB. 
Allocations per second went from 40 GB to roughly 10 GB.


Christopher-Chianelli (approved):
LGTM. I wonder if the JVM could eventually detect this and optimize it.

PR #2646, opened 2026-09-08
refactor: standardize handling of BigDecimal in scores

triceo:
Fully backwards compatible, in some cases even marginally improving performance.

Includes an unrelated change to integer comparisons, which are now safer and together bring Sonar rating back to A.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/api/score/HardMediumSoftBigDecimalScore.java
+++ b/core/src/main/java/ai/timefold/solver/core/api/score/HardMediumSoftBigDecimalScore.java
@@ -32,19 +32,11 @@ public record HardMediumSoftBigDecimalScore(long structuralScore, BigDecimal har
             BigDecimal.ZERO, BigDecimal.ZERO);
     public static final HardMediumSoftBigDecimalScore ONE_HARD = new HardMediumSoftBigDecimalScore(BigDecimal.ONE,
             BigDecimal.ZERO, BigDecimal.ZERO);
-    private static final HardMediumSoftBigDecimalScore MINUS_ONE_HARD =
-            new HardMediumSoftBigDecimalScore(BigDecimal.ONE.negate(),
-                    BigDecimal.ZERO, BigDecimal.ZERO);
     public static final HardMediumSoftBigDecimalScore ONE_MEDIUM =
             new HardMediumSoftBigDecimalScore(BigDecimal.ZERO,
                     BigDecimal.ONE, BigDecimal.ZERO);
-    private static final HardMediumSoftBigDecimalScore MINUS_ONE_MEDIUM =
-            new HardMediumSoftBigDecimalScore(BigDecimal.ZERO,
-                    BigDecimal.ONE.negate(), BigDecimal.ZERO);
     public static final HardMediumSoftBigDecimalScore ONE_SOFT = new HardMediumSoftBigDecimalScore(BigDecimal.ZERO,
             BigDecimal.ZERO, BigDecimal.ONE);
-    private static final HardMediumSoftBigDecimalScore MINUS_ONE_SOFT =
-            new HardMediumSoftBigDecimalScore(BigDecimal.ZERO, BigDecimal.ZERO, BigDecimal.ONE.negate());
 
     public HardMediumSoftBigDecimalScore(BigDecimal hardScore, BigDecimal mediumScore,
             BigDecimal softScore) {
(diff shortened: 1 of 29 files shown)


core/src/main/java/ai/timefold/solver/core/api/score/HardMediumSoftBigDecimalScore.java
```diff
@@ -176,7 +140,7 @@ public HardMediumSoftBigDecimalScore power(double exponent) {
         var exponentBigDecimal = BigDecimal.valueOf(exponent);
         // The (unspecified) scale/precision of the exponent should have no impact on the returned scale/precision
         // TODO FIXME remove .intValue() so non-integer exponents produce correct results
-        // None of the normal Java libraries support BigDecimal.pow(BigDecimal)
+        //  None of the normal Java libraries support BigDecimal.pow(BigDecimal)
```
Christopher-Chianelli: Seems an extra space was introduced here?

core/src/main/java/ai/timefold/solver/core/api/score/HardSoftBigDecimalScore.java
```diff
@@ -131,7 +121,7 @@ public HardSoftBigDecimalScore power(double exponent) {
         var exponentBigDecimal = BigDecimal.valueOf(exponent);
         // The (unspecified) scale/precision of the exponent should have no impact on the returned scale/precision
         // TODO FIXME remove .intValue() so non-integer exponents produce correct results
-        // None of the normal Java libraries support BigDecimal.pow(BigDecimal)
+        //  None of the normal Java libraries support BigDecimal.pow(BigDecimal)
```
Christopher-Chianelli: Ditto

core/src/main/java/ai/timefold/solver/core/impl/score/ScoreUtil.java
```diff
@@ -209,6 +216,76 @@ public static <Score_ extends IBendableScore<Score_>> String buildBendableShortS
...
+     * although both are zero.
+     * <p>
+     * Zero is the only value that a score factory is permitted to replace with a shared constant.
+     * A shared constant has scale 0, and zero keeps its meaning at every scale,
+     * because zero multiplied or divided stays zero.
+     * Every other value must keep the scale that the caller supplied, because {@code multiply()},
+     * {@code divide()} and {@code power()} take the scale of their result from the scale of the input.
+     * If {@code SimpleBigDecimalScore.of(new BigDecimal("1.0"))} returned the shared {@code ONE}, which has scale 0,
+     * then {@code multiply(1.2)} would give 1 instead of 1.2.
+     *
```
Christopher-Chianelli, line 236: The default `multiply` of `BigDecimal` is full precision, with a scale of `left.scale + right.scale`: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/math/BigDecimal.html#multiply(java.math.BigDecimal) ; If I am intrepreting the above Javadoc correctly (take the scale of their result from the scale of the input), then `100 (scale 2, precision 1) * 1.5 = 100`?

Christopher-Chianelli (commented):
LGTM. Although if the Javadoc are correct, the unchanged `BigDecimal...Score`'s `multiply`/`divide` have surprising behaviour.

core/src/main/java/ai/timefold/solver/core/api/score/HardSoftBigDecimalScore.java
triceo: Both intentional. The extra space makes it treated as part of the TODO, which it is.

Christopher-Chianelli: Isn't the standard to indent it by 5 then; i.e. like this?
```
// TODO Line 1
//      Line 2
//      Line 3
```

triceo: Is this actually written down anywhere? Maybe I just missed something.
I was going off of what my IDE does - the extra space treats the whole thing as a TODO.

Christopher-Chianelli: Nope; FWIW, it is configurable in IDEA, and as far as I know, not a standard of any kind: https://www.jetbrains.com/help/idea/using-todo.html

core/src/main/java/ai/timefold/solver/core/impl/score/ScoreUtil.java
triceo: Even though this was a pre-existing issue, I decided to fix it as well. 100 * 1.5 simply should not be 100. :-)

Christopher-Chianelli (approved):

PR #2648, opened 2026-09-08
chore: remove demand-supply logic of basic and list variables

### LINKED ISSUE #2633: Remove list variables supply-demand logic

zepfred:


core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/DelegatingListVariableState.java
```diff
@@ -68,6 +66,11 @@ public ListVariableDescriptor<Solution_> getSourceVariableDescriptor() {
         return delegate.getSourceVariableDescriptor();
     }
 
+    @Override
+    public void resetWorkingSolution(InnerScoreDirector<Solution_, ?> scoreDirector) {
```
triceo, line 70: Why is this added?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
```diff
@@ -21,14 +21,19 @@ public record ChangedVariableNotifier<Solution_>(BiConsumer<VariableDescriptor<S
             },
             null);
 
-    public CollectionInverseVariableSupply getCollectionInverseVariableSupply(VariableMetaModel<?, ?, ?> variableMetaModel) {
+    public CollectionInverseVariableState getCollectionInverseVariableSupply(VariableMetaModel<?, ?, ?> variableMetaModel) {
         if (innerScoreDirector == null) {
-            return entity -> Collections.emptyList();
+            return new CollectionInverseVariableState() {
```
triceo: Arguably this can be a singleton.
(Why do we need this case anyway? Why is score director valid and null?)

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedBasicVariableState.java
```diff
@@ -15,18 +16,18 @@
 import org.jspecify.annotations.Nullable;
 
 @NullMarked
-public final class ExternalizedBasicVariableStateSupply<Solution_> implements BasicVariableStateSupply<Solution_> {
+public final class ExternalizedBasicVariableState<Solution_> implements BasicVariableState<Solution_> {
```
triceo: Why not just `Default...`? Externalization is something that happens _to it_, maybe - it isn't externalized automatically.

Also, why does the interface even exist? Can we not live with just the class? Is there anything we need to hide?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedListVariableState.java
```diff
@@ -9,64 +9,63 @@
...
 import ai.timefold.solver.core.preview.api.domain.metamodel.ElementPosition;
+import ai.timefold.solver.core.preview.api.domain.metamodel.PositionInList;
 
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
 @NullMarked
-final class ExternalizedListVariableStateSupply<Solution_>
-        implements ListVariableStateSupply<Solution_, Object, Object> {
+public final class ExternalizedListVariableState<Solution_> implements ListVariableState<Solution_, Object, Object> {
```
triceo: Same comment regarding `Default`.
The interface here may be necessary, on account of the delegating counterpart.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableProcessor.java
```diff
@@ -0,0 +1,390 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.inverserelation.InverseRelationShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.nextprev.NextElementShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.domain.variable.nextprev.PreviousElementShadowVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.impl.util.CollectionUtils;
+import ai.timefold.solver.core.preview.api.domain.metamodel.ElementPosition;
+import ai.timefold.solver.core.preview.api.domain.metamodel.PositionInList;
+
+final class ListVariableProcessor<Solution_> {
```
triceo: IMO a more fitting name would be something like `ListVariableStateCarrier`. It doesn't as much process as it carries/holds the information.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/SolverVariableSupport.java
```diff
@@ -47,13 +46,11 @@
  * @param <Solution_> the solution type, the class with the {@link PlanningSolution} annotation
  */
 @NullMarked
-public final class ShadowVariableSupport<Solution_> implements SupplyManager {
+public final class SolverVariableSupport<Solution_> implements SupplyManager {
```
triceo: Why no longer "shadow"?
And if that went away, does the word "Solver" actually help us here? Why not just `VariableSupport`?

triceo (commented):
First read-through.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/ChangedVariableNotifier.java
zepfred: The EMPTY variable passes null to the score director. Additionally, the method `ChangedVariableNotifier::empty()` returns the EMPTY value, which is utilized by both `ConsistencyTracker` and `ShadowVariableUpdateHelper`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedBasicVariableState.java
zepfred: For basic variables, there is a single implementation of `BasicVariableState`, allowing us to eliminate the interface. However, `ListVariableState` has multiple implementations, so I have chosen to maintain this contract for basic variables to ensure consistency with list variables.

zepfred: There is nothing we need to hide for this class, and the concrete implementation can be shared. If we choose to share the concrete implementation, I can rename it to `BasicVariableState`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedListVariableState.java
zepfred: For the `ListVariableState`, we have two child classes, and I confirmed that both need to exist. Therefore, we need to keep the interface, but the class will be renamed to `DefaultListVariableState`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ExternalizedBasicVariableState.java
triceo: For the sake of simplicity, I would do that.

core/src/main/java/ai/timefold/solver/core/impl/heuristic/selector/move/generic/list/kopt/DelegatingListVariableState.java
zepfred: The `ListVariableState` is stateful by definition. We have a default implementation in the interface that does nothing. I removed this implementation and added logic to make it predictable instead of quietly ignoring it.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/SolverVariableSupport.java
zepfred: I was also unhappy with the name. `ShadowVariableSupport` suggests it is limited to shadow variables only, which is misleading since we also keep change handlers for genuine variables and solution trackers. I'll change it to `VariableSupport`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/inverserelation/CollectionInverseVariableState.java
```diff
@@ -0,0 +1,23 @@
+package ai.timefold.solver.core.impl.domain.variable.inverserelation;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public interface CollectionInverseVariableState {
+
+    CollectionInverseVariableState EMPTY = new CollectionInverseVariableState() {
```
triceo, line 8: I wonder why we need this.
When exactly is `scoreDirector` null, so that this gets returned?

What use is an empty collection here, who is it going to help? Shouldn't this have been a fail-fast instead?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/VariableSupport.java
```diff
@@ -213,22 +212,83 @@ public <Supply_ extends Supply> Supply_ demand(Demand<Supply_> demand) {
         }
     }
 
+    @Override
+    public <Supply_ extends Supply> boolean cancel(Demand<Supply_> demand) {
```
triceo, line 216: We're keeping the supply here? Why?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/VariableSupport.java
```diff
@@ -213,22 +212,83 @@ public <Supply_ extends Supply> Supply_ demand(Demand<Supply_> demand) {
...
+        } else {
+            supplyMap.put(demand,
+                    new SupplyWithDemandCount(supplyWithDemandCount.supply, supplyWithDemandCount.demandCount - 1L));
+        }
+        return true;
+    }
+
+    public BasicVariableState<Solution_> getBasicVariableState(VariableDescriptor<Solution_> variableDescriptor) {
+        BasicVariableState<Solution_> state =
+                findBasicHandler(variableDescriptor, handler -> handler instanceof BasicVariableState<Solution_> handlerState
```
triceo, line 232: Isn't this inefficient? It leads to a loop. Is it ever cached in any way?

triceo (commented):
A few more comments. We're getting there!
(I am a bit surprised that, even though we're removing unnecessary functionality, we're not actually removing lines of code. That's not what should normally happen.)

zepfred:
> A few more comments. We're getting there! (I am a bit surprised that, even though we're removing unnecessary functionality, we're not actually removing lines of code. That's not what should normally happen.)

Essentially, we are removing demand definitions such as this one:
```
    @Override
    public BasicVariableStateDemand<Solution_> getStateDemand() {
        return stateDemand;
    }
```

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/inverserelation/CollectionInverseVariableState.java
zepfred: `ConsistencyTracker#setUnknownConsistencyValues` initializes with an empty `ChangedVariableNotifier` that passes a null score director. This path can lead to a request for `CollectionInverseVariableState#getInverseCollection` with a null score director. So, fail-fast would break the logic as this is valid path for `ConsistencyTracker`.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/VariableSupport.java
zepfred: `VariableSupport` still implements `SupplyManager`, and we still need `SupplyManager` in the enterprise repo. Moving this implementation to a separate class would improve separation of concerns. Is that what you have in mind?

triceo: Nevermind, I misunderstood the method.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateCarrier.java
```diff
@@ -0,0 +1,443 @@
...
+                }
+                if (externalizedPreviousElementProcessor != null) {
+                    externalizedPreviousElementProcessor.unsetElement(scoreDirector, unassignedValue);
+                }
+            }
+        }
+    }
+
+    private void addElement(Object entity, List<Object> elements, Object element, int index) {
+        if (requiresPositionMap) {
```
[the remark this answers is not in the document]
triceo: Technically correct, but practically identical to `main`. Not changing anything.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/ListVariableStateCarrier.java
```diff
@@ -0,0 +1,443 @@
...
+
+    private ExternalizedIndexVariableProcessor<Solution_> externalizedIndexProcessor = null;
+    private ExternalizedListInverseVariableProcessor<Solution_> externalizedInverseProcessor = null;
+    private ExternalizedNextPrevElementVariableProcessor<Solution_> externalizedPreviousElementProcessor = null;
+    private ExternalizedNextPrevElementVariableProcessor<Solution_> externalizedNextElementProcessor = null;
+
+    private boolean requiresPositionMap = true;
+    private InnerScoreDirector<Solution_, ?> scoreDirector;
+    private int unassignedCount = 0;
+    private Map<Object, MutablePosition> elementPositionMap;
```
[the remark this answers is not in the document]
triceo: Technically correct, but identical to `main`. Not changing anything.

triceo (approved):

PR #2649, opened 2026-09-08
perf: recording fewer, cheaper move undo steps

triceo:
Lazily allocate the variable-change list, merge a list-change's before/after undo steps into one, and reduce MoveDirector.executeTemporary() overhead.

Brings 5-10 % of extra throughput in score director benchmarks.


core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
```diff
@@ -676,32 +676,63 @@ void executeAllowingStructurallyFlawedSolutions(Move<Solution_> move) {
...
 
-    public final InnerScore<Score_> executeTemporary(Move<Solution_> move) {
+    /**
+     * Executes {@code move} temporarily,
+     * hands its score to {@code postprocessor} while the move is still applied,
+     * then undoes the move and restores the previous working score.
+     * <p>
+     * If either the move or the postprocessor throws, the move is <strong>not</strong> undone;
+     * the move's recorded actions may be half-applied,
+     * and every caller of this method treats such a failure as fatal.
```
Christopher-Chianelli: Confusing wording; it makes it sound like something is returned when the move/postprocessor throws, when really what happens is the exception is re-raised. Maybe change the wording to say callers should not try to handle any exceptions from this method/treat all exceptions from this method as fatal?

core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
```diff
@@ -676,32 +676,63 @@ void executeAllowingStructurallyFlawedSolutions(Move<Solution_> move) {
...
+     * Exists for exhaustive search backtracking; nothing else needs an undo move.
+     * The returned move <strong>has already been applied</strong> by the time this method returns -
+     * it is what undid {@code move}.
+     * Replaying it is therefore only correct after {@code move} has been re-applied,
+     * which is what {@code restoreWorkingSolution} does by composing the stored undo moves
+     * with the forward moves that descend the tree.
+     * Replaying it without that re-application corrupts the working solution:
+     * for a list variable, the merged undo indices no longer match the list and this throws
+     * {@link IndexOutOfBoundsException}; for a basic variable, it silently restores the wrong value
+     * instead of throwing.
```
Christopher-Chianelli: Am I right in this is describing what happens if you apply the (undo) move incorrectly? Can be confused for what happens if you call this method.

core/src/main/java/ai/timefold/solver/core/impl/move/PendingListChangeTracker.java
```diff
@@ -0,0 +1,91 @@
...
+                Maybe close each beforeListVariableChanged with its afterListVariableChanged \
+                before changing the same entity again."""
+                .formatted(entity, openAction.fromIndex(), openAction.toIndex(),
+                        newAction.fromIndex(), newAction.toIndex()));
+    }
+
+    @Nullable
+    ListVariableBeforeChangeAction<?, ?, ?> remove(Object entity) {
+        if (overflowMap != null) {
+            return overflowMap.remove(entity);
```
Christopher-Chianelli, line 74: Should you make `overflowMap` null if it become empty after a removal, so the more efficient single slot case can be used?

core/src/main/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirector.java
```diff
@@ -73,30 +56,53 @@ public Move<Solution_> createUndoMove() {
...
+            variableChangesEscaped = false;
+        } else {
+            variableChangeList.clear();
         }
     }
 
     @Override
     public void beforeVariableChanged(VariableDescriptor<Solution_> variableDescriptor, Object entity) {
-        variableChanges.add(new VariableChangeAction<>(entity, variableDescriptor.getValue(entity), variableDescriptor));
+        getVariableChangeList()
```
Christopher-Chianelli, line 103: Not so sure how I feel about this; it would be very easy to use the field instead of the getter (which creates a new list if the field is null) instead, especially since we are inside the class that owns the field.

core/src/main/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirector.java
```diff
@@ -112,14 +118,13 @@ public void afterVariableChanged(VariableDescriptor<Solution_> variableDescripto
...
-            cache.put(entity, fromIndex);
-        }
         var list = variableDescriptor.getValue(entity);
-        variableChanges.add(new ListVariableBeforeChangeAction<>(entity,
-                List.copyOf(list.subList(fromIndex, toIndex)), fromIndex, toIndex,
-                variableDescriptor));
+        var action = new ListVariableBeforeChangeAction<>(entity,
+                List.copyOf(list.subList(fromIndex, toIndex)), fromIndex, toIndex, list.size(),
+                variableDescriptor);
+        // Rejects a second bracket for an entity whose previous one is still open.
```
Christopher-Chianelli:
```suggestion
        // pendingListChangeTracker will fail fast if a second beforeListVariableChanged call on the same entity occurs before a corresponding afterListVariableChanged call
```

core/src/main/java/ai/timefold/solver/core/impl/move/ListVariableBeforeChangeAction.java
```diff
@@ -3,30 +3,113 @@
...
+        return originalListSize;
+    }
+
+    /**
+     * Called by {@link VariableChangeRecordingScoreDirector#afterListVariableChanged}
+     * once that call has been validated against this bracket.
+     *
+     * @param mutatedToIndex the {@code toIndex} of that {@code afterListVariableChanged} call
+     */
+    void merge(int mutatedToIndex) {
```
Christopher-Chianelli: This is a confusing method name; what is it merging with (the afterListVariableChange call)? Maybe `setPostToIndex`?

core/src/test/java/ai/timefold/solver/core/impl/move/MoveDirectorTest.java
```diff
@@ -2497,12 +2502,115 @@ void restoreWorkingScoreWithPostprocessorAndFailure() {
...
+        // Replaying it is therefore only correct after the forward move has been re-applied,
+        // which is the invariant exhaustive search backtracking relies on.
+        moveDirector.execute(move);
+        assertThat(entity.getValue()).isEqualTo(badValue);
+        moveDirector.execute(undoMove);
+        assertThat(entity.getValue()).isEqualTo(goodValue);
+    }
+
+    @Test
+    void restoreWorkingScoreWhenHandlingStructurallyFlawedSolutions() {
```
Christopher-Chianelli, line 2592: Confusing test name; the solution is never structurally flawed in this test

Christopher-Chianelli (changes requested):
Code for the most part looks good, but argubly some methods and comments are misleading.

core/src/main/java/ai/timefold/solver/core/impl/move/PendingListChangeTracker.java
triceo: That's a question I struggle with a lot with these improvements.

- It is generally better not to have to allocate or access the map, that much I know.
- But once you've already paid for allocating the map, is it _still_ better to throw it away only to possibly reallocate it all over again just a little later? I do not currently have benchmarks that could prove or disprove.

In this case, I decided to keep the map. This is very much a corner case - outside of K-Opt or other super-complex moves, the map won't show up. If it ever shows up in profiling, we can adjust.

core/src/main/java/ai/timefold/solver/core/impl/move/MoveDirector.java
triceo: You're right that this describes how the returned undo move can or can not be handled. In this case, I do not see the confusion; this method is right to describe how its return should be used. (In other comments on this PR, the case for confusion is IMO stronger.)

That said, I have made it a bit shorter by removing some details which perhaps are not necessary.

core/src/main/java/ai/timefold/solver/core/impl/move/ListVariableBeforeChangeAction.java
triceo: Leftover from a previous impl where this name actually made sense. :-) Didn't spot this.

core/src/test/java/ai/timefold/solver/core/impl/move/MoveDirectorTest.java
triceo: Well spotted. The test now includes a structurally flawed solution.

core/src/main/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirector.java
triceo: IMO the sizing of the list is important enough to warrant reuse within a method.
I do not see much simplification between this and `variableChangeList == null ? newPreSizedList() : variableChangeList`.

As a side note, these before/after methods are called so often that I actually try to make them as small as possible to guarantee inlining. The conditional call would have made for a slightly larger method, although I did not test if this would get us over the inlining threshold.

triceo:
I have addressed the comments. (Some of the things pointed out are arguably judgment calls.)
Please remember to also review the enterprise counterpart. (It is trivial.)

core/src/test/java/ai/timefold/solver/core/testutil/AbstractMeterTest.java
```diff
@@ -20,10 +20,14 @@
...
- * completely separate from the other tests.
+ * Every solver publishes its meters to the same global registry under the same meter names,
+ * and {@link TestMeterRegistry#publish()} keys its measurements by meter name only.
+ * A solver in a concurrently running test therefore overwrites these measurements,
+ * which makes a test read one meter from its own solver and another meter from a foreign solver.
+ * <p>
+ * The global resource lock makes these tests run in isolation, without any other test running.
  */
-@Execution(ExecutionMode.SAME_THREAD)
+@ResourceLock(Resources.GLOBAL)
```
triceo, line 30: Probably on account of the recent meter fix, the flaky tests have become quite a bit more flaky. Since it was screwing up the CI runs here, it made me finally find a real fix.

core/src/main/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirector.java
```diff
@@ -128,49 +133,88 @@ public void beforeListVariableChanged(ListVariableDescriptor<Solution_> variable
...
+        var reportedLengthDelta = toIndex - pendingBeforeAction.originalToIndex();
+        if (actualLengthDelta != reportedLengthDelta) {
+            throw new IllegalArgumentException("""
+                    The afterListVariableChanged (%d, %d) of entity (%s) reports a length change of (%d), \
+                    but its list variable actually changed length by (%d).
+                    Maybe check implementation of your %s; \
+                    its beforeListVariableChanged/afterListVariableChanged range must cover everything it changed."""
+                    .formatted(fromIndex, toIndex, entity, reportedLengthDelta, actualLengthDelta, Move.class.getSimpleName()));
+        }
+        // pendingBeforeAction mutated in place by merge();
```
Christopher-Chianelli: stale comment

```suggestion
        // pendingBeforeAction mutated in place by updateToIndex();
```

core/src/test/java/ai/timefold/solver/core/impl/move/VariableChangeRecordingScoreDirectorTest.java
```diff
@@ -0,0 +1,556 @@
...
+import ai.timefold.solver.core.impl.domain.variable.descriptor.ListVariableDescriptor;
+import ai.timefold.solver.core.impl.score.director.InnerScoreDirector;
+import ai.timefold.solver.core.testdomain.list.TestdataListEntity;
+import ai.timefold.solver.core.testdomain.list.TestdataListSolution;
+import ai.timefold.solver.core.testdomain.list.TestdataListValue;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Covers the merge mechanism in {@link VariableChangeRecordingScoreDirector}:
```
Christopher-Chianelli: might be stale reference to the old method name; maybe move `{@code beforeListVariableChanged}/{@code afterListVariableChanged}` before "merge".

Christopher-Chianelli: Also, `VariableChangeRecordingScoreDirector` has other methods for basic variables competely unrelated to the merge mechanism, and arguably we would eventually want to add tests for those if they were not present before.

Christopher-Chianelli (approved):

PR #2650, opened 2026-09-08
fix: use the same tag keyset everywhere

Christopher-Chianelli:
Some meter registries require the same keyset to be used for meters with the same name.

### DIFF
--- a/core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/SolverTags.java
+++ b/core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/SolverTags.java
@@ -1,0 +1,37 @@
+package ai.timefold.solver.core.impl.solver.monitoring;
+
+import java.util.Objects;
+
+import ai.timefold.solver.core.api.solver.SolverManager;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import io.micrometer.core.instrument.Tags;
+
+/**
+ * Some Micrometer registries like Prometheus require each meter
+ * to be registered with the same tag key set. In order to enforce
+ * each {@link Tags} instance having the same keys, this class is
+ * to be used instead of creating the {@link Tags} manually.
+ * <p>
+ * Each field is nullable; when null, the value of the tag will be "null".
+ *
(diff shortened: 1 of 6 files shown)


core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/SolverTags.java
```diff
@@ -0,0 +1,37 @@
...
+        return Tags.of(
+                "problem.id", Objects.requireNonNullElse(
+                        problemId, "null").toString());
+    }
+
+    public static SolverTags withProblemId(Object problemId) {
+        return new SolverTags(problemId);
+    }
+
+    public static SolverTags withoutProblemId() {
```
triceo: Wouldn't it make more sense to just do a `UUID.randomUUID()`?
I don't like that we export `null` to 3rd party systems, who knows what they'll do with that. Also, `null` fails as a unique ID.

triceo (commented):
I don't like the `null`, otherwise LGTM.

core/src/main/java/ai/timefold/solver/core/impl/solver/monitoring/SolverTags.java
Christopher-Chianelli: My main concern with that is reproducibility.  Unique id is not necessarily required; something that contains the date-time when it started (like `OffsetDateTime.now(ZoneOffset.UTC)`) is arguable more useful to identify the run.

triceo: I'll be OK with some sort of timestamp, although I don't understand how that solves reproducibility.

Christopher-Chianelli: It doesn't; for reproducibility we would need a constant string like "null". But compared to a bunch of random letters and numbers derived from the date (`UUID.randomUUID()`), the actual date timestamp is more useful for identifying a run.

triceo: +1 for some timestamp. Let's make sure we document that, if ID is not provided, that this will happen.

PR #2651, opened 2026-09-09
refactor: improve RNG

triceo:
Replaces JDK-standard `SplittableRandom` with its more advanced implementation, also provided by the JDK; not only does it give us better and more resilient randomness, it may also be a bit faster. While I was there, I also closed some loopholes.

Turtle tests are passing, meaning no new corruption issues, and the solver reaches the same score as it always has. Microbenchmark results are split:

- Score Director shows no changes; that is not surprising, as it barely ever deals with randomness.
- Move providers show significant changes; 5 improvements, 10 regressions, 35 unchanged. Improvements up to 10 %, regressions up to 20 % (mean 6 %). This benchmark is over-sensitive to random numbers, it does them far too much compared to a standard solver run; some moves (such as ChangeMove) are so simple that the benchmark pretty much degenerates to just picking random numbers. Even though it is clear that there will be _some_ regression from these changes, it will not be nearly as significant. In Neighborhoods, we can take the hit, preview API - for move selectors, I'll run some actual models to see what happens.

This is an experiment. Not to be reviewed or merged (yet).


core/src/main/java/ai/timefold/solver/core/impl/solver/random/DelegatingSplittableRandomGenerator.java
```diff
@@ -18,22 +20,36 @@
...
+
+    /**
+     * {@code SplittableRandom} belongs to the legacy group of generators, next to {@code Random}.
+     * LXM replaces it, and the solver splits the generator
+     * per usage, per step, per child thread and per partition.
+     *
+     * @see <a href="https://openjdk.org/jeps/356">JEP 356: Enhanced Pseudo-Random Number Generators</a>
+     */
+    private static final RandomGeneratorFactory<RandomGenerator.SplittableGenerator> DELEGATE_FACTORY =
+            RandomGeneratorFactory.of("L64X128MixRandom");
```
triceo, line 32: This here is the core change.

core/src/main/java/ai/timefold/solver/core/config/heuristic/selector/move/NearbyUtil.java
```diff
@@ -11,28 +11,26 @@
...
     }
 
-    private static EntitySelectorConfig configureEntitySelector(EntitySelectorConfig entitySelectorConfig,
-            RandomGenerator random) {
+    private static EntitySelectorConfig configureEntitySelector(EntitySelectorConfig entitySelectorConfig) {
         if (entitySelectorConfig == null) {
             entitySelectorConfig = new EntitySelectorConfig();
         }
-        var entitySelectorId = ConfigUtils.addRandomSuffix("entitySelector", random);
+        var entitySelectorId = "entitySelector-" + UUID.randomUUID();
```
triceo, line 33: Needs confirming this is safe.
Was this ID ever accessible to the users? Do they have any business depending on it?

triceo:
Proven to be slower without bringing better solution quality.
Closing.

PR #2660, opened 2026-09-14
perf(shadows): replace the declarative change queue with an int heap

larsbeck:
## What

`AffectedEntitiesUpdater` walks the affected nodes of a declarative shadow variable update in topological order, using a `PriorityQueue<BaseTopologicalOrderGraph.NodeTopologicalOrder>`. Profiling a declarative model put a fifth of the solver's CPU time inside that queue. This replaces it with `NodeTopologicalOrderQueue`, a binary min-heap of `int` node ids.

## Why it was expensive

Two reasons, both addressed here.

**Every comparison re-read the topological order through the interface.** `NodeTopologicalOrder.compareTo` calls `BaseTopologicalOrderGraph.getTopologicalOrder(int)` twice, which is an interface dispatch plus — for the union-find backed implementation — a walk to the component representative. A sift does `O(log n)` of those. The new queue reads a node's order **once**, when the node enters, and compares plain ints from there on.

This is sound because a node's order cannot change while nodes sit in the queue: `commitChanges()` has already run by then, and nothing the updater does between `offer` and `poll` touches the graph's edges. That is the same assumption `NodeTopologicalOrder`'s own javadoc already documents.

**Nodes were enqueued once per incoming edge.** The updater pushed each forward neighbour and discarded the duplicates on the way out, after `visited` had already been set — so the heap sifted entries it would go on to throw away. The new queue drops a node that is already in it. That also bounds it at one entry per node, so the two backing arrays never grow.

Neither change affects which nodes are processed, or in which order. Because a node is only ever offered when unvisited, and now only once, the `visited` check on poll became unreachable and was removed.

## Benchmarks

Two models, two machines-worth of configuration, always interleaved A/B with a rebuild between arms so drift hits both sides equally. Apple M4 Pro, JDK 25.0.2 Temurin, `PHASE_ASSERT`.

### A 266-entity declarative model, single threaded

Three interleaved pairs, 120 s each:

| | main | int heap |
|---|---:|---:|
| mean moves/sec | 3 984 | 4 664 |

**+17.1 %**, with no overlap: the slowest patched run beats the fastest baseline run.

A back-to-back profiled pair (async-profiler, 1 ms CPU, `-XX:+DebugNonSafepoints`, 180 s each) shows where it comes from:

| inclusive share of the solver thread | main | int heap |
|---|---:|---:|
| change queue | 19.49 % | 9.59 % |
| `NodeTopologicalOrder.compareTo` | 16.05 % | 0.00 % |
| `MNRDirectedGraph.getTopologicalOrder` | 1.92 % | 0.10 % |
| queue CPU per move evaluation | 50.9 µs | 22.1 µs |
| **everything else per move evaluation** | **210.5 µs** | **208.0 µs** |

The last row is the control: outside the queue, a move evaluation costs the same. The gain is the queue and nothing else.

### A 1 000-job model in release mode, 9 move threads

Packaged Quarkus application, 1 000 jobs / 100 machines / 990 dependencies, five-minute budget, three interleaved pairs:

| | main | int heap |
|---|---:|---:|
| mean moves/sec | 5 865 | 7 575 |
| mean local search steps | 27 410 | 33 343 |

**+29.2 %**, again with no overlap on either metric.

### The same dataset with the #2630 fast path active

Repeated on a variant of that model that does not declare `@ShadowVariablesInconsistent`, so `ignoreInconsistentSolutions` is true and the looped-status bookkeeping added in #2630 is skipped (4 baseline runs, 3 int-heap runs):

| | main | int heap |
|---|---:|---:|
| mean moves/sec | 20 185 | 26 955 |
| time to converge | 222 s | 184 s |

**+33.5 %.** Both arms converge to an identical score, so here the change buys wall clock rather than score. Worth noting that the two optimisations compose, and that the queue's share of the remaining time grows once the bookkeeping is gone.

## Memory

Same dataset and recipe as the release-mode benchmark above, one run per arm at nine move threads, with the heap pinned to `-Xms2g -Xmx2g` on both sides and no profiler attached. Figures come from the GC log, so allocation is measured rather than sampled.

Per unit of work, both are the same:

- **allocation per move evaluation: 0.2 % lower** with the int heap
- **GC pause per move evaluation: 0.7 % higher** — i.e. unchanged

| | main | int heap |
|---|---:|---:|
| move evaluations/sec | 5 996 | 7 384 |
| young collections | 1 041 | 1 279 |
| concurrent cycles | 6 | 6 |
| mixed collections | 4 | 4 |
| full collections | 0 | 0 |
| total GC pause | 2 436 ms | 3 008 ms |
| old gen after GC, trough | 520 MB | 528 MB |
| old gen after GC, mean | 574 MB | 589 MB |

The higher absolute totals are the 23 % of extra work the int-heap arm gets through in the same five minutes; normalised, nothing moves. Old generation occupancy is within 1.5 % at the trough, which is single-run variation.

That matches what the change does structurally. Neither queue allocates in steady state — the `PriorityQueue` is pre-sized to the node count and reused, and the new heap never grows, because deduplication bounds it at one entry per node. An allocation profile puts the queue at 0.0000 % of sampled bytes on **both** arms. What the new queue retains is two `int[nodeCount]` plus a `BitSet` per score director, a few tens of KB each.

One methodology note, in case anyone reproduces this. The heap has to be pinned. Run with a default heap and the int-heap arm looks like it uses 24 % more: G1 sizes the heap up in response to the higher allocation rate, which raises the IHOP threshold in absolute bytes, so it runs half as many concurrent cycles and old generation sits higher between sweeps. Fixing `-Xmx` removes that: the cycle and mixed-collection counts above become identical, and the apparent gap collapses.

## Tests

- Full `core` suite green: 5 947 tests, 0 failures.
- `NodeTopologicalOrderQueueTest` adds 7 tests, including one that cross-checks the queue against `java.util.PriorityQueue` over 50 random workloads with deliberately many tied topological orders.

## Not in this PR

`FixedVariableReferenceGraph` uses a `PriorityQueue<NodeTopologicalOrder>` for its own change tracker and would suit the same treatment, but the benchmarks above do not exercise that path, so I left it alone rather than change something I could not measure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GXxe7Qn7NdmfYYUcsdWtLH


triceo:
Thanks, @larsbeck! Been meaning to do something about that queue for a while.
I'll run my own benchmarks and if the benefits are confirmed (which I expect they will), we'll review and merge.

triceo:
@Christopher-Chianelli Please review.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * A binary min-heap of graph node ids, ordered by {@link BaseTopologicalOrderGraph#getTopologicalOrder(int)}.
+ * It replaces a {@code PriorityQueue<NodeTopologicalOrder>} on the declarative shadow variable update path,
+ * where profiling showed the heap to be a fifth of the solver's CPU time.
+ * It differs from that queue in two ways, both of which leave the processing order intact.
+ * <p>
+ * <b>The topological order is read once, when a node enters the queue, and is then held in {@link #orders}.</b>
+ * {@code NodeTopologicalOrder.compareTo} instead reads it through the {@link BaseTopologicalOrderGraph} interface
+ * twice per comparison, which costs an interface dispatch plus, for the union-find backed implementations,
+ * a walk to the component representative - and a sift touches {@code O(log n)} comparisons.
```
Christopher-Chianelli: sift?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+            index = childIndex;
+        }
+        nodeIds[index] = nodeId;
+        orders[index] = order;
+    }
+
+    @Override
+    public String toString() {
+        return "size = " + size;
+    }
```
Christopher-Chianelli: Undescriptive toString; not really needed.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+            }
+            nodeIds[index] = nodeIds[parentIndex];
+            orders[index] = orders[parentIndex];
+            index = parentIndex;
+        }
+        nodeIds[index] = nodeId;
+        orders[index] = order;
+    }
+
+    private void siftDown(int startIndex) {
```
Christopher-Chianelli: It seems you invented new terminology for `upheap`/`downheap`?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+ * and bounds the queue at one entry per node, so the backing arrays never have to grow.
+ * <p>
+ * Ties are broken arbitrarily, as they are by a {@link java.util.PriorityQueue}. Two nodes share a topological order
+ * only when neither is a predecessor of the other, so neither one's update can feed the other's, and the order in
+ * which the two are processed does not affect the outcome.
+ * <p>
+ * This class is not thread safe.
+ */
+@NullMarked
+final class NodeTopologicalOrderQueue {
```
Christopher-Chianelli, line 38: I personally dislike `sift`; its usage is mostly historical; upheaping/downheaping does not relate to sifting in any way.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private void siftUp(int startIndex) {
+        var index = startIndex;
+        var nodeId = nodeIds[index];
+        var order = orders[index];
+        while (index > 0) {
+            var parentIndex = (index - 1) >>> 1;
+            if (orders[parentIndex] <= order) {
+                break;
+            }
```
Christopher-Chianelli, line 126: I would add a comment saying "this node comes before its parent, so swap with its parent in the heap"

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+        while (index < half) {
+            var childIndex = (index << 1) + 1;
+            var rightIndex = childIndex + 1;
+            if (rightIndex < size && orders[rightIndex] < orders[childIndex]) {
+                childIndex = rightIndex;
+            }
+            if (order <= orders[childIndex]) {
+                break;
+            }
+            nodeIds[index] = nodeIds[childIndex];
```
Christopher-Chianelli, line 157: I would add a comment saying "this node come after at least one of its children, so swap the parent with an earlier child."

core/src/test/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueueTest.java
```diff
@@ -0,0 +1,184 @@
...
+        queue.offer(0);
+        queue.offer(1);
+        queue.offer(2);
+        queue.offer(3);
+
+        var polled = drain(queue);
+        assertThat(polled.getFirst()).isEqualTo(2);
+        assertThat(polled.subList(1, 4)).containsExactlyInAnyOrder(0, 1, 3);
+    }
+
```
Christopher-Chianelli, line 149: Missing a test that draining a queue where only some nodes are offered does not include the nodes in the graph that were not offered.

i.e.
```
  var queue = new NodeTopologicalOrderQueue(graphWithOrders(0, 10, 20, 30, 40), 4);
  queue.offer(1);
  queue.offer(3);
  var polled = drain(queue);
  assertThat(polled).containsExactly(1, 3);
```

Christopher-Chianelli (changes requested):
Code largely looks good, although historical, I believe `sift` is a terrible name for describing `upheap`/`downheap`; `upheap`/`downheap` is not a filter (while sifting is). The name is no where near a standard, with many implementations using different names. I argue it is more closer to bubbling up/sinking down to a layer with similar density. I would use `upheap`/`downheap` instead of trying to tie the operation to a real world process.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * A binary min-heap of graph node ids, ordered by {@link BaseTopologicalOrderGraph#getTopologicalOrder(int)}.
+ * It replaces a {@code PriorityQueue<NodeTopologicalOrder>} on the declarative shadow variable update path,
+ * where profiling showed the heap to be a fifth of the solver's CPU time.
+ * It differs from that queue in two ways, both of which leave the processing order intact.
+ * <p>
+ * <b>The topological order is read once, when a node enters the queue, and is then held in {@link #orders}.</b>
+ * {@code NodeTopologicalOrder.compareTo} instead reads it through the {@link BaseTopologicalOrderGraph} interface
+ * twice per comparison, which costs an interface dispatch plus, for the union-find backed implementations,
```
Christopher-Chianelli: Union-find is not used here or anywhere. Remove this part of the Javadoc.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
larsbeck: I guess it is a matter of taste. I have no strong opinion, just that Java's [PriorityQueue](https://github.com/openjdk/jdk25u-dev/blob/268b211e79c6e8a49db4cd2a1be979b76811c4f0/src/java.base/share/classes/java/util/PriorityQueue.java#L95-L103) which is implemented as a balanced heap also uses [siftUp](https://github.com/openjdk/jdk25u-dev/blob/268b211e79c6e8a49db4cd2a1be979b76811c4f0/src/java.base/share/classes/java/util/PriorityQueue.java#L635) and [siftDown](https://github.com/openjdk/jdk25u-dev/blob/268b211e79c6e8a49db4cd2a1be979b76811c4f0/src/java.base/share/classes/java/util/PriorityQueue.java#L668-L718).

I am fine with changing it if you have strong feelings. Wdyt @triceo ?

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
larsbeck: I wrote "move into the hole", not "swap", because the code does not swap: it shifts parents or children into the hole and writes the node once, after the loop.

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
larsbeck: same as above

core/src/main/java/ai/timefold/solver/core/impl/domain/variable/declarative/NodeTopologicalOrderQueue.java
triceo: No strong opinion either. It's true that the word "sift" doesn't speak to me, I don't even know what it means and I had never seen it before I looked at `PriorityQueue` source code.

larsbeck: Ok, I'll rename them.

triceo (approved):
I have verified Chris' comments are resolved, and I have none of my own. I will merge this once CI is green. (With the exception of the models tests, which are consistently broken.)

PR #2662, opened 2026-09-15
perf(bavet): speed up ifExists+join preload

triceo:
Commit fd13b5210d5412d38d0c53b7917cbf4737b1081a introduces "prepareForSettle", which performs certain iterative operations before nodes propagate, in order to solve a data staleness issue. This step is unnecessary in some cases, and this commit allows for skipping it.

Most importantly, this step will be skipped with a clean session, to which fact will only be inserted, never updated or retracted. Those facts cannot become stale, and therefore this overhead can be eliminated.

Identified during Neighborhoods work, where a single super-expensive stream would show this behavior where previously we had no idea it existed.


core/src/main/java/ai/timefold/solver/core/impl/bavet/common/AbstractIfExistsNode.java
```diff
@@ -157,9 +176,46 @@ protected final boolean isPendingLeft(LeftTuple_ leftTuple) {
...
     public final boolean canDeferWork() {
-        return isFiltering;
+        if (isFiltering && inputLayerDelta < 0) {
+            throw new IllegalStateException("Impossible state: the input layer delta of node (%s) was never set."
+                    .formatted(this));
+        }
+        // A filtering node whose inputs are at most one layer apart can only ever read a tuple
+        // that has already been told it is doomed, which the per-read isActive() guards catch.
+        // Two or more layers apart, the doom may not have propagated that far yet; only then is deferral needed.
+        return isFiltering && inputLayerDelta >= 2;
```
Christopher-Chianelli: I wonder if the 2 in here and in the join should be a static constant that can be referred by the tests. Maybe something like `MAX_EAGER_DELTA` or `MIN_DEFER_DELTA`?

core/src/test/java/ai/timefold/solver/core/impl/bavet/bi/UnindexedJoinBiNodeTest.java
```diff
@@ -98,4 +107,54 @@ void updateRight_refreshesRightFact() {
...
+            var preload = preloadSequence[i];
+            if (i == 0 && !preload) {
+                continue; // Do not disable preload which is already disabled.
+            }
+            if (preload) {
+                node.preloadStarted();
+            } else {
+                node.preloadEnded();
+            }
+        }
```
Christopher-Chianelli: Does this test anything? It would be a `preloadedStarted()` and `preloadEnded()` with nothing inbetween.

core/src/test/java/ai/timefold/solver/core/impl/bavet/uni/UnindexedIfExistsUniNodeRoundsTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private static final BiPredicate<Integer, Integer> GREATER_THAN = (left, right) -> left > right;
+
+    private final LiveTupleTracker downstream = new LiveTupleTracker();
+    private final SequentialPositionTracker positionTracker = new SequentialPositionTracker();
+    private final UnindexedIfExistsUniNode<Integer, Integer> node =
+            new UnindexedIfExistsUniNode<>(true, downstream, GREATER_THAN, positionTracker);
+    private final Propagator propagator = node.getPropagator();
+    private final List<UniTuple<Integer>> leftTupleList = new ArrayList<>();
+    private final List<UniTuple<Integer>> rightTupleList = new ArrayList<>();
```
Christopher-Chianelli: I am a bit concerned about tests sharing state with these as final fields without a corresponding `setUp` or `tearDown`.

core/src/test/java/ai/timefold/solver/core/impl/score/stream/bavet/BavetRegressionTest.java
```diff
@@ -420,6 +420,170 @@ public void filteringJoinNullConflictRightQuadJoinUnassignOne() {
...
+            scoreDirector.beforeListVariableChanged(variableDescriptor, entity, 0, 2);
+            entity.getValueList().remove(value1);
+            scoreDirector.afterListVariableChanged(variableDescriptor, entity, 0, 1);
+            scoreDirector.afterListVariableElementUnassigned(variableDescriptor, value1);
+
+            assertScore(scoreDirector,
+                    assertMatch(value2, value2, value2));
+        }
+    }
+
```
Christopher-Chianelli, line 874: I would also add a regression test for `ifExists` and `ifNotExists`, since those nodes were also affected.

Christopher-Chianelli (changes requested):
The fix looks good, although I think the tests should be improved.

triceo:
As I was making changes based on the review, I got finally fed up with how many things I have to copy-paste between joins and ifExists. The classes now share a common parent, extracting the shared logic and hiding some internals.

Christopher-Chianelli (approved):

PR #2665, opened 2026-09-15
feat: Create profile set custom seed

diogodanielsoaresferreira:
- Add execution profile to set random seed in run (using the var QUARKUS_TIMEFOLD_SOLVER_DEFAULT_RANDOM_SEED)
- Add possibility to add parameters to execution profiles

Fixes: https://github.com/TimefoldAI/timefold-platform/issues/5335


cristianonicolai (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/impl/executionprofile/SeedExecutionProfile.java
```diff
@@ -0,0 +1,68 @@
...
+        return Map.of(PARAMETER_SEED, seed);
+    }
+
+    @Override
+    public Map<String, String> toEnvironment(Map<String, String> resolvedParameters) {
+        String seed = resolvedParameters.get(PARAMETER_SEED);
+        if (seed == null) {
+            return Map.of();
+        }
+        return Map.of(ENV_QUARKUS_RANDOM_SEED, seed);
```
mswiderski, line 57: so the expectation is that gateway will call this method before scheduling pod and set these variables on pod spec? 

Maybe we could consider a custom implementation of ConfigSource that we could back it by the configuration from storage - that way we could load execution profile and the options in that config source which would then deliver anything that quarkus config properties might expect.

diogodanielsoaresferreira: The idea is good, and we could read it from storage, but the problem is that ConfigSources initialize very early in the lifecycle, and the storage clients are configured via CDI, so we would have to connect to the storages manually, which would be fickly.

Another option would be to rely on a configurations file mounted by the gateway that could be read by the pod. But that would again increase the complexity.

As I think that this options works for now and it's easier to maintain, I would keep it as is for now.

mswiderski (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/impl/executionprofile/SeedExecutionProfile.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+
+import ai.timefold.solver.service.definition.internal.executionprofile.ExecutionProfile;
+
+/**
+ * Runs the solver with a fixed random seed, making a run reproducible.
+ * <p>
+ * The seed is an optional {@code seed} parameter; when it is not supplied, a random seed is generated and persisted with
```
zepfred: Note that the seed generation depends on the environment mode. By default, we use the `PHASE_ASSERT` mode, which has a seed value of `0`. In contrast, the `NON_REPRODUCIBLE` mode uses a random number as its seed.

diogodanielsoaresferreira: So that means that for this to be useful, we also need to set the environment mode to `NON_REPRODUCIBLE`?

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
```diff
@@ -21,13 +21,11 @@ public record RunConfiguration(
                 description = "Optional maximum number of threads to be used for solving.",
                 minimum = "1") @JsonInclude(JsonInclude.Include.NON_EMPTY) @Positive Integer maxThreadCount,
         @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(
-                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags) {
+                description = "Optional tags to be assigned to the dataset.") @Size(max = 100) Set<String> tags,
+        @JsonInclude(JsonInclude.Include.NON_NULL) @Schema(hidden = true) Map<String, String> options) {
```
zepfred, line 25: Can we pass arbitrary values to `options`?

service/test-model/src/test/java/ai/timefold/solver/service/testmodel/EmployeeScheduleResourceTest.java
```diff
@@ -337,6 +347,55 @@ void getIssueTypeByCode() {
...
+                new ModelRequest<EmployeeSchedule, EmptyModelConfigOverrides>(createInputEmployeeSchedule())
+                        .withConfiguration(Configuration.<EmptyModelConfigOverrides> empty()
+                                .withRun(new RunConfiguration("unknown-property-e2e")));
+
+        ObjectNode body = objectMapper.valueToTree(modelRequest);
+        JsonNode runNode = body.path(ModelRequest.ModelRequestAttribute.CONFIG.value()).path("run");
+        assertThat(runNode.isObject()).as("run configuration should be serialized as an object").isTrue();
+        ((ObjectNode) runNode).put("notARealOption", "boom");
+
+        // Permitting unknown properties in the schema must not weaken the strict Jackson mapping.
```
zepfred, line 388: I think this test answers my previous question in `RunConfiguration`.

service/definition/src/main/java/ai/timefold/solver/service/definition/impl/executionprofile/SeedExecutionProfile.java
zepfred: No, it just means the seed `0` will be used with the default environment. But I see you already updated the documentation, which makes more sense to me now. I believe we want to keep the runs reproducible wherever possible, and we can now change the seed if necessary. Using `NON_REPRODUCIBLE` does not seem necessary.

zepfred (approved):

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
diogodanielsoaresferreira: Yes, but only the needed values will be used by the Execution Profile; other ones will be discarded

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
```diff
@@ -11,7 +12,6 @@
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 
-@Schema(additionalProperties = Schema.False.class)
```
rsynek, line 14: Since the solver-service is now guarded by Revapi, this change requires a justification in the `src/build/revapi-differences.json`.

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfile.java
```diff
@@ -29,11 +24,13 @@ public interface ExecutionProfile {
...
 
     /**
-     * Additional configuration contributed by this profile, applied to the run's environment - each entry is injected as
-     * an environment variable into the solver pod. Keys must be valid environment-variable names. When multiple profiles
-     * are activated and define the same key, the resulting value is unspecified. Defaults to no extra configuration.
+     * Reads the values this profile recognizes from the run's options (as supplied via {@code RunConfiguration.options}),
+     * validates them, and maps them to environment variables injected into the solver pod. The profile picks out only the
+     * keys it recognizes and ignores the rest, since the options map is shared with other run configuration. It must not
+     * invent values the caller did not supply. Keys must be valid environment-variable names. When multiple profiles are
+     * activated and define the same key, the resulting value is unspecified. Defaults to no environment variables.
```
rsynek, line 31: I assume debugging will be difficult in such a situation. Should we rather fail fast if multiple active profiles define the name key?

service/definition/src/main/java/ai/timefold/solver/service/definition/api/domain/RunConfiguration.java
diogodanielsoaresferreira: Done 👍

service/definition/src/main/java/ai/timefold/solver/service/definition/internal/executionprofile/ExecutionProfile.java
diogodanielsoaresferreira: Yes, applied in timefold-orbit, since there we have an overall view of all execution profiles.

PR #2667, opened 2026-09-15
fix: do not add a structurally flawed move as a finalist

### LINKED ISSUE #2666: Local search executes a structurally flawed move when a step accepts nothing
**Describe the bug**

A local search step whose moves were **all rejected** still executes one of them. When the move it picks leaves the solution structurally flawed, `MoveDirector.execute` throws and the solve dies:

```
java.lang.IllegalArgumentException: The move (...) caused the solution to become structurally flawed.
	at ai.timefold.solver.core.impl.move.MoveDirector.execute(MoveDirector.java:667)
	at ai.timefold.solver.core.impl.move.MoveDirector.execute(MoveDirector.java:654)
	at ai.timefold.solver.core.impl.score.director.AbstractScoreDirector.executeMove(AbstractScoreDirector.java:459)
	at ai.timefold.solver.core.impl.localsearch.DefaultLocalSearchPhase.doStep(DefaultLocalSearchPhase.java:121)
	at ai.timefold.solver.core.impl.localsearch.DefaultLocalSearchPhase.solve(DefaultLocalSearchPhase.java:112)
	at ai.timefold.solver.core.impl.solver.AbstractSolver.runPhases(AbstractSolver.java:100)
	at ai.timefold.solver.core.impl.solver.DefaultSolver.solve(DefaultSolver.java:158)
```

This affects any model with declarative shadow variables whose graph can become cyclic — a job scheduling model where a job may not start before the job it depends on has finished, and where placing a job ahead of its own dependency on the same machine closes a loop. Such moves are rejected constantly and correctly; the problem is only what happens at the end of a step in which nothing was accepted.

**Expected behavior**

A structurally flawed move is never executed as a step. If a step has no accepted move, the phase should end (or pick nothing) rather than execute a rejected move.

**Actual behavior**

The forager falls back to the best *selected* move, which may be a rejected and structurally flawed one, and `doStep` executes it. Instrumenting `DefaultLocalSearchPhase.doStep` at the point of failure shows:

```
stepIndex=10289 recordedScore=-1structural/0hard/0soft recordedFlawed=true selected=2 accepted=0
```

The score recorded for the winning move is already marked structurally flawed, and `accepted=0`.

**To Reproduce**

Branch with a reproducer test, based on `main` at `87ba4c0b41aea74dd024b19c116e609f83508fb0`:

https://github.com/larsbeck/timefold-solver/tree/repro-structurally-flawed-step

[`core/src/test/java/ai/timefold/solver/core/impl/localsearch/StructurallyFlawedStepTest.java`](https://github.com/larsbeck/timefold-solver/blob/repro-structurally-flawed-step/core/src/test/java/ai/timefold/solver/core/impl/localsearch/StructurallyFlawedStepTest.java)

```
mvn -pl core test -Dtest=StructurallyFlawedStepTest
```

Two tests, both failing on `main` by design, whole class runs in about 4 seconds:

- `foragerDoesNotPickARejectedStructurallyFlawedMove` — pins the mechanism in half a second with no solve at all: a single rejected, structurally flawed move is the only thing the forager is given, and `pickMove` hands it back as the step.
- `solveDoesNotExecuteAStructurallyFlawedMove` — end to end on the existing `TestdataDependencyNoInconsistentField*` test domain (5 entities, 5 dependency chains of 8 values), construction heuristic plus `LATE_ACCEPTANCE` over `listChange` + `listSwap`.

It needs **no Enterprise, no move threads and no diversified late acceptance** — it reproduces single threaded in community core with plain late acceptance.

## Environment

**Timefold Solver Version or Git ref**: `main` @ `87ba4c0b41aea74dd024b19c116e609f83508fb0`

```
openjdk version "25.0.2" 2026-01-20 LTS
OpenJDK Runtime Environment Temurin-25.0.2+10 (build 25.0.2+10-LTS)
OpenJDK 64-Bit Server VM Temurin-25.0.2+10 (build 25.0.2+10-LTS, mixed mode, sharing)
```

```
Darwin 25.6.0 Darwin Kernel Version 25.6.0: Fri Jul 31 19:17:26 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T6041 arm64
```

## Additional information

Three places combine to allow it, each reasonable on its own:

1. `AbstractAcceptor.isAccepted` rejects every structurally flawed move, as it should:
   ```java
   if (moveScope.getScore().isStructurallyFlawed()) {
       return false;
   }
   ```
2. `AcceptedLocalSearchForager.addMove` passes the podium every move regardless of acceptance, and `HighestScoreFinalistPodium.addMove` keeps an unaccepted one as long as no accepted move has turned up (`if (finalistIsAccepted && !accepted) return;`) — a fallback so that a step always has something to do.
3. `AcceptedLocalSearchForager.pickMove` returns that fallback finalist, and `DefaultLocalSearchPhase.doStep` executes it.

A step ends with nothing accepted mainly when **termination fires part way through it**. That is why the end-to-end reproducer hinges on the move count limit rather than on the problem alone: `190_000` finishes cleanly, `200_000` does not. It also means the failure is not step-deterministic in a time-boxed solve — it depends on the instant the spent or unimproved limit lands, so the same dataset crashes in one run and not in another, and a faster build hits it on a different step. On a real 10 000 job dataset at 9 move threads we saw it in roughly one run in eight.

The podium's ranking itself is fine: it compares `moveScope.getScore().raw()`, and `Score.compareTo` orders by `structuralScore()` first, so a flawed move always ranks below a non-flawed one. The fallback only surfaces a flawed move when *every* move the step got to see was rejected — which is exactly what a mid-step termination produces, as in the `selected=2 accepted=0` case above.

Christopher-Chianelli:
Fixes #2666


triceo (changes requested):
Shouldn't we rather ask who's sending those moves to the podium, and stop it?
The solver already knows perfectly well how to handle a situation where no moves were accepted, it's not a new situation - only one that will perhaps happen more often now with the specific JS usage pattern. 

If the move got as far as finalist, IMO we failed much earlier in filtering out the move. This is further confirmed for me by two other finalist impls having to make the same check.

(We agreed that the behavior of structurally flawed solutions would be to treat the move as if it never existed; which means having it leak anywhere outside of the one place which needs to execute it should not be an acceptable solution.)

core/src/main/java/ai/timefold/solver/core/impl/localsearch/decider/forager/finalist/AbstractFinalistPodium.java
```diff
@@ -30,6 +30,10 @@ public void stepStarted(LocalSearchStepScope<Solution_> stepScope) {
 
     protected void clearAndAddFinalist(LocalSearchMoveScope<Solution_> moveScope) {
         finalistList.clear();
+        if (moveScope.getScore().isStructurallyFlawed()) {
+            throw new IllegalArgumentException("Finalist (%s) is structurally flawed."
```
triceo: Should these be "Impossible state:" now?

triceo (approved):

PR #2673, opened 2026-09-17
perf: fixed graph update once

fodzal:
**Issue.**
In a fixed variable reference graph, when two shadow variables share a source and one
also depends on the other, the shared source enqueues both directly, and updating the first
re-enqueues the second through their edge before dedup catches it. Its supplier then runs twice for
nothing:

```java
@ShadowVariable(supplierName = "durationSupplier") // @ShadowSources("value")
Integer duration;

@ShadowVariable(supplierName = "endTimeSupplier")  // @ShadowSources({ "value", "duration" })
Integer endTime;
```

`value` marks `duration` and `endTime` directly; `duration` also marks `endTime` via the
`duration → endTime` edge. `endTimeSupplier` runs twice for the same pass.

**Fix.** The dedup check happened at enqueue time (`visited`, seeded from only the first queued
node) instead of at dequeue time. Track updated nodes and check when a node comes *out* of the
queue instead of when it goes in:

```java
if (updated.get(changedNodeId)) {
    continue;
}
```

**Benchmark.** 200 entities, each a chain of 20 shadow ints all sourced from the same genuine
variable and each also depending on the previous link; every move marks all 20 directly and
re-marks 19 of them through the chain:

| | before | after |
|---|---|---|
| supplier calls / move | ~35 | 20 (exact) |
| move time, trivial supplier | ~3.4 µs | ~3.5 µs (noise; overhead dominates) |
| move time, realistic supplier | ~28.3 µs | ~18.6 µs (**-34%**) |


Christopher-Chianelli (changes requested):
There is a simpler way of doing this;  modify the original code to use the field `isChanged` (which already have all nodeIds in  `changeTracker` and thus will not readd it to the queue)  and delete the local variable `visited`. This would eliminate the need of the `updated.get` call in the loop and will additionally prevent allocating a new bitset on every update. You may need to modify the constructor so isChanged is updated when nodes are added to changeTracker.

fodzal:
> There is a simpler way of doing this; modify the original code to use the field `isChanged` (which already have all nodeIds in `changeTracker` and thus will not read it to the queue) and delete the local variable `visited`. This would eliminate the need of the `updated.get` call in the loop and will additionally prevent allocating a new bitset on every update. You may need to modify the constructor so isChanged is updated when nodes are added to changeTracker.

Thanks, it's way simpler and more efficient like this.
Same logic in the constructor (through `markChanged`) even if it's less important there.

Christopher-Chianelli (approved):
LGTM when green. Thanks for your contribution @fodzal!

fodzal:
@Christopher-Chianelli
In fact, I just realized we can do it even simpler (and more performant on big graphs) by reusing the `NodeTopologicalOrderQueue` of https://github.com/TimefoldAI/timefold-solver/pull/2660
Sorry for the back and forth!

Christopher-Chianelli (approved):
LGTM when green

